PR #1, opened 2017-01-18
OSE rebase proposal

kargakis:
@jupierce @stevekuznetsov @tdawson @bbguimaraes @smunilla

### DIFF
--- a/docs/proposals/ose-rebase.md
+++ b/docs/proposals/ose-rebase.md
@@ -1,0 +1,90 @@
+## Rebase OSE on Origin
+
+### Abstract
+
+Today, code in OSE lands by fetching and merging Origin manually. OSE-specific changes are
+scattered across the repo and it is hard to reason about any conflicts. This doc proposes
+moving all the changes in OSE that diverge from Origin into a set of commits that will sit
+on top of the Origin code. A Jenkins job will be rebasing OSE master on top of Origin master
+on a daily basis and rebases will continue to happen only for the master branch so the rest
+of this doc does not touch on updates in release branches.
+
+
+### Naming requirements
+
+There are different types of commits that need to be carried on top of Origin master:
+* a long-term carry related to Oauth template branding
+* a long-term carry related to release tooling
+* a commit that is generated in every build of OSE and holds the web console branding
+* a long-term carry that includes all the generated diff from old tito commits
+* a tito generated commit that tags the latest build of OSE
+
+In order to properly handle the additional commits, we need to set naming requirements
+for the messages of those commits. In particular:
+* long-term carries should start with a `[CARRY]` prefix followed by a prefix that denotes
+the type of the commit. For example, "[CARRY][BUILD] Tooling updates" denotes that this
+commit needs to be carried long-term and is specific to the building process of the repo.
+* updates to any existing [CARRY] commit should start with a `[SQUASH]` prefix followed
+by the type of the commit that they are supposed to squash to. For example, if the commit
+from the example above needs to be updated, a developer can simply create a PR that starts
+with "[SQUASH][BUILD]" followed by any message.
+* commits that are not going to be carried long-term but are necessary until the next rebase
+lands, should use a `[DROP]` prefix. Note that these commits are going to be dropped
+automatically by the rebase process so you should be sure when to use this prefix.
+
+### The process
+
+During the actual rebase, `GIT_SEQUENCE_EDITOR` is going to be used by passing it a script that
+will automatically handle all the logic of dropping and squashing commits in long-term carries
+as necessary. If merge conflicts arise, they will need to be manually resolved as it already is
+the case today. In the future, we will extend the conflict resolution mechanism in Jenkins to
+send an e-mail to the author and committer of the conflicting commits.
+
+A nightly build of OSE should follow the steps below in order to rebase on top of Origin:
+```sh
+git clone git@github.com:openshift/ose.git
+pushd ose
+git remote add upstream git@github.com:openshift/origin.git
+git fetch upstream
+git checkout master
+
+# Note down the current latest tag because squashing the tito commit into specfile updates
+# will drop it and tito tag needs it further down the build process in order to determine
+# the next (new) tag.
+last_tag="$( git describe --abbrev=0 --tags )"
+
+# rebase.py handles all the logic of squashing and dropping commits.
+# The most common state of master is going to be the following but
+# rebase.py shouldn't really care about the order but only make sure
+# that all the additional commits on top of Origin master follow the
+# naming requirements this proposal sets.
+#
+#
+#         .-Other [CARRY] commit
+#        /  .-[CARRY] Specfile updates
+#       /  /  .-[DROP] webconsole bump
+#      /  /  /  .-Tito tag commit
+#     /  /  /  /
+# m--c--c--w--t
+#  \           \
+#   \           `-ose/master/HEAD,v3.5.0.23
+#    `-origin/master/HEAD
+#
+GIT_SEQUENCE_EDITOR=rebase.py git rebase -i upstream/master
+
+# We need to retag, because the previous squash removed the latest tito tag.
+git tag "${last_tag}" HEAD
+```
+
+Common conflicts can rise when new changes step on branding or tooling code. Manual resolution is the
+only thing we can do in such cases. Nothing should change from what happens today when a conflict needs
+to be resolved but we can reduce the amount of conflicts by using build flags that will generate branding
+code depending on the repository that is being built. We should also investigate whether we can stop
+checking in various code like tito diffs, generated docs, bindata, and the ose-images script. Avoiding
+those diffs will reduce even further the number of conflicts.
+
+
+In case a new [CARRY] commit needs to be introduced or an existing one needs to be updated, developers
+do not need to force-push to master. Instead, construct your commit messages using the naming requirements
+specified above and the rebase process should be responsible for collapsing [SQUASH] commits into existing
+[CARRY] commits.


docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,39 @@
...
+
+Today, code in OSE lands by fetching and merging Origin manually. OSE-specific changes
+are scattered in the repo and it's hard to reason about any conflicts. This doc proposes
+moving all the changes in OSE that diverge from Origin into a set of commits that
+will sit on top of the Origin code[0] and a Jenkins job that will be rebasing master and
+all active release branches on a daily basis.
+
+There are two types of commits that need to be carried on top of Origin: 
+1. long-term carries related to branding and release tooling (master and release branches)
+2. CVEs (release branches only)
```
stevekuznetsov: Technically there is another type: bug-fixes that are not CVEs but that we do not back-port to Origin releases (since we do not want to incur the maintenance cost).

stevekuznetsov: [from @kargakis]
Not sure I undestand this. Is the separation of Origin CI from OSE CI not a maintainace cost per se? Or is it less expensive to distribute the load brought up by testing/merging PRs across the two repos? I would expect us to have a clear flow as to how commits flow from one repo to the other. The less bidirectional this thing is the better.

stevekuznetsov: This is not really our decision to make here, I don't think -- it's a PM/product thing. We don't want to maintain Origin releases and be expected to re-release them with more bug fixes. I guess we could ask, if we take all fixes to OSE and automatically start bringing them back, and get an automated process for build&release for both Origin and OSE, then maybe the argument is weaker against it. However, today it's a product distinction.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,39 @@
...
+Rebases are going to be simpler for the master branch. The OSE master will be hard-reset to
+the Origin master and any carry commits will be applied on top. If any conflicts are hit, manual
+resolution will be required. In the future, we can extend the conflict resolution mechanism in
+Jenkins to send an e-mail either to the author and committer of the carry commit or we can store
+e-mails in the commit description and use those instead.
+
+Rebases for release branches are less trivial because they may contain backported fixes for CVEs
+that do not exist in the respective Origin branches. In such a case, backports should continue
+to be carried until we identify that the backport exists in the Origin branch and drop the OSE commit.
+How? Assuming the backport ends up w/o any additional change in Origin then cherry-picking
```
stevekuznetsov: We can't use the `UPSTREAM: NUMBER: MESSAGE` syntax for commits but we could do `UPSTREAM: MESSAGE` and enforce that the commit messages line up, to allow us to detect the commit in Origin. We should also have some validation to ensure the diffs are identical in the Origin release branch and the OSE release branch. If not, we'd need a different carry commit to bring us from the Origin release CVE commit to the OSE release CVE commit, but this should either be rare or disallowed.

stevekuznetsov: [from @kargakis]
I didn't propose anywhere in the proposal using the "UPSTREAM: NUMBER: MESSAGE" pattern. I expected that the rebase would work like this:
(assuming we know the hashes of our carries / bugfixes)
```
$ git reset --hard upstream/release-1.*
$ git cherry-pick $commit_hash
```

If the cherry-pick succeeds, we can move to the next. If not, there are two reasons why it can fail:

1. either there are conflicts that need to be resolved or
2. a bugfix (even with a different commit message) with the exact diff cannot be applied because the diff already exists
    so you get something like

```
On branch test
You are currently cherry-picking commit 665673d.

nothing to commit, working directory clean
The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty

Otherwise, please use 'git reset'
```

Unfortunately I may end up parsing output in case of an error to separate between the two cases but at least this solves the naming problem.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,39 @@
...
+the OSE commit will produce no additional commit. Conflicts in backports need to be resolved manually.
+
+Code that already exists in Origin master and needs to be backported in an enterprise branch can be simply
+backported in the respective Origin release branch and the Jenkins job should be responsible for merging
+the Origin release branch into the OSE release branch.
+
+Naming conventions we will have to enforce in commit messages in OSE:
```
stevekuznetsov: We need to make a mechanism for master PRs whereby someone can go through the normal PR process to add to the long-term carries, maybe with a `[SQUASH]` tag and a matching commit message. This would allow for a normal PR where the author is not required to rewrite history, but then on the next rebase their commit would be squashed into the parent carry commit.

stevekuznetsov: [from @kargakis]
SGTM

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,39 @@
...
+* commits that need to be carried long-term such as branding or release tooling changes will start with [CARRY]
+* commits for CVEs may contain a [CVE] label. These commits should be the only fixes that developers
+would need to open pull requests for in the OSE repo (apart from rebases, assuming an automated rebase
+fails, and it needs to run manually due to conflicts).
```
stevekuznetsov: Some of the diffs look like simple `sed 's/Origin/Container Platform/g'` type diffs -- do we want to be applying git patches or do we want to be running diff scripts?

stevekuznetsov: [from @kargakis]
I am not sure about about the difference between the two or why this distinction matters.

stevekuznetsov: Well, if we were doing diffs, it matters what the diff context is, right? So

```diff
  foo
- origin
+ container platform
  bar
```

Would not necessarily apply over something like

```diff
- foo
+ foobar
  origin
- bar
+ barfoo
```

So we could reduce the conflicts potentially by just doing the `sed` without context. Similarly, if we were doing the `sed` we would be able to catch new occurrences of things like "Origin" automatically as they get introduced.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,39 @@
...
+
```
stevekuznetsov: We need to determine if we really want generated stuff from tito in master carry commits, or if we can just auto-gen the entire changelog and specfile in release branches. This is impacted by the choice of doing releases from master or not. If we are branching out for Online releases, maybe master won't do any.

stevekuznetsov: [from @kargakis]
I don't see this as a prerequisite for being addressed in this proposal. We can consolidate all tito diffs in a commit and once we decide that we want it autogenerated we simply remove the carry?

stevekuznetsov: Agreed.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,39 @@
...
+We will need to repurpose the commitchecker tool[1] to work with OSE and take into account our tags and
```
stevekuznetsov: To make use of commitchecker we may need to move things like `hack/ose_images.sh` to a separate dir (as commitchecker is directory based).

stevekuznetsov: [from @kargakis]
Can't comment unless I actually get to work on the commitchecker. This is more of an implementation detail I guess.

stevekuznetsov: Agreed.

docs/proposals/ose-rebase.md
stevekuznetsov: In the case where a bug-fix touches code that has branding, for instance, or any other part of the codebase where we carry a patch, the bug-fix as it lands in Origin will not be identical to that which hits OSE. By using a naming convention, we would not have to rely on the diffs being identical for the cherry-pick to fall out. Do we want that? I'm not sure.

I was also thinking we could just set up `github/openshift/origin.git` as a remote and literally `git rebase` onto that `master` branch. No need to manually `cherry-pick` anything. The rebase should fail in the same manner. We can hash out impls later but I think we can get porcelain output from `git` and not need to do any output munging.

stevekuznetsov: For context, if a commit goes into Origin as `UPSTREAM: xxx: whatever` and `xxx` PR has merged into k8s by the time the next rebase happens, the Origin `UPSTREAM` is dropped, but it's done as a process, not because the diff is identical (it can be identical, but it does not need to be).

jupierce:
@stevekuznetsov @kargakis 
I was discussing this with Jon Toppins (kernel dev) and he has seen a successful pattern that leverages a "branding branch" instead of using cherry-picks/tags. Not sure if this would be easier to reason about, but thought I'd bring it up.

Something like this:
git clone git://../openshift/ose --single-branch -b master
cd ose
git remote add upstream git://../openshift/origin
git fetch --all
git reset upstream/master --hard
git checkout -b branding/osev3
git rebase upstream/master
git checkout master
git merge branding/osev3
// branding has now been applied on top of clean upstream master
// pushes

With some process around it, seems like we could have per-release branding as things evolve. For CVEs, perhaps another release specific branch which could managed as a "layer" on top of branding.

docs/proposals/ose-rebase.md
kargakis: I am probably lacking some context here. What is the process *today* with respect to rebasing release branches?

stevekuznetsov:
@jupierce I'm not clear as to what that achieves that having branding commits in `master` and doing a `rebase` does not. Seems like it is roughly equivalent, but will also generate merge commits? The process I was considering was:

```sh
git clone git@github.com/openshift/ose.git
pushd ose
git remote add upstream git@github.com/openshift/origin.git
git fetch upstream
git checkout master
git rebase upstream/master
```

docs/proposals/ose-rebase.md
kargakis: Yes, I think running `sed` for fixing branding w/o context makes sense.

docs/proposals/ose-rebase.md
kargakis: > For context, if a commit goes into Origin as UPSTREAM: xxx: whatever and xxx PR has merged into k8s by the time the next rebase happens, the Origin UPSTREAM is dropped, but it's done as a process, not because the diff is identical (it can be identical, but it does not need to be).

Last time I did a Kube rebase this was a manual process. Has it changed ever since?

kargakis: > In the case where a bug-fix touches code that has branding, for instance, or any other part of the codebase where we carry a patch, the bug-fix as it lands in Origin will not be identical to that which hits OSE. By using a naming convention, we would not have to rely on the diffs being identical for the cherry-pick to fall out. Do we want that? I'm not sure.

When you talk about naming convention here does it apply only for OSE commits or for commits in both repos?

> I was also thinking we could just set up github/openshift/origin.git as a remote and literally git rebase onto that master branch. No need to manually cherry-pick anything. The rebase should fail in the same manner. We can hash out impls later but I think we can get porcelain output from git and not need to do any output munging.

This also sounds good.

jupierce:
@stevekuznetsov .. yeah, merge commits aside, it looks like Jon's pattern collapses to yours.

docs/proposals/ose-rebase.md
stevekuznetsov: There are two cases I think -- `s/Origin/Container Platform/g` and the doc URL one.

docs/proposals/ose-rebase.md
kargakis: Ok, I think for this proposal it's enough to take into account the additional bugfixes in release branches and assume that they don't exist in Origin release branches.

stevekuznetsov: AFAIK `enterprise-3.x` branches in OSE are never synced with the Origin `release-1.x` branches after they're cut. But yes, agreed -- the distinction is important as CVEs are probably going to be in Origin branches as well, the bug-fixes, no.

docs/proposals/ose-rebase.md
stevekuznetsov: > Last time I did a Kube rebase this was a manual process. Has it changed ever since?

We should ask Andy, but I thought the format, etc was for machine reading?

> When you talk about naming convention here does it apply only for OSE commits or for commits in both repos?

Name in OSE could be `UPSTREAM: <message>` and we would ask authors to use `<message>` in Origin as well.

jupierce:
@kargakis Since it seems like the rebase approach is workable, can the proposal be written to include "standard operating procedures" for the various eventualities? This would include what the automated and manual git steps are.

(1) What the nightly build looks like in terms of git steps.
(2) What cutting an OSE release looks like / corresponding steps in Origin (ongoing or one-time).
(3) What introducing a new branching change into OSE looks like.
(4) What introducing a CVE into OSE looks like.
(5) How to handle that CVE moving into origin (cherry picking from OSE?).
(6) What normal conflicts can arise / how to reconcile the conflict.
(7) What might go wrong if normal procedures aren't followed (e.g. someone doesn't follow the right procedure for (5)) / how to reconcile the conflict.

kargakis:
@jupierce I just pushed an update addressing most if not all comments from @stevekuznetsov. I agree that even more granularity in this proposal is needed so as part of my next push I will try to address your points, thanks.

stevekuznetsov:
@jupierce the card was focused on `master` -- do we want to address the CVE stuff and releases as well right now?

kargakis:
> @jupierce the card was focused on master -- do we want to address the CVE stuff and releases as well right now?

IIUC the card is about proposing a mechanism that will address rebases on both master and the release branches. I don't mind exploring this in greater detail for the release branches.

docs/proposals/ose-rebase.md
kargakis: > Name in OSE could be UPSTREAM: <message> and we would ask authors to use <message> in Origin as well.

The problem with this approach is that it is not enforcable in any way (or at least I cannot think of any) so inevitably some commit names will diverge and slip in Origin in subtle different ways that will break our flow. Assuming we address branding before-the-fact, 95% of the CVEs should end up with the same diff in Origin. Or even better how about a mechanism that will automatically open PRs with the CVEs in Origin as soon as we can have them in Origin?

jupierce:
@stevekuznetsov Unless it somehow explodes the scope, this seems like a good moment to ensure the entire OSE lifecycle is clean / automation friendly.

kargakis:
Thinking about this more, I don't think we need to do rebases for OSE release branches from Origin release branches. Since we don't need to update Origin release branches, we can simply:

* rebase OSE master onto Origin master
* rebase OSE latest_release_branch onto OSE master
* the rebase process doesn't care about old release branches, fixes would continue to land there as today

stevekuznetsov:
> Since we don't need to update Origin release branches,

Origin release branches get updated with CVEs and important bug-fixes. I'm not sure this holds.

stevekuznetsov:
> Origin release branches get updated with CVEs and important bug-fixes

e.g.

https://github.com/openshift/origin/pull/12634
https://github.com/openshift/origin/pull/12606
https://github.com/openshift/origin/pull/12470
https://github.com/openshift/origin/pull/12263
https://github.com/openshift/origin/pull/12388
...

kargakis:
These land first in OSE though, right? So a mechanism that pulls them back
to Origin should be a separate thing and it's still not required to rebase
branches?

stevekuznetsov:
They land first in OSE, but I would expect us to rebase OSE release branches onto Origin release branches to keep the diff valid. Pulling them back is a manual process and will remain a manual process. Why would we not want to rebase release branches? I thought in all cases we were wanting to keep the diff on top of Origin minimal.

kargakis:
Since fixes land in OSE first, creating an admittedly complex process that
involves mapping commits by names and resolving conflicts seems to
outweight what we would gain by keeping the diffs minimal. I am not sure
that divergence (a commit in OSE lands in Origin slightly different) in
release branches is a real problem at all. I think simply rebasing the
latest OSE release branch against OSE master should do what we need.

stevekuznetsov:
A situation I would like to avoid would look like:
 1. bob adds a CVE fix to OSE release branch
 2. bob backports this to Origin but for whatever reason needs to change it a little
 3. we don't rebase the OSE release branch
 4. joe adds a new CVE fix on top of bob's in the OSE release branch
 5. joe can't cleanly backport his fix to Origin since it sits on top of the OSE version of joe's patch


@tdawson what are your thoughts? should we bother rebasing OSE release branches at all?

tdawson:
I don't see what rebasing released branches from Origin gives us.
For starters, we currently don't do it.  When was the last time we saw anything in the Origin 1.2 branch.  Yet we just released errata for OCP 3.2.  We have no plans to merge in the Origin 1.3.3 release because it only has a small percentage of the bugs fixed that are in OCP 3.3.
You need to keep in your mind that Origin does not support it's releases for the length of time that OCP does. 
I think  @kargakis says it correctly.  It creates a very large level of complexity that totally throws away any benefit you get from doing what you have planned.

stevekuznetsov:
Cool. That makes it simpler. Only rebase on `master` and forget release branches. No need for complexity around dropping UPSTREAM commit as `master` should have none.

kargakis:
Updated the proposal with all latest comments.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,53 @@
...
+create git commits locally automatically, eg. hack/cherry-pick-from-ose.sh openshift/ose/pull/578.
+
+Common conflicts can rise when new changes step on branding or tooling code. Manual resolution is the
+only thing we can do in such cases but we can reduce the amount of conflicts by using build flags that
+will replace branding instances depending on the repository that is being built. We should also investigate
+whether we can stop checking in various code like tito diffs, generated docs[1], and the ose-images script.
+Avoiding those diffs will reduce even further the number of conflicts.
+
+Naming conventions we can use in commit messages in OSE:
+* commits that need to be carried long-term such as branding or release tooling changes will start with [CARRY]
```
kargakis: If we end up carrying no diffs on top of Origin, this conventions seems redundant..

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,53 @@
...
+* commits that are fixing branding or release tooling changes will start with [SQUASH]. The rebase process
+should be responsible for squashing those commits appropriately into the [CARRY] commits.
+* bugfixes and CVEs can contain a [FIX] tag. We would have to enforce these when developers open PRs against the
```
kargakis: Since bugfixes/CVEs end up in branches and we are not going to rebase branches, this convention is not needed.

docs/proposals/ose-rebase.md
stevekuznetsov: I don't see a way out of carrying `tito` changes and `Dockerfile` changes (we expect `Dockerfiles` in OSE to be identical to those in dist-git)

docs/proposals/ose-rebase.md
stevekuznetsov: I see some cases where we might do a rebase but agree that it's going to be rare and manual so we don't need this

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,51 @@
...
+
+Common conflicts can rise when new changes step on branding or tooling code. Manual resolution is the
+only thing we can do in such cases but we can reduce the amount of conflicts by using build flags that
+will replace branding instances depending on the repository that is being built. We should also investigate
+whether we can stop checking in various code like tito diffs, generated docs[1], and the ose-images script.
+Avoiding those diffs will reduce even further the number of conflicts.
+
+Naming conventions we can use in commit messages in OSE:
+* commits that need to be carried long-term such as branding or release tooling changes will start with [CARRY]
+* commits that are fixing branding or release tooling changes will start with [SQUASH]. The rebase process
```
kargakis: There is this single commit from tito that will always need to be squashed into the main tito diff.
For example, https://github.com/openshift/ose/commit/d55d08ff385afbd67e0aa90e40d5ef77cd370816
If we are going to use [SQUASH], we should probably need to extend tito with an option to prepend suffixes in git commits?
Or is it enough to identify the tito commit and squash it directly in the tito main commit?

stevekuznetsov: `tito` wants those commits to be where the tags are. Can we investigate not having them in `master`? They are useful in figuring out the next `tito tag` version but they are not useful in that all of the other information that they change in the `.spec` file can be generated at any point (for example, without an intermediate tag, the changelog would just be longer on the next tag).

kargakis: @stevekuznetsov are you suggesting removing the tito commit altogether from `master`? I thought that was impossible based on your other comment (https://github.com/openshift/aos-cd-jobs/pull/1/files#r99820921). If squashing the commit is simpler, I would go with it for now.

kargakis: > but they are not useful in that all of the other information that they change in the .spec file can be generated at any point (for example, without an intermediate tag, the changelog would just be longer on the next tag).

Hm, I see what you mean here. We still need to store the previous `tito tag` version somewhere, right?

stevekuznetsov: Yeah, we need to store the previous tag somewhere -- I'm not sure what can be done here but would like you and @tdawson to try to hash it out. A rebase won't be possible if we have a tito-injected commit every day in `master`. We have `tito tag --use-version VERSION --use-release RELEASE` these days which might be a workaround, but yeah still unclear.

kargakis: > A rebase won't be possible if we have a tito-injected commit every day in master. 

Why is this a problem if tito tagging applies the new diff after the rebase on top of the already carried main tito commit?

stevekuznetsov: If we remove the `tito` commits via squash, how do we keep the tags that used to point to those commits?

kargakis: > If we remove the tito commits via squash, how do we keep the tags that used to point to those commits?

Can you elaborate more? Tito tags point to the commits that produce them? It's not clear by looking at the changelog why this is a problem.

stevekuznetsov: Today we have this situation:

```
          .-carry-.            .-ose/master
         /         \          /
--o--o--c--c--c--c--c--x--o--o
      \                 \
       `-origin/master    `-v3.4.0.2
```

Where commits `c` are carries on top of `origin/master`. As you can see, commit `x` is generated by `tito` and the tag that `tito` creates is pointing to that commit. If we squash commit `x` into some other commit ... where does that tag go?

kargakis: Where is the relationship between x and the tag stored? The commit stored in the changelog during `tito tag` is not x but x-1.
https://github.com/openshift/ose/commit/4f84c83229824d4727810a9dc5d5f49e83e1d145#diff-3477fbcab168ba3ee773ffeb9323e53eR24

stevekuznetsov: Yeah so we want the `x-1` commit to build Origin off of, as the Origin build cites the last source code commit in it's build metadata. The tag from `tito`, however, sits on the commit that updates the specfile (`x` in my example).

stevekuznetsov:
After talking with @danmcp  today we now know we need some brand-less generic `master` and other branches that follow it for Online, OCP, Dedicated, maybe more. Sorry this comes really late in the game but I don't think we were aware of this until now.

danmcp:
@stevekuznetsov Make sure we loop in @tdawson on this.  It's possible some of these can be shared.

kargakis:
We ended up agreeing that this proposal is just for master and we already
strive for stripping away whatever we can. What do the new branches mean
for this proposal? I would expect them to be orthogonal to how we update
master.

docs/proposals/ose-rebase.md
kargakis: Now I see, thanks. I can't see a way out of having these commits in master since we build off of master...

kargakis: Do we need just the latest tag generated by tito or the assumption is that people can access any built version at any point in time?

stevekuznetsov:
But that is also just my view I think in reality we should choose whatever makes the process easiest.

docs/proposals/ose-rebase.md
stevekuznetsov: I'm not really sure. The point of _tito_ is to create a tree where you can access any given build and re-build it deterministically, but the only place where I see that being useful for us is re-creating builds like 3.4, for which we have branches. If we don't keep any `tito` commits, we can keep track of versions some other way and instruct `tito` to use a specific version and release on each tag ... but that seems like a massive workaround and we don't get to use `tito` for what `tito` actually wants to be doing. In the end, too, what we do here has to make sense from a process perspective for Troy. 

Especially with the revelation about daily builds of OCP and Online ... let's try to set up a chat this morning.

kargakis:
Updated the proposal with feedback from the latest discussions with @stevekuznetsov @tdawson and @jupierce - the git commands are tentative but at least there is agreement into what we want to do.

docs/proposals/ose-rebase.md
```diff
@@ -39,11 +45,9 @@ Avoiding those diffs will reduce even further the number of conflicts.
 
 Naming conventions we can use in commit messages in OSE:
 * commits that need to be carried long-term such as branding or release tooling changes will start with [CARRY]
-* commits that are fixing branding or release tooling changes will start with [SQUASH]. The rebase process
```
kargakis: Removed because I don't think automatically squashing is trivial - the initial rebase proposal can assume the git history is as expected: Origin commits + carry commits + tito commit, this should always be the git state of the repo between rebases w/o any manual intervention.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,55 @@
...
+```
+git clone git@github.com/openshift/ose.git
+pushd ose
+git remote add upstream git@github.com/openshift/origin.git
+git fetch upstream
+git checkout master
+# The following two commands assume the last two commits in git history are the main tito
+# diff and the last tito tag commit.
+git reset --soft HEAD~2
+git commit -m "[CARRY][BUILD] Specfile updates"
```
stevekuznetsov: Also want to carry the last tag forward

```sh
last_tag="$( git describe --abbrev=0 --tags )"
git reset --soft HEAD~2
git tag "${last_tag}"
```

kargakis: The last tag will be produced by `tito tag` further down the build process:
https://github.com/openshift/aos-cd-jobs/blob/bfa5857ec56fa55031ed6f29a6c9f69b66c7a2fb/scripts/merge-and-build.sh#L97-L141
Or am I missing something?

stevekuznetsov: `tito` looks at the current tag to generate the new tag -- so if `tito tag` sees nothing it will use version `0` or something

stevekuznetsov: I'm not 100% on it -- `tito` also keeps versions in the file at `.tito/packages` but I don't really want to make this change require a specific new version build of `tito`, so if we can try to fulfill as many of it's assumptions that is better -- if you do it without re-tagging and it works, ok, even better

kargakis:
@stevekuznetsov @jupierce @tdawson any final comments - this should be ready now

stevekuznetsov:
Proposal looks fine, lets get a MVP together

kargakis:
Updated the proposal with comments from https://github.com/openshift/aos-cd-jobs/pull/40

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,120 @@
...
+will continue to be backported back to Origin manually but we may eventually build a script that will
+create git commits locally automatically, eg. hack/cherry-pick-from-ose.sh openshift/ose/pull/578.
+
+Common conflicts can rise when new changes step on branding or tooling code. Manual resolution is the
+only thing we can do in such cases but we can reduce the amount of conflicts by using build flags that
+will replace branding instances depending on the repository that is being built. We should also investigate
+whether we can stop checking in various code like tito diffs, generated docs[1], and the ose-images script.
+Avoiding those diffs will reduce even further the number of conflicts.
+
+In case of manual conflict resolution, or when a new [CARRY] commit needs to be introduced in master, or
```
stevekuznetsov: We'll need to send out some e-mails as in general, working with OSE is going to become more of a PITA -- you'll need to 

```sh
git checkout master
git fetch upstream
git reset --hard upstream/master
```

Every time now, `pull` won't help you.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,120 @@
...
+#      /  /  /  .-Tito tag commit
+#     /  /  /  /
+# m--c--c--s--t
+#  \           \
+#   \           `-ose/master/HEAD,v3.5.0.23
+#    `-origin/master/HEAD
+#
+# In the first case, we simply want to reset the last two commits and commit them as one.
+# In the second case, we want the same thing as in the first case plus to completely drop
+# the webconsole bump since a new one may be created during the current build.
```
stevekuznetsov: If we drop the existing Web Console carry every time one exists, aren't we guaranteed to get a new one with the script? OCP will never track what's in Origin for the web console.

kargakis: No. Imagine a scenario where in the previous rebase we got a bump because the console wasn't updated in Origin and in the current rebase what we get from Origin includes HEAD of console.

kargakis: I have already seen that a couple of times with the current process. Nothing should change really as we transition from merge to rebase.

stevekuznetsov: I thought OSE always tracked a non-master version of the console with OCP branding in it? And Origin always tracked master?

docs/proposals/ose-rebase.md
kargakis: Not clear to me what you are saying here. Where `pull` gets into the picture? Because of the interactive rebase? I think it will be easier to resolve conflicts by using `git rebase -i` rather than resetting and cherry-picking the carries back.

docs/proposals/ose-rebase.md
kargakis: Interesting. That is the case but it seems that as branches close down for the release, the scenario above may happen since we switch rebasing from upstream/master to upstream/release-new. So in a normal sprint cycle when we rebase from master we should always carry a commit that rollbacks the console from what's in Origin master to what's in the console branch?

docs/proposals/ose-rebase.md
kargakis: But I can be convinced otherwise

stevekuznetsov: No, I meant for developers that may be familiar with using `git pull` in normal workflows in other repositories ... won't work here anymore.

docs/proposals/ose-rebase.md
stevekuznetsov: @tdawson could you please weigh in here

docs/proposals/ose-rebase.md
kargakis: 1. I hope people are not using `git pull` in general.
2. Can you describe in more detail what workflow would not work and where is that workflow useful today?

stevekuznetsov: I'm a developer. I checked out `ose refs/heads/master` at some point in the past. I need to do a backport PR, so I go do:

```sh
cd "${GOPATH}/github.com/openshift/ose"
git checkout master
git fetch upstream
git merge upstream/master
git checkout -b my-feature-branch
# git work done
```

But now the `git merge` (or `git pull` or whatever people are using to get to the latest state before working) is not going to work, since we are re-writing history in `master` all the time. It is not a blocker but is worth letting people know that they should be `reset --hard` for what's in `master`.

stevekuznetsov: "Not a blocker" === benefits we get from doing the automated rebase far, far, *far* outweigh the added mental effort people need to have to deal with it. But they should be alerted, or there's going to be a bunch of confusion!

kargakis: Backports not related to branding are not landing in master but worth letting people know of our new process, that's for sure.

docs/proposals/ose-rebase.md
tdawson: What Steve said is correct.
origin-web-console/master => origin
origin-web-console/enterprise-3.5 => ose-3.5
origin-web-console/enterprise-3.4 => ose-3.4
...etc.
If the ose branding gets blown away, such as accidentally running the script pointing to master, and then you rerun the branding script, pointing to the correct branch, then the branding get's put back.

So, if the branding get's blown away, we just need to make sure it get's put back before the final tag.

stevekuznetsov: What is `ose/master` tracking usually? Latest enterprise branch in web console?

kargakis: @tdawson this is not related to the branding commit (do we have a script that generates the branding diff?)

When we merge Origin master into OSE master, hack/vendor.sh is rolling back the console from console/master (what is vendored inside Origin master) to console/enterprise-3.x (where x is the new release for which no branch has been cut yet).

When a release approaches and we cut the new branch and switch from master to enterprise-3.x rebases, we merge Origin release-1.x into OSE enterprise-3.x, hack/vendor.sh may or may not produce any additional diff for the console.

For example:
https://github.com/openshift/ose/commits/enterprise-3.5
Automatic commit of package [atomic-openshift] release [3.5.0.35-1]. 
Automatic commit of package [atomic-openshift] release [3.5.0.33-1]. 

and others have no web console bump followed by them, most likely because there was no diff between the vendored console in Origin and OSE.

Another thing that I just realized is that when we merge Origin master to OSE master, we also vendor the web console **master** even though we are targetting the enterprise branch.
https://github.com/openshift/aos-cd-jobs/blob/626a5f7f31a6d0da8657ed175da603a7aa04e76c/scripts/merge-and-build.sh#L69

kargakis: > When we merge Origin master into OSE master, hack/vendor.sh is rolling back the console from console/master (what is vendored inside Origin master) to console/enterprise-3.x (where x is the new release for which no branch has been cut yet).

So this is not true. When we are doing master rebases, we vendor the web console from master.

stevekuznetsov: From @tdawson on IRC:

```
origin-web-console/master -> origin (plain)
origin-web-console/enteprise-3.6 -> ose/master (right now)
origin-web-console/enterprise-3.5 -> ose/enterprise-3.5
```

kargakis: Note that `origin-web-console/master` is merged to `origin-web-console/enteprise-3.6` before `origin-web-console/enteprise-3.6` is merged to `ose/master` so essentially this is `origin-web-console/master` to `ose/master`.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,105 @@
...
+git fetch upstream
+
+git add -A
+git commit -m "[CARRY] for a reason"
+
+# Rebase against the last Origin HEAD that was brought in during the previous rebase. This should open
+# the git editor of your choice and include the commits we are carrying on top of Origin. Reorder
+# appropriately to ensure the commit order described above is respected.
+git rebase -i $(git merge-base master upstream/master)
+git push origin master -f
```
stevekuznetsov: This requires that contributors have force-push rights to `master`. Can we use a flow by which the rebase script identifies if there are three consecutive commits as described above, and if there is anything else after it, rebases to squash them down or make them come before the last three that we require?

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,105 @@
...
+The tag that points to the last tito commit needs to be retaged because the squash is dropping
+it. If any conflicts are hit, manual resolution will be required as it is already the case
+with `git merge`. In the future, we can extend the conflict resolution mechanism in Jenkins
+to send an e-mail to the author and committer of the carry commits.
+
+A nightly build of OSE should follow the steps below in order to rebase on top of Origin:
+```sh
+git clone git@github.com/openshift/ose.git
+pushd ose
+git remote add upstream git@github.com/openshift/origin.git
```
tdawson: Should this be

```
git clone git@github.com:openshift/ose.git
pushd ose
git remote add upstream git@github.com:openshift/origin.git
```
I'm using git-1.8.3.1 and it like : instead of / between github.com and openshift

stevekuznetsov: Yep, this looks like a typo

docs/proposals/ose-rebase.md
stevekuznetsov: I think it's out of scope of this proposal... for now. But we should look in the future to use `GIT_SEQUENCE_EDITOR` with `git rebase -i`. This would allow us to do more complex things, like allowing people without force-merge access to commit to `master`.

kargakis: Yeah, let's not block on this one but have it as a follow-up. I'll create a new Trello card once we get this out of the door.

docs/proposals/ose-rebase.md
kargakis: Good catch, fixed.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,107 @@
...
+# Squash the tito commit into the specfile updates
+git add .tito origin.spec
+git commit -m "[CARRY][BUILD] Specfile updates"
+
+# Drop the unstaged changes for the web console - a new diff will be regenerated once
+# we revendor the Web Console for OSE.
+git checkout -- pkg/assets/bindata.go pkg/assets/java/bindata.go
+
+# We need to retag, because the previous squash removed the latest tito tag.
+git tag "${last_tag}" HEAD
```
rhcarvalho: At this point, do we expect to have clones of the repository that knew about `last_tag`?

One thing that bit me in the past is that after re-tagging and pushing the tag to a remote, other clones of the repository will NOT get the updated reference of the tag on fetch/pull.
This leads to a situation where different clones of the repo think a certain tag point to a different commit.

kargakis: If you need access to tags other than the latest one, you should be able to get the srpm from the mirrors they are uploaded after every build.

docs/proposals/ose-rebase.md
```diff
@@ -0,0 +1,110 @@
...
+
+In order to properly handle the additional commits, we need to set naming requirements
+for the messages of those commits. In particular:
+* long-term carries should start with a `[CARRY]` prefix followed by a prefix that denotes
+the type of the commit. For example, "[CARRY][BUILD] Tooling updates" denotes that this
+commit needs to be carried long-term and is specific to the building process of the repo.
+* updates to any existing [CARRY] commit should start with a `[SQUASH]` prefix followed
+by the type of the commit that they are supposed to squash to. For example, if the commit
+from the example above needs to be updated, a developer can simply create a PR that starts
+with "[SQUASH][BUILD]" folllowed by any message.
```
stevekuznetsov: `s/folllowed/followed/`

kargakis:
There seems to be consensus over this proposal so I am going to merge it, mainly because I want to use it as a link in sanity_check.py so when developers submit a PR in OSE master that is not following the naming requirements, they will get an error message pointing to this proposal. Additional comments can be addressed in separate PRs.

PR #5, opened 2017-02-02
Update to support openstack and linch-pin

arilivigni:
- Added topology os_origin-3node-cluster.yml
- Changed macros and combined setup of linch-pin and provision cluster
- Don't use ShiningPanda plugin run own Virtual Venv
- Removed Axes for System-CPython-2.7


jhadvig:
@jupierce @stevekuznetsov fyi

jhadvig:
@arilivigni did you tried to run the changes ?

PR #6, opened 2017-02-03
Initial commit of Origin `oct` Jenkins jobs

stevekuznetsov:
The set of Jenkins jobs that are run on a pull request to Origin that
use the `origin-ci-tool` are generated using the tools in this commit.
To change the set of jobs that are included in a full Origin test run,
change `children/test_pull_requests_origin_future.txt`. To change the
commands that are run in any specific test, change the corresponding
file under `commands/`.

Re-generate the set of job XMLs with `generate.sh`.

Push the new job XMLs to the Jenkins server by running `push-update.sh`
with `$USERNAME` and `$PASSWORD` set.

Verify that any new commit you make is checking in the accompanying
generated changes with `verify.sh`.


stevekuznetsov:
@jupierce

stevekuznetsov:
@smarterclayton PTAL `commands/*`

commands/test_pull_requests_origin_check_future.sh
```diff
@@ -0,0 +1,9 @@
+# run commitchecker outside release container as it needs
+# access to git; also explicitly force godeps verification
+symbolic_ref="$( git symbolic-ref HEAD )"
+branch="${symbolic_ref##refs/heads/}"
+if [[ "${branch}" == "master" ]]; then
```
stevekuznetsov: @smarterclayton so before here we were using `${BRANCH}` which comes from the Jenkins environment.

I wasn't super happy with this -- so the API here would be, write the script in this file as if you were SSH-ed in. I'll escape `$` chars, etc. Don't try to make use of Jenkins envars, so for instance here I got the branch name from the git state.

PR #11, opened 2017-02-07
Set workspace selinux label manually.

bbguimaraes:
@jupierce: PTAL.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,6 +46,11 @@ git merge ${merge_opts} '${commit}' -m '${commit_msg}'
 """)
 }
 
+// https://issues.jenkins-ci.org/browse/JENKINS-37069
+def fix_workspace_label() {
+    sh "chcon -Rt svirt_sandbox_file_t '${env.WORKSPACE}'*"
+}
+
 node('buildvm-devops') {
     properties([[
         $class: 'ParametersDefinitionProperty',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -46,6 +46,11 @@ git merge ${merge_opts} '${commit}' -m '${commit_msg}'
 """)
 }
 
+// https://issues.jenkins-ci.org/browse/JENKINS-37069
+def fix_workspace_label() {
+    sh "chcon -Rt svirt_sandbox_file_t '${env.WORKSPACE}'*"
```
stevekuznetsov: Can you mount the volume into the container with `:Z` relabeling?

bbguimaraes: I should have linked the other PR where we discussed this: https://github.com/jupierce/aos-cd-jobs/pull/5

The problem is jenkins is the one adding the mount flag for the workspace, and looking through the code (linked in other PR) there's no way to change that.

stevekuznetsov: I'd be interested in what @pmorie opinion is of the relabel

pmorie: It is better to use the `:z` (note, lowercase) bindmount flag for this, if possible.  The name of `svirt_sandbox_file_t` is changing soon, so if we hard-code the SELinux type, we'll have to update it when the policy is updated.

bbguimaraes: We really have our hands tied by the way docker plugin executes the `docker run` command:

https://github.com/jenkinsci/docker-workflow-plugin/blob/master/src/main/java/org/jenkinsci/plugins/docker/workflow/client/DockerClient.java#L110

The problem is not so much about the workspace, we could copy it to another directory and mount it with the `:Z` suffix. Jenkins also mounts another directory (that's why we have the `*` at the end) to keep some control files (logs, pid, etc.) and the mount flags for that are also hard-coded.

stevekuznetsov: I submitted a PR to the docker workflow plugin repository, but for the time being let's use this workaround.

PR #19, opened 2017-02-09
Update build ose branch

kargakis:
@jupierce @tdawson @bbguimaraes

### DIFF
--- a/merge-build-push/merge-and-build.sh
+++ b/merge-build-push/merge-and-build.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
-# Setup
-#set -o xtrace
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
 OSE_MASTER="3.5"
 if [ "$#" -ne 2 ]; then
   MAJOR="3"
@@ -15,13 +18,19 @@ OSE_VERSION="${MAJOR}.${MINOR}"
 if [ "${OSE_VERSION}" != "${OSE_MASTER}" ] ; then
   PUSH_EXTRA="--nolatest"
 fi
-BUILDPATH="${HOME}/go"
+
+# Use the directory relative to this Jenkins job.
+BUILDPATH="${WORKSPACE}/go"
+mkdir -p $BUILDPATH
 cd $BUILDPATH
-export GOPATH=`pwd`
+export GOPATH="$( pwd )"
 WORKPATH="${BUILDPATH}/src/github.com/openshift/"
+mkdir -p $WORKPATH
 echo "GOPATH: ${GOPATH}"
 echo "BUILDPATH: ${BUILDPATH}"
 echo "WORKPATH ${WORKPATH}"
+
+go get github.com/jteeuwen/go-bindata
 
 if [ "${OSE_VERSION}" == "3.2" ] ; then
   echo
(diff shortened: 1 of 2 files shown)


merge-build-push/merge-and-build.sh
```diff
@@ -15,7 +18,10 @@ OSE_VERSION="${MAJOR}.${MINOR}"
 if [ "${OSE_VERSION}" != "${OSE_MASTER}" ] ; then
   PUSH_EXTRA="--nolatest"
 fi
-BUILDPATH="${HOME}/go"
+
+# Use the directory relative to this Jenkins job.
+BUILDPATH="${WORKSPACE}/go"
+mkdir -p $BUILDPATH
 cd $BUILDPATH
 export GOPATH=`pwd`
```
tdawson: Because you are starting with a fresh workspace, you have to add the following after you export your GOPATH

`go get github.com/jteeuwen/go-bindata`

stevekuznetsov: Also don't use back-ticks -- they're a deprecated feature. Use the `$()` subshell expansion.

And don't mix scoping statements (`export`, `readonly`, `local`, `declare`) with something else that has a return code on the same line.

```sh
GOPATH="$( pwd )"; export GOPATH
```

kargakis:
```
./merge-build-push/merge-and-build.sh: line 24: WORKSPACE: unbound variable
```
https://atomic-e2e-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/aos-cd-builds/job/build%252Fose-rebase-test/9/console
@bbguimaraes any idea why this is happening? Do I need to tinker things more in the Jenkinsfile?

bbguimaraes:
On Fri, Feb 10, 2017 at 01:12:02AM -0800, Michail Kargakis wrote:
> ```
> ./merge-build-push/merge-and-build.sh: line 24: WORKSPACE: unbound variable
> ```
> https://atomic-e2e-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/aos-cd-builds/job/build%252Fose-rebase-test/9/console
> @bbguimaraes any idea why this is happening? Do I need to tinker
> things more in the Jenkinsfile?

@kargakis: works on my jenkins instance.  I noticed only now that the version
there is 1.651.3, while the openshift image is 2.32.2.  Maybe that's the
reason, I can't find documentation for that version.

You can pass it in from the pipeline (assuming `env.WORKSPACE` exists
there, it does on version 2) or just use `$PWD` at the beginning of
the script.

kargakis:
@jupierce @tdawson based on the state of https://github.com/openshift/aos-cd-jobs/tree/build/ose-pipeline and I guess the intention is to switch to it, do you think it's worth maintaining a separate branch with duplicate work?

jupierce:
@kargakis It is duplicate work for now, but I think it also mitigates some (admittedly small) risks. The bash based method is much closer to what works reliably today. We should move to the pipeline version only once we have a reliable fallback in place.

kargakis:
@jupierce makes sense

kargakis:
@jupierce ptal, these changes should work now. You can compare the diff with what I have in https://github.com/kargakis/aos-cd-jobs/tree/build/ose-rebase-test which now passes in Jenkins.

jupierce:
LGTM

PR #22, opened 2017-02-10
Populate `env.WORKSPACE` if not defined

bbguimaraes:
`env.WORKSPACE` is not available in pipelines.  This is fixed on newer versions
of jenkins/pipeline-plugin.  The solution here was tested and works for the
newer version and is a combination of suggestions from the issue linked in the
code, hopefully compatible with the environment we have.

@jupierce: PTAL, very much open to suggestions for improvements.

@kargakis, @tdawson: FYI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,6 +46,13 @@ git merge ${merge_opts} '${commit}' -m '${commit_msg}'
 """)
 }
 
+// https://issues.jenkins-ci.org/browse/JENKINS-33511
+def set_workspace() {
+    if(env.WORKSPACE == null) {
+        env.WORKSPACE = pwd()
+    }
+}
+
 // https://issues.jenkins-ci.org/browse/JENKINS-37069
 def fix_workspace_label() {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -46,6 +46,13 @@ git merge ${merge_opts} '${commit}' -m '${commit_msg}'
 """)
 }
 
+// https://issues.jenkins-ci.org/browse/JENKINS-33511
+def set_workspace() {
+    if(env.WORKSPACE.toString() == 'null') {
```
bbguimaraes: `env.WORKSPACE` seems to be a piece of magic, `!env.WORKSPACE` and
`env.WORKSPACE == null` do not work.  This is the closest that works.

kargakis: How about using `mktemp -d`?

kargakis: We will need to manually cleanup those directories though.

kargakis: If `pwd` works while we are in the workspace of the job then its' probably the best solution.

bbguimaraes:
English =)

PR #30, opened 2017-02-15
Add static localhost.yml

jhadvig:
@stevekuznetsov PTAL


stevekuznetsov:
Let's also get the `group_vars/OSEv3/general.yml` and `nested_group_mappings` in this PR -- the `host_vars` you have look right.

jhadvig:
@stevekuznetsov updated.. thanks

inventory/group_vars/OSEv3/general.yml
```diff
@@ -0,0 +1,11 @@
+---
+openshift_pkg_version: "-0.0.1"
```
stevekuznetsov: We just merged a PR into Origin that means we don't always build `origin-0.0.1` RPMs -- you'll want to test this out and see if you just want to get rid of this line, or figure out how to get the correct version in here.

jhadvig: Based on what we pick the version ?

stevekuznetsov: Previously it was _only_ possible to build Origin RPMs at version `0.0.1`. Now we build more correct versions. I think we should set `openshift_release=v1.5` and make sure we are installing the RPMs that we built and not some other RPMs on accident

jhadvig:
@stevekuznetsov I've updated the PR.. this iventory configuration work, though some additional openshift-ansible tinkering has to be done, but will submit a separate pr for that. 
PTAL

stevekuznetsov:
Do we not want 
```yml
origin_release: 1.5
```

Somewhere in the `OSEv3` configuration?

jhadvig:
Wasnt sure if its necessary, since the version was obtained from built packages

stevekuznetsov:
I think it just pulls latest, which happens to be what we just built. However, the value of this may determine some installer logic -- @sdodson could you clarify if we are losing anything by not specifying `origin_release` ?

sdodson:
> I think it just pulls latest, which happens to be what we just built. However, the value of this may determine some installer logic -- @sdodson could you clarify if we are losing anything by not specifying origin_release ?

origin_release doesn't exist, openshift_release does and it's primarily used to set the version of containerized installations because those aren't bound by yum repo content so we need to know your intended version when we pull the first image. After the origin or ose image is pulled we determine it's full version and then match all subsequent image versions to that image. Unless you care about containerized installs I don't think this matters right now.

sdodson:
PR LGTM

PR #31, opened 2017-02-15
added upgrade for cicd

mwoodson:


Jenkinsfile
```diff
@@ -0,0 +1,7 @@
+node( 'buildvm-devops' ) {
+	input 'The wrong key is currently being used for create. Fix this once keys have been setup correctly. Do not proceed unless you know what this message means.'
```
jupierce: This input line will be outdated now that the keys have been associated correctly. Perhaps something like:
```
input 'Are you certain you want to perform an upgrade on the cicd test cluster?'
```

Jenkinsfile
```diff
@@ -0,0 +1,7 @@
...
+	sshagent(['3fc8b6ea-5188-45f3-a401-c28efd4887e4']) { // build-int-key
```
jupierce: @themurph has now established named keys for us using human readable logical names. You can see those logical names here: https://atomic-e2e-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/aos-cd/credentials/

So this line should be:
```
sshagent(['cicd']) { 
```

jupierce:
lgtm

PR #37, opened 2017-02-17
Initial sprint-control Jenkins pipeline concept

jupierce:
ptal @mwoodson @themurph

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,334 @@
+sshkey_dev_preview_int = "test-key"
+sshkey_dev_preview_stg = "test-key"
+sshkey_dev_preview_prod = "test-key"
+
+last_action_day = 0
+year=getCurrentYear()
+
+// TODO: as these fields mature and stop changing, change into be build parameters so that a simple script can trigger a build
+def dates = input(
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,334 @@
...
+    phase = "dev-preview-int/create"
+    if ( isToday( phase, todayFields, dates.int_recreate ) ) {
+
+        def skip_build = false
+        while ( true ) {
+
+            try {
+                if ( !skip_build ) {
+                    stage( "${phase} openshift/ose build" ) {
+                        runOSEBuild()
```
stevekuznetsov: Do we necessarily need this to happen here? Async daily builds may make more sense if we can trigger them in the morning EST and deal with issues before this process runs at 6pm EST.

jupierce: @stevekuznetsov Yep. That is actually exactly what @tdawson has requested. Moving it to async at 6am with the job build/scheduled/ose-3.5 . The only drawback is that the sprint-control script will send out a "success" note if the build fails -- because the old level is already installed and upgrade will do nothing. Lots of room for improvement.

PR #40, opened 2017-02-20
Switch from merge to rebase

kargakis:


kargakis:
@jupierce updated this PR with the newer test for the rebase and will switch the target branch to build/ose-rebase-test once you update it.

scripts/merge-and-build.sh
```diff
@@ -84,17 +84,29 @@ git clone git@github.com:openshift/ose.git
...
-  git checkout -q enterprise-${OSE_VERSION}
+  git checkout -q fake-master
   git remote add upstream git@github.com:openshift/origin.git --no-tags
   git fetch --all
 
+  # TODO: Once we hard-reset master to contain Origin commits, the carry commits, and the latest tito tag commit
+  # all previous tags are going to be dropped. During the migration, I think we will need to manually tag the
+  # last tito commit or maybe not.
+  last_tag="$( git describe --abbrev=0 --tags )"
+  # Assumes the two latest commits are the tito diffs - won't work otherwise
```
stevekuznetsov: Since we're using `tito` to generate one of these commits and the comment two lines down to generate the other, we can do a sanity check here:

```sh
last_commit_subject="$( git log HEAD~2..HEAD~1 --pretty=%s )"
if [[ ! "${last_commit_subject}" ~= Automatic commit of package \[atomic\-openshift\] release \[[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+\.\]\. ]]; then
    echo "[FATAL] The last commit doesn't look like a commit from \`tito\`!"
    echo "[FATAL]   ${last_commit_subject}"
    exit 1
fi

squash_commit_subject="$( git log HEAD~1..HEAD --pretty=%s )"
if [[ "${squash_commit_subject}" != "[CARRY][BUILD] Specfile updates" ]]; then
    echo "[FATAL] The second to last commit doesn't look like a squashed specfile commit!"
    echo "[FATAL]   ${squash_commit_subject}"
    exit 1
fi
```

kargakis: Agreed, will add.

jupierce:
@kargakis build/ose-rebase-test has been reset.

kargakis:
@jupierce @stevekuznetsov ready

scripts/merge-and-build.sh
```diff
@@ -114,6 +139,8 @@ else
   fi
 fi # End check if we are master
 
+# TODO: Origin already vendors the web console. Should we stop updating it separately for OSE?
```
kargakis: @tdawson any opinions on this one?

tdawson: No we should not stop.
This updating of the web console is done in conjunction with the upstream vendoring the web console.  The OCP stuff has to be put in at some time, and this is when.

kargakis: @stevekuznetsov this means that we will need a bit more git-fu to ensure the latest commits are from tito.

scripts/merge-and-build.sh
```diff
@@ -84,17 +84,42 @@ git clone git@github.com:openshift/ose.git
...
   git remote add upstream git@github.com:openshift/origin.git --no-tags
   git fetch --all
 
+  # TODO: Once we hard-reset master to contain Origin commits, the carry commits, and the latest tito tag commit
+  # all previous tags are going to be dropped. During the migration, I think we will need to manually tag the
+  # last tito commit or maybe not.
+  last_tag="$( git describe --abbrev=0 --tags )"
+  regex="Automatic commit of package [atomic-openshift] release"
+  last_commit_subject="$( git log HEAD~1..HEAD --pretty=%s )"
+  if [[ ! "${last_commit_subject}" =~ "${regex}" ]]; then
```
stevekuznetsov: nit: technically `${regex}` is a literal string and not a regex here
bug: don't we need a `*` at the end?

```sh
if [[ ! "${last_commit_subject}" =~ "${prefix}"* ]]; then
```

scripts/merge-and-build.sh
```diff
@@ -141,6 +168,7 @@ echo "=========="
 tito tag --accept-auto-changelog
 export VERSION="v$(grep Version: origin.spec | awk '{print $2}')"
 echo ${VERSION}
+exit 0
 git push
 git push --tags
```
stevekuznetsov: We probably want this to be one command to make sure we can't crash between -- `git push --tags` should do both the objects and the refs

scripts/merge-and-build.sh
stevekuznetsov: Looks like it may make sense to just remove entirely the last commit for web console and re-create it with the below code, but do it in the right order.

kargakis: As is, the following will be the most common case of a rebase:
```
12345 last_tito_commit (HEAD)   <- needs a soft reset
67890 webconsole_bump   <- needs to be dropped
11223 tito_carry_commit   <- needs a soft reset
```
I am not sure there is a sane way to deal with the above case before a `git rebase` .. 

Idea: 
1. Create e a stage branch from which we can cherry-pick in the right order.
2. `git reset --hard upstream/master`
3. Cherry-pick in the right order

stevekuznetsov:
```sh
last_tag="$( git describe --abbrev=0 --tags )"

#         .-OAuth client branding
#        /  .-Web client branding
#       /  /  .-Specfile squashed
#      /  /  /  .-Tito tag commit
#     /  /  /  /
# m--o--w--s--t
#  \           \
#   \           `-ose/master/HEAD,v3.5.0.23
#    `-origin/master/HEAD

git rebase origin/master

#                  .-OAuth client branding
#                 /  .-Web client branding
#                /  /  .-Specfile squashed
#               /  /  /  .-Tito tag commit
#              /  /  /  /
# x--x--x--m--o--w--s--t
#           \           \
#            \           `-ose/master/HEAD
#             `-origin/master/HEAD

git reset HEAD~3

#                  .-OAuth client branding
#                 /
#                /
#               /
#              /
# x--x--x--m--o
#           \  \
#            \  `-ose/master/HEAD
#             `-origin/master/HEAD

rm -f pkg/assets/bindata.go pkg/assets/java/bindata.go
go-bindata -nocompress -nometadata -prefix "${CONSOLE_REPO_PATH}/dist"      -pkg "assets" -o "pkg/assets/bindata.go"      "${CONSOLE_REPO_PATH}/dist/..."
go-bindata -nocompress -nometadata -prefix "${CONSOLE_REPO_PATH}/dist.java" -pkg "java"   -o "pkg/assets/java/bindata.go" "${CONSOLE_REPO_PATH}/dist.java/..."

git add pkg/assets/bindata.go pkg/assets/java/bindata.go
git commit -m "[CARRY][BRANDING] Web console branding"

#                  .-OAuth client branding
#                 /  .-Web client branding
#                /  /
#               /  /
#              /  /
# x--x--x--m--o--w
#           \     \
#            \     `-ose/master/HEAD
#             `-origin/master/HEAD

git add .tito/ origin.spec

#                  .-OAuth client branding
#                 /  .-Web client branding
#                /  /  .-Specfile squashed
#               /  /  /
#              /  /  /
# x--x--x--m--o--w--s
#           \        \
#            \        `-ose/master/HEAD,v3.5.0.23
#             `-origin/master/HEAD

tito tag --accept-auto-changelog

#                  .-OAuth client branding
#                 /  .-Web client branding
#                /  /  .-Specfile squashed
#               /  /  /  .-Tito tag commit
#              /  /  /  /
# x--x--x--m--o--w--s--t
#           \           \
#            \           `-ose/master/HEAD,v3.5.0.24
#             `-origin/master/HEAD
```

tdawson:
When this goes live, how do we make changes to ose, that are ose specific?
Everything we used to do, now get's wiped out.

scripts/merge-and-build.sh
kargakis: With some extra validation in place, this can work. It is also a better idea to stick with rebasing (blacklisting) instead of resetting+cherry-picking (whitelisting) - should address Troy's concern.

kargakis:
> When this goes live, how do we make changes to ose, that are ose specific?
Everything we used to do, now get's wiped out.

You will still be able to check in OSE specific commits, but you will need to make sure the last three commits are 1. the web console bump 2. specfile 3. tito tag, for automated rebases to work. You should be able to reorder them manually with: 

git rebase -i upstream/master

This unfortunately means we will need to manually force-push when something is added in between rebases.

scripts/merge-and-build.sh
stevekuznetsov: Sure -- before we do the rebase to `origin/master` we can do a rebase onto ourselves -- so we can make the ordering of commits correct (i.e. the last four commits are as we want above)

stevekuznetsov:
We can layer user PRs on top of HEAD and do the reorder in the rebase script. That way we can continue to use the normal PR flow

stevekuznetsov:
Actually -- just looking back to the proposal, @kargakis @tdawson we disallow PRs into `ose/master` -- if you want it into `master`, you get it into Origin and we get it from there automatically. Carries otherwise live only in release branches, which are not automatically rebased and do not have this problem.

kargakis:
@stevekuznetsov comments addressed, ptal

kargakis:
The changelog seems like it's going to have the additional carry commits in every new tito tag...

scripts/merge-and-build.sh
kargakis: I just realized that your `Web client branding` commit is different from the one I referred to in my initial comment. So we can end up with something like:
```
# x--x--x--m--o--w--s--b--t
```
where `b` is a bump commit for the web console (additional diff between console HEAD and the vendored console in Origin).

kargakis: It doesn't seem to change anything. We can still soft-reset everything, recreate the web client branding commit, recreate the specfile commit, maybe add a new bump for the webconsole, tito tag.

kargakis: Do you think we will need to recreate `Web client branding` on every rebase?

scripts/merge-and-build.sh
```diff
@@ -84,17 +115,37 @@ git clone git@github.com:openshift/ose.git
...
   git fetch --all
 
+  # TODO: Once we hard-reset master to contain Origin commits, the carry commits, and the latest tito tag commit
+  # all previous tags are going to be dropped. During the migration, I think we will need to manually tag the
+  # last tito commit or maybe not.
+  last_tag="$( git describe --abbrev=0 --tags )"
+
+  sanity_check
+  maybe_webconsole_commit="$( git log HEAD~2..HEAD~1 --pretty=%s )"
+  if [[ ${maybe_webconsole_commit} =~ "[DROP] bump origin-web-console"* ]]; then
```
stevekuznetsov: We should always expect to have this commit

stevekuznetsov:
Hmm that's a good point. Maybe we can do some nice and functional `sed` to remove them :)

scripts/merge-and-build.sh
kargakis: If there is no change in the web console repo for more than the timeframe between two rebases, there is no commit for the web console.

stevekuznetsov: I see

tdawson:
Double checking the following scenario.
We need to make a change to ose master.  Let's say it's a tito change, because that's currently the only change I see needing to be done.
We need to add the 3.6 build section into .tito/releasers and/or we need to update .tito/lib/origin/common.py because we have another release specific tag.
Will @kargakis 's proposal with "git rebase -i upstream/master" and commit order, allow someone to do that?

kargakis:
> Double checking the following scenario.
> We need to make a change to ose master. Let's say it's a tito change, because that's currently the only change I see needing to be done.
> We need to add the 3.6 build section into .tito/releasers and/or we need to update .tito/lib/origin/common.py because we have another release specific tag.
> Will @kargakis 's proposal with "git rebase -i upstream/master" and commit order, allow someone to do that?

Yes, the flow would be something like this:
```sh
$ vi .tito/releasers
# update ./tito/releasers
$ git add .tito/releasers
$ git commit -m "message does't matter because this will be squashed"
$ git rebase -i upstream/master
# squash new commit into the "[CARRY] specfile changes" commit
$ git push origin master -f
```
If you add a net-new commit that is specific to OSE and does not belong to any of the rest `CARRY` commits, you should still use the flow above but simply rearrange the new commit below the latest three commits in master (assuming HEAD=`latest tito tag commit`, HEAD\~2=`[DROP] webconsole bump`, HEAD\~3=`[CARRY] specfile changes`) instead of squashing it. I will update the rebase proposal to mention this.

stevekuznetsov:
Oh, right -- a specific change to one of the carry commits. There was a section for annotating a commit with `[SQUASH]<commit-name-to-squash-into>` but I don't see that in the proposal today. Good catch @tdawson

tdawson:
Thanks for the detailed clarification.  It makes me feel better.

kargakis:
> Oh, right -- a specific change to one of the carry commits. There was a section for annotating a commit with [SQUASH]<commit-name-to-squash-into> but I don't see that in the proposal today. Good catch @tdawson

I don't want to get into the business of squashing out-of-band commits initially but we can definitely reiterate on this later if it happens frequently.

kargakis:
> I don't want to get into the business of squashing out-of-band commits initially but we can definitely reiterate on this later if it happens frequently.

@stevekuznetsov hm, rebasing changes the hashes for all the child commits, meaning any commit reordering will also change the hash of the latest tito commit. Will tito care about this?

kargakis:
> Hmm that's a good point. Maybe we can do some nice and functional sed to remove them :)

I have ended up with some functional sed commands but they are not nice... 
Initially I had:
```
cat origin.spec | sed 's/- \[CARRY\].*//' | sed 's/- \[DROP\].*//' > origin.spec
```
The problem is that the empty lines are not cleaned up. Thanks to sed one-liners I found out in the wild, I have ended up with the following command:
```
cat origin.spec | sed -e :a -e '$!N;s/\n- \[CARRY\].*/ /;ta' -e 'P;D' | sed -e :a -e '$!N;s/\n- \[DROP\].*/ /;ta' -e 'P;D' > origin.spec
```
But another problem is that tito breaks down commits into multiple lines if one exceeds a specific size.
So even with sed removing the lines we don't care about, we may still end up in a weird state. For example, starting from:
```
* Mon Feb 27 2017 Michail Kargakis <mkargaki@redhat.com>
- Automatic commit of package [atomic-openshift] release [3.5.0.32-1].
  (tdawson@redhat.com)
- [CARRY][BUILD] Specfile updates (skuznets@redhat.com)
- [DROP][UPSTREAM] Migration script for etcd (skuznets@redhat.com)
- [CARRY][BRANDING] OAuth server branding (skuznets@redhat.com)
- [DROP] bump origin-web-console (mkargaki@redhat.com)
- net-new (mkargaki@redhat.com)
- [CARRY][BRANDING] Web client branding (skuznets@redhat.com)
- [DROP][FIXME] Change image build and drop this cluster up flag
  (skuznets@redhat.com)
- [DROP][INVESTIGATE] Weird shell diffs (skuznets@redhat.com)
- [DROP][BRANDING] Manual page updates for base image name
  (skuznets@redhat.com)
```
and using the second `sed` command, we end up with:
```
* Mon Feb 27 2017 Michail Kargakis <mkargaki@redhat.com>
- Automatic commit of package [atomic-openshift] release [3.5.0.32-1].
  (tdawson@redhat.com)
- net-new (mkargaki@redhat.com)
  (skuznets@redhat.com)
  (skuznets@redhat.com)
```
Additional sed or should I stop this madness? Would a flag in `tito tag` that drops commits from the changelog that match a provided pattern be an overkill?

stevekuznetsov:
> I don't want to get into the business of squashing out-of-band commits initially but we can definitely reiterate on this later if it happens frequently.

Aren't we screwed the first time someone makes a change to a carry commit otherwise?

> rebasing changes the hashes for all the child commits, meaning any commit reordering will also change the hash of the latest tito commit. Will tito care about this?

I don't think so

> Additional sed or should I stop this madness?

No more madness in my opinion

> Would a flag in `tito tag` that drops commits from the changelog that match a provided pattern be an overkill?

`tito tag` has `--changelog` to supply a custom changelog... maybe we want to just take upstream approach, make our own changelog and supply it?

stevekuznetsov:
We could update `tito` but that becomes kind of a PITA... will take like a week or more to get released.

kargakis:
> tito tag has --changelog to supply a custom changelog... maybe we want to just take upstream approach, make our own changelog and supply it?

Yes, I think this is what we really want. I will have a look tomorrow.

kargakis:
Updated to set a custom changelog in tito that does not take into account carry commits. Hopefully this is the last problem for the rebase. ptal

scripts/merge-and-build.sh
```diff
@@ -190,7 +192,8 @@ echo
 echo "=========="
 echo "Tito Tagging"
 echo "=========="
-tito tag --accept-auto-changelog
+CHANGELOG=$(git log $PREVIOUS_ORIGIN_HEAD..$CURRENT_ORIGIN_HEAD --pretty="%s (%ae)" --no-merges)
```
stevekuznetsov: That's cool!

stevekuznetsov:
@kargakis and/or @tdawson can you use some `git log` fu to determine the frequency with which changes to carry commits happen in `master`? If it's frequent I do not want to push this in if we are going to need a manual resolution very soon for a PR to master.

kargakis:
> @kargakis and/or @tdawson can you use some git log fu to determine the frequency with which changes to carry commits happen in master? If it's frequent I do not want to push this in if we are going to need a manual resolution very soon for a PR to master.

Aren't those conflicts already happening with the current state (thus needing manual resolution)?

stevekuznetsov:
I guess that is a good point

kargakis:
@stevekuznetsov @jupierce fake-master was updated by @tdawson, can we get this merged so we can have a test run in Jenkins?

kargakis:
First run: https://atomic-e2e-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/aos-cd-builds/job/build%252Fose-rebase-test/24/console

kargakis:
We we will also want to test with the webconsole bump commit in there.

PR #47, opened 2017-02-24
Add openshift-ansible builds

smunilla:
Add in builds of openshift ansible after to building any other OCP packages


smunilla:
@sdodson PTAL

scripts/merge-and-build.sh
```diff
@@ -51,6 +51,47 @@ if [ "${OSE_VERSION}" == "3.2" ] ; then
...
+git clone git@github.com:openshift/openshift-ansible.git
+cd openshift-ansible/
+git checkout release-1.${MINOR}
+
+# Check to see if there have been any changes since the last tag
+HEAD_COMMIT="$(git log -n1 --oneline | awk '{print $1}')"
+echo $HEAD_COMMIT
+LAST_TAG=$(git describe --abbrev=0 --tags)
+git checkout -q ${LAST_TAG}
+LAST_COMMIT="$(git log -n1 --oneline | awk '{print $1}')"
```
stevekuznetsov: Use porcelain commands 
```sh
git log -n 1 --pretty=%h
```

scripts/merge-and-build.sh
```diff
@@ -51,6 +51,47 @@ if [ "${OSE_VERSION}" == "3.2" ] ; then
...
+echo $LAST_COMMIT
+
+if [ "${HEAD_COMMIT}" == "${LAST_COMMIT}" ]; then
```
stevekuznetsov: You can also just do this:

```sh
if ! git describe --abbrev=0 --tags --exact-match HEAD 2>&1 >/dev/null; then
```

That asks git for the last tag that exactly matches `HEAD` and will error if there is none

PR #62, opened 2017-02-27
[wip] Catch-up with merge-and-build.sh

bbguimaraes:
Figured I could open this while unfinished, since it's mostly last week's work
that hasn't changed much.  Catching up with the changes for 1.5/3.5 in
`merge-and-build.sh`; some corrections to the way the web console was vendored
(the way we were using `vendor_console.sh` wasn't 100% compatible with how it
should be done); other minor fixes.

More to come.


bbguimaraes:
@jupierce: this is finally taking shape.  It needs more testing (which I'll do
tomorrow) before being merged, but PTAL.

jupierce:
@bbguimaraes This is looking really good!

bbguimaraes:
Final updates (hopefully), depend on #110.

PR #65, opened 2017-03-02
Add openshift-ansible builds

smunilla:
Add in builds of openshift ansible after to building any other OCP packages.


smunilla:
@stevekuznetsov Better now?

PR #67, opened 2017-03-02
Initial commit of openshift-online build automation

smunilla:
New script in a new branch to all for daily builds of
openshift-scripts rpm.


smunilla:
@stevekuznetsov I think this does all the needful?

PR #70, opened 2017-03-06
Updates to the rebase script

kargakis:
Soft reset does not drop changes from the staging area. Instead use
simple reset and drop the changes manually.

@stevekuznetsov @bbguimaraes

### DIFF
--- a/scripts/merge-and-build.sh
+++ b/scripts/merge-and-build.sh
@@ -78,29 +78,28 @@ function sanity_check() {
   echo "Checking if the last commit is the last tito tag commit..."
   last_commit_subject="$( git log HEAD~1..HEAD --pretty=%s )"
   if [[ ! ${last_commit_subject} =~ "Automatic commit of package [atomic-openshift] release"* ]]; then
+    set +o xtrace
     echo "[FATAL] The last commit doesn't look like a commit from \`tito\`!"
     echo "[FATAL]   ${last_commit_subject}"
     exit 1
   fi
 
-  echo "Checking if the second to last commit is the specfile squashed commit..."
-  squash_commit_subject="$( git log HEAD~2..HEAD~1 --pretty=%s )"
-  if [[ ${squash_commit_subject} != "[CARRY][BUILD] Specfile updates" ]]; then
-    # The second to last commit may be the web console bump
-    echo "Checking if the second to last commit is a webconsole bump commit..."
-    squash_commit_again="$( git log HEAD~3..HEAD~2 --pretty=%s )"
-    if [[ ${squash_commit_again} != "[CARRY][BUILD] Specfile updates" ||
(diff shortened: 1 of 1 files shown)


scripts/merge-and-build.sh
```diff
@@ -175,10 +175,9 @@ if [ "${OSE_VERSION}" != "3.2" ] ; then
   VC_COMMIT="$(GIT_REF=enterprise-${OSE_VERSION} hack/vendor-console.sh 2>/dev/null | grep "Vendoring origin-web-console" | awk '{print $4}')"
   git add pkg/assets/bindata.go
   git add pkg/assets/java/bindata.go
-  set +e # Temporarily turn off errexit. THis is failing sometimes. Check with Troy if it is expected.
-  # This fails only when there is nothing new to commit, which is normal.
-  git commit -m "[DROP] bump origin-web-console ${VC_COMMIT}"
-  set -e
+  if [ "$(git status --porcelain)" ]; then
```
stevekuznetsov: I would prefer this to be more explicit:

```sh
if [[ -n "$( git status --porcelain )" ]]; then
```

Also, the basic "does `git status` have output" check doesn't make a distinction between staged and unstaged changes, which does matter for `git commit`. If this is in the rebase script, though, again, won't we always have changes since we're blowing away old changes and OCP never tracks Origin Web Console master?

kargakis: > If this is in the rebase script, though, again, won't we always have changes since we're blowing away old changes and OCP never tracks Origin Web Console master?

I am stashing the previous web console diff so up to the point we revendor the console, the directory should be clean. Also, [OCP master is tracking Origin Web Console master](https://github.com/openshift/aos-cd-jobs/blob/fa928f42996b1566f2f06e39cba1f277509ab402/scripts/merge-and-build.sh#L108).

stevekuznetsov: OCP Master tracking non-enterprise web console seems wrong ... @tdawson could you please weigh in here

kargakis: @stevekuznetsov ptal

scripts/merge-and-build.sh
```diff
@@ -120,21 +119,17 @@ if [ "${OSE_VERSION}" == "${OSE_MASTER}" ] ; then
...
-  else
-    git reset --soft HEAD~2
-  fi
-
+  # Reset the last three commits and pick up only the tito diff.
+  git reset HEAD~3
   git add .tito/ origin.spec
   git commit -m "[CARRY][BUILD] Specfile updates"
+  # Drop the previous web console diff - will be regenerated below.
+  git stash
```
stevekuznetsov: Not sure this is the best approach, instead

```sh
git checkout -- pkg/assets/bindata.go pkg/assets/java/bindata.go
```

kargakis:
More scoped, I like it. Updated

PR #73, opened 2017-03-07
Don't rebuild integration and extended test binaries; use the ones al…

jim-minter:
This PR should be applied after https://github.com/openshift/origin/pull/13275 has merged.

This PR should ensure that our tests use integration and extended test binaries that have already been compiled within the release container as part of the `make release` process.


jim-minter:
@smarterclayton 
@stevekuznetsov

PR #81, opened 2017-03-08
run make test under hack/env

jim-minter:
@stevekuznetsov 
@smarterclayton

### DIFF
--- a/config/test_cases/test_branch_origin_integration.yml
+++ b/config/test_cases/test_branch_origin_integration.yml
@@ -6,4 +6,5 @@ extensions:
       title: "run integration tests"
       repository: "origin"
       script: |-
-        JUNIT_REPORT='true' make test -o check -k
+        OS_BUILD_ENV_DOCKER_ARGS="-v /tmp:/openshifttmp " OS_BUILD_ENV_PRESERVE=_output/local hack/env JUNIT_REPORT='true' make test-tools test-integration -o check -k
+        JUNIT_REPORT='true' make test-end-to-end -o check -k
--- a/config/test_cases/ami_build_origin_int_rhel_base.yml
+++ b/config/test_cases/ami_build_origin_int_rhel_base.yml
@@ -66,7 +66,8 @@ actions:
     title: "run integration tests"
     repository: "origin"
     script: |-
-      JUNIT_REPORT='true' make test -o check -o build -k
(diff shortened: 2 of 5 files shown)


config/test_cases/test_branch_origin_integration.yml
```diff
@@ -6,4 +6,4 @@ extensions:
       title: "run integration tests"
       repository: "origin"
       script: |-
-        JUNIT_REPORT='true' make test -o check -k
\ No newline at end of file
+        OS_BUILD_ENV_PRESERVE=_output/local hack/env JUNIT_REPORT='true' make test -o check -k
```
stevekuznetsov: Does this work? I would not expect `test/end-to-end.sh` to work inside `hack/env`. No `docker` client there.

Also you'll want to mount `/tmp:/openshifttmp` like in the `test_branch_origin_check` so we can grab artifacts back from the container.

jim-minter:
Updated

stevekuznetsov:
Launched https://ci.openshift.redhat.com/jenkins/view/Origin%20Branch%20Tests/job/test_branch_origin_integration/136/ to test this

config/test_cases/test_branch_origin_integration.yml
```diff
@@ -6,4 +6,5 @@ extensions:
       title: "run integration tests"
       repository: "origin"
       script: |-
-        JUNIT_REPORT='true' make test -o check -k
\ No newline at end of file
+        OS_BUILD_ENV_DOCKER_ARGS="-v /tmp:/openshifttmp " OS_BUILD_ENV_PRESERVE=_output/local hack/env JUNIT_REPORT='true' make test-tools test-integration -o check -k
+        JUNIT_REPORT='true' make test-tools test-end-to-end -o check -k
```
stevekuznetsov: We don't need the `test-tools` target down here, unless you expect there to be a difference between inside and outside of `hack/env`?

stevekuznetsov:
@jim-minter can you make the small edit above and also change the `config/test_cases/ami_build_origin_int_rhel_base.yml` today? Tests look fine.

jim-minter:
@stevekuznetsov done

stevekuznetsov:
Thanks!

stevekuznetsov:
@jim-minter for whatever reason this is causing https://github.com/openshift/origin/issues/12661 with 100% certainty -- I'm going to need to revert this. Weird that it did not trigger that failure when I tested it earlier.

PR #111, opened 2017-03-14
Move publish-oc-binary.sh to build-scripts

bbguimaraes:
Depends on #110.

### DIFF
--- a/scripts/merge-and-build.sh
+++ b/scripts/merge-and-build.sh
@@ -249,32 +249,8 @@ echo "=========="
 echo "Publish the oc binary"
 echo "=========="
 ssh ocp-build@rcm-guest.app.eng.bos.redhat.com \
-    sh -s "$OSE_VERSION" "${VERSION#v}" <<-'EOF'
-	set -eux
-	OSE_VERSION=$1
-	VERSION=$2
-	rpm=/mnt/rcm-guest/puddles/RHAOS/AtomicOpenShift/$OSE_VERSION/latest/x86_64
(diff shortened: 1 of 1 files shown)


scripts/merge-and-build.sh
```diff
@@ -243,32 +243,7 @@ echo "=========="
...
-	rsync \
-		-av --delete-after --progress --no-g --omit-dir-times --chmod=Dug=rwX \
-		-e "ssh -l jenkins_aos_cd_bot -o StrictHostKeyChecking=no" \
-		"$outdir" \
-		use-mirror-upload.ops.rhcloud.com:/srv/pub/openshift-v3/clients/
-	ssh -l jenkins_aos_cd_bot -o StrictHostKeychecking=no \
-		use-mirror-upload.ops.rhcloud.com \
-		/usr/local/bin/push.pub.sh openshift-v3 -v
-EOF
+    sh -s "$OSE_VERSION" "${VERSION#v}" < "$WORKSPACE/publish-oc-binary.sh"
```
jupierce: @bbguimaraes It's not clear to me how you expect publish-oc-binary.sh to be in your $WORKSPACE . 

ose_images.sh is only accessible because the Jenkinsfile sets up PATH to include the extracted branch.

bbguimaraes: @jupierce: you are absolutely right, I forgot to fix this.  We can't just add
this script to the path, as it is not executed directly, but passed as input to
the ssh command.

Corrected (and rebased).

jupierce:
ltgm. Thanks @bbguimaraes!

PR #112, opened 2017-03-14
Use rebase.py to do interactive rebases

kargakis:
@stevekuznetsov we can give this a spin today


scripts/merge-and-build.sh
```diff
@@ -124,25 +124,19 @@ if [ "${OSE_VERSION}" == "${OSE_MASTER}" ] ; then
...
-  set -e
 
   echo
   echo "=========="
   echo "Merge origin into ose stuff"
   echo "=========="
 ## Switch back once master is 3.6
 #  git merge -m "Merge remote-tracking branch upstream/master" upstream/master
-  git rebase upstream/release-1.5
+  GIT_SEQUENCE_EDITOR=rebase.py git rebase -i upstream/release-1.5
```
stevekuznetsov: Don't we want two rounds of `git rebase`?

```sh
# rebase once to get current set of carries in order
GIT_SEQUENCE_EDITOR=rebase.py git rebase  -i $( git merge-base master upstream/master)
# should be clean rebase to bring in new changes
git rebase upstream/master
```

kargakis: Doesn't seem like it's needed. I am testing it locally and it works fine as is. If you feel it's cleaner to separate the two steps I can do it.

stevekuznetsov: Oh, yeah that is right, the rebase onto `release-1.5` or whatever should do both. Disregard

kargakis:
@stevekuznetsov updated rebase.py to squash based only on the second label and not require the subject to be the same and also added some tests.

kargakis:
@jupierce @smunilla @bbguimaraes for moar eyes

scripts/merge-and-build.sh
```diff
@@ -74,35 +74,6 @@ if [ "${OSE_VERSION}" != "3.2" ] ; then
   ssh-add ${HOME}/.ssh/id_rsa
 fi # End check if we are version 3.2
 
-function sanity_check() {
```
stevekuznetsov: I could also see an argument for breaking this out...

```
sanity_check.py "$( git log "$( git merge-base origin/master )"..HEAD --pretty='%h %s' )"
```

kargakis: This will require to break up your python script and I don't think it's required now. If you don't have any other comments, I would like to move forward with this (it's still on the test branch) and test the flow.

stevekuznetsov: I don't know if we have the best way to report errors from within the `GIT_SEQUENCE_EDITOR` -- a verification step that sits before it would let us signal errors better IMO. That flow would also allow us to not need to do error checks in the rebase script -- one thing I did not check in there was that no `[SQUASH]` commit existed earlier in time than the commit it was being squashed into.

kargakis: > I don't know if we have the best way to report errors from within the GIT_SEQUENCE_EDITOR -- a verification step that sits before it would let us signal errors better IMO. That flow would also allow us to not need to do error checks in the rebase script 

Probably nothing will change by breaking this out of `GIT_SEQUENCE_EDITOR` because it all depends on what the script returns to you as an error. So this is more of an argument for making sure we have informative errors. More on this later

> one thing I did not check in there was that no [SQUASH] commit existed earlier in time than the commit it was being squashed into.

I will add a test case for this and make sure we handle it correctly.

kargakis:
Updated rebase.py to not care about the order of SQUASHes and added additional test cases.

scripts/merge-and-build.sh
kargakis: Here is the error I am getting back from the editor now:
```
$ GIT_SEQUENCE_EDITOR=../aos-cd-jobs/scripts/rebase.py git rebase -i public/release-1.5 
Commit `Change image build and drop this cluster up flag` (c122009) invalid!
Could not execute editor
```
I will update the error messages but at this stage I don't see breaking out the validation as a big win.

kargakis:
Broke validation out into a separate script because I figured we will need to use it in the OSE test job

scripts/rebase.py
```diff
@@ -174,27 +174,40 @@ def squash_named_commits(commits):
...
     :return: updated commits
     """
+
+    # Consolidate all but SQUASH commits.
     carry_commits = []
+    reordered_commits = []
     for commit in commits:
         if commit.subject.startswith(_CARRY_PREFIX):
             carry_commits.append(commit)
+        if commit.subject.startswith(_SQUASH_PREFIX) == False:
```
stevekuznetsov:
```py
if not commit.subject.startswith(_SQUASH_PREFIX):
```

scripts/rebase.py
```diff
@@ -1,116 +1,10 @@
...
-        print('No carry commit for `.spec` file found!')
-        exit(1)
-
-    if tito_index < spec_index:
-        print('Found the `tito tag` commit before the `.spec` file  commit!')
-        exit(1)
-
+import os
+import re
+import git
```
stevekuznetsov: Imports of code in a project should ideally be relative so `$PYTHONPATH` can't screw it up

scripts/rebase.py
```diff
@@ -172,32 +66,63 @@ def squash_named_commits(commits):
...
     """
+
+    # Consolidate all but SQUASH commits.
     carry_commits = []
+    reordered_commits = []
     for commit in commits:
-        if commit.subject.startswith(_CARRY_PREFIX):
+        if commit.subject.startswith(git._CARRY_PREFIX):
             carry_commits.append(commit)
+        if commit.subject.startswith(git._SQUASH_PREFIX) == False:
```
stevekuznetsov:
```py
if not commit.subject.startswith(git._SQUASH_PREFIX):
```

scripts/rebase.py
```diff
@@ -172,32 +66,63 @@ def squash_named_commits(commits):
...
+
+        # Walk through all SQUASH commits and add them in the right order in reordered_commits.
+        carry_commit_index = reordered_commits.index(carry_commit)
+        for index, squash_commit in enumerate(squash_commits):
+            squash_commit.action = git.Action.fixup
+            reordered_commits.insert(carry_commit_index + index + 1, squash_commit)
+
+    return reordered_commits
+
+def same_type_commits(commit1, commit2):
```
stevekuznetsov:
```py
def commit_types_match(*commits):
```

maybe?

scripts/rebase.py
```diff
@@ -172,32 +66,63 @@ def squash_named_commits(commits):
...
+    Return whether they are of the same type thus can
+    be squashed together.
+
+    :param commit1: first commit to check for its type
+    :param commit2: second commit to check for its type
+    :return: whether the commits are of the same type
+    """
+    prefix = os.path.commonprefix([commit1, commit2])
+    match = re.match(r"\[.*\]", prefix)
+    if match is None:
```
stevekuznetsov:
```py
return match is not None
```

scripts/sanity_check.py
```diff
@@ -0,0 +1,85 @@
...
+            carry_types.append(match.group(0))
+
+    # Validate that CARRY types are unique across all CARRIES.
+    if len(carry_types) != len(set(carry_types)):
+        print('Found more than one carry commit with the same type!')
+        exit(1)
+
+    # Validate that SQUASHES have a type.
+    squash_types = []
+    for commit in commits:
```
stevekuznetsov: Could DRY this out... seems like very similar logic to carry commit above

testdata/1_input
```diff
@@ -0,0 +1,8 @@
+pick 4365d5d [DROP][INVESTIGATE] Weird shell diffs
```
stevekuznetsov: Would be nice to associate intent with these tests.

scripts/rebase.py
kargakis: I am not sure how to make relative imports work and there is a PEP that suggest against them
https://www.python.org/dev/peps/pep-0008/#imports

testdata/1_input
kargakis: I can add comments on top of these files - basically 1 is the example you had in Trello, 2 and 3 are common scenarios and 4+5 are testing SQUASHes before their CARRY counterparts (you asked for that:) ).

kargakis: Hm, comments in the files will not work. Not as long as I have the current setup.

kargakis: I can add a separate doc - will probably need one anyway.

scripts/rebase.py
```diff
@@ -101,21 +101,20 @@ def squash_named_commits(commits):
 
     return reordered_commits
 
-def same_type_commits(commit1, commit2):
+def commit_types_match(commits):
```
stevekuznetsov: You haven't made this a variadic argument `

kargakis: fixed

PR #123, opened 2017-03-16
Allow GCE jobs to run different suites so we can test overlay

smarterclayton:
Should really be extended_gce.yml, but will change in a subsequent spot

### DIFF
--- a/config/test_cases/test_branch_origin_extended_conformance_gce.yml
+++ b/config/test_cases/test_branch_origin_extended_conformance_gce.yml
@@ -4,6 +4,12 @@ extensions:
   sync_repos:
     - name: "release"
   parameters:
+    - name: "SUITE"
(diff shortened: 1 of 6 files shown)


smarterclayton:
@stevekuznetsov

smarterclayton:
Rebased

smarterclayton:
Meeeerge

config/test_cases/test_branch_origin_extended_conformance_gce.yml
```diff
@@ -4,6 +4,12 @@ extensions:
   sync_repos:
     - name: "release"
   parameters:
+    - name: "SUITE"
```
stevekuznetsov: Yeah, this should not be in the `conformance` job ... but the tools today do not let you change that simply so ... whatever

PR #133, opened 2017-03-21
Use correct repo for building openshift-images

smunilla:


smunilla:
@jupierce This should finally get us building oso images.

PR #134, opened 2017-03-21
[wip] multidirectory.sh

bbguimaraes:
@jupierce, @stevekuznetsov: these are just the script, a small test suite, and
a temporary script to simulate the migration.  I'll start working on
restructuring the jobs on the master branch.  In the meantime, PTAL.


stevekuznetsov:
I will review this with Origin Bash style guide as well as correctness.

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
+# Creates/removes/updates branches according the `jobs/` directory on the
+# master branch.  Local files and branches will be modified, so this script
+# should be executed on a clean, dedicated clone.
+set -eu
+
+set_difference() {
```
stevekuznetsov: We prefer:

```sh
# set_difference does ....
function set_difference() {
```

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
...
+    printf '%s\n%s\n%s\n' "$1" "$1" "$2" | sort | uniq -u
+}
+
+list_branches() {
+    git branch --list --remotes 'origin/*' \
```
stevekuznetsov: It's a little hard for me to decode this, but I think you're taking a list of refs like `origin/branchname` and getting just `branchname`?

Always prefer to write non-Bash like Bash:

```sh
# list_branches lists the branch names for the `upstream' remote
function list_branches() {
    for branch in $( git branch --list --remotes ); do
        if [[ "${branch}" == upstream/* ]]; then # how does your script handle entries like upstream/HEAD -> upstream/master
            echo "${branch:9}" # if string slicing isn't your thing, could use sed
        fi
    done
}
```

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
...
+        | awk '$1 ~ /^origin\/.*\// {print gensub("^origin/", "", 1, $1)}'
+}
+
+list_jobs() {
+    [ -e jobs/ ] || return 0
```
stevekuznetsov: Here you could write:

```sh
# list_jobs lists directories that contain Jenkinsfiles
function list_jobs() {
    for jenkinsfile in $( find jobs/ -name Jenkinsfile ); do
        dirname "${jenkinsfile}"
    done
}
```

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
...
+    (cd jobs/ && find * -name Jenkinsfile | sed 's,/Jenkinsfile$,,')
+}
+
+filter_unchanged() {
```
stevekuznetsov: This doesn't seem super useful -- use `git` to your advantage here, always do the copy/clobber and check if there is anything to commit.

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
...
+        | awk -v "m=$msg" '$0 !~ m {print $1}'
+}
+
+create_branches() {
+    local rev x
+    rev=$1
+    shift
+    for x; do
+        git checkout --quiet --orphan "$x" "$rev"
+        find "jobs/$x/" -mindepth 1 -maxdepth 1 -exec mv -t . {} +
```
stevekuznetsov: Can't this just be

```sh
mv "jobs/${branch_name}"/* .
```

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
...
+    shift
+    for x; do echo "$x $(git show --format=%s --no-patch "origin/$x")"; done \
+        | awk -v "m=$msg" '$0 !~ m {print $1}'
+}
+
+create_branches() {
+    local rev x
+    rev=$1
+    shift
+    for x; do
```
stevekuznetsov: This could be a little more descriptive in what it does:

```sh
for branch_name in ${branch_names}; do
```

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
```
stevekuznetsov: Should be

```sh
#!/bin/bash
```

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
...
+        | awk '$1 ~ /^origin\/.*\// {print gensub("^origin/", "", 1, $1)}'
+}
+
+list_jobs() {
+    [ -e jobs/ ] || return 0
+    (cd jobs/ && find * -name Jenkinsfile | sed 's,/Jenkinsfile$,,')
+}
+
+filter_unchanged() {
+    local msg x
```
stevekuznetsov: The `local` keyword is undefined in POSIX-compliant `#!/bin/sh`

build-scripts/multidirectory.sh
```diff
@@ -0,0 +1,57 @@
...
+
+master_rev=$(git rev-parse --short master)
+branches=$(list_branches | sort)
+jobs=$(list_jobs | sort)
+create=$(set_difference "$branches" "$jobs")
+delete=$(set_difference "$jobs" "$branches")
+update=$(filter_unchanged "$master_rev" $(set_difference "$create" "$jobs"))
+create_branches "$master_rev" $create $update
+git checkout --quiet master
+if [ "$create" -o "$delete" ]; then
```
stevekuznetsov: Always prefer not to use POSIX shell, and to use `[[` over `[`. Here's a good example of why:

The `-a` and `-o` operators in `[` are not well-defined, and their precedence is not strictly defined anywhere. They are obsolescent as defined in [IEEE](pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html). Furthermore here you're using an implicit check in that `[` will consider non-empty string "true". Much better to be explicit:

```sh
if [[ -n "${create}" || -n "${delete}" ]]; then
```

Although for this specific case, you should be OK doing a push always, as if there are no changes then there will be no problem pushing nothing.

build-scripts/multidirectory_test.sh
```diff
@@ -0,0 +1,229 @@
...
+    (cd work/ && "$MULTIDIRECTORY_SH")
+    git clone --quiet origin.git/ check/
+    cd check/
+    branches=$(git ls-remote --heads origin \
+        | awk '{print gensub("^refs/heads/", "", 1, $2)}')
+    echo "$branches" | diff -u ../expected_branches.txt -
+    branches=$(echo "$branches" | { grep -x / || true; })
+    for x in $branches; do
+        ls "jobs/$x/Jenkinsfile" > /dev/null
+        (find * ! \( -wholename jobs -a -prune \) && cd "jobs/$x" && find *) \
```
stevekuznetsov: Probably not an issue but interesting to note that `find *` will interpret file names with dashes in them as command-line arguments to `find`. Prefer to use `find ./*`

build-scripts/multidirectory_test.sh
```diff
@@ -0,0 +1,229 @@
...
+    git init --quiet init/
+    cd init/
+    git commit --quiet --allow-empty --message first
+    git push --quiet ../origin.git master
+    cd -> /dev/null
+    git clone --quiet origin.git/ setup/
+}
+
+branches() {
+    (
```
stevekuznetsov: Why do you need subshells for actions in here?

build-scripts/multidirectory_test.sh
```diff
@@ -0,0 +1,229 @@
...
+export GIT_AUTHOR_NAME GIT_AUTHOR_EMAIL GIT_COMMITTER_NAME GIT_COMMITTER_EMAIL
+test_dir=$PWD/tests
+rm -rf "$test_dir"
+[ "$#" -gt 0 ] || set -- $(seq -f 'test%.0f' 0 12)
+for x; do
+    echo "$x"
+    mkdir -p "$test_dir/$x"
+    set +e
+    (set -e; cd "$test_dir/$x"; "$x")
+    [ "$?" -eq 0 ] || failed=1
```
stevekuznetsov: Prefer to use the following to achieve this effect:

```sh
mkdir -p "${test_dir}/${x}"
if ! cd "${test_dir}/${x}"; then
    # handle the error case
fi
```

The `set -o errexit` will not apply to commands executing inside of the `if` statement check, as the exit code of those commands is used to determine which branch of the `if` is executed. Although here, why are you checking to see if the `cd` failed? Would you expect `mkdir` to fail before that?

build-scripts/multidirectory.sh
bbguimaraes: There's a lot going on here, mainly because there isn't a good way to list
branches that's easy for a script to parse (at least that I could find):

- we list all remote branches that match the pattern `origin/*`
- the awk pattern (`$1 ~ /…/`) filters out branches that don't have a `/` in
  their names
- `gensub(…)` removes the `origin/` prefix
- we take out `git branch`'s superfluous information (such as those two stupid
  spaces at the beginning of each line, and the `-> origin/master` you
  mentioned) by doing the substitution on `$1` instead of `$0`.

I just realized the restriction "only branches that have a `/` in their names"
can be embedded in the pattern for `git branch`. That'll make things simpler:

```
git branch --list --remotes 'origin/*/*' \
    | awk '{print gensub("^origin/", "", 1, $1)}'
```

In a perfect world, that'd be:

```
git give-me-the-list-of-branches-from-this-remote-without-extra-junk \
    origin '*/*'
```

build-scripts/multidirectory.sh
bbguimaraes: `cd jobs/ && find …` is mainly to get the output in the correct form (i.e.
without having to remove the `jobs/` prefix).  You're right about `dirname`
though, I completely forgot about it, but I knew there *had* to be a better way
to do this.

build-scripts/multidirectory.sh
bbguimaraes: You learn something everyday…

build-scripts/multidirectory.sh
bbguimaraes: That won't move files that start with a dot.  I didn't know if there are/will be
any, so I decided to play safe.

build-scripts/multidirectory.sh
stevekuznetsov: The basic review I have for you does not change. Get general output from `git branch` and use a `for` loop to cleanly and iteratively apply filters. Don't make someone decode complicated `awk` expressions (that's two programming languages you need to know now, not just one!).

build-scripts/multidirectory.sh
bbguimaraes: The problem with a simple `git push` here is we have to force-push branches we
are updating, but not those we are creating (and I don't even know if
"force-pushing a deletion" has a different meaning).

Playing safe again, we could just force-push everything.

build-scripts/multidirectory.sh
stevekuznetsov: Let's not worry about optimizing this for that case until we hit it (this is usually set with `shopt` settings for globbing anyway and may be different on different shells).

build-scripts/multidirectory_test.sh
bbguimaraes: Shorter and a little "safer" than `cd $x; … cd -> /dev/null`.  Same as bellow.

build-scripts/multidirectory_test.sh
bbguimaraes: This is again to get the right output.  Since these are tests where all files
are under our control, I didn't think it was worth complicating this (already
non-trivial) line.

build-scripts/multidirectory_test.sh
bbguimaraes: These lines are tricky.  The original was:

```
(cd "$test_dir/" && "$x") || failed=1
```

but that taught me that when `set -e` is in effect, not only an error is not
generated if any command but the last in a chain of `&&` and/or `||` fails
(which I already knew) but it also cancels the effect of `set -e` for all those
commands.

So here, because `"$x"` (the test we are running) is on the left side of `||`,
`set -e` is disable when executing it.  That led to this more complicated code,
where we need to `set +e` on the parent shell, start a subshell, `set -e`,
execute the test, test the result on the parent, and then `set -e` again.

Not the prettiest thing in the world (as any test suite written in shell is
probably going to be), but not *that* bad.

bbguimaraes:
@stevekuznetsov: thanks a lot for the thorough review.  I've explained some of
the hairy bits, I'll address the other comments and update the PR tomorrow.

build-scripts/multidirectory_test.sh
stevekuznetsov: >  (as any test suite written in shell is probably going to be)

Be careful....

stevekuznetsov:
I can see some use for the test suite, but looking at it again I would have to suggest we not commit it. Having written a Bash-based testing framework for Origin, even one without bugs and with good usefulness, I am really, really, really weary of doing that again. In all honesty I may not have chosen Bash for this script, especially as it would allow us to run sane tests on it, but I don't think it's reasonable to ask for this to be re-written. Instead of a complicated test suite, I think it is  more reasonable to make a streamlined script and locally test it. I'm not sure the script needs a lot of complexity.

stevekuznetsov:
If we split apart some of the actions it becomes a little easier to "mentally validate" the script without tests:

```sh
#!/bin/bash

# prune_branches.sh removes branches from the remote if they
# do not have corresponding Jenkinsfiles in the master branch

remote="${1:-"origin"}"

git checkout master
for branch in $( git branch --list --remotes --no-merged HEAD "${remote}"/*); do
	# a branch a/b should only exist if a/b/Jenkinsfile exists in master
	jenkinsfile_dir="${branch#${remote}}"
	if [[ ! -e "${jenkinsfile_dir}/Jenkinsfile" ]]; then
		git push "${remote}" --delete "${branch}"
	fi
done
```

```sh
#!/bin/bash

# create_or_update_branches.sh creates or updates branches on
# the remote if a Jenksnsfile exists in the master branch at
# a sub-path from jobs/

remote="${1:-"origin"}"

git checkout master
current_head="$( git log -n 1 --pretty=%h )"

# we can clobber the state of any local branches as
# the master branch will have the authoritative state
for branch in $( git branch --list --no-merged master ); do
	git branch --delete --force "${branch}"
done

for jenkinsfile_path in $( find jobs/ -name Jenkinsfile ); do
	# a branch a/b should exist if a/b/Jenkinsfile exists in master
	jenkinsfile_dir=$( dirname "${jenkinsfile_path}" )
	branch_name="${jenkinsfile_dir#jobs/}"

	git checkout --quiet --orphan "${branch_name}" master
	mv "${jenkinsfile_dir}"/* .
	rm -rf jobs/
	git commit --all --message "Auto-generated job branch for ${branch_name} from ${current_head}"
	git push "${remote}" --force "${branch_name}"
done
```

bbguimaraes:
Ok then, I leave it up to you to decide what to do with this PR.  If you want
to close it and add that instead, that's fine for me.

stevekuznetsov:
Up to you - if you want to stay with Bash, let's pare down the LOC so we don't need tests. If you want to keep the test suite you wrote (which was good!!) Let's write this in Python and write unit tests that way.

bbguimaraes:
@stevekuznetsov: updated, this fails only one of the old tests, the one that
guaranteed no changes were made to existing branches if they were up to date.
I'm not sure if that's relevant, so leaving it like that.

build-scripts/multidirectory.py
```diff
@@ -0,0 +1,44 @@
...
+import os
+import shutil
+import subprocess
+import sys
+
+def remove_branches(remote):
+    branches = subprocess.check_output([
+        'git', 'branch', '--list', '--remotes', '--no-merged', 'HEAD',
+        remote + '/*'])
+    for branch in [x[len(remote) + 1:] for x in branches.split()]:
```
stevekuznetsov: Use an explanatory variable here, not obvious what you're looping over.

```py
branch_refs = subprocess.check_output() #...
branch_names = [ ref.lstrip(remote + '/') for ref in branch_refs.split()]
for branch in branch_names:
    # ...
```

build-scripts/multidirectory.py
```diff
@@ -0,0 +1,44 @@
...
+        if not os.path.exists(os.path.join('jobs', branch, 'Jenkinsfile')):
+            subprocess.check_call([
+                'git', 'push', '--quiet', '--delete', remote, branch])
+
+def create_or_update_branches(remote):
+    if not os.path.exists('jobs/'):
+        return
+    msg = 'Auto-generated job branch from {{}} from {}'.format(
+        subprocess.check_output(['git', 'log', '-n', '1', '--pretty=%h']))
+    jobs = [d for (d, _, files) in os.walk('jobs/') if 'Jenkinsfile' in files]
```
stevekuznetsov: `job_directories` may be more clear

build-scripts/multidirectory.py
```diff
@@ -0,0 +1,44 @@
...
+    for job in jobs:
```
stevekuznetsov: We should try to keep the level of abstraction high on this method so someone can take a cursory look at the script and understand what's going on:

```py
for branch in job_branches():
    if branch_exists_locally(branch):
        delete_local_branch(branch)

    create_local_branch(branch)
    push_branch(branch)
````

**Comment by bbguimaraes** [comment]
Updated, with changes discriminated on a separate commit.

All these commits should be squashed into one before merging.

File: build-scripts/multidirectory.py [hunk 16]
```diff
@@ -0,0 +1,67 @@
...
+        'git', 'branch', '--list', '--remotes', '--no-merged', 'HEAD',
+        remote + '/*'])
+    branch_names = [ref[len(remote) + 1:] for ref in branch_refs.split()]
+    for branch in branch_names:
+        prune_branch(branch)
+
+def create_or_update_branches(remote):
+    if not os.path.exists('jobs/'):
+        return
+    job_directories = \
```
stevekuznetsov: 120 characters per line is fine, this stuff is only hurting the readability.

or:

```py
for directory, _, files in os.walk('jobs/'):
    if 'Jenkinsfile' in files:
        branch = job[len('jobs/'):]
        create_or_update_branch(branch)
```

build-scripts/multidirectory.py
```diff
@@ -0,0 +1,67 @@
...
+
+def prune_branch(branch):
+    if not jenkinsfile_exists(branch):
+        delete_remote_branch(branch)
+
+def jenkinsfile_exists(branch):
+    return os.path.exists(os.path.join('jobs', branch, 'Jenkinsfile'))
+
+def delete_remote_branch(branch):
+    subprocess.check_call([
```
stevekuznetsov: Same on these, tolerate more chars per line, 120 won't kill anyone :)

stevekuznetsov:
I think what you have written now in Python is a good first step, translating it from
the previous approach in Bash. However, I think we need to not only transliterate the
code from shell to Python but also change the mindset with which we approach the problem.
This is what I meant to say earlier. Either we write a very small shell script and revisit
the problem later or we write a Python program that is meant to not only handle the current
needs of this repository but also scale to the needs that it will have in the future.

So, in that vein I would like to see shell code like so:

```sh
#!/bin/bash

# prune_branches.sh removes branches from the remote if they
# do not have corresponding Jenkinsfiles in the master branch

remote="${1:-"origin"}"
for branch in $( git branch --list --remotes --no-merged HEAD "${remote}"/*); do
    jenkinsfile_dir="jobs/${branch#${remote}}"
    if [[ ! -e "${jenkinsfile_dir}/Jenkinsfile" ]]; then
        git push "${remote}" --delete --force "${branch}"
    fi
done
```

Translated into Python code like so:

```py
from os import getenv
from os.path import join, exists

from git import Repo, RemoteReference


def initialize_repo():
    repo_dir = join(getenv('WORKSPACE'), 'aos-cd-jobs')
    return Repo.clone_from('git@github.com:openshift/aos-cd-jobs.git', repo_dir)


def prunable_remote_refs(repo):
    candidates = []
    for ref in repo.remotes.origin.refs:
        if not (ref.name == 'HEAD' or ref.remote_head == 'master'):
            candidates.append(ref)
    return candidates


def remote_ref_needs_pruning(ref):
    return not exists(jenkinsfile_for_ref(ref))

def jenkinsfile_for_ref(ref):
    return join(
        ref.repo.working_dir,
        ref.remote_head,
        'Jenkinsfile'
    )

def prune_remote_ref(ref):
    RemoteReference.delete(ref.remote_name, ref, force=True)


def prune_remote_refs(repo):
    for ref in prunable_remote_refs(repo):
        if remote_ref_needs_pruning(ref):
            prune_remote_ref(ref)
            
if __name__ == '__main__':
    repo = initialize_repo()
    prune_remote_refs(repo)
```

This approach does not use `subprocess` and has a couple advantages:
 - we are not parsing strings at any point in the interaction with the repo
 - we are interacting with a fully featured `git` interface, so extension is easier

Furthermore, as mentioned above, the methods do not mix _how_ with _what_. For instance,
`prune_remote_refs()` and the `__main__` entrypoint both list _what_ is going on whereas
the other methods explain _how_ it's being done. This separation of different levels of
abstraction keeps the barrier to entry for this program small and lets us chop and change
the underlying implementations without having to touch the higher level logic.

We can now write tests for this functionality, focusing on the two methods that contain
our own logic:

```py
from unittest import TestCase, main
from unittest.mock import MagicMock, patch

from aos_cd_jobs.pruner import jenkinsfile_for_ref, prunable_remote_refs, prune_remote_refs

class TestPruneMethods(TestCase):
    def test_jenkinsfile_for_ref(self):
        """ Jenksinsfile should be under ref path from repo root """
        ref = MagicMock()
        ref.repo.working_dir = 'base'
        ref.remote_head = 'path'

        self.assertEqual(jenkinsfile_for_ref(ref), 'base/path/Jenkinsfile')

    def test_prunable_remote_refs_exclude_head(self):
        """ HEAD should not be considered prunable """
        ref = MagicMock()
        ref.name = 'HEAD'
        repo = MagicMock()
        repo.remotes.origin.refs = [ref]

        self.assertEqual(prunable_remote_refs(repo), [])

    def test_prunable_remote_refs_exclude_head(self):
        """ master should not be considered prunable """
        ref = MagicMock()
        ref.remote_head = 'master'
        repo = MagicMock()
        repo.remotes.origin.refs = [ref]

        self.assertEqual(prunable_remote_refs(repo), [])

    def test_prunable_remote_refs_contain_rest(self):
        """ normal refs should be considered prunable """
        repo = MagicMock()
        repo.remotes.origin.refs = [MagicMock()] * 10

        self.assertEqual(prunable_remote_refs(repo), repo.remotes.origin.refs)

    @patch('pruner.remote_ref_needs_pruning', return_value=True)
    @patch('pruner.prunable_remote_refs', return_value=[MagicMock()])
    @patch('pruner.prune_remote_ref')
    def test_prune_remote_refs_positive(self, prune_remote_ref_mock):
        """ refs that need pruning should be pruned """
        prune_remote_refs(MagicMock())

        self.assertTrue(prune_remote_ref_mock.called)

    @patch('pruner.remote_ref_needs_pruning', return_value=False)
    @patch('pruner.prunable_remote_refs', return_value=[MagicMock()])
    @patch('pruner.prune_remote_ref')
    def test_prune_remote_refs_positive(self, prune_remote_ref_mock):
        """ refs that do not need pruning should not be pruned """
        prune_remote_refs(MagicMock())

        self.assertFalse(prune_remote_ref_mock.called)


if __name__ == '__main__':
    unittest.main()
```

bbguimaraes:
@stevekuznetsov

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,47 @@
...
+        create_remote_branch(repo, job)
+
+def list_jobs(directory, walker):
+    return (
+        name[len(directory) + 1:]
+        for (name, _, files) in walker
+        if 'Jenkinsfile' in files
+    )
+
+def get_branch_by_name(branches, name):
```
stevekuznetsov: Not sure why this is necessary, you can do:

```py
if 'name' in repo.branches:
```

 and

```py
repo.branches['name']
```

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,47 @@
...
+    return next((b for b in branches if b.name == name), None)
+
+def create_remote_branch(repo, name):
+    branch = repo.active_branch.checkout(orphan=name)
+    directory = join(repo.working_dir, 'jobs', name)
```
stevekuznetsov: Make a constant for `'jobs'` if you're using it in more than one place

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,47 @@
...
+    create_branch(directory, listdir(directory))
+    repo.index.add(['.'])
+    repo.index.commit(
+        'Auto-generated job branch from {} from {}'.format(
+            name, repo.heads.master.commit.hexsha[:7]))
+    repo.remotes.origin.push(name, force=True)
+
+def create_branch(directory, files):
```
stevekuznetsov: This is a misleading method name

aos_cd_jobs/updater_test.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
+from unittest import TestCase, main
+from mock import Mock, call, patch
+
+from aos_cd_jobs.updater import create_branch, create_remote_branch, get_branch_by_name, list_jobs, update_branches
+
+class TestUpdateMethods(TestCase):
+    def test_list_jobs(self):
```
stevekuznetsov: Should be two tests, one positive and one negative.

aos_cd_jobs/updater_test.py
```diff
@@ -0,0 +1,70 @@
...
+        walker = (
+            ('dir/job0', None, ['Jenkinsfile']),
+            ('dir/notajob', None, ['']),
+            ('dir/job1', None, ['Jenkinsfile']),
+        )
+        self.assertEqual(tuple(list_jobs('dir', walker)), ('job0', 'job1'))
+
+    def test_get_branch_by_name(self):
+        branches = (Mock(), Mock())
+        branches[0].name = 'branch0'
```
stevekuznetsov: There doesn't seem to be any added benefit to testing the positive case twice here, so you do not need both of these

aos_cd_jobs/updater_test.py
```diff
@@ -0,0 +1,70 @@
...
+        branches[1].name = 'branch1'
+        self.assertEqual(get_branch_by_name(branches, 'branch0'), branches[0])
+        self.assertEqual(get_branch_by_name(branches, 'branch1'), branches[1])
+        self.assertIsNone(get_branch_by_name(branches, 'branch2'))
```
stevekuznetsov: This should be a separate test. Try to keep each test down to one assertion

aos_cd_jobs/updater_test.py
```diff
@@ -0,0 +1,70 @@
...
+        directory = 'jobs/build/ose'
+        create_branch(directory, ('Jenkinsfile', 'README'))
+        rename_mock.assert_has_calls((
+            call('jobs/build/ose/Jenkinsfile', 'Jenkinsfile'),
+            call('jobs/build/ose/README', 'README')))
+        rmtree_mock.assert_called_once_with(directory)
+
+    @patch('aos_cd_jobs.updater.listdir', lambda *_: ())
+    @patch('aos_cd_jobs.updater.create_branch')
+    def test_create_remote_branch(self, create_mock):
```
stevekuznetsov: This test over-specifies the implementation specifics of the `create_remote_branch` call

aos_cd_jobs/updater_test.py
```diff
@@ -0,0 +1,70 @@
...
+        branches = (Mock(), Mock())
+        branches[0].name = 'branch0'
+        branches[1].name = 'branch1'
+        self.assertEqual(get_branch_by_name(branches, 'branch0'), branches[0])
+        self.assertEqual(get_branch_by_name(branches, 'branch1'), branches[1])
+        self.assertIsNone(get_branch_by_name(branches, 'branch2'))
+
+    @patch('aos_cd_jobs.updater.rename')
+    @patch('aos_cd_jobs.updater.rmtree')
+    def test_create_branch(self, rmtree_mock, rename_mock):
```
stevekuznetsov: When you write more targeted functions like this one, it may make sense to write specific implementation tests (in the case of behavior verification, as is necessary here). Those types of tests are less appropriate for functions with larger scopes.

bbguimaraes:
@stevekuznetsov: updated.

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+        if branch is not None:
+            branch.delete()
+        create_remote_branch(repo, job)
+
+def list_jobs(repo):
+    jobs = []
+    jobs_directory = join(repo.working_dir, JOBS_DIRECTORY)
+    for (dirpath, _, filenames) in walk(jobs_directory):
+        if 'Jenkinsfile' in filenames:
+            jobs.append(dirpath[len(jobs_directory) + 1:])
```
stevekuznetsov: Don't use string transformations when you mean to use path transformations.

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+from shutil import rmtree
+
+from git import Repo
+
+from aos_cd_jobs.common import JOBS_DIRECTORY, initialize_repo
+
+def update_branches(repo):
+    for job in list_jobs(repo):
+        branch = get_branch_by_name(repo.branches, job)
+        if branch is not None:
```
stevekuznetsov: I don't think there is a lot of benefit to this. Without the method, you have

```py
if branch in repo.branches:
    branch.delete()
```

with the method you have

```py
if branch in repo.branches:
    branch = repo.branches[branch]
else:
    branch = None

if branch is not None:
    branch.delete()
```

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+    directory = join(repo.working_dir, JOBS_DIRECTORY, name)
+    create_job_file_tree(directory, listdir(directory))
+    repo.index.add(['.'])
+    repo.index.commit(
+        'Auto-generated job branch from {} from {}'.format(
+            name, repo.heads.master.commit.hexsha[:7]))
+    repo.remotes.origin.push(name, force=True)
+
+def create_job_file_tree(directory, files):
+    for file in files:
```
stevekuznetsov: `file` is a builtin -- don't alias it as a method name

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+    return jobs
+
+def get_branch_by_name(branches, name):
+    if name in branches:
+        return branches[name]
+
+def create_remote_branch(repo, name):
+    branch = repo.active_branch.checkout(orphan=name)
+    directory = join(repo.working_dir, JOBS_DIRECTORY, name)
+    create_job_file_tree(directory, listdir(directory))
```
stevekuznetsov: Why not pass in just `directory` and do the `listdir` call inside?

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+    repo.index.add(['.'])
+    repo.index.commit(
+        'Auto-generated job branch from {} from {}'.format(
+            name, repo.heads.master.commit.hexsha[:7]))
+    repo.remotes.origin.push(name, force=True)
+
+def create_job_file_tree(directory, files):
+    for file in files:
+        rename(join(directory, file), file)
+    rmtree(directory)
```
stevekuznetsov: Have you considered `shutil.move()` instead of `rename` + `rmtree`?

bbguimaraes:
@stevekuznetsov

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+
+from aos_cd_jobs.common import JOBS_DIRECTORY, initialize_repo
+
+def update_branches(repo):
+    for job in list_jobs(repo):
+        if job in repo.branches:
+            branch = repo.branches[job]
+        else:
+            branch = None
+        if branch is not None:
```
stevekuznetsov: I'm still not really sure we need this -- the example I had from before was:

```py
if job in repo.branches:
    repo.branches[job].delete()
create_remote_branch(repo, job)
```

Still keeps us from having to do `None` checks, etc

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+        if 'Jenkinsfile' in filenames:
+            jobs.append(relpath(dirpath, jobs_directory))
+    return jobs
+
+def create_remote_branch(repo, name):
+    branch = repo.heads.master.checkout(orphan=name)
+    directory = join(repo.working_dir, JOBS_DIRECTORY, name)
+    create_job_file_tree(repo.working_dir, directory)
+    repo.index.remove(repo.index.entries.values())
+    repo.index.add([f for f in listdir(repo.working_dir) if f != '.git'])
```
stevekuznetsov: I think the suggestions in [this thread](https://github.com/gitpython-developers/GitPython/issues/292) are a little clearer and more correct than this.

aos_cd_jobs/updater.py
```diff
@@ -0,0 +1,46 @@
...
+
+def list_jobs(repo):
+    jobs = []
+    jobs_directory = join(repo.working_dir, JOBS_DIRECTORY)
+    for (dirpath, _, filenames) in walk(jobs_directory):
+        if 'Jenkinsfile' in filenames:
+            jobs.append(relpath(dirpath, jobs_directory))
+    return jobs
+
+def create_remote_branch(repo, name):
```
stevekuznetsov: You previously had a test for this function but have deleted it. The rationale behind my comment previously was as such:

The method as is written has a lot of scope. I understand why you were using behavior validation in your test, and I agree that it is the correct means by which you should be testing this, but the previous test was overly specific. Once the test was committed, it would mean that even the slightest change to the method in question would necessarily cause a test to fail and thereby generate more work. If we try to write this as a "what" method instead of a "why" method, we get something like this (arguments elided, but you get the point):

```py
def create_remote_branch():
    initialize_orphan_branch()
    populate_branch()
    publish_branch()
```

Then, we can write very specific implementation-validation tests for the "how" methods like `publish_branch()`. Even though any one test is still very specific and refactoring any one of the underlying "how" methods still leads to test failure, it's scoped. A small code change leads to a small test change.

I would much rather see this than see us drop the test entirely.

bbguimaraes:
@stevekuznetsov

bbguimaraes:
@stevekuznetsov: the way I test this is using two local repositories.  The
first is a mirror of https://github.com/openshift/aos-cd-jobs.git, the second a
clone of the first (see the scripts in the first commit).

We can then execute the script on the second clone (which will result in pushes
to the first) and check that the state of the first clone is correct.  Would
that be enough or do you think a "dry-run mode" is still necessary?

And if it is necessary, what should it do?  The two main operations are
deleting branches on the remote and creating local branches that follow a
certain structure and pushing them to the remote.  How should the "output" of a
dry run be?

stevekuznetsov:
A dry-run would have just changed local git state and printed out where it would have pushed. However, your alternative sounds just as good. You need #145 to really test it, right?

bbguimaraes:
To get the repository to the final state described in the card, yes.  I have
to review the changes that happened to the branches since I worked on that.

bbguimaraes:
@jupierce, @stevekuznetsov: rebase to the latest master, added a jenkinsfile
for the job.

Since we have the job already in place, *do not merge* this.  Once this is
reviewed and approved, we can do a final update to #145 and then get this in.

aos_cd_jobs/common.py
```diff
@@ -1,11 +1,13 @@
...
+from os.path import exists, join
 
 from git import Repo
 
 JOBS_DIRECTORY = 'jobs'
 
 def initialize_repo():
     repo_dir = join(getenv('WORKSPACE'), 'aos-cd-jobs')
+    if exists(repo_dir):
+        return Repo(repo_dir)
```
stevekuznetsov: I think we want to clear out the working directory on the job and start from scratch here so we have fewer headaches about what local git state does and does not exist. Thoughts?

Jenkinsfile
```diff
@@ -4,8 +4,42 @@ properties(
...
+
 node('buildvm-devops') {
-  sshagent(['openshift-bot']) { // git repo privileges stored in Jenkins credential store
-    echo "Bruno's branch split pipeline"
+  try {
+    deleteDir()
+    set_workspace()
+    stage('clone') {
+      dir('aos-cd-jobs') {
+        checkout scm
```
stevekuznetsov: if we do a scm trigger, will this be done for us?

Jenkinsfile
```diff
@@ -4,8 +4,42 @@ properties(
...
+        sh 'git checkout master'
+      }
+    }
+    stage('run') {
+      sshagent(['openshift-bot']) { // git repo privileges stored in Jenkins credential store
+        sh '''\
+virtualenv env/
+env/bin/pip install gitpython
+export PYTHONPATH=$PWD/aos-cd-jobs
+env/bin/python aos-cd-jobs/aos_cd_jobs/pruner.py
```
stevekuznetsov: If the scripts have a shebang and are executable, just run them

Jenkinsfile
bbguimaraes: No, we have to explicitly tell jenkins to clone the repository.  I guess that's
because you might not want to.

What we do get with the scm trigger is the ability to clone it without giving
the source url, which is nice for testing with a local mirror.

Jenkinsfile
bbguimaraes: This is forcing the interpreter to be the one inside the virtualenv.  To
execute the scripts directly, we'd have to do:

```sh
source env/bin/activate
aos-cd-jobs/aos_cd_jobs/pruner.py
aos-cd-jobs/aos_cd_jobs/updater.py
```

aos_cd_jobs/common.py
bbguimaraes: The first command in the job clears the existing workspace, so we already get a
fresh clone at every run.  This change is to reuse the repository between the
execution of `pruner.py` and `updater.py`.  The intersection between the two is
very small, so there's not much risk of one interfering with the other.

Jenkinsfile
stevekuznetsov: Why don't we want to activate the virtualenv?

Jenkinsfile
stevekuznetsov: That is super weird. On the freestyle jobs if you do a SCM trigger on a commit, it does the cloning for you before your steps run. They really turned everything on it's head with the Jenkinsfiles :\

Jenkinsfile
bbguimaraes: Sourcing the `activate` file and executing the virtualenv's `python` binary are
two equivalent (in this case, at least) ways to use the virtualenv.

I don't have a preference for any, other than three lines versus two.

stevekuznetsov: I think it's more appropriate to activate it. I don't know for certain that it's equivalent to use the `python` symlink in the `virtualenv` dir, as the activation also changes `$PATH` and `$PYTHON_HOME`. You may find the "normal" activate script to fail in Jenkins as the shell has no `$PS1`, so in other jobs we do:

```
export VIRTUAL_ENV="$( pwd )/env"
export PATH="${VIRTUAL_ENV}/bin:${PATH}"
unset PYTHON_HOME
```

bbguimaraes: To me that's one more reason to use the binary, not the `activate` script.
From virtualenv's docs:

> This will change your `$PATH` so its first entry is the virtualenv’s `bin/`
> directory. (You have to use `source` because it changes your shell
> environment in-place.) This is all it does; it’s purely a convenience. If you
> directly run a script or the python interpreter from the virtualenv’s `bin/`
> directory (e.g. `path/to/ENV/bin/pip` or `/path/to/ENV/bin/python-script.py`)
> there’s no need for activation.

https://virtualenv.pypa.io/en/stable/userguide/#activate-script

bbguimaraes:
@stevekuznetsov: jenkins job updated.

PR #136, opened 2017-03-22
Only keep 7 days worth of job history for any one job

stevekuznetsov:


smarterclayton:
We may want to go back to 14 days if we figure out the other problem

stevekuznetsov:
Easy change to make if we see that this does not help us with allocations

PR #138, opened 2017-03-22
Switch from merge to rebase

kargakis:
Implementation of the rebase proposal: https://github.com/openshift/aos-cd-jobs/pull/1

@jupierce @tdawson @stevekuznetsov @smunilla @bbguimaraes


kargakis:
Still misses unit tests for rebase.py but this is ready for reviews

kargakis:
@bbguimaraes @jhadvig @smunilla could also help out with reviewing

scripts/sanity_check_test.py
```diff
@@ -0,0 +1,139 @@
...
+    """ Tests for commit validation """
+
+    def assertFailure(self, commits, expected):
+        from StringIO import StringIO
+
+        real_commits = convert_commits(commits)
+        out = StringIO()
+
+        self.assertRaises(SystemExit, sanity_check.validate_commits, real_commits, out)
+        assert out.getvalue().strip() == expected
```
stevekuznetsov:
```py
self.assertEquals(out.getvalue().strip(), expected)
```

for parallelism

scripts/sanity_check_test.py
```diff
@@ -0,0 +1,139 @@
...
+        "pick ba426b9 Automatic commit of package [atomic-openshift] release [3.5.0.50-1].",
+        "pick 4365d5d [DROP][INVESTIGATE] Weird shell diffs",
+        "pick 8ff2968 [DROP][FIXME] Change image build and drop this cluster up flag",
+        "pick 5hfos3s [SQUASH][BUILD] Check in new shiny script for OSE",
+        "pick c21ea22 [SQUASH][BRANDING] Additional OAuth server branding",
+        ]
+        self.assertSuccess(commits)
+
+
+def convert_commits(commits):
```
stevekuznetsov: Should just have commit literals

scripts/sanity_check_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+    # Test cases that excercise invalid scenarios.
+
+    def test_squash_not_matching_carry(self):
+        """
+        [SQUASH] commit has a type that is not matching any [CARRY].
+        """
+
+        commits = [
```
stevekuznetsov: This just needs one commit, right? No need for others in here.

stevekuznetsov: Also do we need to differentiate between `[SQUASH]` without any `[CARRY]` at all and `[SQUASH]` without any matching `[CARRY]`?

scripts/sanity_check_test.py
```diff
@@ -0,0 +1,139 @@
...
+        ]
+        expected = "Not all SQUASH commits correspond to a CARRY commit!"
+        self.assertFailure(commits, expected) 
+
+    def test_carry_without_type(self):
+        """
+        [CARRY] commit does not have a type.
+        """
+
+        commits = [
```
stevekuznetsov: again, just need the one input commit

scripts/sanity_check_test.py
```diff
@@ -0,0 +1,139 @@
...
+        "pick c21ea21 [SQUASH][BRANDING] OAuth server branding",
+        "pick c21ea22 [SQUASH][DIFFERENT] Different kind of stuff",
+        "pick 975a829 [CARRY][BUILD_GEN] Specfile updates",
+        "pick ba426b9 Automatic commit of package [atomic-openshift] release [3.5.0.50-1].",
+        "pick 975a829 [SQUASH] without a type because I am lazy",
+        ]
+        expected = "Found a [SQUASH] commit without a type: `[SQUASH] without a type because I am lazy` (975a829)"
+        self.assertFailure(commits, expected)
+
+    def test_missing_tito_tag(self):
```
stevekuznetsov: the fact that the rest of the test cases need this commit not to fail is a symptom of these being integration tests, not unit tests

scripts/sanity_check_test.py
kargakis: We need to validate that every SQUASH matches a CARRY, otherwise what's the reason for having SQUASH in the first place?

scripts/sanity_check_test.py
kargakis: All the logic is found in validate_commits. Are you suggesting spliting commit message validation out of cross-commit validation?

scripts/sanity_check_test.py
kargakis: Enforcing intent in our carries is a good way to maintain as few as possible so I think there is value in cross-commit validation (why would I submit a SQUASH w/o a matching CARRY?)

stevekuznetsov: I was talking about the logic for the test, not for the business logic -- I mean that there is extraneous data in this list that is not necessary for the correct execution of the test. I was also asking if, in order to test this feature, we need two tests -- one that tested that if there were no `[CARRY]` commits with the same subject as the `[SQUASH]` it would fail as well as testing if there were no `[CARRY]` commits at all it would fail.

scripts/sanity_check_test.py
stevekuznetsov: If you test at a more granular level than `validate_commits()` it makes the test suite more robust. There's value to the top-level tests but these sorts of inherent dependencies between tests just make it harder to refactor the code in the future, as well as making for lots of false-flag test failures when someone breaks this functionality, as the rest of the tests have not isolated the unit under test.

scripts/sanity_check_test.py
kargakis: The input is always all the commits on top of Origin.

scripts/sanity_check_test.py
kargakis: Totally agreed. I split the validation yesterday and will scope the unit tests today.

kargakis:
@stevekuznetsov unit tests are ready once again

kargakis:
> What is the bootstrapping we need to get master into the right state before this lands?

Force-tag the latest existing tag once we reset master. Note that we will need to do the same thing for `stage` but we can roll this in master first (I will need to update merge-and-build.sh in this PR) and then roll it in stage too after a while.

stevekuznetsov:
Do we want to script out the bootstrapping steps?

kargakis:
I am not sure we want to automate the bootstraping phase... @jupierce @tdawson thoughts?

stevekuznetsov:
If not automate just write out the steps explicitly so it's hard for whoever is doing it to screw it up

tdawson:
+1 to writting out the steps and doing the initial bootstrap manually

kargakis:
Ok, I will revisit the process once more on my fork and write the steps down. Last thing we need to decide on is do we roll the new process for the stage branch too or only for master initially? If the former, we need to reset stage alongisde the master reset. Also I will probably need to create a fake-stage branch (can exist in my fork) because the generated CARRY commit will differ between master and stage.

kargakis:
Here are the steps (verified once more in my fork).

We may want to backup master before making the switch but I expect the backup to become obsolete after the first successful build using the new process.

git checkout master
last_tag=$( git describe --tags --abbrev=0 )
git reset --hard fake-master
git tag -f $last_tag HEAD
git push origin master -f refs/tags/${last_tag}

stevekuznetsov:
Do you have steps for bootstrapping `fake-master`? I thought we would do those steps on `master` instead of just cutting over from `fake`-master`? Or if we are cutting over do you have steps to update `fake-master` to the last merge/rebase?

kargakis:
> Or if we are cutting over do you have steps to updatefake-master` to the last merge/rebase?

There is a job that syncs fake-master with master. If we want to cut from upstream/master then the steps are:
```sh
git clone git@github.com:openshift/ose
cd ose
git remote add upstream git@github.com:openshift/origin
git fetch upstream
git checkout fake-master
CURRENT_HEAD=$(git merge-base fake-master upstream/master)
picks=$(git log --oneline --reverse $CURRENT_HEAD..HEAD --pretty=%h)
git checkout master
git reset --hard upstream/master
for c in ${picks}; do
   git cherry-pick $c
done
```

kargakis:
Or we could avoid fake-master commits altogether by resetting (neither --hard nor --soft) master to upstream/master and constructing the cherry-picks from scratch.

kargakis:
> Or we could avoid fake-master commits altogether by resetting (neither --hard nor --soft) master to upstream/master and constructing the cherry-picks from scratch.

Correction: that should be a soft reset.

Another thing I just realized while reworking the rebase proposal in https://github.com/openshift/aos-cd-jobs/pull/172 is that developers cannot do
```
git fetch upstream
git merge upstream/master
```
anymore .... :/

They will always need to
```
git reset --hard upstream/master
```

Also I still need to think through the scenario of merge conflicts. Today, the build is restarted once the conflicts are resolved. This will not work with the current state of the code because the script will fail with an unbound variable for `changelog` - all commits already exist in master after the conflict is resolved so tito cannot find any new commit to use in the changelog.

stevekuznetsov:
> `git reset --hard upstream/master`

Yeah, that was an early thing I noticed. We will just need to blast aos-devel. Such is the harm of rewriting history.

> Also I still need to think through the scenario of merge conflicts. Today, the build is restarted once the conflicts are resolved. This will not work with the current state of the code because the script will fail with an unbound variable for changelog - all commits already exist in master after the conflict is resolved so tito cannot find any new commit to use in the changelog.

If you're looking for suggestions here can you lay out some example commit trees to make this a little clearer?

kargakis:
> If you're looking for suggestions here can you lay out some example commit trees to make this a little clearer?

This does not depend on a specific example. Today, when the build fails due to a conflict, we resolve the conflict manually, update master (which means all of the new commits are checked-in outside of the build process), and restart the build. I think the changelog is recorded correctly because tito uses --accept-auto-changelog which is looking in the diff from the previous tag? 

This process will not work with the way I am doing things here. The changelog is based on the previous HEAD and the current HEAD of Origin. After a conflict is resolved and we update master, `tito tag` will fail because the `changelog` variable will be unset (previous HEAD will equal current HEAD). The only solution I see out of this, is that whoever resolves a conflict, also needs to tito tag and the build process should kick off from that point.

Is it possible to have a Jenkins pipeline where we can skip a step?

stevekuznetsov:
> Is it possible to have a Jenkins pipeline where we can skip a step?

Yes absolutely.

kargakis:
We could also add an option in the current merge-and-build.sh that short-circuits to the point the tito tag has been done.

stevekuznetsov:
We can also begin to unravel the script so we have a pipeline with more than one stage ;)

kargakis:
> We can also begin to unravel the script so we have a pipeline with more than one stage ;)

This is what @bbguimaraes has been working on in https://github.com/openshift/aos-cd-jobs/tree/build/ose-pipeline

kargakis:
I just updated the master in my fork once again - now by reconstructing the commits from scratch. Possibly the best approach: https://pastebin.com/a247uHYL (same thing as https://github.com/openshift/aos-cd-jobs/pull/138#issuecomment-292194764).

scripts/git.py
```diff
@@ -0,0 +1,75 @@
...
+
+    return commits
+
+def create_commit(line, action):
+    length = len(action)+1
+    return Commit(
+        action=action,
+        hash=line[length:length+7],
+        subject=line[length+8:]
+    )
```
rhcarvalho: Suggestion:

```python
def create_commit(line):
    action, hash, subject = line.split(maxsplit=2)
    return Commit(
        action=action,
        hash=hash,
        subject=subject,
    )
```

The above reads the action from the `line`, otherwise, calling `create_commit` with an action that does not match the line would cause strange behavior: `create_commit('fixup 32df334 This is the subject', Action.pick)` -> `Commit('pick', ' 32df33', ' This is the subject')` (off-by-one char).

kargakis: Thanks although this is Python 3 and the buildvm is using 2 atm so I had to drop the maxsplit arg

rhcarvalho: No difference between Python 2 or 3 here, both have the `maxsplit` argument ;)
Though, yeah, you need to write it like this:

```python
line.split(None, 2)
```

Not as a keyword argument...

PR #139, opened 2017-03-22
Determine install/upgrade version script

jhadvig:
This script will determine the GA and latest packages for previous version of package that will be on input.
EG:
input: ` origin-1.4.1-1.el7`
output:
 - ORIGIN_INSTALL_VERSION file that includes previous GA version-> `1.3.0-3.el7.x86_64`
 - ORIGIN_UPGRADE_VERSION file that includes previous latest version-> `1.3.3-1.el7.x86_64`

The script will return pre-release versions of packages if there are no released ones or fail.

### DIFF
--- a/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
+++ b/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
@@ -6,45 +6,123 @@ extensions:
     - name: "aos-cd-jobs"
   actions:
     - type: "script"
-      title: "install origin GA and update"
+      title: "determine the release commit for origin images"
+      repository: "origin"
+      script: |-
+        jobs_repo="/data/src/github.com/openshift/aos-cd-jobs/"
+        git log -1 --pretty=%h > "${jobs_repo}/ORIGIN_COMMIT"
+        source hack/lib/init.sh
+        os::build::rpm::format_nvra > "${jobs_repo}/ORIGIN_BUILT_VERSION"
+    - type: "script"
+      title: "build an openshift-ansible release"
+      repository: "openshift-ansible"
+      script: |-
+        tito_tmp_dir="tito"
+        mkdir -p "${tito_tmp_dir}"
+        tito tag --offline --accept-auto-changelog
+        tito build --output="${tito_tmp_dir}" --rpm --test --offline --quiet
+        createrepo "${tito_tmp_dir}/noarch"
+        cat << EOR > ./openshift-ansible-local-release.repo
+        [openshift-ansible-local-release]
+        baseurl = file://$( pwd )/${tito_tmp_dir}/noarch
+        gpgcheck = 0
+        name = OpenShift Ansible Release from Local Source
+        EOR
+        sudo cp ./openshift-ansible-local-release.repo /etc/yum.repos.d
+        basename "${tito_tmp_dir}"/noarch/atomic-openshift-utils*.rpm .rpm > /data/src/github.com/openshift/aos-cd-jobs/OPENSHIFT_ANSIBLE_BUILT_VERSION
+    - type: "script"
+      title: "install ansible atomic-openshift-utils"
       repository: "aos-cd-jobs"
       script: |-
         pkg_name="origin"
-        sudo yum install -y ansible-2.2.0.0
         git checkout sjb
-        general_yml_file=inventory/group_vars/OSEv3/general.yml
-        if [ $pkg_name = "atomic-openshift" ]
+        echo $pkg_name > PKG_NAME
+        echo "openshift-ansible openshift-ansible-callback-plugins openshift-ansible-docs openshift-ansible-filter-plugins openshift-ansible-lookup-plugins openshift-ansible-playbooks openshift-ansible-roles" > OPENSHIFT_ANSIBLE_PKGS
+        sudo python hack/determine_install_upgrade_version.py $( cat OPENSHIFT_ANSIBLE_BUILT_VERSION ) > OPENSHIFT_ANSIBLE_VARS
+        source OPENSHIFT_ANSIBLE_VARS
+        versioned_packages_to_install=""
+        if [[ ${ATOMIC_OPENSHIFT_UTILS_INSTALL_MINOR_VERSION} -le 4 ]]
         then
-          sed -i "s/deployment_type:.*/deployment_type: \"openshift-enterprise\" /g" $general_yml_file
+          sudo yum erase -y ansible
+          versioned_packages_to_install="ansible-2.2.0.0"
         fi
-        openshift_pkgs=( $pkg_name openshift-ansible-playbooks )
-        for pkg in "${openshift_pkgs[@]}"
+        packages_to_install=( $( cat ./OPENSHIFT_ANSIBLE_PKGS ) )
+        for pkg in "${packages_to_install[@]}"
         do
-          sudo python hack/get_available_pkgs.py $pkg > "all_${pkg}_versions"
-          sed -i '/^[a-zA-Z]/d' all_${pkg}_versions
-          sort -u -o all_${pkg}_versions all_${pkg}_versions
-          cat all_${pkg}_versions
+          versioned_packages_to_install+=" ${pkg}-${ATOMIC_OPENSHIFT_UTILS_INSTALL_VERSION}"
         done
-        current_openshift_version=$(tail -n1 all_${pkg_name}_versions | awk -F'-' '{print $1}')
-        echo "current version -> $current_openshift_version"
-        current_openshift_major_version=$(echo $current_openshift_version | awk -F'.' '{print $1}')
-        current_openshift_minor_version=$(echo $current_openshift_version | awk -F'.' '{print $2}')
-        latest_version=$current_openshift_major_version.$(($current_openshift_minor_version + 1))
-        all_current_openshift_versions_array=($(cat "all_${pkg_name}_versions" | grep "$current_openshift_major_version.$current_openshift_minor_version"))
-        all_current_openshift_ansible_versions_array=($(cat "all_openshift-ansible-playbooks_versions" | grep "^3.$current_openshift_minor_version"))
-        set_openshift_version=$(echo ${all_current_openshift_versions_array[0]})
-        echo "=== Installing openshift-ansible-${all_current_openshift_ansible_versions_array[0]} ==="
-        sudo yum install -y openshift-ansible-playbooks-${all_current_openshift_ansible_versions_array[0]}
-        echo "=== Installing ${pkg_name}-${set_openshift_version} ==="
-        ansible-playbook --become --become-user root --connection local -i inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-node/network_manager.yml
-        ansible-playbook --become --become-user root --connection local -i inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml -e set_openshift_version="${set_openshift_version}"
-        if [ ${#all_current_openshift_versions_array[@]} -gt 1 ]
+        echo "=== Installing atomic-openshift-utils-${ATOMIC_OPENSHIFT_UTILS_INSTALL_VERSION} packages ==="
+        sudo yum install -y ${versioned_packages_to_install}
+    - type: "script"
+      title: "install Ansible plugins"
+      repository: "origin"
+      script: |-
+        sudo yum install -y python-pip
+        sudo pip install junit_xml
+        sudo chmod o+rw /etc/environment
+        echo "ANSIBLE_JUNIT_DIR=$( pwd )/_output/tests/ansible_junit" >> /etc/environment
+        sudo mkdir -p /usr/share/ansible/plugins/callback
+        for plugin in 'default_with_output_lists' 'generate_junit'; do
+           wget "https://raw.githubusercontent.com/openshift/origin-ci-tool/master/oct/ansible/oct/callback_plugins/${plugin}.py"
+           sudo mv "${plugin}.py" /usr/share/ansible/plugins/callback
+        done
+        sudo sed -r -i -e 's/^#?stdout_callback.*/stdout_callback = default_with_output_lists/' -e 's/^#?callback_whitelist.*/callback_whitelist = generate_junit/' /etc/ansible/ansible.cfg
+    - type: "script"
+      title: "install origin GA and update"
+      repository: "aos-cd-jobs"
+      script: |-
+        pkg_name=$( cat ./PKG_NAME )
+        if [[ "${pkg_name}" == "origin" ]]; then
+            deployment_type="origin"
+        elif [[ "${pkg_name}" == "atomic-openshift" ]]; then
+            deployment_type="openshift-enterprise"
+        else
+            echo "Can't determine deployment type"
+            exit 1
+        fi
+        echo "${deployment_type}" > DEPLOYMENT_TYPE
+        sudo python hack/determine_install_upgrade_version.py $( cat ORIGIN_BUILT_VERSION ) > ORIGIN_VARS
+        source ORIGIN_VARS
+        source OPENSHIFT_ANSIBLE_VARS
+        echo "=== Installing ${pkg_name}-${ORIGIN_INSTALL_VERSION} ==="
+        ansible-playbook  -vv                \
+                          --become           \
+                          --become-user root \
+                          --connection local \
+                          -i inventory/      \
+                          /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-node/network_manager.yml \
+                          -e deployment_type=$( cat ./DEPLOYMENT_TYPE)
+        # Remove "enable_excluders" variable after openshift-ansbile pkgs are rebuilded with https://github.com/openshift/openshift-ansible/pull/3788
+        ansible-playbook  -vv                \
+                          --become           \
+                          --become-user root \
+                          --connection local \
+                          -i inventory/      \
+                          /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml \
+                          -e openshift_pkg_version="-${ORIGIN_INSTALL_VERSION}"         \
+                          -e deployment_type=$( cat ./DEPLOYMENT_TYPE)                  \
+                          -e enable_excluders=false
+        if [[ -v ORIGIN_UPGRADE_VERSION  && -v ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION ]]
         then
-          echo "=== Updating openshift-ansible to openshift-ansible-${all_current_openshift_ansible_versions_array[${#all_current_openshift_ansible_versions_array[@]}-1]} ==="
-          sudo yum erase -y openshift-ansible
-          sudo yum install -y openshift-ansible-playbooks-${all_current_openshift_ansible_versions_array[${#all_current_openshift_ansible_versions_array[@]}-1]}
-          echo "=== Updating ${pkg_name} to ${pkg_name}-${all_current_openshift_versions_array[${#all_current_openshift_versions_array[@]}-1]} ==="
-          ansible-playbook --become --become-user root --connection local --inventory inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_${current_openshift_minor_version}/upgrade.yml -e openshift_upgrade_target="${all_current_openshift_versions_array[${#all_current_openshift_versions_array[@]}-1]}"
+          echo "=== Updating atomic-openshift-utils-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION} ==="
+          versioned_packages_to_install=""
+          packages_to_install=( $( cat ./OPENSHIFT_ANSIBLE_PKGS ) )
+          for pkg in "${packages_to_install[@]}"
+          do
+            versioned_packages_to_install+=" ${pkg}-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION}"
+          done
+          sudo yum upgrade -y ${versioned_packages_to_install}
+          echo "=== Updating ${pkg_name} to ${ORIGIN_UPGRADE_VERSION} ==="
+          upgrade_playbook="/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_${ORIGIN_UPGRADE_MINOR_VERSION}/upgrade.yml"
+          ansible-playbook  -vv                    \
+                            --become               \
+                            --become-user root     \
+                            --connection local     \
+                            --inventory inventory/ \
+                            "${upgrade_playbook}"  \
+                            -e openshift_pkg_version="-${ORIGIN_UPGRADE_VERSION}" \
+                            -e deployment_type=$( cat ./DEPLOYMENT_TYPE)          \
+                            -e oreg_url='openshift/origin-${component}:'"${ORIGIN_UPGRADE_VERSION_PKG_VERSION}"
         fi
     - type: "script"
       title: "build an origin release"
@@ -57,45 +135,34 @@ extensions:
         sed -i 's|go/src|data/src|' _output/local/releases/rpms/origin-local-release.repo
         sudo cp _output/local/releases/rpms/origin-local-release.repo /etc/yum.repos.d/
     - type: "script"
-      title: "build an openshift-ansible release"
-      repository: "openshift-ansible"
-      script: |-
-        tito_tmp_dir="tito"
-        mkdir -p "${tito_tmp_dir}"
-        tito tag --offline --use-version 3.6.0 --accept-auto-changelog
-        tito build --output="${tito_tmp_dir}" --rpm --test --offline --quiet
-        createrepo "${tito_tmp_dir}/noarch"
-        cat << EOR > ./openshift-ansible-local-release.repo
-        [openshift-ansible-local-release]
-        baseurl = file://$( pwd )/${tito_tmp_dir}/noarch
-        gpgcheck = 0
-        name = OpenShift Ansible Release from Local Source
-        EOR
-        sudo cp ./openshift-ansible-local-release.repo /etc/yum.repos.d
-    - type: "script"
-      title: "install the openshift-ansible release"
-      script: |-
-        sudo yum install -y atomic-openshift-utils
-    - type: "script"
-      title: "check out the correct branch of aos-cd-jobs"
+      title: "upgrade openshift-ansible to release"
       repository: "aos-cd-jobs"
       script: |-
-        git checkout sjb
+        source ORIGIN_VARS
+        source OPENSHIFT_ANSIBLE_VARS
+        versioned_packages_to_upgrade=""
+        packages_to_upgrade="$( cat ./OPENSHIFT_ANSIBLE_PKGS )"
+        for pkg in "${packages_to_upgrade[@]}"
+        do
+          versioned_packages_to_upgrade+=" ${pkg}-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_RELEASE_VERSION}"
+        done
+        sudo yum upgrade -y ${versioned_packages_to_upgrade}
     - type: "script"
-      title: "determine the release commit for origin images"
-      repository: "origin"
-      script: |-
-        git log -1 --pretty=%h >> /data/src/github.com/openshift/aos-cd-jobs/ORIGIN_COMMIT
-    - type: "script"
-      title: "update origin"
+      title: "update origin to release"
       repository: "aos-cd-jobs"
       script: |-
-        sudo python hack/get_available_pkgs.py origin > "all_origin_versions"
-        sed -i '/^[a-zA-Z]/d' all_origin_versions
-        sort -u -o all_origin_versions all_origin_versions
-        upgrade_version=$(tail -1 all_origin_versions)
-        upgrade_minor_version=$(echo "${upgrade_version}" | awk -F'.' '{print $2}')
-        ansible-playbook --become --become-user root --connection local --inventory inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_${upgrade_minor_version}/upgrade.yml -e openshift_image_tag="$( cat ./ORIGIN_COMMIT )" -e openshift_upgrade_target="${upgrade_version}"
+        pkg_name=$( cat ./PKG_NAME )
+        source ORIGIN_VARS
+        upgrade_playbook="/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_${ORIGIN_UPGRADE_RELEASE_MINOR_VERSION}/upgrade.yml"
+        ansible-playbook  -vv                    \
+                          --become               \
+                          --become-user root     \
+                          --connection local     \
+                          --inventory inventory/ \
+                           "${upgrade_playbook}" \
+                          -e openshift_pkg_version="-${ORIGIN_UPGRADE_RELEASE_VERSION}" \
+                          -e deployment_type=$( cat ./DEPLOYMENT_TYPE) \
+                          -e oreg_url='openshift/origin-${component}:'"$( cat ./ORIGIN_COMMIT )"
     - type: "script"
       title: "expose the kubeconfig"
       script: |-
@@ -106,6 +173,9 @@ extensions:
       repository: "origin"
       script: |-
         KUBECONFIG=/etc/origin/master/admin.kubeconfig TEST_ONLY='true' JUNIT_REPORT='true' make test-extended SUITE=conformance
+  generated_artifacts:
+    origin_package_history.log: "sudo yum history info origin"
+    openshift_ansible_package_history.log: "sudo yum history info openshift-ansible"
   system_journals:
     - origin-master.service
     - origin-node.service
--- a/hack/determine_install_upgrade_version.py
+++ b/hack/determine_install_upgrade_version.py
@@ -1,0 +1,87 @@
+import yum
+import sys
+import os.path
+import rpmUtils.miscutils as rpmutils
+
+# Get all the matching `MAJOR.MINOR` versions of provided input package nerva
+# Returned list of package versions is in `'pkg_version'-'pkg_release'` format ,eg: `1.3.0-3.el7`
+def get_matching_versions(input_pkg_name, available_pkgs, search_version):
+	release_pkgs = []
+	pre_release_pkgs = []
+	for pkg in available_pkgs:
+		if pkg.version.startswith(search_version):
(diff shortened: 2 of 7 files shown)


hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,55 @@
...
+	sys.exit(1)
+except:
+	print "[ERROR] Incorrect package nevra format: " + input_pkg
+	sys.exit(1)
+
+input_pkg_name = str(input_pkg_obj.name)
+input_pkg_version_list = map(int, input_pkg_obj.version.split("."))
+search_version_list = input_pkg_version_list[:]
+search_version_list[1] = search_version_list[1] -1
+search_version = ".".join(map(str, search_version_list)[:2])
```
stevekuznetsov: Maybe the `rpm` package has utilities for this?

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,55 @@
...
+
+yb = yum.YumBase()
+yb.conf.disable_excludes = ["all"]
```
stevekuznetsov: Add a comment why this is necessary

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,55 @@
...
+for pkg in available_pkgs:
+    if pkg.version.startswith(search_version):
+    	if pkg.release.startswith("0"):
+    		pre_release_pkgs.append(pkg.vra)
+    	else:
+    		release_pkgs.append(pkg.vra)
+
+def determine_install_upgrade_versions(version_list):
+	write_package_version(version_list[0], "install")
+	if (len(version_list) > 1) and (version_list[0] is not version_list[len(version_list)-1]):
```
stevekuznetsov: Why do we need the second check?

stevekuznetsov: Also, do not use `is not` -- it will check to see if those two items are literally in the same memory address, not if they are equal.

hack/determine_install_upgrade_version.py
jhadvig: Yeah I was checking particularly for that but haven't found anything. But will double check :+1:

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,55 @@
+import yum
+import sys
+import os.path
+from version_utils import rpm
```
jhadvig: They are... at least I was using them and needed to install the `version_utils` via pip

hack/determine_install_upgrade_version.py
jhadvig: Yeah will add bunch of comments

hack/determine_install_upgrade_version.py
jhadvig: well I thought that it might happen that there will be pkg duplicity when two repos for some reason would contain the same version of pkg. But not 100% sur eif that could occur...

hack/determine_install_upgrade_version.py
stevekuznetsov: Hmm... I think we need to be using what's installed with

```
yum install rpm-python
```

hack/determine_install_upgrade_version.py
stevekuznetsov: Let's worry about it if it happens -- for now I don't think we need to.

hack/determine_install_upgrade_version.py
stevekuznetsov: If not, I think this could be written in such a way to be a little more clear:

```py
major, minor, *rest = input_pkg.version.split('.')
previous_version = '.'.join(major, minor+1, *rest)
```

jhadvig:
@stevekuznetsov uptedate.. tested it with most of the versions of `origin` and `openshift-ansible`
PTAL

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,56 @@
+import yum
+import sys
+import os.path
+import rpmUtils.miscutils as rpmutils
+
+if len(sys.argv) != 2:
+    print "[ERROR] No NEVRA provided!"
+    sys.exit(3)
+input_pkg = sys.argv[1]
+input_pkg_list = rpmutils.splitFilename(input_pkg)
```
stevekuznetsov: break this out

```py
name, version, release, epoch, arch =  rpmutils.splitFilename(input_pkg)
```

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,56 @@
...
+if any(char.isalpha() for char in input_pkg_list[1]):
+	print "[ERROR] Incorrect package nevra format: " + input_pkg
+	sys.exit(2)
+input_pkg_name = str(input_pkg_list[0])
+input_pkg_version_list = map(int, input_pkg_list[1].split("."))
+search_version_list = input_pkg_version_list[:]
+search_version_list[1] = search_version_list[1] -1
+search_version = ".".join(map(str, search_version_list)[:2])
```
stevekuznetsov: still think the readability approach is better 

```py
major, minor, *rest = version.split('.')
search_version = '.'.join(major, minor - 1, *rest)
```

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,56 @@
...
+	if pkg.version.startswith(search_version):
+		if pkg.release.startswith("0"):
+			pre_release_pkgs.append(pkg.vra)
+		else:
+			release_pkgs.append(pkg.vra)
+
+# Write install and upgrade versions of desired package into a file called "package_name"_INSTALL_UPGRADE_VERSION
+# The upgrade version will be written into the file only in case there is more then one version of packages previous
+# minor release. 
+def determine_install_upgrade_versions(version_list):
```
stevekuznetsov:
```py
def write_version_vars(path, install_version, upgrade_version=None):
```

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,55 @@
...
+yb = yum.YumBase()
+# Need make sure that all the packages are listed, cause if excluders are enabled for some reason,
+# listing of origin or docker pkgs will be disabled, based on the enabled excluder.
+yb.conf.disable_excludes = ["all"]
+
+release_pkgs = []
+pre_release_pkgs = []
+
+available_pkgs = yb.doPackageLists('available', patterns=[pkg_name], showdups=True).available
+available_pkgs.sort(key=lambda x: x.version)
```
stevekuznetsov: Shouldn't this use the `rpm.versionCompare()` ?

jhadvig:
@stevekuznetsov updated

stevekuznetsov:
Only thing we're missing is a suite of test cases

jhadvig:
@stevekuznetsov TCs added

hack/determine_install_upgrade_version_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+class DetermineInstallUpgradeVersionTestCase(unittest.TestCase):
+	"Test for `determine_install_upgrade_version.py`"
+	PKG_NAME = "origin"
+	MATCHING_1_5 = ["1.5.0-0.4.el7.x86_64"]
+	MATCHING_1_4 = ["1.4.1-1.el7.x86_64"]
+	MATCHING_1_3 = ["1.3.1-1.el7.x86_64", "1.3.2-3.el7.x86_64"]
+	MATCHING_1_2 = ["1.2.0-0.el7.x86_64", "1.2.5-0.el7.x86_64"]
+
+	def test_get_matching_versions(self):
```
stevekuznetsov: Prefer to hard-code test data in tests rather than in more global scope.

hack/determine_install_upgrade_version_test.py
```diff
@@ -0,0 +1,41 @@
...
+		self.assertEqual(get_matching_versions(self.PKG_NAME, test_pkgs_objs, "1.5"), self.MATCHING_1_5)
```
stevekuznetsov: Let's break these up into specific tests and name them by the intent that they surface. For instance, this test could be:

```py
def test_get_matching_versions_only_prelease(self):
    """ when only pre-release versions exist, they are returned """
    self.assertEqual(
        get_matching_versions('origin', ['1.5.0-0.4.el7.x86_64'], '1.5'),
        ['1.5.0-0.4.el7.x86_64']
    )
```

Then you could pair it with:
```py
def test_get_matching_versions_release_and_prelease(self):
    """ when both release and pre-release versions exist, only release versions are returned """
    self.assertEqual(
        get_matching_versions('origin', ['1.5.0-0.4.el7.x86_64','1.5.0-1.4.el7.x86_64'], '1.5'),
        ['1.5.0-1.4.el7.x86_64']
    )
```

And follow up with the other tests you have written -- when more than one version exists, all are returned, etc.

hack/determine_install_upgrade_version_test.py
```diff
@@ -0,0 +1,41 @@
...
+		self.assertEqual(get_matching_versions(self.PKG_NAME, test_pkgs_objs, "1.4"), self.MATCHING_1_4)
+		self.assertEqual(get_matching_versions(self.PKG_NAME, test_pkgs_objs, "1.3"), self.MATCHING_1_3)
+		self.assertEqual(get_matching_versions(self.PKG_NAME, test_pkgs_objs, "1.2"), self.MATCHING_1_2)
+
+
+if __name__ == '__main__':
+	test_pkgs = ["origin-1.2.0-0.el7.x86_64",
```
stevekuznetsov: Always prefer to place test data as a literal object in the test scope. Otherwise, you start giving tests more than one reason to fail, and that makes it harder to debug failures.

hack/determine_install_upgrade_version_test.py
stevekuznetsov: I guess instead of the strings you'd need literal `TestPackage` objects or `dict`s

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,65 @@
...
+		return [version_list[0],version_list[len(version_list)-1]]
+	return [version_list[0]]
+
+# Write install and upgrade versions of desired package into a file called "package_name"_INSTALL_UPGRADE_VERSION
+# The upgrade version will be written into the file only in case there is more then one version of packages previous
+# minor release. 
+def write_version_vars(version_list):
+	file_name = pkg_name.upper() + "_INSTALL_UPGRADE_VERSION"
+	if os.path.isfile(file_name):
+		os.remove(file_name)
```
kargakis: Wondering if this works since you only import `os.path`...

jhadvig:
@stevekuznetsov @kargakis 
did:
- refactored the TC's  
- added updated `test_branch_origin_extended_conformance_install_update.yml` that will work with new `determine_install_upgrade_version.py`
- removed old `get_available_pkgs.py`

PTAL

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,100 @@ extensions:
...
     - type: "script"
-      title: "check out the correct branch of aos-cd-jobs"
+      title: "install origin GA and update"
       repository: "aos-cd-jobs"
       script: |-
-        git checkout sjb
+        pkg_name=$( cat ./PKG_NAME )
+        if [ $pkg_name = "atomic-openshift" ]
+        then
+          sed -i "s/deployment_type:.*/deployment_type: \"openshift-enterprise\" /g" inventory/group_vars/OSEv3/general.yml
```
stevekuznetsov: pass as `-e` to `ansible-playbook`

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,100 @@ extensions:
...
+        fi
+        sudo python hack/determine_install_upgrade_version.py $( cat ORIGIN_BUILT_VERSION )
+        source ORIGIN_INSTALL_UPGRADE_VERSION
+        source ATOMIC-OPENSHIFT-UTILS_INSTALL_UPGRADE_VERSION
+        ORIGIN_INSTALL_MINOR_VERSION=$( echo ${ORIGIN_INSTALL_VERSION} | awk -F'.' '{print $2}' )
+        echo "=== Installing ${pkg_name}-${ORIGIN_INSTALL_VERSION} ==="
+        ansible-playbook -vv --become --become-user root --connection local -i inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-node/network_manager.yml
+        ansible-playbook -vv --become --become-user root --connection local -i inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml -e openshift_pkg_version="-${ORIGIN_INSTALL_VERSION}" -e enable_excluders=false
```
stevekuznetsov: Why no excluders?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,100 @@ extensions:
...
+        if [[ -v ORIGIN_UPGRADE_VERSION  && -v ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION ]]
+        then
+          echo "=== Updating openshift-ansible to openshift-ansible-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION} ==="
+          sudo yum upgrade -y atomic-openshift-utils-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION}
+          echo "=== Updating ${pkg_name} to ${ORIGIN_UPGRADE_VERSION} ==="
+          ansible-playbook -vv --become --become-user root --connection local --inventory inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_${ORIGIN_INSTALL_MINOR_VERSION}/upgrade.yml -e openshift_pkg_version="-${upgrade_version}" -e oreg_url='openshift/origin-${component}:'"$( echo ${ORIGIN_UPGRADE_VERSION} | awk -F'-' '{print $1}' )"
```
stevekuznetsov: Again if you're putting `NVR`together in the var or whatever, also provide the specific fields that we need -- the Bash should be entirely minimal.

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,100 @@ extensions:
...
+      repository: "aos-cd-jobs"
+      script: |-
+        upgrade_pkgs="ansible atomic-openshift-utils"
+        source ORIGIN_INSTALL_UPGRADE_VERSION
+        ORIGIN_INSTALL_MINOR_VERSION=$( echo ${ORIGIN_INSTALL_VERSION} | awk -F'.' '{print $2}' )
+        if [[ ${ORIGIN_INSTALL_MINOR_VERSION} == 4 ]]
+        then
+          upgrade_pkgs="$( cat ./OPENSHIFT_ANSIBLE_PKGS )"
+        fi
+        sudo yum upgrade -y ${upgrade_pkgs}
```
stevekuznetsov: Why aren't these version-pinned to what we built?

hack/determine_install_upgrade_version_test.py
```diff
@@ -0,0 +1,95 @@
...
+	def __init__(self, version, release, vra):
+		self.version = version
+		self.release = release
+		self.vra = vra
+
+class GetMatchingVersionTestCase(unittest.TestCase):
+	"Test for `determine_install_upgrade_version.py`"
+
+	def setUp(self):
+		test_pkgs = ["origin-1.2.0-0.el7.x86_64",
```
stevekuznetsov: Prefer not to use global state, list the specific things you need in each test as necessary

hack/determine_install_upgrade_version_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+	def test_determine_search_for_openshift_ansible_legacy_version(self):
+		""" when openshift-ansible, which doesnt have different versioning schema is in 3.4 version """
+		self.assertEqual(determine_search_version("openshift-ansible", "3.5.0"), "3.4")
+
+class GetInstallUpgradeVersionTestCase(unittest.TestCase):
+	"Test for `determine_install_upgrade_version.py`"
+
+	def test_get_install_upgrade_with_multiple_matching_release_versions(self):
+		""" when install and upgrade version are picked from multiple matching versions of released versions are """
```
stevekuznetsov: did you forget the end of this comment?

hack/determine_install_upgrade_version_test.py
```diff
@@ -0,0 +1,95 @@
...
+		matching_versions = ["1.2.0-1.el7.x86_64", "1.2.2-1.el7.x86_64", "1.2.5-1.el7.x86_64"]
+		install_upgrade_versions = ["1.2.0-1.el7.x86_64", "1.2.5-1.el7.x86_64"]
+		self.assertEqual(get_install_upgrade_version(matching_versions), install_upgrade_versions)
+
+	def test_get_install_upgrade_with_single_matching_release_version(self):
+		""" when install and upgrade version are picked from single matching versions of released versions are """
```
stevekuznetsov: copy pasta

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,100 @@ extensions:
         name = OpenShift Ansible Release from Local Source
         EOR
         sudo cp ./openshift-ansible-local-release.repo /etc/yum.repos.d
+        basename $(ls ${tito_tmp_dir}/noarch/ | grep atomic-openshift-utils ) .rpm >> /data/src/github.com/openshift/aos-cd-jobs/OPENSHIFT_ANSIBLE_BUILT_VERSION
```
jhadvig: `basename "${tito_tmp_dir}"/noarch/atomic-openshift-utils*.rpm .rpm >> /data/src/github.com/openshift/aos--cd-jobs/OPENSHIFT_ANSIBLE_BUILT_VERSION`

have to add the adidtional `.rpm` argument canuse I need only the name without the extension

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
jhadvig: forgot that we need to test specific verions

hack/determine_install_upgrade_version_test.py
jhadvig: no, the comments should differ from the rest of them.. check them

hack/determine_install_upgrade_version_test.py
jhadvig: but we need to test that right versions of install/upgrade pkgs are picked, from a list of diferent versions and not just releases. For that reason I've created a list of pkg versions that simulates the output that would be provided by yum. For picking the right instlal/upgrade version from the matching versions I've created the `GetInstallUpgradeVersionTestCase`

hack/determine_install_upgrade_version.py
```diff
@@ -0,0 +1,74 @@
...
+	# Cause of the origin version schema change we had to manually change the version just for this 
+	# case where we should look for 1.5.x versions if the built version is origin-3.6.x 
+	if pkg_name == "origin" and major == "3" and minor == "6":
+		major = "1"
+	search_version = '.'.join([major, str(int(minor) - 1)])
+	return search_version
+
+if __name__ == "__main__":
+	if len(sys.argv) != 2:
+	    print "[ERROR] No NEVRA provided!"
```
kargakis: print errors to stderr

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -6,63 +6,19 @@ extensions:
...
       repository: "origin"
       script: |-
-        hack/build-base-images.sh
-        OS_BUILD_ENV_PRESERVE=_output/local hack/env OS_ONLY_BUILD_PLATFORMS='linux/amd64' hack/build-rpm-release.sh
-        sudo systemctl restart docker
-        hack/build-images.sh
-        sed -i 's|go/src|data/src|' _output/local/releases/rpms/origin-local-release.repo
-        sudo cp _output/local/releases/rpms/origin-local-release.repo /etc/yum.repos.d/
+        source hack/lib/build/rpm.sh; source hack/lib/log/output.sh; source hack/common.sh; source hack/lib/build/rpm.sh; source hack/lib/util/text.sh
+        git log -1 --pretty=%h >> /data/src/github.com/openshift/aos-cd-jobs/ORIGIN_COMMIT
```
kargakis: Why is this appending instead of replacing?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
stevekuznetsov: ack

hack/determine_install_upgrade_version_test.py
stevekuznetsov: hm -- very hard to see the difference. I'd suggest one of two things:

 - your method names / test names are verbose and expressive. if you think you captured the entirety of the test reason in the name, get rid of the comment
 - emphasize the differences... so this might read "when only a single matching version is present..."

jhadvig:
@stevekuznetsov check & merge :grimacing:

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,102 @@ extensions:
...
+        then
+          echo "openshift-enterprise" > DEPLOYMENT_TYPE
+          sed -i "s/deployment_type:.*/deployment_type: \"openshift-enterprise\" /g" inventory/group_vars/OSEv3/general.yml
+        fi
+        sudo python hack/determine_install_upgrade_version.py $( cat ORIGIN_BUILT_VERSION )
+        source ORIGIN_INSTALL_UPGRADE_VERSION
+        source ATOMIC-OPENSHIFT-UTILS_INSTALL_UPGRADE_VERSION
+        echo "=== Installing ${pkg_name}-${ORIGIN_INSTALL_VERSION} ==="
+        ansible-playbook -vv --become --become-user root --connection local -i inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-node/network_manager.yml -e deployment_type=$( cat ./DEPLOYMENT_TYPE)
+        ansible-playbook -vv --become --become-user root --connection local -i inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml -e openshift_pkg_version="-${ORIGIN_INSTALL_VERSION}" -e enable_excluders=false -e deployment_type=$( cat ./DEPLOYMENT_TYPE)
```
stevekuznetsov: Why no excluders here?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,102 @@ extensions:
...
+        if [[ -v ORIGIN_UPGRADE_VERSION  && -v ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION ]]
+        then
+          echo "=== Updating openshift-ansible to openshift-ansible-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION} ==="
+          sudo yum upgrade -y atomic-openshift-utils-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_VERSION}
+          echo "=== Updating ${pkg_name} to ${ORIGIN_UPGRADE_VERSION} ==="
+          ansible-playbook -vv --become --become-user root --connection local --inventory inventory/ /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_${ORIGIN_UPGRADE_MINOR_VERSION}/upgrade.yml -e openshift_pkg_version="-${ORIGIN_UPGRADE_VERSION}" -e deployment_type=$( cat ./DEPLOYMENT_TYPE) -e oreg_url='openshift/origin-${component}:'"${ORIGIN_UPGRADE_VERSION_PKG_VERSION}"
```
stevekuznetsov: Nit, but we should see if this works to make multiline shell -- would read better:

```sh
upgrade_playbook="/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_${ORIGIN_UPGRADE_MINOR_VERSION}/upgrade.yml"
ansible-playbook -vv --become           \
                 --become-user root     \
                 --connection local     \
                 --inventory inventory/ \
                  "${upgrade_playbook}" \                   \
                 -e deployment_type=$( cat ./DEPLOYMENT_TYPE)          \
                 -e openshift_pkg_version="-${ORIGIN_UPGRADE_VERSION}" \
                 -e oreg_url='openshift/origin-${component}:'"${ORIGIN_UPGRADE_VERSION_PKG_VERSION}" 
```

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,102 @@ extensions:
...
+    - type: "script"
+      title: "upgrade openshift-ansible to release"
+      repository: "aos-cd-jobs"
+      script: |-
+        upgrade_openshift_ansible_pkgs="ansible atomic-openshift-utils"
+        source ORIGIN_INSTALL_UPGRADE_VERSION
+        source ATOMIC-OPENSHIFT-UTILS_INSTALL_UPGRADE_VERSION
+        if [[ -v ATOMIC_OPENSHIFT_UTILS_UPGRADE_MINOR_VERSION && ${ATOMIC_OPENSHIFT_UTILS_UPGRADE_MINOR_VERSION} == 4 ]]
+        then
+          upgrade_openshift_ansible_pkgs="$( cat ./OPENSHIFT_ANSIBLE_PKGS )"
```
stevekuznetsov: `packages_to_upgrade`?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,102 @@ extensions:
...
+        fi
+        for pkg in "${upgrade_openshift_ansible_pkgs[@]}"
+        do
+          upgrade_openshift_ansible_pkgs="${upgrade_openshift_ansible_pkgs} ${pkg}-${ATOMIC_OPENSHIFT_UTILS_UPGRADE_MINOR_VERSION}"
```
stevekuznetsov: `versioned_packages_to_upgrade`?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,102 @@ extensions:
...
+    - type: "script"
+      title: "install ansible atomic-openshift-utils"
+      repository: "aos-cd-jobs"
+      script: |-
+        pkg_name="origin"
+        git checkout sjb
+        echo $pkg_name > PKG_NAME
+        echo "openshift-ansible openshift-ansible-callback-plugins openshift-ansible-docs openshift-ansible-filter-plugins openshift-ansible-lookup-plugins openshift-ansible-playbooks openshift-ansible-roles" > OPENSHIFT_ANSIBLE_PKGS
+        sudo python hack/determine_install_upgrade_version.py $( cat OPENSHIFT_ANSIBLE_BUILT_VERSION )
+        source ATOMIC-OPENSHIFT-UTILS_INSTALL_UPGRADE_VERSION
```
jhadvig: So you suggest just to print out the versions from the script ?
Well Im not the biggest fan of both versions but I sticked to the idea to do less magic with bash and more with python scripts, thats why I'm writing it directly to file, so in other steps we just source the file and we are good.
But I can go with:
```
sudo hack/determine_install_upgrade_version.py "$( cat OPENSHIFT_ANSIBLE_BUILT_VERSION )" > OPENSHIFT_ANSIBLE_VARS
source OPENSHIFT_ANSIBLE_VARS
```

jhadvig: so ?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +28,102 @@ extensions:
...
+        openshift_ansible_pkgs_versions=""
+        openshift_ansible_pkgs="atomic-openshift-utils"
```
jhadvig: yeah but wanted to be more explicit ... wanna rename it ?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
jhadvig: The excluders are fixed and now they are installed and enabled upon the install process. During the upgrade process they are disabled so no need to disable them now

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
stevekuznetsov: Yeah, writing it to `stdout` makes sense -- it should look very similar in the Python script but make the flow of data a little more clear in the Bash

jhadvig:
PR updated
Testing PR: https://ci.openshift.redhat.com/jenkins/view/All/job/test_branch_origin_extended_conformance_install_update/138/console

@stevekuznetsov PTAL

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +29,138 @@ extensions:
...
+                          /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-node/network_manager.yml \
+                          -e deployment_type=$( cat ./DEPLOYMENT_TYPE)
+        ansible-playbook  -vv                \
+                          --become           \
+                          --become-user root \
+                          --connection local \
+                          -i inventory/      \
+                          /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml \
+                          -e openshift_pkg_version="-${ORIGIN_INSTALL_VERSION}"         \
+                          -e enable_excluders=false                                     \
```
stevekuznetsov: Based on our conversation this morning, we can drop this, right?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +29,138 @@ extensions:
...
+        packages_to_install="atomic-openshift-utils"
+        if [[ ${ATOMIC_OPENSHIFT_UTILS_INSTALL_MINOR_VERSION} == 4 ]]
+        then
+          sudo yum erase -y ansible
+          versioned_packages_to_install="ansible-2.2.0.0"
+          packages_to_install=( $( cat ./OPENSHIFT_ANSIBLE_PKGS ) )
+        fi
+        for pkg in "${packages_to_install[@]}"
+        do
+          versioned_packages_to_install="${versioned_packages_to_install} ${pkg}-${ATOMIC_OPENSHIFT_UTILS_INSTALL_VERSION}"
```
stevekuznetsov: nit, this can be 

```sh
versioned_packages_to_install+=" ${pkg}-${ATOMIC_OPENSHIFT_UTILS_INSTALL_VERSION}"
```

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +29,138 @@ extensions:
...
+      script: |-
+        packages_to_upgrade="ansible atomic-openshift-utils"
+        source ORIGIN_VARS
+        source OPENSHIFT_ANSIBLE_VARS
+        if [[ -v ATOMIC_OPENSHIFT_UTILS_UPGRADE_MINOR_VERSION && ${ATOMIC_OPENSHIFT_UTILS_UPGRADE_MINOR_VERSION} == 4 ]]
+        then
+          packages_to_upgrade="$( cat ./OPENSHIFT_ANSIBLE_PKGS )"
+        fi
+        versioned_packages_to_upgrade=""
+        for pkg in "${upgrade_openshift_ansible_pkgs[@]}"
```
stevekuznetsov: Where is `upgrade_openshift_ansible_pkgs` from?

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +29,138 @@ extensions:
...
+        sed -i 's|go/src|data/src|' _output/local/releases/rpms/origin-local-release.repo
+        sudo cp _output/local/releases/rpms/origin-local-release.repo /etc/yum.repos.d/
+    - type: "script"
+      title: "upgrade openshift-ansible to release"
+      repository: "aos-cd-jobs"
+      script: |-
+        packages_to_upgrade="ansible atomic-openshift-utils"
+        source ORIGIN_VARS
+        source OPENSHIFT_ANSIBLE_VARS
+        if [[ -v ATOMIC_OPENSHIFT_UTILS_UPGRADE_MINOR_VERSION && ${ATOMIC_OPENSHIFT_UTILS_UPGRADE_MINOR_VERSION} == 4 ]]
```
stevekuznetsov: Why do we only want to upgrade these if we're on minor .4 ?

jhadvig:
@stevekuznetsov 
I've added two aditional steps into the playbook:
1. that will print all the installed and updated versions of origin and openshift-ansible using `yum history`
2. that will check that proper images are on the cluster and that they are running

Regarding the `enable_excluders` variable that we spoke about... for now we have to keep them until 1.5-release pkg are re-built, backporting this change https://github.com/openshift/openshift-ansible/pull/3693
After that we can remove the variable. Sorry about the misunderstanding

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -72,30 +29,141 @@ extensions:
...
     - type: "script"
-      title: "install the openshift-ansible release"
+      title: "install ansible atomic-openshift-utils"
+      repository: "aos-cd-jobs"
       script: |-
-        sudo yum install -y atomic-openshift-utils
+        pkg_name="origin"
+        git checkout sjb
+        echo $pkg_name > PKG_NAME
+        wget https://raw.githubusercontent.com/jhadvig/aos-cd-jobs/determin_version_test/hack/determine_install_upgrade_version.py -P hack
```
kargakis: s/jhadvig/openshift/

Or this won't work until it merges?

jhadvig: yeah.. it wont until its merged.. wanted to remove after LGTM

config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -57,45 +135,44 @@ extensions:
...
-      title: "determine the release commit for origin images"
-      repository: "origin"
+      title: "get history of the origin package"
+      repository: "aos-cd-jobs"
       script: |-
-        git log -1 --pretty=%h >> /data/src/github.com/openshift/aos-cd-jobs/ORIGIN_COMMIT
+        sudo yum history info $( cat ./PKG_NAME )
     - type: "script"
-      title: "update origin"
+      title: "get history of the openshift-ansible package"
```
stevekuznetsov: Why not as generated artifacts?

jhadvig:
:champagne:

PR #157, opened 2017-03-27
Move all steps under the the diff check

smunilla:


smunilla:
This should fix the unbound variable error we saw in yesterday's run.

PR #159, opened 2017-03-29
Add `$ETCD_DATA_DIR` to system environment

stevekuznetsov:
We need `etcd` to store it's data in a tmpfs to ensure that the weird
semantics around IO on EC2 EBS don't cause `etcd` to fall over in the
middle of our test suites. We don't control the installation of `etcd`
so we try to influence storage via system-wide environment.

/cc @smarterclayton @liggitt does this seem reasonable?

### DIFF
--- a/config/common/test_cases/origin.yml
+++ b/config/common/test_cases/origin.yml
@@ -10,9 +10,12 @@ actions:
   - type: "script"
     title: "use a ramdisk for etcd"
     script: |-
-      sudo mkdir -p /tmp
-      sudo mount -t tmpfs -o size=2048m tmpfs /tmp
-      sudo restorecon -R /tmp
+      sudo su root
+      mkdir -p /tmp
+      mount -t tmpfs -o size=2048m tmpfs /tmp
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
fixes https://github.com/openshift/aos-cd-jobs/pull/159

config/common/test_cases/origin.yml
```diff
@@ -10,9 +10,11 @@ actions:
   - type: "script"
     title: "use a ramdisk for etcd"
     script: |-
-      sudo mkdir -p /tmp
-      sudo mount -t tmpfs -o size=2048m tmpfs /tmp
-      sudo restorecon -R /tmp
+      sudo su root
+      mkdir -p /tmp/etcd
```
liggitt: why the etcd subdir? doesn't this get shadowed by the mount?

stevekuznetsov: Good point

stevekuznetsov:
@smarterclayton are you OK with this approach?

PR #160, opened 2017-03-30
Add architecture type to the os::build::rpm::format_nvr function

jhadvig:
@stevekuznetsov PTAL


stevekuznetsov:
@jhadvig so you pushed this already, yes?

PR #162, opened 2017-03-31
Adding initial sops

jupierce:
@mwoodson @tdawson 
The starting point for some simple standard operating procedures. Please let me know your thoughts. 

@tdawson Planning to fully automate the stage cut at some point with a Jenkins job, but until then, I wanted something we can reference. I also have "stage-sprint###" for the backup of each stage branch even though we discussed stage-<last ocp version> . The rationale being that it is easier and allows us to correlate  origin/stage-sprint130  and origin/stage-sprint130

### DIFF
--- a/docs/sop/checklist-cluster-upgrade.md
+++ b/docs/sop/checklist-cluster-upgrade.md
@@ -1,0 +1,9 @@
+- [ ] If the cluster has not been upgraded before
+  - [ ] Create the template directory for the cluster (including upgrading the private AWS information for the cluster)
+  - [ ] If the job is an upgrade, log into the cluster and remove any legacy yum repos
+- [ ] Re-vendor appropriate openshift-ansible if it is still not dynamic  (origin/openshift-ansible:stage? release-1.5?)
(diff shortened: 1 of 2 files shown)


tdawson:
Are we missing openshift-ansible on these, or will that be done by the other group?

stevekuznetsov:
Do we want to try to write out the SOP as shell?

```sh
sprint=123
for repo in origin origin-web-console ose openshift-ansible online; do
    git "clone git@github.com:openshift/${repo}.git"
    pushd "${repo}"
    git checkout stage
    git checkout -b "stage-${sprint}"
    git reset --hard master
    for ref in master stage "stage-${sprint}"; do
        git push --force origin "${ref}"
    done
    popd
done
```

jupierce:
@stevekuznetsov If there is truly nothing else to it, then yes, we should just write a Jenkins job. This is mean to find and document any lingering details before we do so (e.g. Scott seemed interested when you talked about stage being a tracking branch for release-X.Y).

jupierce:
@tdawson Thanks for catching that. I've added openshift-ansible.

PR #166, opened 2017-04-03
Skuznets/deployment type

stevekuznetsov:
@jhadvig FYI -- inventory is shared we need to be careful


jhadvig:
well we can still pass custom variables to the playbook directly

stevekuznetsov:
@jhadvig yeah, that's how I fixed it in https://github.com/openshift/aos-cd-jobs/pull/166/commits/1e4aae20245d305d4520b49967f8a5c2752ef1d1

jhadvig:
@stevekuznetsov :+1:

PR #167, opened 2017-04-03
Bump Golang on the AMI to 1.7, install from CentOS PaaS SIG

stevekuznetsov:

/cc @bparees we can source it from CentOS so we should switch `base_ami` to do this too

### DIFF
--- a/config/test_cases/ami_build_origin_int_rhel_base.yml
+++ b/config/test_cases/ami_build_origin_int_rhel_base.yml
@@ -13,7 +13,7 @@ actions:
   - type: "host_script"
     title: "install golang"
     script: |-
-      oct prepare golang --version '1.6.3' --repo 'oso-rhui-rhel-server-releases-optional'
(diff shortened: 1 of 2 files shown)


bparees:
I prefer us staying on a path where the golang version in the base ami is irrelevant.

stevekuznetsov:
I agree, but I'm not ready to do the potentially unbounded amount of work to figure out who depends on it being there

bparees:
sure, i'm just saying we should not propagate this pattern to the base ami job.

bparees:
the work to find out who depends on it being there is:  is base_ami broken?  is it because someone wants golang1.7 and doesn't have it?  done.

stevekuznetsov:
That presumes that a user of `golang` package on `base_ami` downstream jobs has some means by which to block the publishing of new `base_ami` versions ... which they do not. Especially if they are evangelists, qe, etc.

PR #177, opened 2017-04-10
Add the `tox` unit test and linting job for OpenShift Ansible

stevekuznetsov:

/cc @sdodson this just runs `tox` in your repo dir -- not quite what the other job does but @rhcarvalho said this was the correct way to do it?

### DIFF
--- a/config/common/test_suites/openshift_ansible.yml
+++ b/config/common/test_suites/openshift_ansible.yml
@@ -1,4 +1,5 @@
 ---
 children:
+  - "test_pull_request_openshift_ansible_tox"
   - "test_pull_request_openshift_ansible_extended_conformance_install_with_status_check"
   - "test_pull_request_openshift_ansible_extended_conformance_install_update"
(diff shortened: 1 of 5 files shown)


rhcarvalho:
A discussion to be had is that perhaps we can use just Travis for that (as we already do), instead of running the same tests in both Travis and our own infrastructure.

stevekuznetsov:
We have limited Travis slots for the organization, right? Why not just remove the Travis job and let it work like the rest of the CI>?

rhcarvalho:
> We have limited Travis slots for the organization, right? Why not just remove the Travis job and let it work like the rest of the CI>?

The "rest of CI" is source of constant frustration... a black box only a few people can handle.

I don't see what discussion is to be had if even though I mentioned on IRC and in a comment here you go ahead and self-merge the PR without anybody else's input... why even bother creating a PR then?!

stevekuznetsov:
> The "rest of CI" is source of constant frustration... a black box only a few people can handle.

Working to make it better is an infinitely better option than using Travis.

> I don't see what discussion is to be had if even though I mentioned on IRC and in a comment here you go ahead and self-merge the PR without anybody else's input... why even bother creating a PR then?!

For visibility ... your comment assumes that there was a question of whether or not we were going to use Jenkins for our CI. Unfortunately I don't think that level of conversation is up for debate.

PR #182, opened 2017-04-12
openshift ansible integration tests

sosiouxme:
add openshift-ansible integration tests on PRs

### DIFF
--- a/config/test_cases/test_pull_request_openshift_ansible_tox.yml
+++ b/config/test_cases/test_pull_request_openshift_ansible_tox.yml
@@ -18,6 +18,20 @@ actions:
     repository: "openshift-ansible"
     script: |-
       tox
+  - type: "script"
+    title: "build images for integration tests"
+    repository: "openshift-ansible"
+    script: |-
+      if [ -x test/integration/build-images.sh ]; then
+        test/integration/build-images.sh
+      fi
+  - type: "script"
+    title: "run integration tests"
+    repository: "openshift-ansible"
+    script: |-
+      if [ -x test/integration/run-tests.sh ]; then
+        test/integration/run-tests.sh
+      fi
 post_actions: []
 artifacts: []
 generated_artifacts: {}
--- a/generated/test_pull_request_openshift_ansible_tox.xml
+++ b/generated/test_pull_request_openshift_ansible_tox.xml
@@ -112,6 +112,34 @@ tox
 EOF
 } | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel</command>
         </hudson.tasks.Shell>
+        <hudson.tasks.Shell>
+          <command>#!/bin/bash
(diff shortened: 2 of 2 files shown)


sosiouxme:
@stevekuznetsov once you get this in we can begin testing with https://github.com/openshift/openshift-ansible/pull/3816

config/test_cases/test_pull_request_openshift_ansible_tox.yml
```diff
@@ -18,6 +18,20 @@ actions:
     repository: "openshift-ansible"
     script: |-
       tox
+  - type: "script"
+    title: "build images for integration tests"
+    repository: "openshift-ansible"
+    script: |-
+      if [ -x test/integration/build-images.sh ]; then
```
stevekuznetsov: Will this fail as expected if the file doesn't exist? `-x` just checks for the user's execute bit, right?

sosiouxme: It also checks for existence.
http://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html
So it silently passes if the file is missing.

config/test_cases/test_pull_request_openshift_ansible_tox.yml
```diff
@@ -18,6 +18,20 @@ actions:
     repository: "openshift-ansible"
     script: |-
       tox
+  - type: "script"
```
rhcarvalho: This should not be in the same job as tox, should be an independent job.

sosiouxme: I don't think it's the long-term location we want this, but the tox job is pretty short and @stevekuznetsov thought it would be easiest to just start out by extending it rather than create a new job. As integration tests get unwieldy, they could be moved to a new job(s).

Note that there's a build run for every test... right now builds are pretty quick, but I could see them getting a lot longer in the future. We didn't do the builds separately at this time as there's no really good place to push them to (private registry hopefully coming "soon").

rhcarvalho: I would put them separate for logical separation.

And for now (and probably upcoming weeks) it would be nice if we could trigger the job independent of the other job -- we don't want bugs and experimentation here to mess up with PRs from other people.

rhcarvalho: I propose we start with an opt-in flow. Is there an easy way to parse a comment like [test:integration] or similar to trigger a job? IOW, is the logic that parses [merge] easy to extend?

stevekuznetsov: We do not want to pay the extra money to start two VMs on EC2 and pay by the hour there -- no reason not to run this here. If these jobs do not pass today and are not ready to be in the queue, we will have to make this a separate job. However, if they are blocking today there is no scenario under which breaking changes will get into master due to experimentation, so I see no reason not to include them.

sosiouxme: I'd like to be able to start running these tests. How about we merge this and just agree not to actually merge my PR supplying the tests until we've either established confidence it won't break the build or decided to do something else.

stevekuznetsov:
@sosiouxme can you run `./generate.sh` and check that in, then I will merge

sosiouxme:
@stevekuznetsov done.

PR #183, opened 2017-04-12
Check the right set of statuses based on the branch

jhadvig:
So now the set of the blocking statuses will be determined based on the branch against which the PR is sent.
Based on the sonversation with @sdodson, master tests current devel (3.6) and last release (3.5). Other branches only test their respective versions. So when testing against master 3.6 and 3.5 blocking statuses shall be checked.

@stevekuznetsov PTAL

### DIFF
--- a/check-pull-request-status.py
+++ b/check-pull-request-status.py
@@ -6,12 +6,14 @@ import sys
 from yaml import load
 from github3 import login
 
-if len(sys.argv) != 3:
-    print("USAGE: {} PULL_REQUEST_ID CREDENTIAL_STORE".format(sys.argv[0]))
+if len(sys.argv) != 5:
+    print("USAGE: {} PULL_REQUEST_ID OPENSHIFT_ANSIBLE_TARGET_BRANCH CREDENTIAL_STORE STATUS_CONFIG".format(sys.argv[0]))
     exit(1)
 
 pull_request_id = int(sys.argv[1])
-credential_file = sys.argv[2]
+target_branch = sys.argv[2]
+credential_file = sys.argv[3]
+status_config_file = sys.argv[4]
 
 with open(credential_file) as credential_store:
     credentials = load(credential_store)
@@ -23,13 +25,14 @@ client = login(token=oauth_token)
 pull_request = client.pull_request(owner='openshift', repository='openshift-ansible', number=pull_request_id)
 pull_request_statuses = client._json(client._get(pull_request.statuses_url.replace('statuses', 'status')), 200)['statuses']
 
-blocking_statuses = [
-    'aos-ci-jenkins/OS_3.5_NOT_containerized',
-    'aos-ci-jenkins/OS_3.5_NOT_containerized_e2e_tests',
-    'aos-ci-jenkins/OS_3.5_containerized',
-    'aos-ci-jenkins/OS_3.5_containerized_e2e_tests',
-    'continuous-integration/travis-ci/pr'
-]
+blocking_statuses = []
+
+with open(status_config_file) as status_config:
+    config = load(status_config)
+    if target_branch not in config:
+        print("[ERROR] No GitHub PR statuses have been configured for branch '{}'.".format(target_branch))
+    blocking_statuses.extend(config[target_branch])
+    blocking_statuses.extend(config["common_status"])
 
 found_statuses = {}
 
--- a/config/test_cases/test_pull_request_openshift_ansible_extended_conformance_install_with_status_check.yml
+++ b/config/test_cases/test_pull_request_openshift_ansible_extended_conformance_install_with_status_check.yml
@@ -7,8 +7,12 @@ overrides:
       type: "pull_request"
     - name: "origin"
 extensions:
+  sync_repos:
+    - name: "aos-cd-jobs"
   actions:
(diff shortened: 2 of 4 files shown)


check-pull-request-status.py
```diff
@@ -23,14 +24,34 @@
...
     'continuous-integration/travis-ci/pr'
 ]
 
+# Master tests current devel (3.6) and last release (3.5)
+if target_branch == "master":
+    for version in versions[:2]:
+        for suffix in suffix_variations:
+            blocking_statuses.append(prefix + version + suffix)
+# Other branches eg: release-1.5 only test their respective versions
+elif target_branch.startswith("release-"):
```
stevekuznetsov: This presumes that the statuses exist at the same moment that we break out a new branch and run a `[test]` against it. I don't think that generally this is true. I think a better approach would be to accept a external config that would have a static mapping of branch to tests. This would allow for us to be certain that we are not changing what tests we look for unless someone explicitly adds or removes one from the config.

jhadvig:
@stevekuznetsov PTAL

check-pull-request-status.py
```diff
@@ -6,12 +6,14 @@
 from yaml import load
 from github3 import login
 
-if len(sys.argv) != 3:
-    print("USAGE: {} PULL_REQUEST_ID CREDENTIAL_STORE".format(sys.argv[0]))
+if len(sys.argv) != 5:
+    print("USAGE: {} PULL_REQUEST_ID OPENSHIFT_ANSIBLE_TARGET_BRANCH CREDENTIAL_STORE".format(sys.argv[0]))
```
stevekuznetsov: update doc text for config file

check-pull-request-status.py
```diff
@@ -23,13 +25,21 @@
...
+
+with open(status_config_file) as status_config:
+    config = load(status_config)
+    versions = []
+    if target_branch == "master":
+        versions.extend(config["master"])
+    elif target_branch.startswith("release-"):
+        versions.append(target_branch)
+    else:
+        print("[ERROR] Can't determine set of statuses that shall be checked for the '{}' branch.".format(target_branch))
```
stevekuznetsov: I can see why this approach may save some space in the YAML config, but I think it may be better to just have an entirely simple mapping:

```yml
master:
  - test1
  - test2
otherbranch:
  - test1
  - test3
```

We will change the file very rarely and from that I think it's fine for us to repeat the common stuff as much as necessary. Also, when we get branches like `staging` and `staging-sprint123` this current code will break but if it's just a simple mapping we can add all the config we need.

jhadvig:
@stevekuznetsov updated

check-pull-request-status.py
```diff
@@ -23,13 +25,12 @@
...
-    'aos-ci-jenkins/OS_3.5_NOT_containerized_e2e_tests',
-    'aos-ci-jenkins/OS_3.5_containerized',
-    'aos-ci-jenkins/OS_3.5_containerized_e2e_tests',
-    'continuous-integration/travis-ci/pr'
-]
+blocking_statuses = []
+
+with open(status_config_file) as status_config:
+    config = load(status_config)
+    blocking_statuses.extend(config[target_branch])
```
stevekuznetsov: Tiny tiny nit: add an error message:

```py
if target_branch not in config:
    print '[ERROR] No GitHub PR statuses have been configured for branch {}.'.format(target_branch)
```

jhadvig:
@stevekuznetsov nit addressed

stevekuznetsov:
@jhadvig OK! Let me know when and I can show you how to change the Rosie-bot config

sdodson:
This is causing

```
########## STARTING STAGE: CHECK OTHER JOB STATUSES ##########
+ [[ -s /var/lib/jenkins/jobs/test_pull_request_openshift_ansible_extended_conformance_install_with_status_check/workspace/activate ]]
+ source /var/lib/jenkins/jobs/test_pull_request_openshift_ansible_extended_conformance_install_with_status_check/workspace/activate
++ export VIRTUAL_ENV=/var/lib/jenkins/origin-ci-tool/aa51983c42af91fd00b2a39c162b0b9017148462
++ VIRTUAL_ENV=/var/lib/jenkins/origin-ci-tool/aa51983c42af91fd00b2a39c162b0b9017148462
++ export PATH=/var/lib/jenkins/origin-ci-tool/aa51983c42af91fd00b2a39c162b0b9017148462/bin:/sbin:/usr/sbin:/bin:/usr/bin
++ PATH=/var/lib/jenkins/origin-ci-tool/aa51983c42af91fd00b2a39c162b0b9017148462/bin:/sbin:/usr/sbin:/bin:/usr/bin
++ unset PYTHON_HOME
++ export OCT_CONFIG_HOME=/var/lib/jenkins/jobs/test_pull_request_openshift_ansible_extended_conformance_install_with_status_check/workspace/.config
++ OCT_CONFIG_HOME=/var/lib/jenkins/jobs/test_pull_request_openshift_ansible_extended_conformance_install_with_status_check/workspace/.config
+ git checkout sjb
fatal: Not a git repository (or any of the parent directories): .git
++ export status=FAILURE
++ status=FAILURE
+ set +o xtrace
```

PR #185, opened 2017-04-12
Set CPU and Memory accounting in the base AMI

stevekuznetsov:

/cc @derekwaynecarr FYI
/cc @sdodson PTAL

### DIFF
--- a/config/test_cases/ami_build_origin_int_rhel_base.yml
+++ b/config/test_cases/ami_build_origin_int_rhel_base.yml
@@ -22,6 +22,17 @@ actions:
     title: "prepare repositories"
     script: |-
       oct prepare repositories
+  - type: "script"
+    title: "install system accounting rules"
+    script: |-
+      cat <<EOF >origin-ami-accounting.conf
+      [Manager]
+      DefaultCPUAccounting=yes
+      DefaultMemoryAccounting=yes
+      EOF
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Removed conditional on the `daemon-reexec` ... works AFAICT:

```console
$ cat <<EOF >origin-ami-accounting.conf
> [Manager]
> DefaultCPUAccounting=yes
> DefaultMemoryAccounting=yes
> EOF
$ mkdir -p /etc/systemd/system.conf.d/
$ mv origin-ami-accounting.conf /etc/systemd/system.conf.d/
$ systemctl daemon-reexec
$ systemctl show docker.service --property CPUAccounting --property MemoryAccounting
CPUAccounting=yes
MemoryAccounting=yes
```

sdodson:
Yeah, if you're doing this in the base image build and you're sure that the host will be rebooted you don't need any of the systemctl daemon-reexec stuff

stevekuznetsov:
> you don't need any of the systemctl daemon-reexec stuff

We run tests on the AMI after we mint it but without a reboot AFAICT(???) so I'll keep it in, can't hurt.

stevekuznetsov:
https://ci.openshift.redhat.com/jenkins/job/ami_build_origin_int_rhel_base/52/ running with these changes

PR #190, opened 2017-04-17
Archive binary journal

sdodson:
We need to create the persistent journal directory first, so as one of the last steps when building your AMI do `mkdir -p /var/log/journal`. I didn't see where you were building the AMI.

Then, to read the contents of this archive extract it and `journalctl -D .`


sdodson:
I'm not sure how portable this format is, but I'd assume a fedora journalctl can read a rhel7 journal.

PR #193, opened 2017-04-19
Install and upgrade versions should be different

jhadvig:
It might happen the `determine_install_upgrade_versions.py` will determine the same versions of the install and upgrade pkg. This happens when there are the same versions of pkgs but in different repositories.
As an example we can use [this install_upgrade test](https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_extended_conformance_install_update/193/consoleFull#197040384158b6e51eb7608a5981914356), where when you search for `source OPENSHIFT_ANSIBLE_VARS` you will see that the install and upgrade versions are the same. 
@stevekuznetsov PTAL

### DIFF
--- a/hack/determine_install_upgrade_version.py
+++ b/hack/determine_install_upgrade_version.py
@@ -25,8 +25,9 @@ def get_matching_versions(input_pkg_name, available_pkgs, search_version):
 
 # Get only install(first) and upgrade(last) versions from version list
 def get_install_upgrade_version(version_list):
-	if len(version_list) > 1: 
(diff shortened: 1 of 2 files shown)


hack/determine_install_upgrade_version.py
```diff
@@ -25,8 +25,9 @@ def get_matching_versions(input_pkg_name, available_pkgs, search_version):
 
 # Get only install(first) and upgrade(last) versions from version list
 def get_install_upgrade_version(version_list):
```
stevekuznetsov: `get_matching_versions()` should return a de-duped list or a set

PR #194, opened 2017-04-19
Add sjb/ prefix for sjb files and scripts

jhadvig:
Since we move to multibranch directory we have add the `sjb/` prefix for the files and scripts that we are using from that directory.
@stevekuznetsov I guess more fixes will be needed ,but now aware of any yet. Thoughts ?

@bbguimaraes @jupierce fyi


stevekuznetsov:
@jhadvig i think i handled this in a commit

jhadvig:
closing this one since @stevekuznetsov fixed the issue in https://github.com/openshift/aos-cd-jobs/commit/e380644c18207c8c92ffb3a89bc4d44c59e92a7c

PR #198, opened 2017-04-20
Remove duplicate packages

jhadvig:
Rules out the possibility that install GA and upgrade to latest will point to the same version.
Additional TC's added.
@stevekuznetsov PTAL

### DIFF
--- a/sjb/hack/determine_install_upgrade_version.py
+++ b/sjb/hack/determine_install_upgrade_version.py
@@ -3,6 +3,16 @@ import sys
 import os.path
 import rpmUtils.miscutils as rpmutils
 
+# Remove duplicate packages from different repositories
+def remove_duplicate_pkgs(available_pkgs):
+	available_pkgs_tups = []
+	uniq_available_pkgs = []
+	for pkg in available_pkgs:
+		if pkg.pkgtup not in available_pkgs_tups:
+			available_pkgs_tups.append(pkg.pkgtup)
+			uniq_available_pkgs.append(pkg)
+	return uniq_available_pkgs
+
 # Get all the matching `MAJOR.MINOR` versions of provided input package nerva
 # Returned list of package versions is in `'pkg_version'-'pkg_release'` format ,eg: `1.3.0-3.el7`
 def get_matching_versions(input_pkg_name, available_pkgs, search_version):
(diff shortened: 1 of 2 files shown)


sjb/hack/determine_install_upgrade_version.py
```diff
@@ -3,6 +3,16 @@
 import os.path
 import rpmUtils.miscutils as rpmutils
 
+# Remove duplicate packages from different repositories
+def remove_duplicate_pkgs(available_pkgs):
```
stevekuznetsov: Will it work just to use a `set()` here?

jhadvig: unfortunately no. The thing is that the pkgs that point to the same verions are pointing to a different repo, that why `set()` wont treat them as dups.

stevekuznetsov: ack

sjb/hack/determine_install_upgrade_version.py
```diff
@@ -80,8 +90,9 @@ def determine_search_version(pkg_name, pkg_version):
 	yb.conf.disable_excludes = ["all"]
```
stevekuznetsov: Unrelated to this PR but we should check to see if we can remove this -- was this due to some excluder bug with them being on all the time or are we always running this script when they are on and we need to keep it?

stevekuznetsov:
LGTM

sjb/hack/determine_install_upgrade_version.py
```diff
@@ -80,8 +90,9 @@ def determine_search_version(pkg_name, pkg_version):
...
 	sys.stdout = old_stdout
 	available_pkgs = rpmutils.unique(yb.doPackageLists('available', patterns=[pkg_name], showdups=True).available)
-	available_pkgs.sort(lambda x, y: rpmutils.compareEVR((x.epoch, x.version, x.release), (y.epoch, y.version, y.release)))
 	search_version = determine_search_version(pkg_name, pkg_version)
+	available_pkgs = remove_duplicite_pkgs(available_pkgs)
```
smarterclayton: typo, `duplicate`

smarterclayton:
One comment about typo

jhadvig:
@smarterclayton comment addressed

smarterclayton:
LGTM as well

PR #200, opened 2017-04-20
Update rebase instructions for conflicts

kargakis:

@tdawson @jupierce @stevekuznetsov an update on the instructions in case of conflicts. I tried to make to automate the process with a script but we depend on envs that are set prior and post-rebase so I don't think we can  script this out.

### DIFF
--- a/docs/proposals/ose-rebase.md
+++ b/docs/proposals/ose-rebase.md
@@ -192,18 +192,13 @@ arise, they will need to be manually resolved as it already is the case today. I
 we will extend the conflict resolution mechanism in Jenkins to send an e-mail to the author
 and committer of the conflicting commits.
 
-A nightly build of OSE will follow the steps below in order to rebase on top of Origin:
+The steps below will be followed to rebase OSE on top of Origin.
 ```sh
 git clone git@github.com:openshift/ose.git
 pushd ose
 git remote add upstream git@github.com:openshift/origin.git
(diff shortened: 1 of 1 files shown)


docs/proposals/ose-rebase.md
```diff
@@ -256,12 +253,13 @@ git push origin ${CURRENT_BRANCH} -f refs/tags/${last_tag}
...
-resolved manually ie. master is updated with the conflict resolved and a new build is
-kicked off. This is not going to be possible with the suggested process because the
-changelog for tito needs to be constructed based on the previous and the current HEAD of
-Origin that got pulled in with the rebase. All the steps above need to run manually and
-then the build can be restarted. This also means that the build process is idempotent and
-it will skip the above part.
+resolved manually by simply using `git merge`, resolving any conflicts, and then kick off
+a new build. The process of updating master with resolved conflicts needs to slightly
+change because the changelog for tito needs to be constructed based on the previous and
+the current HEAD of Origin that got pulled in with `git rebase`. All the steps above need to
```
stevekuznetsov: So the "downside" here we are trying to work around is that one rebase gets two changelog entries?

kargakis: The "downside" is that we can't simply resolve the conflicts and rerun the job as it was the case with `git merge`.

PR #204, opened 2017-04-20
Disable exluders in inventory

jhadvig:
@sdodson here Im desabling the excluders globally for intall/upgrade plyabooks. I guess this should be enough. Thoughts ?

ccing @stevekuznetsov

### DIFF
--- a/sjb/inventory/group_vars/OSEv3/general.yml
+++ b/sjb/inventory/group_vars/OSEv3/general.yml
@@ -12,3 +12,4 @@ openshift_master_identity_providers:
     challenge: "true"
     kind: "AllowAllPasswordIdentityProvider"
 ansible_ssh_user: "ec2-user"
+enable_excluders: "false"


sdodson:
Yes, assuming that's used for all playbooks that should work.

stevekuznetsov:
If we are shipping with them on by default I do not see why this is something we want to do.

sdodson:
Until we have 1.5 builds without broken excluders I think this is the right thing to do.

jhadvig:
Yeah, that was my point in https://github.com/openshift/origin/issues/13814#issuecomment-295644998 
But I though we want to get rid of flakes that are cause by issue in the openshift-ansible-3.5.46

stevekuznetsov:
I confirmed with @smarterclayton yesterday that we can probably drop the GA install for good -- we get 80% of the benefit from last install + upgrade, and we don't get the whole host of flakes from it. Haven't had time to communicate that to @jhadvig  yet, sorry. I'd rather do that first, if that doesn't fix it disable them ONLY for the install, if that doesn't work it may be reasonable to disable globally, but that needs to be a last resort.

jhadvig:
@stevekuznetsov so just to sum it up... we are going to install the latest release of previous verions and then upgrade to the one we build ? is that it ?
So not:  install GA -> upgrade latest -> upgrade to build
But:  install latest -> upgrade to build

Also you said `probably` ... how much probable is that ? since we will need to modify few things(jobs/scripts/...) :smile:

stevekuznetsov:
Yeah, it's going to be a bunch of work for you... I would confirm this morning with @smarterclayton but that "probably" was 99% as I understand it. We should see if it's _ever_ going to be a long term plan to do GA -> latest -> upgrade, or if it's never going to be, since that will color how you change up the scripts.

jhadvig:
Ok so shall I start to work on it ? or creating an issue should be enough for now ?

stevekuznetsov:
Let's wrap up the card for this sprint -- email, etc -- then we can start a WIP for the changes/

stevekuznetsov:
Alternatively if the bug is severe (happening on more than 1/10 jobs) let's do this change for now and get them back in later. @sdodson your call on frequency

sdodson:
Looks like about 7/12 test_pull_request_origin jobs since 10pm last night failed in this way.

sdodson:
2/2 merge_pull_request_origin failed in this way too

stevekuznetsov:
OK, let's get this in and clean up the rest of it later.

PR #213, opened 2017-04-24
Gather logs prototype

jupierce:
@themurph Might be a starting point.


themurph:
👍  Thanks for sending this along.

PR #216, opened 2017-04-25
Stop logging useless information about the system in AMI job

stevekuznetsov:


kargakis:
Is the expectation here that developers are not going to debug failures in the AMI? Because to me it seems that these are useful information to know.

stevekuznetsov:
It's captured in the artifacts section, this predated that section and is no longer necessary as a build step.

kargakis:
ack

PR #229, opened 2017-04-28
Updated script to only pull each service log for the listed services.…

themurph:
… Also variablized as much as possible and documented.


jupierce:
lgtm

PR #231, opened 2017-05-02
Return verbose status info and events

kargakis:
@themurph @jupierce 


jupierce:
Thanks @kargakis  .

PR #237, opened 2017-05-05
Containerize release builds

kargakis:
@stevekuznetsov I managed to get around https://github.com/openshift/origin/pull/13642#issuecomment-299178579 by running rpmbuild inside a container. See https://ci.openshift.redhat.com/jenkins/view/All/job/test_kargakis_debug/1/console where I reproduce the failure and https://ci.openshift.redhat.com/jenkins/view/All/job/test_kargakis_debug/4/ where I move the build into a container and the test passes.

### DIFF
--- a/sjb/config/common/test_cases/origin_release.yml
+++ b/sjb/config/common/test_cases/origin_release.yml
@@ -8,4 +8,6 @@ extensions:
       script: |-
         export OS_BUILD_IMAGE_ARGS=''
         hack/build-base-images.sh
-        make release
+        OS_BUILD_ENV_PRESERVE="_output/local" hack/env make build-rpms
+        hack/build-images.sh
+        hack/extract-release.sh
--- a/sjb/generated/push_origin_release.xml
+++ b/sjb/generated/push_origin_release.xml
@@ -108,7 +108,9 @@ set -o errexit -o nounset -o pipefail -o xtrace
 cd &#34;\${GOPATH}/src/github.com/openshift/origin&#34;
 export OS_BUILD_IMAGE_ARGS=&#39;&#39;
 hack/build-base-images.sh
-make release
+OS_BUILD_ENV_PRESERVE=&#34;_output/local&#34; hack/env make build-rpms
+hack/build-images.sh
+hack/extract-release.sh
 SCRIPT
 chmod +x &#34;${script}&#34;
 scp -F ./.config/origin-ci-tool/inventory/.ssh_config &#34;${script}&#34; openshiftdevel:&#34;${script}&#34;
--- a/sjb/generated/test_branch_origin_extended.xml
+++ b/sjb/generated/test_branch_origin_extended.xml
@@ -118,7 +118,9 @@ set -o errexit -o nounset -o pipefail -o xtrace
 cd &#34;\${GOPATH}/src/github.com/openshift/origin&#34;
 export OS_BUILD_IMAGE_ARGS=&#39;&#39;
 hack/build-base-images.sh
-make release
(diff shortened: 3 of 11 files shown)


stevekuznetsov:
Changes seem reasonable but I am worried about breaking `make release`

kargakis:
@stevekuznetsov if the goal is to transition to rpm installs inside the containers isn't this essentially `make release` after the Origin PR is merged?

stevekuznetsov:
I thought `make release` was broken in your PR and this was a workaround to getting tests to run for it?

kargakis:
Switching `make release` to use `rpmbuild` seems that breaks `make release` in case it is run atop the base ami which is what our integration and networking suites are doing. Moving rpmbuild to run into a container seems that it solves the issue, I am still trying to understand why the original failure and if there is another way of fixing this.

kargakis:
I just found out that we are doing exactly the same thing in the `build` ami that is based on `base`.
https://github.com/openshift/aos-cd-jobs/blob/20f7e57fccc8960a664ae39a925fe4b913c1e14a/sjb/config/test_cases/ami_build_origin_int_rhel_build.yml#L19

kargakis:
@smarterclayton @stevekuznetsov can we move forward with this? @smarterclayton do you think static compilation[0] will solve the issue I was seeing[1]?

[0] https://github.com/openshift/origin/pull/14127
[1]
```
++ Building go targets for linux/amd64: images/pod cmd/dockerregistry cmd/gitserver pkg/sdn/plugin/sdn-cni-plugin vendor/github.com/containernetworking/cni/plugins/ipam/host-local vendor/github.com/containernetworking/cni/plugins/main/loopback examples/hello-openshift
++ Building go targets for linux/amd64: cmd/openshift cmd/oc
# github.com/openshift/origin/cmd/oc
/usr/lib/golang/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
collect2: error: ld returned 1 exit status

# github.com/openshift/origin/cmd/openshift
/usr/lib/golang/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
collect2: error: ld returned 1 exit status
```

stevekuznetsov:
> can we move forward with this?

Yes, why not?

smarterclayton:
Static compilation doesn't seem related to that, but not sure

kargakis:
> Yes, why not?

Do you want me to update the jobs then and merge this one then? I don't think there is another way to move forward with the rpm installs.

stevekuznetsov:
@kargakis if we are happy with the state of this PR and it works, yes let's do it

sjb/config/common/test_cases/origin_release.yml
```diff
@@ -8,4 +8,6 @@ extensions:
       script: |-
         export OS_BUILD_IMAGE_ARGS=''
         hack/build-base-images.sh
-        make release
\ No newline at end of file
+        OS_BUILD_ENV_PRESERVE="_output/local" hack/env make build-rpms
+        hack/build-images.sh
```
stevekuznetsov: We decided not to use `hack/env make release` because it interacted poorly with this script, right?

kargakis: The current `make release` cannot run inside `hack/env` because it uses `hack/env` already I think. `make release` with `build-rpms` works fine inside `hack/env`.

kargakis: build-images also works fine inside hack/env

stevekuznetsov: Oh of course. Ack ack ack

kargakis:
If there are no objections, I would like to roll this out tomorrow.

stevekuznetsov:
None from me. Let's do it.

PR #238, opened 2017-05-05
Make use of the `build` stage for jobs that don't need to compile

stevekuznetsov:
Jobs that run on a timer can use the latest available `build` image
stage for their EC2 VM and not incur the cost of compiling Origin
binaries and building images, etc.


kargakis:
What happens now if I open a PR with a new feature accompanied with a test in the extended build suite?

stevekuznetsov:
You run it using `[testextended][extended: core(\[builds\])]` which triggers a different job -- the ones that got updated here are daily jobs only.

kargakis:
ack

PR #242, opened 2017-05-08
Improving emails and mode names

jupierce:
@tdawson This started by trying to clean up the success email (the wrong puddle URLs were being sent out for older levels). Ended up moving OSE_MASTER to the Jenkinsfile and changing the BUILD_MODE options for clarity. 

I tried to factor out OSE_MASTER_BRANCHED with "enterprise:pre-release" build mode. Please let me know your thoughts.


tdawson:
LGTM
I like it.  It makes things easier to understand instead of all these different types of masters.

PR #207, opened 2017-04-20
Remove extended_conformance_install from testing origin PR

jhadvig:
@stevekuznetsov @smarterclayton PTAL
as you spoke I've removed extended_conformance_install from testing origins PR.

PR #245, opened 2017-05-09
Install-Upgrade job should do single upgrade

jhadvig:
Got rid of the first upgrade in the install_upgrade job so now we install the latest instead of the GA of the previous version of origin
@stevekuznetsov PTAL

This PR is prior to https://github.com/openshift/aos-cd-jobs/pull/207

### DIFF
--- a/sjb/hack/determine_install_upgrade_version.py
+++ b/sjb/hack/determine_install_upgrade_version.py
@@ -33,11 +33,9 @@ def get_matching_versions(input_pkg_name, available_pkgs, search_version):
 		print("[ERROR] Can not determine install and upgrade version for the `" + input_pkg_name + "` package", sys.stderr)
 		sys.exit(1)
 
-# Get only install(first) and upgrade(last) versions from version list
-def get_install_upgrade_version(version_list):
-	if len(version_list) > 1: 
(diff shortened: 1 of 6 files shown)


sjb/hack/determine_install_upgrade_version.py
```diff
@@ -34,10 +34,8 @@ def get_matching_versions(input_pkg_name, available_pkgs, search_version):
 		sys.exit(1)
 
 # Get only install(first) and upgrade(last) versions from version list
```
stevekuznetsov: Comment out of date

jhadvig:
@stevekuznetsov comment addressed.
Here is the job that was running with those changes https://ci.openshift.redhat.com/jenkins/view/All/job/test_branch_origin_extended_conformance_install_update/350/  ... install & upgrade process passed without any issues, but the conformance test failed. dont think it was related to the change. PTAL

stevekuznetsov:
Failures were expected.

jhadvig:
@stevekuznetsov ok.. will upgrade the jobs in the Jenskins once this is merged. Thanks !

stevekuznetsov:
@jhadvig you can do the sync

PR #252, opened 2017-05-12
Adjust action title

jhadvig:
Just action title adjustment. Noticed that the title is not valid since we removed the GA to latest upgrade.
@stevekuznetsov PTAL


stevekuznetsov:
@jhadvig good to sync

PR #262, opened 2017-05-17
Update the Docker testing job to run the Ansible installer

stevekuznetsov:

### DIFF
--- a/jobs/package-dockertested/Jenkinsfile
+++ b/jobs/package-dockertested/Jenkinsfile
@@ -92,21 +92,106 @@ node('buildvm-devops') {
 			stage ('Build an Origin release') {
 				sh """{
 cat <<EOF
-sudo su origin
 set -o errexit -o nounset -o pipefail -o xtrace
 cd /data/src/github.com/openshift/origin
 hack/build-base-images.sh
-make release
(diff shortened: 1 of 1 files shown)


jobs/package-dockertested/Jenkinsfile
```diff
@@ -92,21 +92,106 @@ node('buildvm-devops') {
 			stage ('Build an Origin release') {
 				sh """{
 cat <<EOF
-sudo su origin
 set -o errexit -o nounset -o pipefail -o xtrace
 cd /data/src/github.com/openshift/origin
 hack/build-base-images.sh
-make release
+OS_BUILD_ENV_PRESERVE=_output/local hack/env OS_ONLY_BUILD_PLATFORMS='linux/amd64' hack/build-rpm-release.sh
+sudo systemctl restart docker
```
kargakis: why is this needed?

stevekuznetsov: Funny story. `docker cp` takes RAM proportional to what you copy, which in the `hack/env` case often is > a couple GB. The daemon also holds onto that RAM for about ten minutes after the `docker cp` client call is finished. This was killing our jobs, so we just restart the daemon because a dead process can't sit on memory ;)

kargakis: Good to know, can you add it as a comment?

stevekuznetsov: I thought there was a commit message in there somewhere with the explanation ... hmmm ... yes we can add it

PR #269, opened 2017-05-18
Run daily extended tests on a cluster installed by OpenShift-Ansible

stevekuznetsov:
The current daily extended test jobs run off of an AMI that contains the
pre-build Origin and OpenShift Ansible binaries, RPMs and images. The
tests, however, set up a small OpenShift cluster to run on themselves.
In contrast, the end-to-end tests in the main Origin merge queue run on
top of a cluster installed by OpenShift-Ansible. Switching the parent
job configuration on these daily jobs will cause them to run like the
Origin end-to-end tests do.

/cc @bparees @csrwng @gabemontero @oatmealraisin @coreydaley @jim-minter thoughts? Any reason to think this will not be acceptable for the tests?

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_builds.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_builds.yml
@@ -1,5 +1,5 @@
 ---
-parent: 'common/test_cases/origin_built_release.yml'
+parent: 'common/test_cases/origin_installed_release.yml'
 overrides:
   timer: 'H H * * *'
   email:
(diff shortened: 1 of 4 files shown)


bparees:
nothing springs to mind.  will this make it faster, or slower? :)

stevekuznetsov:
> will this make it faster, or slower?

The Ansible cluster install takes about ten minutes. The Origin `hack/` install takes about one minute.

bparees:
> The Ansible cluster install takes about ten minutes. The Origin hack/ install takes about one minute.

so that'd be "slower" then. got it. :)

are we/can we not share a pre-ansible cluster installed AMI among the extended test jobs?

coreydaley:
So, are we switching over to this just to be more consistent/standardize on one setup?

stevekuznetsov:
There is great value to us to run on a cluster set up with the installer. As that's how production clusters are installed, we need to run on top for our tests to be able to catch regressions in the clusters that we really care about instead of the toy set up we have in Origin `hack/`. A side effect is that our tests look more homogenous.

stevekuznetsov:
Any final comments before merge?

bparees:
not from me

PR #283, opened 2017-05-24
configure template service broker in gce test environments

jim-minter:
this is needed to add the template service broker tests to conformance (https://github.com/openshift/origin/pull/14320).

### DIFF
--- a/sjb/inventory/group_vars/OSEv3/general.yml
+++ b/sjb/inventory/group_vars/OSEv3/general.yml
@@ -11,5 +11,7 @@ openshift_master_identity_providers:
     login: "true"
     challenge: "true"
     kind: "AllowAllPasswordIdentityProvider"
+openshift_template_service_broker_namespaces:
+  - "openshift"
 ansible_ssh_user: "ec2-user"
 enable_excluders: "false"


jim-minter:
@stevekuznetsov ptal

stevekuznetsov:
Is this required only in GCE?

stevekuznetsov:
Ah, it is manifesting in GCE but should be fine in all environments to have this option. If so, please add it as a static YAML entry to the inventory at `sjb/inventory/group_vars/OSEv3/general.yml`

jim-minter:
Thanks @stevekuznetsov - updated & pushed, ptal

stevekuznetsov:
LGTM

jim-minter:
@stevekuznetsov this additional setting isn't taking effect on GCE.  AFAICS launch.yaml in origin-gce includes /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml .  Does it really use the general.yml file - which, for example, sets ansible_ssh_user to "ec2-user"?  There must be somewhere else where I should be setting openshift_template_service_broker_namespaces - do you know where that would be?

stevekuznetsov:
Hmm, you're right -- in GCE we'd need to also add options to the inventory at https://github.com/openshift/release/blob/master/cluster/test-deploy/data/vars.yaml

PR #296, opened 2017-05-26
Pass openshift_hosted_router_create_certificate as a boolean

stevekuznetsov:
While the other `"true"` values in the inventory do need to be strings
due to the way OpenShift ends up interpreting them, this specific value
should be a real YAML boolean as it gets used by Ansible.


stevekuznetsov:
@ewolinetz @sdodson the UX here kinda sucks -- we should try to always cast to `bool` in the playbooks when we validate input

PR #300, opened 2017-05-30
Continuous upgrade job

jhadvig:
Continuous upgrade will consist from three jobs:
1, Provision - install OCP on a single cluster node and trigger *Upgrade* job
2, Upgrade - upgrade OCP to latest and trigger *Upgrade* job on sucess. On failure trigger *Terminate* job
3, Terminate - terminate installed OCP node and trigger *Provision* job

Jobs are written consist from multiple scripts and will generate a single running script for each job. *Provision* and *Upgrade* jobs are `trap`ed on error and will cause triggering the *Terminate* job

@stevekuznetsov PTAL

### DIFF
--- a/continuous-upgrade/jobs/install-ocp.sh
+++ b/continuous-upgrade/jobs/install-ocp.sh
@@ -1,0 +1,36 @@
+script="$( mktemp )"
+cat <<SCRIPT >"${script}"
+#!/bin/bash
+set -o errexit -o nounset -o pipefail -o xtrace
+
+cat << EOR > ./openshift-int.repo
+[openshift-int]
+baseurl = https://mirror.openshift.com/enterprise/online-int/latest/x86_64/os/
+enabled = 1
+gpgcheck = 0
+gpgkey = https://mirror.ops.rhcloud.com/libra/keys/RPM-GPG-KEY-redhat-release https://mirror.ops.rhcloud.com/libra/keys/RPM-GPG-KEY-redhat-beta https://mirror.ops.rhcloud.com/libra/keys/RPM-GPG-KEY-redhat-openshifthosted
+name = OpenShift Enterprise int Builds
+sslclientcert = /var/lib/yum/client-cert.pem
+sslclientkey = /var/lib/yum/client-key.pem
+sslverify = 0
+EOR
+
+sudo cp ./openshift-int.repo /etc/yum.repos.d
+sudo yum --disablerepo=* --enablerepo=openshift-int,oso-rhui-rhel-server-releases install -y atomic-openshift-utils
+cd /data/src/github.com/openshift/aos-cd-jobs/
+ansible-playbook  -vv          \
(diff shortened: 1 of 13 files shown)


continuous-upgrade/jobs/provision-job.yml
```diff
@@ -0,0 +1,12 @@
+- job:
+    name: continuos-upgrade/install-job
```
stevekuznetsov: you should

```
sed -i 's/continuos/continuous/g' $( find continuous-upgrade/ -type f )
```

continuous-upgrade/jobs/install-ocp.sh
```diff
@@ -0,0 +1,37 @@
+script="$( mktemp )"
+cat <<SCRIPT >"${script}"
```
stevekuznetsov: We had to do this because we were not using JJB -- I think we should be good to just use a script block in the YAML

continuous-upgrade/jobs/provision-job.yml
```diff
@@ -0,0 +1,12 @@
...
+    project-type: freestyle
+    auth-token: origin1
+    defaults: global
+    builders:
+      - shell: 
+          !include-raw:
+              - install-oct.sh
+              - terminate-lib.sh
```
stevekuznetsov: I am not sure this is right -- we should only terminate on user ask. If something goes wrong we want it to stick around until someone comes to clean it up.

continuous-upgrade/jobs/terminate-cloud-resources.sh
```diff
@@ -0,0 +1,13 @@
...
+touch $latest
+cp $latest/bin/activate $WORKSPACE/activate
+cat >> $WORKSPACE/activate <<EOF
+export OCT_CONFIG_HOME=~/continuous-upgrade/
+EOF
+
+source $WORKSPACE/activate
+oct deprovision
+
+curl -kL "https://ci.openshift.redhat.com/jenkins/job/continuous-upgrade_provision-job/build?token=origin1"
```
stevekuznetsov: We should describe these as Jenkins job dependencies https://docs.openstack.org/infra/jenkins-job-builder/triggers.html#triggers.reverse

continuous-upgrade/jobs/terminate-job.yml
```diff
@@ -0,0 +1,10 @@
+- job:
+    name: continuos-upgrade/terminate-job
+    project-type: freestyle
+    test-var: testik
```
stevekuznetsov: should remove?

jhadvig:
@stevekuznetsov comments addressed as a separate commit.. will squash before merge.. PTAL

stevekuznetsov:
Talked about regeneration, decided to eat the cost of drift for now. We can reconsider later if we find ourselves playing catch-up.

PR #318, opened 2017-06-07
[ops] Maintain ssh key to gather cluster logs

codificat:
A script that rotates the ssh key in `~opsmedic/.ssh/authorized_keys` that is associated to the log collection script, and a Jenkins pipeline to drive it.

The script maintains ssh keys whose comment looks like `logs_access_key_week_XX-timestamp`, where *XX* is the current week's number. This is because by default it checks if a key for this week is already present and aborts (can be `--force`'d).

Once a week a new key is generated and pushed to openshift's shared-secrets repo.

The currently active key and the previous one are kept in `authorized_keys`.

### DIFF
--- a/jobs/cluster/rotate-log-access-key/Jenkinsfile
+++ b/jobs/cluster/rotate-log-access-key/Jenkinsfile
@@ -1,0 +1,122 @@
+#!/usr/bin/env groovy
+
+node('buildvm-devops') {
+
+    properties([
+            [   $class: 'ParametersDefinitionProperty',
+                parameterDefinitions: [
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        name: 'BASTION_HOST',
+                        description: 'Bastion host\'s hostname',
+                        defaultValue: 'use-tower2.ops.rhcloud.com'
+                    ],
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        name: 'BASTION_USER',
+                        description: 'Bastion host\'s username',
+                        defaultValue: 'opsmedic'
+                    ],
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        name: 'BASTION_CREDENTIALS',
+                        description: 'Jenkins Credentials to ssh to bastion host for key gen',
+                        defaultValue: 'rotate_logs_key'
+                    ],
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        name: 'SECRETS_REPO',
+                        description: 'Git repo for shared secrets',
+                        defaultValue: 'git@github.com:openshift/shared-secrets.git'
+                    ],
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        name: 'OSE_CREDENTIALS',
+                        description: 'Credentials for the secrets repository',
+                        defaultValue: 'openshift-bot'
+                    ],
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        name: 'KEY_FILE',
+                        description: 'Relative path to the key file inside the secrets repo',
+                        defaultValue: 'online/logs_access_key'
+                    ],
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        defaultValue: 'pep@redhat.com',
+                        description: 'Failure Mailing List',
+                        name: 'MAIL_LIST_FAILURE'
+                    ],
+                ]
+            ],
+            pipelineTriggers([[
+                        $class: 'TimerTrigger',
+                        spec: 'H 0 * * 1'  // every Monday around midnight
+                    ]])
+        ]
+    )
+
+    // Force Jenkins to fail early if this is the first time this job has been run/and or new parameters have not been discovered.
+    echo '''Parameters:
+BASTION_HOST:[${BASTION_HOST}],
+BASTION_USER:[${BASTION_USER}],
+BASTION_CREDENTIALS:[${BASTION_CREDENTIALS}],
+SECRETS_REPO:[${SECRETS_REPO}],
+OSE_CREDENTIALS:[${OSE_CREDENTIALS}],
+KEY_FILE:[${KEY_FILE}],
+MAIL_LIST_FAILURE:[${MAIL_LIST_FAILURE}]
+'''
+
+    try {
+
+        stage('Checkout secrets repo') {
+            checkout(
+                $class: 'GitSCM',
+                userRemoteConfigs: [
+                    [
+                        name: 'origin',
+                        url: SECRETS_REPO,
+                        credentialsId: OSE_CREDENTIALS
+                    ]
+                ],
+                branches: [[ name: '*/master' ]]
+            )
+        }
+
+        stage('Generate and install/rotate Key') {
+            def new_key = null
+            sshagent([BASTION_CREDENTIALS]) {
+                echo "Generate new key and rotate authorized_keys"
+                // This is performed by the command associated to this user/key, see
+                // tower-scripts/bin/rotate-log-access-key.sh
+                new_key = sh (
+                    script: "ssh -o StrictHostKeyChecking=no ${BASTION_USER}@${BASTION_HOST}",
+                    returnStdout: true
+                )
+            }
+            writeFile file: KEY_FILE, text: new_key
+        }
+
+        stage('Push new key to shared secrets repo') {
+            sshagent([OSE_CREDENTIALS]) {
+                sh '''
+git add ${KEY_FILE}
+git commit -m "New SSH key to gather cluster logs"
+git push origin HEAD:master
+'''
+            }
+        }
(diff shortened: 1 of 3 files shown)


codificat:
@jupierce PTAL.

I have some doubts around how the script will be installed and the pipeline step that runs it probably needs adjustment.

The script options are probably a bit useless, but added them just in case - again depending on how the script can be invoked it might need some adjustment in parsing them.

Also, the commit details probably should be something else but not sure what.

jobs/cluster/rotate-log-access-key/Jenkinsfile
```diff
@@ -0,0 +1,96 @@
...
+	    )
+	}
+	writeFile file: KEY_FILE, text: new_key
+    }
+
+    stage('Push new key to shared secrets repo') {
+	sshagent([OSE_CREDENTIALS]) {
+	    sh '''
+git add ${KEY_FILE}
+git config user.name jenkins
```
jupierce: user.name and user.email should already be globally configured on buildvm-devops node. Hopefully no need for these lines.

tower-scripts/bin/rotate-log-access-key.sh
```diff
@@ -0,0 +1,117 @@
...
+KEY_COMMAND="${KEY_COMMAND:-/usr/local/bin/gather-logs.sh \${SSH_ORIGINAL_COMMAND\}}"
+
+# Prepare the comment part of the ssh key to help us identify the key:
+# Keep in separate vars because "prefix_WEEKNUM" is what lets us look for the
+# presence of a key for this week
+KEY_COMMENT="${KEY_COMMENT_PREFIX}_$(date +%V)"
+KEY_TIMESTAMP=$(date +%Y%m%d-%H:%M:%S)
+
+# Exit if a key for this week is already there
+if grep -q ${KEY_COMMENT} ${AUTHKEYS_FILE} && [ -z "$FORCE" ]; then
```
jupierce: Since Jenkins will be the primary user here, would it make sense to just echo out the existing key and exit 0? That would just let us run the task again if anything happened during the push.

codificat: That would be nice - unfortunately at this point the script only knows about the public SSH key there, it can't output the existing private key because it's only kept in the shared-secrets repo.

The `--force` flag should hopefully do the trick. However, so far the script's options are not exposed through Jenkins - will look into that, at least for that one.

tower-scripts/bin/rotate-log-access-key.sh
```diff
@@ -0,0 +1,117 @@
...
+# The 'log_access_key_week' part of the comment can be changed via the
+# --prefix= option or the KEY_COMMENT_PREFIX env var.
+# The command associated with the key can be changed with the --command=
+# option or the KEY_COMMAND env var
+#
+# The new SSH private key is printed to stdout, meant to be collected by
+# a Jenkins pipeline running this script.
+
+# TODO: this script will be typically run from a hardcoded command= in
+# use-tower's authorized_keys. We should parse SSH_ORIGINAL_COMMAND
```
jupierce: I actually don't think the security team will allow us to parse anything passed from Jenkins/ssh. If you look at https://github.com/openshift/openshift-tools/blob/prod/scripts/cicd/verify-cicd-operation.py , you can see how tightly our operations are limited.

tower-scripts/bin/rotate-log-access-key.sh
jupierce: Good point. See comments above on why parameter parsing from Jenkins/ssh needs to be extremely limited.

tower-scripts/bin/rotate-log-access-key.sh
```diff
@@ -0,0 +1,117 @@
...
+	   		    (default: logs_access_key_week)
+	   --force	    Do not check if there's an existing key for this week
+EOF
+    exit 0
+fi
+
+# Default values for options
+AUTHKEYS_FILE="${AUTHKEYS_FILE:-/home/opsmedic/.ssh/authorized_keys}"
+KEY_COMMENT_PREFIX="${COMMENT_PREFIX:-logs_access_key_week}"
+KEY_COMMAND="${KEY_COMMAND:-/usr/local/bin/gather-logs.sh \${SSH_ORIGINAL_COMMAND\}}"
```
jupierce: Why add SSH_ORIGINAL_COMMAND here? When we implement gather-logs.sh, we can have it inspect the environment variable directly.

tower-scripts/bin/rotate-log-access-key.sh
jupierce: In fact, the requirements have been so strict that we will likely need to implement a pre-filter script which executes this script (similar to verify-cicd-operation.sh). In any case, don't worry about SSH_ORIGINAL_COMMAND for now.

tower-scripts/bin/rotate-log-access-key.sh
codificat: The `gather-logs.sh` script that is currently in the repo expects an argument with the cluster name, and the proposed workflow was that developers would run `ssh -i the_key opsmedic@use-tower2.ops.rhcloud.com the_cluster_name > logs.tgz`, hence why I added this here.

Happy to move the logic as needed, although it seems to make more sense for `gather-logs.sh` to be independent of being run as an authorized_keys command: the entry point here is the `rotate-log-access-key.sh` script (or a wrapper around it if necessary)...

jobs/cluster/rotate-log-access-key/Jenkinsfile
```diff
@@ -0,0 +1,96 @@
...
+git config user.email jenkins@example.com
+git commit -m "New SSH key to gather cluster logs"
+git push origin HEAD:master
+'''
+	}
+    }
+
+    // TODO: revert changes to opsmedic's authorized_keys if the git push of the new key fails
```
codificat: ack

jobs/cluster/rotate-log-access-key/Jenkinsfile
codificat: ack, thanks

jobs/cluster/rotate-log-access-key/Jenkinsfile
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env groovy
+
+node() {
```
codificat: ack

tower-scripts/bin/rotate-log-access-key.sh
codificat: The `gather-logs.sh` script that is currently in the repo expects an argument with the cluster name, and the proposed workflow was that developers would run `"ssh -i the_key opsmedic@use-tower2.ops.rhcloud.com the_cluster_name > logs.tgz"`, hence why I added this here.

Happy to move the logic as needed, although it seems to make more sense for `gather-logs.sh` to be independent of being run as an authorized_keys command: the entry point here is the `rotate-log-access-key.sh` script (or a wrapper around it if necessary)...

tower-scripts/bin/rotate-log-access-key.sh
codificat: I see. So, the alternatives here would be to either:

- skip the check for an existing key for the week and just generate a new one every time it's invoked
- have 2 separate keys to run the script, one of them invoking the force option. Then the way to override from Jenkins is to choose different credentials.

I'm not sure which one I dislike the most, but I guess we should just remove this check for now and just always generate/rotate keys.

tower-scripts/bin/rotate-log-access-key.sh
codificat: Ok, I will just remove option parsing from the script and just leave the variables with pre-defined values.

Then I believe we don't need a pre-filter script, at least not to check arguments here - maybe for the logging side of things?

tower-scripts/bin/rotate-log-access-key.sh
jupierce: I'd side with generating the key every time.

tower-scripts/bin/rotate-log-access-key.sh
jupierce: We will definitely need something strict when wiring the logging invocation. Likely a regex to ensure that a permitted cluster has been specified as the one at only argument.

tower-scripts/bin/rotate-log-access-key.sh
jupierce: I see - all good reasoning. For security reasons, I think /usr/local/bin/gather-logs.sh will be a filter script (like verify-cicd-operation.py is today) which will invoke the aos-cd-jobs/tower-scripts/gather-logs.sh . To disambiguate the names, maybe something like: /usr/bin/verify-gather-logs-operations.py as the filter calling aos-cd-jobs/tower-scripts/gather-logs.sh ?

/usr/bin/verify-gather-logs-operations.py would be committed to a different repo and we will need to get security to take a look.

PR #326, opened 2017-06-09
Forcing ose_images.sh to exit on image push errors

jupierce:
@smunilla FYI


kargakis:
@jupierce can we get a techdebt card to start breaking ose_images.sh out into testable pieces?

jupierce:
@kargakis Definitely. I think we need to rewrite it in python. It's on my list, but I'll make a card too. Thanks.

PR #329, opened 2017-06-12
Add docker image cleanup script

smunilla:


build-scripts/buildvm/docker_image_cleanup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+THREEWEEKSAGO="$(date -Ins --date="3 Weeks Ago")"
+#NOW="$(date)"
+NOW="$(date --date="2017-06-12T07:11:40.909671Z")"
```
jupierce: eh?

jupierce:
fyi @stevekuznetsov

build-scripts/buildvm/docker_image_cleanup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+THREEWEEKSAGO="$(date -Ins --date="3 Weeks Ago")"
+
+docker images -q | while read IMAGE_ID; do
+    IMAGE_DATE = docker inspect --format='{{.Created}}' --type=image ${IMAGE_ID}
```
stevekuznetsov: Am I missing something or does this need a subshell?

```sh
IMAGE_DATE="$( docker inspect --format='{{.Created}}' --type=image "${IMAGE_ID}" )"
```

build-scripts/buildvm/docker_image_cleanup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+THREEWEEKSAGO="$(date -Ins --date="3 Weeks Ago")"
+
+docker images -q | while read IMAGE_ID; do
```
stevekuznetsov: Don't really need the line-reading capabilities of `while read` here, could just

```sh
for image in $( docker images -q ); do

done
```

jupierce:
Wondering if something like this might be more robust for us.

Trying to keep around the newest 50 images in this case:

```
# List all images and print "<create_date> <image_id>"
#       | sort by date
#       | awk out only the image id
#       | uniq so that images ids aren't duplicates
#       | pass through all but newest 50
#       | pass everything to docker rmi using xargs
{
        for image in $( docker images -q ); do
                docker inspect --format='{{.Created}} {{.Id}}' --type=image ${image}
        done
} | sort | awk '{ print $2 }' | uniq | head -n -50 | xargs --no-run-if-empty docker rmi -f
```

jupierce:
@smunilla Thanks for starting this, I'm moving this over to a new Jenkins task for easier debug in the future. https://github.com/openshift/aos-cd-jobs/pull/331

PR #334, opened 2017-06-16
Allow merges with both blocker and bug severities

eparis:
Today rosie will not merge a BZ if it has
```
[merge][severity: bug]
[merge][severity: blocker]
```
because "bug,blocker" is not a valid state.

This PR will instead look for the severity anywhere in the string.

It does mean that a PR with:
```
[merge][severity: elephantblockermonkey]
```
Will get counted as a blocker. We could make it better, but it just
didn't seem worth it to me...

### DIFF
--- a/approvers/devcut_approver.sh
+++ b/approvers/devcut_approver.sh
@@ -8,19 +8,20 @@ if [[ $# -ne 1 ]]; then
 	echo "[ERROR] Usage: $0 SEVERITY"
 	exit 127
 else
-	severity="$1"
+	# make severity lowercase
+	severity="${1,,}"
 	case "${severity}" in
 		"none")
 			echo "[ERROR] Only bugs and blocker bugs are allowed to merge during dev-cut."
 			exit 1
 			;;
-		"bug")
+		*"tooling"*)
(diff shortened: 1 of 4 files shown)


approvers/devcut_approver.sh
```diff
@@ -8,23 +8,24 @@ if [[ $# -ne 1 ]]; then
...
-	severity="$1"
+	# make severity lowercase
+	severity="${1,,}"
 	case "${severity}" in
-		"none")
-			echo "[ERROR] Only bugs and blocker bugs are allowed to merge during dev-cut."
-			exit 1
-			;;
-		"bug")
+		"*tooling*")
```
stevekuznetsov: Globs inside of quotes are treated as string literals -- you want `*"tooling"*`

eparis:
$ ./approvers/stagecut_approver.sh blocker,Bug,monkey
$ echo $?
0

stevekuznetsov:
@eparis Thanks -- do you want to own the push or should i?

stevekuznetsov:
Sounds like a no. I'll push them.

PR #349, opened 2017-06-21
Check for network_manager.yml

jhadvig:
For openshift-ansible 1.3 and lower we dont have the `network_manager.yml` playbook. For that reason are some jobs (that are installing those version) are failing, eg: https://ci.openshift.redhat.com/jenkins/view/All/job/test_pull_request_openshift_ansible_extended_conformance_install_update/444/consoleFull#-172450789458b6e51eb7608a5981914356

Added a check for that playbook.
@stevekuznetsov  PTAL


stevekuznetsov:
We should backport the playbook -- it is necessary to actually run the installer and have it work.

mmahut:
It probably doesn't make sense to backport it all the way down to 1.1, just for the stable releases.

jhadvig:
@mmahut by stable you mean 1.3 and higher ?

PR #357, opened 2017-06-23
allow setting a timeout on script actions

bparees:
fixes https://github.com/openshift/origin/issues/14765

### DIFF
--- a/sjb/actions/script.py
+++ b/sjb/actions/script.py
@@ -19,7 +19,7 @@ cd "\${HOME}"
 SCRIPT
 chmod +x "${script}"
 scp -F ./.config/origin-ci-tool/inventory/.ssh_config "${script}" openshiftdevel:"${script}"
-ssh -F ./.config/origin-ci-tool/inventory/.ssh_config -t openshiftdevel "bash -l -c \\"${script}\\"" """)
+ssh -F ./.config/origin-ci-tool/inventory/.ssh_config -t openshiftdevel "bash -l -c \\"timeout {{ timeout }} ${script}\\"" """)
 
 
 class ScriptAction(Action):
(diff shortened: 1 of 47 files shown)


bparees:
@stevekuznetsov ptal and then point me to where these script actions are created so i can go update some of them to pass real timeout values.  (guessing at least some of them are going to need more than an hour).

bparees:
@stevekuznetsov to make this less risky, i've made the default 6 hours.  we can just go enforce lower timeouts as we find good places to introduce them.

sjb/actions/script.py
```diff
@@ -43,5 +43,6 @@ def generate_build_steps(self):
             command=_SCRIPT_ACTION_TEMPLATE.render(
                 repository=self.repository,
                 command=self.script,
+                timeout=60*60*6 if self.timeout is None
```
stevekuznetsov: Add `timeout=3600` as a param to the `__init__` for `ScriptAction`

stevekuznetsov:
Thanks for the work @bparees -- actions are labeled `script:` or `host_script:` and look like [this](https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_origin_check.yml#L5)

bparees:
@stevekuznetsov i found them, added a second commit w/ a few first proposals.

bparees:
@stevekuznetsov updated.  if there are other low hanging fruit you can think of for me to slap timeouts on, let me know.

bparees:
@stevekuznetsov test job running here:
https://ci.openshift.redhat.com/jenkins/view/All/job/zzz_test_branch_origin_extended_conformance_install_update/1/

any other comments?

stevekuznetsov:
Once the job passes I'll merge and you can push up the changes.

bparees:
@stevekuznetsov job passed:
https://ci.openshift.redhat.com/jenkins/view/All/job/zzz_test_branch_origin_extended_conformance_install_update/1/

PR #358, opened 2017-06-26
List all packages

jhadvig:
even those that are already installed, so in case the determining the upgrade version, the latest version is already installed just print the same version.
Unfortunately the yumBase cant provide all the packages so we have to do union of all the available + the installed.

@stevekuznetsov PTAL

### DIFF
--- a/sjb/hack/determine_install_upgrade_version.py
+++ b/sjb/hack/determine_install_upgrade_version.py
@@ -114,7 +114,10 @@ if __name__ == "__main__":
 	# listing of origin or docker pkgs will be disabled, based on the enabled excluder.
 	yb.conf.disable_excludes = ["all"]
 	sys.stdout = old_stdout
-	available_pkgs = rpmutils.unique(yb.doPackageLists('available', patterns=[pkg_name], showdups=True).available)
+
+	generic_holder = yb.doPackageLists('all', patterns=[pkg_name], showdups=True)
+	available_pkgs = rpmutils.unique(generic_holder.available + generic_holder.installed)
+
 	available_pkgs = remove_duplicate_pkgs(available_pkgs)
 	available_pkgs = sort_pkgs(available_pkgs)
 
--- a/sjb/hack/determine_install_upgrade_version_test.py
+++ b/sjb/hack/determine_install_upgrade_version_test.py
@@ -153,8 +153,8 @@ class SortPackagesTestCase(unittest.TestCase):
 
 	def test_sort_packages_with_exceptional_origin_pkg(self):
 		""" when sorting origin packages with exceptional origin-3.6.0-0.0.alpha.0.1 package """
-		test_pkgs = ["origin-3.6.0-0.0.alpha.0.1", "origin-3.6.0-0.alpha.0.2"]
-		properly_sorted_pkgs = ["origin-3.6.0-0.alpha.0.2"]
+		test_pkgs = ["origin-3.6.0-0.0.alpha.0.1.el7", "origin-3.6.0-0.alpha.0.2.el7"]
+		properly_sorted_pkgs = ["origin-3.6.0-0.alpha.0.2.el7"]
 		test_pkgs_obj = TestPackage.create_test_packages(test_pkgs)
 		properly_sorted_pkgs_obj = TestPackage.create_test_packages(properly_sorted_pkgs)
 		sorted_test_pkgs_obj = sort_pkgs(test_pkgs_obj)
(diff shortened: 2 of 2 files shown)


sjb/hack/determine_install_upgrade_version_test.py
```diff
@@ -158,7 +158,7 @@ def test_sort_packages_with_exceptional_origin_pkg(self):
 		test_pkgs_obj = TestPackage.create_test_packages(test_pkgs)
 		properly_sorted_pkgs_obj = TestPackage.create_test_packages(properly_sorted_pkgs)
 		sorted_test_pkgs_obj = sort_pkgs(test_pkgs_obj)
-		self.assertEqual(sorted_test_pkgs_obj, properly_sorted_pkgs_obj)
+		self.assertTrue(sorted_test_pkgs_obj[0], properly_sorted_pkgs_obj[0])
```
stevekuznetsov: Why?

sjb/hack/determine_install_upgrade_version.py
```diff
@@ -114,7 +114,10 @@ def sort_pkgs(available_pkgs):
 	# listing of origin or docker pkgs will be disabled, based on the enabled excluder.
 	yb.conf.disable_excludes = ["all"]
 	sys.stdout = old_stdout
-	available_pkgs = rpmutils.unique(yb.doPackageLists('available', patterns=[pkg_name], showdups=True).available)
+
+	generic_holder = yb.doPackageLists('all', patterns=[pkg_name], showdups=True)
+	available_pkgs = rpmutils.unique(generic_holder.available + generic_holder.installed)
```
stevekuznetsov: How much output will this be? Would it be better if we write to a debug log from this script and write down more of the steps that are taken?

jhadvig: not really sure what you mean... the thing is that the upgrade version should be the latest (if the `--dependency-branch` flag is not pointing to a release branch)

sjb/hack/determine_install_upgrade_version_test.py
jhadvig: casue `assetEqual` will check if the comparing variables also has the same memory address. and since `test_pkgs_obj` and `properly_sorted_pkgs_obj` are basically diffewrent object with the same semantics you have we have to comapare them like this

stevekuznetsov: That's not true. `unittest.assertEqual` will (for lists) fall back to [`unitttest.assertListEqual`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertListEqual) which [uses the `==` operator](http://asynctest.readthedocs.io/en/latest/_modules/unittest/case.html) and not the `is` operator:

```py
    def assertListEqual(self, list1, list2, msg=None):
        """A list-specific equality assertion.

        Args:
            list1: The first list to compare.
            list2: The second list to compare.
            msg: Optional message to use on failure instead of a list of
                    differences.

        """
        self.assertSequenceEqual(list1, list2, msg, seq_type=list)

    def assertSequenceEqual(self, seq1, seq2, msg=None, seq_type=None):
        """An equality assertion for ordered sequences (like lists and tuples).

        For the purposes of this function, a valid ordered sequence type is one
        which can be indexed, has a length, and has an equality operator.

        Args:
            seq1: The first sequence to compare.
            seq2: The second sequence to compare.
            seq_type: The expected datatype of the sequences, or None if no
                    datatype should be enforced.
            msg: Optional message to use on failure instead of a list of
                    differences.
        """
        if seq_type is not None:
            seq_type_name = seq_type.__name__
            if not isinstance(seq1, seq_type):
                raise self.failureException('First sequence is not a %s: %s'
                                        % (seq_type_name, safe_repr(seq1)))
            if not isinstance(seq2, seq_type):
                raise self.failureException('Second sequence is not a %s: %s'
                                        % (seq_type_name, safe_repr(seq2)))
        else:
            seq_type_name = "sequence"

        differing = None
        try:
            len1 = len(seq1)
        except (TypeError, NotImplementedError):
            differing = 'First %s has no length.    Non-sequence?' % (
                    seq_type_name)

        if differing is None:
            try:
                len2 = len(seq2)
            except (TypeError, NotImplementedError):
                differing = 'Second %s has no length.    Non-sequence?' % (
                        seq_type_name)

        if differing is None:
            if seq1 == seq2:
                return
            // more code if they are not equal 
```

In Python, `==` checks for object equality while `is` will do [object identity](https://docs.python.org/2/library/stdtypes.html#comparisons) as well.

jhadvig:
@stevekuznetsov you were right regarding the assertion.. figured out that the test values in the `test_sort_packages_with_exceptional_origin_pkg` were missing architecture type. Thats why they they where failing. Added them

sjb/hack/determine_install_upgrade_version.py
stevekuznetsov: I just meant it might be easier to look at the output from debugging info if we took a debugging log file and put it all in there -- rather than looking through the job log. If this is outputting a lot of stuff or if you expect to add more debug info maybe that would be appropriate? Your call.

jhadvig: ok.. will add additional debug info using the [logging](https://docs.python.org/2.7/library/logging.html) library, if thats what you mean

stevekuznetsov: Up to you, depending on how much debugging you think you need. This LGTM to merge now if you want.

jhadvig: @stevekuznetsov ok, lets merge this... will open additional PR. Please merge since I dont have merge rights

Issue #361, opened 2017-06-26
package-dockertested: skopeo version skew breaks pre-flight checks


stevekuznetsov:
We install `docker` using the `rhel7next*` suite of repos, which are synced from the latest RHEL 7 Extras compose in Brew. During this process, recently we are bringing in `skopeo-containers` as a dependency:

```
$ sudo yum --disablerepo=\* --enablerepo=rhel7next\* install docker
Loaded plugins: amazon-id, rhui-lb, search-disabled-repos
Resolving Dependencies
--> Running transaction check
---> Package docker.x86_64 2:1.12.6-32.git88a4867.el7 will be installed
--> Processing Dependency: docker-client = 2:1.12.6-32.git88a4867.el7 for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Processing Dependency: docker-common = 2:1.12.6-32.git88a4867.el7 for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Processing Dependency: docker-rhel-push-plugin = 2:1.12.6-32.git88a4867.el7 for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Processing Dependency: container-selinux >= 2:2.12-2 for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Processing Dependency: oci-register-machine >= 1:0-3.10 for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Processing Dependency: oci-systemd-hook >= 1:0.1.4-9 for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Processing Dependency: skopeo-containers for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Processing Dependency: libseccomp.so.2()(64bit) for package: 2:docker-1.12.6-32.git88a4867.el7.x86_64
--> Running transaction check
---> Package container-selinux.noarch 2:2.15-1.git583ca40.el7 will be installed
---> Package docker-client.x86_64 2:1.12.6-32.git88a4867.el7 will be installed
---> Package docker-common.x86_64 2:1.12.6-32.git88a4867.el7 will be installed
---> Package docker-rhel-push-plugin.x86_64 2:1.12.6-32.git88a4867.el7 will be installed
---> Package libseccomp.x86_64 0:2.3.1-2.el7 will be installed
[...]

PR #362, opened 2017-06-26
dockertested: disable docker_image_availability check

sosiouxme:
This check tries to install and use skopeo, which in this environment
does not work due to https://github.com/openshift/aos-cd-jobs/issues/361
Disable the check for now until we can fix the job a better way.

The package_update check is already disabled elsewhere so carry that forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ A custom XML generator lives under the `sjb/` directory. This generator is meant
 bridge the gap between monolithic scripts inside of Freestyle Jenkins Jobs and segmented Jenkins Pipelines driven by source-
 controlled Groovy scripts and libraries.
 
-The generator understand a small set of `action`s, each of which is underpinned by a Python module under
+The generator understands a small set of `action`s, each of which is underpinned by a Python module under
 [`sjb/actions/`](sjb/actions). A configuration YAML file is read in by [`sjb/generate.py`](sjb/generate.py) and used to generate a
 set of input variables to the [Jinja job template XML](sjb/templates/test_case.xml). Jobs can depend on a parent to reuse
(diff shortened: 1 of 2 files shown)


sosiouxme:
@stevekuznetsov thinking this is all that's needed for now?

stevekuznetsov:
I thought we were putting these in `sjb/inventory/group_vars/OSEv3/checks.yml` ?

sosiouxme:
Yes but that affects all the jobs. I thought it was just this job that hit the problem? Or will other jobs use the AMI from this setup?

stevekuznetsov:
Ah, right. OK. We'll need to make sure to update this if we change the variable in the inventory as well.

stevekuznetsov:
Sorry, I wasn't thinking this morning. We can't just disable it for the `dockertested` job. Imagine this scenario:

 - check is disabled in `dockertested`
 - new version of `docker` is vetted and approved
 - new version of `docker` pulls in new version of `skopeo-containers` when installed
 - CI jobs installing the new version of `docker` with appropriate dependencies from our tested repo will pull in the new `skopeo-containers` as well
 - CI jobs that attempt to install an _old_ `skopeo` after installing a new `docker` from the tested repo will fail

stevekuznetsov:
Or, alternatively:

 - check is disabled in `dockertested`
 - new version of `docker` is vetted and approved
 - only the `docker` and `container-selinux` RPM packages are synced from the RHEL 7 Extras compose into our tested repo
 - CI machines can't install Docker from this tested repo because it requires `skopeo-containers` at a higher level than `rhui` repos provide, and we have not synced it into our tested repo

PR #364, opened 2017-06-27
3600 default too short for some jobs

smarterclayton:
Script tasks were failing in origin_check

@stevekuznetsov @bparees

### DIFF
--- a/sjb/actions/script.py
+++ b/sjb/actions/script.py
@@ -30,7 +30,7 @@ class ScriptAction(Action):
     the repository as the working directory.
     """
 
-    def __init__(self, repository, script, title, timeout=3600):
+    def __init__(self, repository, script, title, timeout=7200):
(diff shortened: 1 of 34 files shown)


smarterclayton:
Merged and updated

sjb/actions/script.py
```diff
@@ -30,7 +30,7 @@ class ScriptAction(Action):
     the repository as the working directory.
     """
 
-    def __init__(self, repository, script, title, timeout=3600):
+    def __init__(self, repository, script, title, timeout=7200):
```
bparees: yeah this was intended to be 6 hours but i failed at math.  i'd probably raise it to 21600 if we want to be truly conservative.  but if we're confident 7200 is enough, sure.

bparees:
sorry + thanks

PR #375, opened 2017-06-28
Disable package_availability check

sdodson:
These cause the install to fail early on and it's hard to tell if this would actually cause the install to fail or if this is just a weird artifact of our CI processes.

https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible/293/consoleFull#77170979858b6e51eb7608a5981914356

```
Failure summary:

  1. Host:     localhost
     Play:     Verify Requirements
     Task:     openshift_health_check
     Message:  One or more checks failed
     Details:  check "package_availability":
               Could not perform a yum update.
               Errors from dependency resolution:
                 2:docker-1.12.6-39.1.git6ffd653.el7.x86_64 requires atomic-registries
                 2:docker-1.12.6-39.1.git6ffd653.el7.x86_64 requires container-storage-setup >= 0.3.0-1
               You should resolve these issues before proceeding with an install.
               You may need to remove or downgrade packages or enable/disable yum repositories.

The execution of "/usr/share/ansible/openshift-ansible/playbooks/byo/config.yml"
includes checks designed to fail early if the requirements
of the playbook are not met. One or more of these checks
failed. To disregard these results, you may choose to
disable failing checks by setting an Ansible variable:

   openshift_disable_check=package_availability

Failing check names are shown in the failure details above.
Some checks may be configurable by variables if your requirements
are different from the defaults; consult check documentation.
Variables can be set in the inventory or passed on the
command line using the -e flag to ansible-playbook.
```

### DIFF
--- a/jobs/package-dockertested/install-origin-release.sh
+++ b/jobs/package-dockertested/install-origin-release.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-set -o errexit 
-set -o nounset 
-set -o pipefail 
+set -o errexit
+set -o nounset
+set -o pipefail
 set -o xtrace
 
 cd /data/src/github.com/openshift/origin
@@ -24,7 +24,7 @@ ansible-playbook -vv --become               \
                  -e deployment_type=origin  \
                  -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
                  -e oreg_url='openshift/origin-${component}:'"$( cat ./ORIGIN_COMMIT )" \
-                 -e openshift_disable_check=docker_image_availability,package_update    \
+                 -e openshift_disable_check=docker_image_availability,package_update,package_availability    \
                  /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml
 sudo chmod a+x /etc/ /etc/origin/ /etc/origin/master/
 sudo chmod a+rw /etc/origin/master/admin.kubeconfig
--- a/sjb/inventory/group_vars/OSEv3/checks.yml
+++ b/sjb/inventory/group_vars/OSEv3/checks.yml
@@ -1,4 +1,4 @@
 ---
 openshift_check_min_host_disk_gb: 10
 openshift_check_min_host_memory_gb: 8
-openshift_disable_check: package_update
+openshift_disable_check: package_update,package_availability


sosiouxme:
:disappointed: well we can certainly do this and find out...

This says that if we ran a yum update, it would pull in docker-1.12.6-39.1.git6ffd653.el7.x86_64 (from where?) which has these new dependencies that aren't available. The install doesn't do a yum update per se, and I don't think it tries to update docker to latest if it's already installed, so it seems likely the install would succeed with the check disabled. But is this telling us something useful about the build environment setup?

sdodson:
We'll revert if this doesn't start passing merge jobs.

sdodson:
https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/651/#showFailuresLink

sdodson:
@sosiouxme Installs seem to be working now. I wonder if maybe we should only run this check during an upgrade? In the past we've had a few bugs filed where users have complained that their disconnected environments where they had  packages pre-installed would fail because we were performing repoqueries unnecessarily. This seems like such a case, CI has at least docker and all its deps pre-installed.

sosiouxme:
@sdodson I imagine we'll get that feedback a lot as customers start to hit this.

A lot of the point of this is to prevent weird failures down the line due to whatever is already installed when the deployment begins. Customers are often required to use in-house images with who knows what on there. So it's valuable to ask the question of whether future yum actions will succeed before you start a long install.

This particular check is likely not the best implementation of that. Ideally we'd only test updating/installing exactly what the actual update/install is going to do. But that involves knowing what would be installed, which means having in this check all the logic for every role that installs/updates any RPM. That's unattainable, so we're going to have false positives and/or false negatives.

Given the check can be easily disabled when the user decides it's been overzealous, and given how bad we look when an install fails partway through, I still think it's best to run this at the start, unless we get so much pushback that this is actually making us look worse than failing in RPM hell during the install. May be a decision for PM at that point.

sdodson:
So, just within the context of our CI jobs, do you think it's reasonable to move forward with leaving this check disabled? I think we need to disable it in the logging job too which has a separate definition I have yet to track down.

sosiouxme:
Yes, we need CI to work :) Brenton mentioned trying to adjust some of the
checks to be more accommodating of pre-release setups, something to look
into though I'm not sure how much can be done.

PR #377, opened 2017-06-30
raise default timeout to 4 hours

bparees:


bparees:
@jim-minter fyi 
@stevekuznetsov ptal

sjb/generate.py
```diff
@@ -108,7 +108,7 @@ def load_configuration(config_path):
     # now, the job can define actions to take
     for action in job_config.get("actions", []):
         if action["type"] == "script":
-            actions.append(ScriptAction(action.get("repository", None), action["script"], action.get("title", None), action.get("timeout", 7200)))
+            actions.append(ScriptAction(action.get("repository", None), action["script"], action.get("title", None), action.get("timeout", 14400)))
```
stevekuznetsov: I think the pattern (which you could argue is bad) was to default to `None` here and only have the one default -- in the action definition

bparees: that was what i did originally.  the the action definition then doesn't default unless i explicitly check for None.  which you had me not do.

bparees: (you had me write the ScriptAction init as: `def __init__(self, repository, script, title, timeout=14400):`  so now you can't pass None if you want that default to do anything).

bparees: updated.

sjb/actions/post_script.py
```diff
@@ -12,7 +12,7 @@ class PostScriptAction(Action):
     """
 
     def __init__(self, repository, script, title):
-        self.action = ScriptAction(repository, script, title)
+        self.action = ScriptAction(repository, script, title, None)
```
stevekuznetsov: No reason this can't have a timeout right? Should thread through all the arguments

sjb/actions/script.py
```diff
@@ -30,11 +30,13 @@ class ScriptAction(Action):
     the repository as the working directory.
     """
 
-    def __init__(self, repository, script, title, timeout=7200):
+    def __init__(self, repository, script, title, timeout):
         self.repository = repository
         self.script = script
         if title == None:
             title = _SCRIPT_TITLE
+        if timeout == None:
```
stevekuznetsov: So you're fine to use the `timeout=14400` syntax above -- I didn't use it for `_SCRIPT_TITLE` as when you default to a var it ends up leaking out of scope.

bparees: no i'm not because the caller is explicitly setting the arg to None(by default), which means "timeout=14400" has no effect.

sjb/actions/post_script.py
bparees: threaded.

sjb/actions/script.py
stevekuznetsov: Oh, I see

PR #379, opened 2017-06-30
set default to 4 hours

bparees:


bparees:
@stevekuznetsov corrupted my other branch.  moving here.

PR #389, opened 2017-07-03
Add all artifacts to GCE job

### LINKED ISSUE #360: Ensure GCE Jobs Generate Appropriate Artifacts
Today, we have artifact generation and fetching tasks for all the jobs that run off of AWS, as it only requires one hop from the Jenkins master using the `origin-ci-tool` SSH configuration onto the AWS host. We do not have the same level of artifact gathering for the GCE job, as this job uses the AWS host as a intermediary and connects to GCE from there -- so we grab _e.g._ the `origin-node` log from the AWS host, but that host is not running `origin`.

We need to determine if we can get the normal artifact gathering logic to apply cleanly to the GCE jobs (maybe by using SSH with `-J` jump) or if we just need to hard-code a list of artifacts to get us to a better state quickly, if we think the prior will be hard to do and we do not expect to have more jobs on GCE in the future.

/cc @deads2k @smarterclayton

stevekuznetsov:

Fixes #360

### DIFF
--- a/sjb/config/common/test_cases/origin.yml
+++ b/sjb/config/common/test_cases/origin.yml
@@ -26,7 +26,7 @@ generated_artifacts:
   installed_packages.log: 'sudo yum list installed'
   avc_denials.log: 'sudo ausearch -m AVC -m SELINUX_ERR -m USER_AVC'
   docker.info: 'sudo docker version && sudo docker info && sudo docker images && sudo docker ps -a'
-  filesystem.info: 'sudo df -h && sudo vgs && sudo lvs'
+  filesystem.info: 'sudo df -h && sudo pvs && sudo vgs && sudo lvs'
   pid1.journal: 'sudo journalctl _PID=1 --no-pager --all --lines=all'
 system_journals:
   - docker.service
--- a/sjb/config/test_cases/ami_build_origin_int_rhel_base.yml
+++ b/sjb/config/test_cases/ami_build_origin_int_rhel_base.yml
@@ -101,7 +101,7 @@ generated_artifacts:
   installed_packages.log: 'sudo yum list installed'
   avc_denials.log: 'sudo ausearch -m AVC -m SELINUX_ERR'
   docker.info: 'sudo docker version && sudo docker info && sudo docker images && sudo docker ps -a'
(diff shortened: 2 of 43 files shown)


stevekuznetsov:
test job running: https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_gce/787/console

stevekuznetsov:
@smarterclayton can you PTAL second comit

sjb/config/test_data/gce-artifacts.sh
```diff
@@ -0,0 +1,37 @@
...
+
+set -o xtrace
+
+base_artifact_dir="/tmp/artifacts"
+
+for instance in $( gcloud compute instances list --regexp ".*${INSTANCE_PREFIX}.*" --uri ); do
+    info="$( mktemp )"
+    gcloud compute instances describe "${instance}" > "${info}"
+    name="$( jq '.name' --raw-output "${info}" | tail -c 5 )"
+    if jq '.tags.items | contains("ocp-master")' --exit-status; then
```
smarterclayton: the "ocp-" prefix is configurable but we are unlikely to change it in the near term.  But we might in the future if we do an origin specific one.

smarterclayton:
LGTM if you've tested it.

sjb/config/test_data/gce-artifacts.sh
stevekuznetsov: Aren't we doing an Origin deployment on these anyway, not `atomic-openshift` ?

stevekuznetsov:
Test job: https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_gce/790/console

stevekuznetsov:
Moved everything to the Jenkins master .. https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_gce/791/console

stevekuznetsov:
We don't have `openshift-ansible` installed on those nodes ... new test: https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_gce/804/console

stevekuznetsov:
Success: https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_gce/812/console

PR #390, opened 2017-07-03
Prune exclude list in GCE

stevekuznetsov:

/cc @smarterclayton


smarterclayton:
DNS for services is still broken until my PTR changes land.

PR #396, opened 2017-07-05
Revert "add testdata resources"

bparees:
these are all built into the AMI now.

@stevekuznetsov ptal.


bparees:
bad @openshift-bot !

PR #406, opened 2017-07-07
Document how to test changes in sjb/ literal content

stevekuznetsov:

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,12 @@ sjb/push-update.sh sjb/generated/YOUR_TEST_JOB.xml
 ````
 Cleanup of these jobs post-test is still manual.
 
+If changes are being made to the files under `sjb/` in this repository, it is not enough to copy a job configuration and run it to
+test the changes. Instead, it will be necessary to mark the copied job as syncing a pull request for `aos-cd-jobs` using the `type`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,6 +105,11 @@ sjb/push-update.sh sjb/generated/YOUR_TEST_JOB.xml
 ````
 Cleanup of these jobs post-test is still manual.
 
+If changes are being made to the files under `sjb/` in this repository, it is not enough to copy a job configuration and run it to
+test the changes. Instead, it will be necessary to mark the copied job as syncing a pull request for `aos-cd-jobs` using the `type`
+field on the repository as per [the spec](./sjb/syntax.md#sync_repos). Then, when running your copied job, configure it at run-time
+to merge in your pull request.
```
kargakis: An example would be nice - testing in sjb probably deserves its own doc.

stevekuznetsov: What sort of example would be useful here?

kargakis: It's not clear what "at run-time" means. Most will use the Jenkins UI so calling that out may be enough. Also including a stanza with the exact change that needs to be included in the test job would be nice.

stevekuznetsov: I'll call out the Jenkins UI.

> Also including a stanza with the exact change that needs to be included in the test job

This is going to differ based on how you are configuring your synced repos, so I hope it is OK to assume the dev can follow the link to the syntax doc here.

PR #410, opened 2017-07-07
tests: start blocking on fedora/25/atomic

jlebon:
Make sure the `fedora/25/atomic` test passed before merging the PR. I'll
update this to Fedora 26 soon, though let's at least make sure the
status quo works for now.


sdodson:
@stevekuznetsov i'm fine with this going in.

stevekuznetsov:
Why not `stage` as well?

sdodson:
Sure, that too.

jlebon:
⬆️ Updated to also gate stage.

PR #412, opened 2017-07-10
Changing jenkins plugin update method

jupierce:
@smunilla @bparees @gabemontero


hacks/update-jenkins-plugins/plugins.txt
```diff
@@ -0,0 +1,54 @@
...
+workflow-remote-loader:1.4
+workflow-scm-step:2.4
+workflow-step-api:2.9
+workflow-support:2.14
+
+# kubernetes version 0.11 has been used successfully with jenkins-1, so
+# override its encoded jenkins version requirement (2.32.1).
+kubernetes:0.11;jenkins-version-override
+
+openshift-login:0.12
```
bparees: why isn't this one latest?

gabemontero: Conversely, I think the next two should specify specific versions, and not leverage "latest".

jupierce: I'm open to whatever you guys want. My priority was getting a build at the time this was crafted, so my thinking on "latest" usage was driven by what I assumed was low-risk / necessary. 

If a plugin is under our (devex's) direct control, then I would feel relatively safe with "latest" . 

fwiw, at some point, I would also assume this file (actually, two; one for each version of jenkins we ship) could move into your jenkins repo and re-building the plugin RPMs can be fully automated.  At the moment, this process has to be performed by hand.

gabemontero: I think over time, as we start cutting branches/tags for our jenkins repo, it will be easy to track which plugin versions went with each release by examining the plugin.txt file(s).  "latest" could be  safe say in the master branch during release development, but I'm worried if we start cutting fixes for released versions of our jenkins images, and we inadvertently pull in versions of the plugins targeted for later releases via the use of "latest".

Unless @bparees has a counter I'm missing, I say we move off using "latest".

On the sharing the files, putting in the jenkins repo, automating rebuild, etc. that lines up with the last 1-1 conversation @smunilla and I had.  I'm hopeful we can still achieve that long term goal.  So +1 from me on that one.

bparees: @gabemontero makes sense to me (re: latest)

PR #418, opened 2017-07-11
Move ANSIBLE_JUNIT_DIR to /tmp

jhadvig:
@stevekuznetsov I had to move the ANSIBLE_JUNIT_DIR to `/tmp` cause in the `/usr/share` a non-root user cant create folder. Also tried to put the  `/home/ec2-user/` but the when running the `install_junit.sh` I get permission denied failure when creating the folder cause the script is run by `origin` user. So put into `/tmp` and looks like that solves the issue.

PTAL


stevekuznetsov:
`/tmp` sounds good to me

PR #420, opened 2017-07-12
Dependency branch major version should reflect real package major version

jhadvig:
The issue was cause by the fact that the release branch have major version in their names  set to `1` which in case of non-origin packages caused errors, since the script (in case release branch was set to `release-1.5`) was then looking for eg. atomic-openshift-1.5 which doesnt exists.
Now the major version is set by the package thats on the script input as a first arg.
Also refactored the code a bit and added `schema_change_check` function to have the schema change check separated.
@stevekuznetsov 

Closing https://github.com/openshift/origin/issues/15144

### DIFF
--- a/sjb/hack/determine_install_upgrade_version.py
+++ b/sjb/hack/determine_install_upgrade_version.py
@@ -51,15 +51,20 @@ def print_version_vars(install_version, upgrade_version):
 	print (used_pkg_name + "_UPGRADE_RELEASE_VERSION=" + upgrade_version)
 	print (used_pkg_name + "_UPGRADE_RELEASE_MINOR_VERSION=" + get_minor_version(upgrade_version))
 
-def determine_install_version(pkg_name, pkg_version):
+def determine_search_versions(pkg_name, pkg_version):
 	parsed_version = pkg_version.split('.')
 	major = parsed_version[0]
 	minor = parsed_version[1]
-	# Cause of the origin version schema change we had to manually change the version just for this 
-	# case where we should look for 1.5.x versions if the built version is origin-3.6.x 
-	if pkg_name == "origin" and major == "3" and minor == "6":
-		major = "1"
-	search_version = '.'.join([major, str(int(minor) - 1)])
+	search_install_version = schema_change_check(pkg_name, major, str(int(minor) - 1))
+	search_upgrade_version = schema_change_check(pkg_name, major, minor)
+	return search_install_version, search_upgrade_version
+
+# Cause of the origin version schema change we had to manually change the version just for this
+# case where we should look for 1.5.x versions if the built version is origin-3.6.x
+def schema_change_check(pkg_name, search_major_version, search_minor_version):
+	if pkg_name == "origin" and search_major_version == "3" and int(search_minor_version) <= 5:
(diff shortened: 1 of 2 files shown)


sjb/hack/determine_install_upgrade_version.py
```diff
@@ -51,15 +51,21 @@ def print_version_vars(install_version, upgrade_version):
...
-	if pkg_name == "origin" and major == "3" and minor == "6":
-		major = "1"
-	search_version = '.'.join([major, str(int(minor) - 1)])
+	# case where we should look for 1.5.x versions if the built version is origin-3.6.x
+	search_install_version = schema_change_check(pkg_name, '.'.join([major, str(int(minor) - 1)]))
+	search_upgrade_version = schema_change_check(pkg_name, '.'.join([major, minor]))
+	return search_install_version, search_upgrade_version
+
+def schema_change_check(pkg_name, search_version):
+	search_version_parsed = search_version.split(".")
```
stevekuznetsov: Don't join them just to split them in this method.

```py
def schema_change_check(pkg_name, major, minor):
```

sjb/hack/determine_install_upgrade_version.py
```diff
@@ -96,7 +102,10 @@ def sort_pkgs(available_pkgs):
 	#          ORIGIN_UPGRADE_RELEASE_MINOR_VERSION=5
 	#
 	if args.dependency_branch.startswith("release") or args.dependency_branch.startswith("enterprise"):
-		pkg_version = args.dependency_branch.split("-")[1]
+		input_pkg_version_parsed = pkg_version.split(".")
+		dependency_branch_version_parsed = args.dependency_branch.split("-")[1].split(".")
+		dependency_branch_version_parsed[0] = input_pkg_version_parsed[0]
```
stevekuznetsov: Not following this

jhadvig: So in case the input is:
```
python sjb/hack/determine_install_upgrade_version.py atomic-openshift-utils-3.6.142-1.git.0.90b5f60.el7.noarch --dependency_branch release-1.5
```
It will set the "search version" to `3.5` since, cause we need to check for that version and not the `1.5`, for that reason Im setting the major version, which is the `dependency_branch_version_parsed[0]`, to input package major version, in this case `3.5`

sjb/hack/determine_install_upgrade_version.py
jhadvig: updated :+1:

sjb/hack/determine_install_upgrade_version.py
stevekuznetsov: Add this to the comment above the `if` and use explanatory variables to make it clearer:
```py
if args.dependency_branch.startswith("release") or args.dependency_branch.startswith("enterprise"):
    pkg_major_version = pkg_version.split(".")[0]
    dependency_minor_version = args.dependency_branch.split("-")[1].split(".")[1]
    pkg_version = ".".join(pkg_major_version, dependency_minor_version)
```

jhadvig:
Comments addressed ... @stevekuznetsov  PTAL

PR #425, opened 2017-07-13
extend the sjb syntax description with parent, extensions and overrides keywords description

ingvagabund:
Describing the missing keywords so it is more clear what they are used for.

Signed-off-by: Jan Chaloupka jchaloup@redhat.com


ingvagabund:
@kargakis @stevekuznetsov

ingvagabund:
RIght, PR updated.

stevekuznetsov:
Thanks!

PR #430, opened 2017-07-13
Fetch /metrics from master and kubelet for artifacts

stevekuznetsov:

@deads2k PTAL


sjb/config/common/test_cases/origin.yml
```diff
@@ -91,5 +91,7 @@ generated_artifacts:
   docker.info: 'sudo docker version && sudo docker info && sudo docker images && sudo docker ps -a'
   filesystem.info: 'sudo df -h && sudo pvs && sudo vgs && sudo lvs'
   pid1.journal: 'sudo journalctl _PID=1 --no-pager --all --lines=all'
+  master-metrics.log: 'oc get --raw /metrics --insecure-skip-tls-verify'
+  node-metrics.log: 'oc get --raw /metrics --server=https://localhost:10250 --insecure-skip-tls-verify'
```
stevekuznetsov: We've got the certs but they're not signed for `localhost`, so it's easier to `--insecure-skip-tls-verify` than try to get the hostname because of how we are generating these artifacts right now.

PR #434, opened 2017-07-14
Checkout matching openshift-ansible branch

jhadvig:
@stevekuznetsov PTAL

Running tests with mentioned changes:
- PR against release 1.5 - https://ci.openshift.redhat.com/jenkins/view/All/job/jhadvig/5/
- PR against master        - https://ci.openshift.redhat.com/jenkins/view/All/job/jhadvig/6/

Fixes https://github.com/openshift/origin/issues/15186

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
@@ -23,6 +23,10 @@ extensions:
       script: |-
         tito_tmp_dir="tito"
         mkdir -p "${tito_tmp_dir}"
+        if [[ ${ORIGIN_TARGET_BRANCH} == release-1.* ]]
+        then
+          git checkout -b ${ORIGIN_TARGET_BRANCH} origin/${ORIGIN_TARGET_BRANCH}
+        fi
         tito tag --offline --accept-auto-changelog
         tito build --output="${tito_tmp_dir}" --rpm --test --offline --quiet
         createrepo "${tito_tmp_dir}/noarch"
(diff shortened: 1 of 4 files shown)


sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -23,6 +23,10 @@ extensions:
       script: |-
         tito_tmp_dir="tito"
         mkdir -p "${tito_tmp_dir}"
+        if [[ ${ORIGIN_TARGET_BRANCH} == release-1.* ]]
+        then
+          git checkout -b ${ORIGIN_TARGET_BRANCH} origin/${ORIGIN_TARGET_BRANCH}
```
stevekuznetsov: Isn't there a bunch of logic for translating branch names? Where Origin has `release-1.x` or `release-3.x` and OpenShift Ansible has `enterprise-x`?

jhadvig: I dont think so, cause when you check the origin repo branches you can see what branches are there, in this case only `release-1.x` and `stage-1.x` ... and those reflect to the o-a branches... `enterprise-1.x` branches are only present in OCP

stevekuznetsov: oh, maybe scott moved them

stevekuznetsov: Definitely used to be `enterprise-x`

stevekuznetsov:
I would rather see if we can fix the `oct sync openshift-ansible` step generated by sjb at the start of the job

jhadvig:
Yeah I kinda understand why you want to fix it in the generated step but this is a pretty specific use-case in which we checkout the branch of o-a repository based on the branch against which the PR is sent.  Not sure if this wouldn't be better in the step itself with a proper comment

@stevekuznetsov thoughts ?

sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
jhadvig: Yeah I was surprised as well, cause I remember they have been there

stevekuznetsov:
Any reason why we don't always want to check out the matching branch if it exists? Seems like this would be applicable to `stage`, too (but not `rebase-1.7`)

jhadvig:
Closing in favor of https://github.com/openshift/aos-cd-jobs/pull/434

PR #441, opened 2017-07-18
Fix the origin containerized installation

ingvagabund:
- retag images to v#.#.#[-#.#] form
- set the openshift_image_tag to the closest tag to the current origin commit
- replace rpm -V verification of installed origin to origin image inspection of a given openshift_image_tag

The conformance tests failing due to https://github.com/openshift/origin/issues/15210

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
@@ -12,6 +12,8 @@ extensions:
         hack/build-base-images.sh
         OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=_output/local hack/env OS_ONLY_BUILD_PLATFORMS='linux/amd64' hack/build-rpm-release.sh
         sudo systemctl restart docker
+        git describe --abbrev=0 >> "ORIGIN_TAG"
+        cat ORIGIN_TAG > _output/local/releases/.commit
         hack/build-images.sh
         sed -i 's|go/src|data/src|' _output/local/releases/rpms/origin-local-release.repo
         sudo cp _output/local/releases/rpms/origin-local-release.repo /etc/yum.repos.d/
@@ -58,6 +60,7 @@ extensions:
       script: |-
         jobs_repo="/data/src/github.com/openshift/aos-cd-jobs/"
         git log -1 --pretty=%h >> "${jobs_repo}/ORIGIN_COMMIT"
+        git describe --abbrev=0 >> "${jobs_repo}/ORIGIN_TAG"
         ( source hack/lib/init.sh; os::build::rpm::get_nvra_vars; echo "-${OS_RPM_VERSION}-${OS_RPM_RELEASE}" ) >> "${jobs_repo}/ORIGIN_PKG_VERSION"
     - type: "script"
       title: "install origin"
@@ -76,9 +79,9 @@ extensions:
                          --inventory sjb/inventory/ \
                          -e containerized=true      \
                          -e deployment_type=origin  \
-                         -e openshift_image_tag="$( cat ./ORIGIN_COMMIT )"                      \
-                         -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
-                         -e oreg_url='openshift/origin-${component}:'"$( cat ./ORIGIN_COMMIT )" \
+                         -e openshift_image_tag="$( cat ./ORIGIN_TAG )"                      \
+                         -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"            \
+                         -e oreg_url='openshift/origin-${component}:'"$( cat ./ORIGIN_TAG )" \
                          /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml
     - type: "script"
       title: "expose the kubeconfig"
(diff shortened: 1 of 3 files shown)


ingvagabund:
@stevekuznetsov @jhadvig the cluster was deployed. The job failed on failed e2e tests.

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
```diff
@@ -88,10 +93,11 @@ extensions:
     - type: "script"
       title: "ensure built version of origin is installed"
       timeout: 600
-      repository: "origin"
+      repository: "aos-cd-jobs"
       script: |-
-        origin_package="$( source hack/lib/init.sh; os::build::rpm::format_nvra )"
-        rpm -V "${origin_package}"
```
stevekuznetsov: Why remove this?

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
```diff
@@ -70,15 +71,19 @@ extensions:
                          -e containerized=true      \
                          -e deployment_type=origin  \
                          /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-node/network_manager.yml
+        # add extra tag to docker images in a form of v#.#.#.#
+        for image in $(docker images --format "{{.Repository}}:{{.Tag}}:" | grep ":$( cat ./ORIGIN_COMMIT ):" | cut -d':' -f1 | sort -u); do
+                docker tag ${image}:$( cat ./ORIGIN_COMMIT ) ${image}:$( cat ./ORIGIN_TAG )
```
stevekuznetsov: We didn't like this approach as if it failed, it would silently fail and fall back to pulling the image from a remote registry instead of testing the code we are trying to test.

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
ingvagabund: Since there is no origin rpm installed, just running docker containers.

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
ingvagabund: In that case the building scripts needs to be extended to allow to specify additional extra tag.

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
jhadvig: In fact we are [building rpms](https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml#L13) but @ingvagabund is right that we are not using them since we are using built containers, that we build in "build an origin release" step

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
kargakis: Can you try `OS_BUILD_IMAGE_ARGS="-t yourtag"` and see if that works?

kargakis: When building Origin

kargakis: When building the Origin images

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
kargakis: Well, we need the RPMs to build the images but other than I don't think we have any other use for them here.

ingvagabund: I am thinking of removing them once the images are built to safe some space.

sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml
ingvagabund: Ain't working. That will tag all images with `yourtag` name, not with `${tag}:yourtag`.
Closest thing to do is to override `OS_RELEASE_COMMIT` env. Which is set in `lib/build/archive.sh:143`:
```sh
export OS_RELEASE_COMMIT="$(cat ${OS_OUTPUT_RELEASEPATH}/.commit)"
```

After rewriting the `.commit` file with:
```
cat ../aos-cd-jobs/ORIGIN_TAG > `_output/local/releases/.commit
```

the docker images are properly tagged. I find this minimally hacky. But still, that is something that needs to be fixed in the origin build scripts.

ingvagabund: Removed the docker re-tagging in favour of rewriting the `.commit` file.

ingvagabund:
There are still 5 failing tests which I don't have time to debug. It can be done later. Meantime the failing tests can be skipped if needed. Given the containerized deployment have been failing till now, merging the PR does not break anything. On top of that, the containerized cluster can be deployed now. Just some e2e tests are failing.

@stevekuznetsov let's merge this PR so I can rebase https://github.com/openshift/aos-cd-jobs/pull/442. That one is already working for me.

Issue #437, opened 2017-07-17
Make the `process-pull-requests` (Rosie) calls in the jobs get skipped if $PULL_REFS is set


kargakis:
Extracted from e-mail discussion @stevekuznetsov

stevekuznetsov:
Should be able to do

```sh
if [[ -z "${PULL_REFS:-}" ]]; then
    process-pull-requests ...
fi
```

We'll need to make the change in `sjb/templates/test_case.xml.

kargakis:
Is this template generated from somewhere or we need to do the change directly into it?

kargakis:
Btw I only see `test_pull_requests` being called in there.

stevekuznetsov:
The template is the source of truth, no need to update anywhere. Yes, sorry, `test_pull_requests` is the name of the Ruby script.

PR #448, opened 2017-07-21
Ignore Rosie if PULL_REFS is set

kargakis:
Fixes https://github.com/openshift/aos-cd-jobs/issues/437

### DIFF
--- a/sjb/templates/test_case.xml
+++ b/sjb/templates/test_case.xml
@@ -90,7 +90,9 @@ approve.sh {{ target_repo }} &quot;${{ '{' }}{{ target_repo | replace('-', '_') 
         <hudson.tasks.Shell>
           <command>#!/bin/bash
 set -o errexit -o nounset -o pipefail -o xtrace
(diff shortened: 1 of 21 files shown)


sjb/templates/test_case.xml
```diff
@@ -90,7 +90,9 @@ approve.sh {{ target_repo }} &quot;${{ '{' }}{{ target_repo | replace('-', '_')
         <hudson.tasks.Shell>
           <command>#!/bin/bash
 set -o errexit -o nounset -o pipefail -o xtrace
-test_pull_requests --mark_test_success &quot;${{ '{' }}{{ target_repo | replace('-', '_') | upper }}_PULL_ID}&quot; --repo {{ target_repo }} --config ~/test_pull_requests/{{ target_repo }}.json || true
+if [[ -z "${PULL_REFS:-}" ]]; then
```
stevekuznetsov: you'll need to XML escape these lines too

stevekuznetsov:
Need to regenerate all jobs.

sjb/templates/test_case.xml
kargakis: ok, escaped the quotes. Is there any other characted that I need to escape too?

PR #449, opened 2017-07-21
Checkout matching origin/openshift-ansible branch when syncing repos

jhadvig:
@stevekuznetsov here is the fix for `oct sync ...` which should checkout proper `_TARGET_BRANCH`
So in case of PR to `origin`, OCT will checkout `openshift-ansible` in matching version and vice versa.
response to https://github.com/openshift/aos-cd-jobs/pull/434#issuecomment-315418605

Fixes openshift/origin#15186

### DIFF
--- a/sjb/actions/pull_request_sync.py
+++ b/sjb/actions/pull_request_sync.py
@@ -77,7 +77,7 @@ class PullRequestSyncAction(Action):
 
     def __init__(self, repository):
         self.repository = repository
-        self.parent_sync = SyncAction(repository)
+        self.parent_sync = SyncAction(repository, dependency_repository=repository)
 
     def generate_parameters(self):
         return self.parent_sync.generate_parameters() + [
--- a/sjb/actions/repo_sync.py
+++ b/sjb/actions/repo_sync.py
@@ -12,20 +12,22 @@ _TARGET_BRANCH_PARAMETER_TEMPLATE = Template("""        <hudson.model.StringPara
         </hudson.model.StringParameterDefinition>""")
 
 _SYNC_TITLE_TEMPLATE = Template("SYNC {{ repository | upper }} REPOSITORY")
-_SYNC_ACTION_TEMPLATE = Template("""oct sync remote {{ repository }} --branch "${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}" """)
+_SYNC_ACTION_TEMPLATE = Template("""oct sync remote {{ repository }} --branch "${{ '{' }}{{ dependency_repository | replace('-', '_') | upper }}_TARGET_BRANCH}" """)
 
-_SYNC_DESCRIPTION_TEMPLATE = Template("""Using the <a href="https://github.com/openshift/{{ repository }}/tree/${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}">{{ repository }} ${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}</a> branch.""")
+_SYNC_DESCRIPTION_TEMPLATE = Template("""Using the <a href="https://github.com/openshift/{{ repository }}/tree/${{ '{' }}{{ dependency_repository | replace('-', '_') | upper }}_TARGET_BRANCH}">{{ repository }} ${{ '{' }}{{ dependency_repository | replace('-', '_') | upper }}_TARGET_BRANCH}</a> branch.""")
 
 
 class SyncAction(Action):
     """
     A SyncAction generates a build step that
     synchronizes a repository on the remote
-    host.
+    host and checkout a matching branch based
+    on the dependency repository
     """
 
-    def __init__(self, repository):
+    def __init__(self, repository, dependency_repository):
         self.repository = repository
+        self.dependency_repository = dependency_repository
 
     def generate_parameters(self):
         return [_TARGET_BRANCH_PARAMETER_TEMPLATE.render(repository=self.repository)]
@@ -33,8 +35,8 @@ class SyncAction(Action):
     def generate_build_steps(self):
         return [render_task(
             title=_SYNC_TITLE_TEMPLATE.render(repository=self.repository),
-            command=_SYNC_ACTION_TEMPLATE.render(repository=self.repository)
+            command=_SYNC_ACTION_TEMPLATE.render(repository=self.repository, dependency_repository=self.dependency_repository)
         )]
 
     def description(self):
-        return _SYNC_DESCRIPTION_TEMPLATE.render(repository=self.repository)
+        return _SYNC_DESCRIPTION_TEMPLATE.render(repository=self.repository, dependency_repository=self.dependency_repository)
--- a/sjb/generate.py
+++ b/sjb/generate.py
@@ -99,11 +99,37 @@ if job_type == "test":
 
     # next, repositories will be synced to the remote VM
     sync_actions = []
+    repo_names=[repository["name"] for repository in job_config.get("sync_repos", [])]
+
+    repo_dependencies = {
+        "origin": ["openshift-ansible"],
+        "origin-aggregated-logging": ["openshift-ansible"],
+        "openshift-ansible": ["origin", "origin-aggregated-logging"]
+    }
+
+    is_pull_request = False
+    for repo in job_config.get("sync_repos", []):
+        if repo.get("type") == "pull_request":
+            is_pull_request = True
+
+    # Determine the name of dependency repository that will serve as prefix to 'XXX_TARGET_BRANCH'
+    # variable, so we can set the matching branch for input repository.
+    # Otherwise set the dependency repository to input repository name.
+    def get_parent_repo(dependent_repo_name):
+        if dependent_repo_name in repo_dependencies:
+            for parent in repo_dependencies[dependent_repo_name]:
+                if parent in repo_names:
+                    return parent
+
+        return dependent_repo_name
+
     for repository in job_config.get("sync_repos", []):
         if repository.get("type", None) == "pull_request":
             sync_actions.append(PullRequestSyncAction(repository["name"]))
         else:
-            sync_actions.append(SyncAction(repository["name"]))
+            # if test is a PR, point to dependency repository to the PR's repository branch
+            dependency_repository = get_parent_repo(repository["name"]) if is_pull_request else repository["name"]
+            sync_actions.append(SyncAction(repository["name"], dependency_repository))
 
     if len(sync_actions) > 0:
         actions.append(MultiSyncAction(sync_actions))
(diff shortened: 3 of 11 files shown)


jhadvig:
@kargakis fyi here is the fix for the issue that blocking 1.5 merge queue...

kargakis:
You mean the real fix to your workaround, right? Unfortunately there are two more issues that block release-1.5 merges :(

kargakis:
> Unfortunately there are two more issues that block release-1.5 merges :(

http://post-office.corp.redhat.com/archives/aos-devel/2017-July/msg00736.html

sjb/generate.py
```diff
@@ -99,11 +99,29 @@ def load_configuration(config_path):
...
+        else:
+            return repository_name
+
     for repository in job_config.get("sync_repos", []):
         if repository.get("type", None) == "pull_request":
             sync_actions.append(PullRequestSyncAction(repository["name"]))
         else:
-            sync_actions.append(SyncAction(repository["name"]))
+            # if test is a PR, point to dependency repository to the PR's repository branch
+            repository_target_branch = get_dependency_repo(repository["name"]) if is_pull_request else repository["name"]
```
stevekuznetsov: Resolve the correct branch here, don't pass info down into the actions.

jhadvig: I'm resolving it here... we dont know the branch name until the test is triggered. Upon the generation process we dont know whats the branch name, we have only info about what repos are synced and upon that we can assume what the `_TARGET_BRANCH` should be.
I've removed the additional action and to keep it simple since the `get_dependency_repo` will determine the right `repository_target_branch` ... also not really sure  that it makes sense to check  for the branch name in the `_SYNC_ACTION_TEMPLATE` since there should be always a matching version between `origin` and `o-a`

sjb/actions/repo_sync.py
```diff
@@ -12,9 +12,8 @@
         </hudson.model.StringParameterDefinition>""")
 
 _SYNC_TITLE_TEMPLATE = Template("SYNC {{ repository | upper }} REPOSITORY")
-_SYNC_ACTION_TEMPLATE = Template("""oct sync remote {{ repository }} --branch "${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}" """)
-
-_SYNC_DESCRIPTION_TEMPLATE = Template("""Using the <a href="https://github.com/openshift/{{ repository }}/tree/${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}">{{ repository }} ${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}</a> branch.""")
+_SYNC_ACTION_TEMPLATE = Template("""oct sync remote {{ repository }} --branch "${{ '{' }}{{ repository_target_branch | replace('-', '_') | upper }}_TARGET_BRANCH}" """)
```
stevekuznetsov: `repository_target_branch` is used to create a `${X_TARGET_BRANCH}` var? So it's not a branch, but a repo? This is misleading name -- should have a comment as well in the class doc

jhadvig: yeah thats what the var is for.. sorry my brain was hurting when I was writing it :face_with_head_bandage:  ... will update the name and class docs

sjb/actions/repo_sync.py
```diff
@@ -12,29 +12,27 @@
         </hudson.model.StringParameterDefinition>""")
 
 _SYNC_TITLE_TEMPLATE = Template("SYNC {{ repository | upper }} REPOSITORY")
-_SYNC_ACTION_TEMPLATE = Template("""oct sync remote {{ repository }} --branch "${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}" """)
-
-_SYNC_DESCRIPTION_TEMPLATE = Template("""Using the <a href="https://github.com/openshift/{{ repository }}/tree/${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}">{{ repository }} ${{ '{' }}{{ repository | replace('-', '_') | upper }}_TARGET_BRANCH}</a> branch.""")
+_SYNC_ACTION_TEMPLATE = Template("""oct sync remote {{ repository }} --branch "${{ '{' }}{{ dependency_repository | replace('-', '_') | upper }}_TARGET_BRANCH}" """)
+_SYNC_DESCRIPTION_TEMPLATE = Template("""Using the <a href="https://github.com/openshift/{{ repository }}/tree/${{ '{' }}{{ dependency_repository | replace('-', '_') | upper }}_TARGET_BRANCH}">{{ repository }} ${{ '{' }}{{ dependency_repository | replace('-', '_') | upper }}_TARGET_BRANCH}</a> branch.""")
```
stevekuznetsov: If there is no diff here can we remove the diff please

sjb/actions/pull_request_sync.py
```diff
@@ -77,7 +77,7 @@ class PullRequestSyncAction(Action):
 
     def __init__(self, repository):
         self.repository = repository
-        self.parent_sync = SyncAction(repository)
+        self.parent_sync = SyncAction(repository, repository)
```
stevekuznetsov: Can we use `dependency_repository=repository` to explain what this is here please

sjb/actions/repo_sync.py
```diff
@@ -12,29 +12,27 @@
...
 
     def generate_parameters(self):
         return [_TARGET_BRANCH_PARAMETER_TEMPLATE.render(repository=self.repository)]
 
     def generate_build_steps(self):
-        return [render_task(
-            title=_SYNC_TITLE_TEMPLATE.render(repository=self.repository),
-            command=_SYNC_ACTION_TEMPLATE.render(repository=self.repository)
-        )]
+        return [render_task(title=_SYNC_TITLE_TEMPLATE.render(repository=self.repository),command=_SYNC_ACTION_TEMPLATE.render(repository=self.repository, dependency_repository=self.dependency_repository))]
```
stevekuznetsov: Don't change the formatting here

sjb/generate.py
```diff
@@ -99,11 +99,40 @@ def load_configuration(config_path):
...
+
+    # determine if test is a pull request
+    is_pull_request = False
+    for repository in job_config.get("sync_repos", []):
+        if repository.get("type"):
+            is_pull_request = True
+
+    # Determine the name of dependency repository that will serve as prefix to 'XXX_TARGET_BRANCH'
+    # variable, so we can set the matching branch for input repository.
+    # So far the repository that are dependent to each other are 'origin' and 'openshift-ansible'.
```
stevekuznetsov: Comment is out of date already -- remove it, rest of code is easy enough to follow

sjb/generate.py
```diff
@@ -99,11 +99,40 @@ def load_configuration(config_path):
...
+    # Otherwise set the dependency repository to input repository name.
+    def get_dependency_repo(repository_name):
```
stevekuznetsov: Maybe we could use:


```py
    repo_dependencies = {
        "origin": ["openshift-ansible"],
        "origin-aggregated-logging": ["openshift-ansible"],
        "openshift-ansible": ["origin", "openshift-ansible"]
    }
    
    def get_parent_repo(dependent_repo_name):
        for parent in repo_dependencies[dependent_repo_name]:
            if parent in repo_names:
                return parent
        
        return dependent_repo_name
```

jhadvig:
@stevekuznetsov comments addressed .. PTAL

sjb/generate.py
```diff
@@ -99,11 +99,38 @@ def load_configuration(config_path):
 
     # next, repositories will be synced to the remote VM
     sync_actions = []
+    repo_names=[repository["name"] for repository in job_config.get("sync_repos", [])]
+
+    # determine if test is a pull request
```
stevekuznetsov: This does not matter -- if we are running branch builds and build a release branch of repo `x` we still want the correct branch of other repos as well.

jhadvig:
@stevekuznetsov comments addressed. PTAL

sjb/generated/test_branch_openshift_ansible_logging.xml
```diff
@@ -74,19 +74,19 @@ oct provision remote all-in-one --os &#34;rhel&#34; --stage &#34;build&#34; --pr
...
 &lt;/div&gt;</description>
     </hudson.plugins.descriptionsetter.DescriptionSetterBuilder>
         <hudson.tasks.Shell>
           <command>#!/bin/bash
 SCRIPT_START_TIME=&#34;$( date +%s )&#34; &amp;&amp; export SCRIPT_START_TIME &amp;&amp; echo &#34;########## STARTING STAGE: SYNC ORIGIN-AGGREGATED-LOGGING REPOSITORY ##########&#34; &amp;&amp; trap &#39;export status=FAILURE&#39; ERR &amp;&amp; trap &#39;set +o xtrace; SCRIPT_END_TIME=&#34;$( date +%s )&#34;; ELAPSED_TIME=&#34;$(( SCRIPT_END_TIME - SCRIPT_START_TIME ))&#34;; echo &#34;########## FINISHED STAGE: ${status:-SUCCESS}: SYNC ORIGIN-AGGREGATED-LOGGING REPOSITORY [$( printf &#34;%02dh %02dm %02ds&#34; &#34;$(( ELAPSED_TIME/3600 ))&#34; &#34;$(( (ELAPSED_TIME%3600)/60 ))&#34; &#34;$(( ELAPSED_TIME%60 ))&#34; )] ##########&#34;&#39; EXIT &amp;&amp; set -o errexit -o nounset -o pipefail -o xtrace &amp;&amp; if [[ -s &#34;${WORKSPACE}/activate&#34; ]]; then source &#34;${WORKSPACE}/activate&#34;; fi
-oct sync remote origin-aggregated-logging --branch &#34;${ORIGIN_AGGREGATED_LOGGING_TARGET_BRANCH}&#34; </command>
+oct sync remote origin-aggregated-logging --branch &#34;${OPENSHIFT_ANSIBLE_TARGET_BRANCH}&#34; </command>
```
stevekuznetsov: This job now just switched the branches -- logging uses ansible, and vice versa. This is not what we wanted, is it?

jhadvig: well this is the branch job that I was testing for with the `is_pull_request` variable... I thought that since you told me that it shouldnt matter that we want to checkout the branches the same way as we did with the PR tsts

stevekuznetsov: The logic -- whether it's for a PR or for not -- is wrong, though. We need the branches to converge on one variable, right? Not have them all switch.

jhadvig: No the PR logic should be fine.
eg: `test_pull_request_openshift_ansible_extended_conformance_install`
PR is sent against branch of o-a, so we have to checkout the same branch for origin and thats what the logis is/was (with the `is_pull_request`) doing. So in this case we set the dependency repository to the same branch as the repository's branch that the PR is sent against.

But for the 'branch' TCs I'm not sure how to set the the correct branch since we dont have clue whats the main tested repo and which is the dependency...
For that reason I would rather keep it simple and just cover the PR TCs.

Thoughts ?

stevekuznetsov: I think your approach is valid -- sorry for nixing the PR-only before, it technically should work for the branch jobs too but as you note it is very hard to know which branch should win out.

PR #451, opened 2017-07-21
Switch rotage keys job from buildvm to openshift-build-1

codificat:
buildvm-devops is gone, switching the node of the job to rotate ssh keys for log access.


codificat:
@jupierce PTAL

PR #460, opened 2017-07-25
Cleaner and more verbose failure

adammhaile:
Brew URLs now printed to console
Hard fail on any brew build fail.
Failure logs printed to console if brew fails

### DIFF
--- a/build-scripts/ose_images/ose_images.sh
+++ b/build-scripts/ose_images/ose_images.sh
@@ -395,6 +395,7 @@ check_builds() {
     # 13461148 buildContainer (noarch): free -> open (x86-036.build.eng.bos.redhat.com)
 
     taskid=`cat ${line} | grep -i "Created task:" | head -n 1 | awk '{print $3}'`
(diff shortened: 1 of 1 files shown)


adammhaile:
@jupierce

build-scripts/ose_images/ose_images.sh
```diff
@@ -426,31 +428,32 @@ check_builds() {
 
     if [ "$state" == "open" -o "$state" == "free" ]; then # free state means brew is waiting for a free builder
         echo "brew build for $package is still running..."
-    elif [ "$state" == "closed" ]; then
-        echo "==== ${package} IMAGE COMPLETED ===="
```
stevekuznetsov: misleading diff due to whitespace?

adammhaile: Yeah... even I second guessed myself when I saw the diff. I took the elif/else parts and indented one level so they were their own if/else with a single call to `mv` at the end. I can confirm it's not just a whitespace change.

adammhaile: It's much more clear in the side by side split diff view.

PR #461, opened 2017-07-25
Expand log gathering script with optional node list

codificat:
Allow the user to specify an optional list of nodes from which to collect logs.

This is a minimalist change, aiming to keep the previous functionality intact and only add the additional log collection for node logs: journal for `docker` and `atomic-openshift-node`.

However, some changes have been made to the previous functionality:

- node-related logs (`docker` and `atomic-openshift-node` journal) are *not* collected by default from masters. You can explicitly add the masters as part of the node list to get these logs
- `oc status --all-namespaces` is not collected. I think it's not really useful and is an overhead. I can add it back, but TBH I'd prefer to also remove `oc get {pod,event} --all-namespaces`: on decently sized clusters this might be more problematic than helpful, no?
- using `/var/tmp` instead of `/tmp` for temporary storage, in case data grows a bit too much.

We probably will need more changes to what's being collected, but will wait for feedback about that before changing more.

@jupierce PTAL

### DIFF
--- a/tower-scripts/bin/gather-logs.sh
+++ b/tower-scripts/bin/gather-logs.sh
@@ -14,7 +14,19 @@ NODES="$@"
 MASTER_SERVICES="atomic-openshift-master-api atomic-openshift-master-controllers"
 NODE_SERVICES="docker atomic-openshift-node"
 
-BASEDIR=$(mktemp -p /var/tmp -d gather-logs.XXXXXXXX)
+# Prepare working environment
+TMPDIR="${HOME}/aos-cd/tmp"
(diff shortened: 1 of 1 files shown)


codificat:
There was an existing `rm -rf $BASEDIR` at the end of the script... nonetheless, now I moved that to a trap function and took the opportunity to organize BASEDIR a bit better, matching the `rotate-log-access-key` script's TMPDIR handling. The resulting tarfile now has a nicely named top level dir grouping everything for convenience.

jupierce:
Sorry I missed that. Great work!

PR #468, opened 2017-07-27
Fix inventory name lookup for node log gathering

codificat:
Found a couple of problems in the script's inventory lookup: 

- if you pass a node name like `free-stg` it would match all hosts starting with `free-stg`
- improper handling of the list of hosts

This fixes both.

@jupierce PTAL


jupierce:
Nice catch!

Issue #408, opened 2017-07-07
Investigate breaking down the integration test job into two separate jobs


kargakis:
The integration test job today runs `hack/test-integration.sh` and `hack/test-end-to-end{,-docker}.sh`. While the later test normally takes up to 8min, it may end up hang for more than an hour (see https://github.com/openshift/origin/issues/15093). We need to investigate if we can run the integration test in less than an hour* so it won't make any difference billing-wise to have the additional job.

* also assume that in the future the test will get slower so initially it shouldn't run anywhere near an hour.

PR #472, opened 2017-07-28
create separate jobs for integration and e2e

deads2k:
separate integration tests and e2e tests since the integration ones don't need images.

This creates net new jobs to practice with and ensure success before we change.

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_end_to_end.yml
+++ b/sjb/config/test_cases/test_branch_origin_end_to_end.yml
@@ -1,0 +1,11 @@
+---
+parent: 'common/test_cases/origin_release.yml'
+extensions:
+  actions:
+    - type: "script"
+      title: "run end-to-end tests"
+      repository: "origin"
+      timeout: 7200
+      script: |-
+        OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=_output/local/bin/linux/amd64/etcdhelper hack/env make build WHAT=tools/etcdhelper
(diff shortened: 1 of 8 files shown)


kargakis:
Fixes https://github.com/openshift/aos-cd-jobs/issues/408

deads2k:
> Fixes #408

Does it now. spiffy.  Care to review the matching pull here: https://github.com/openshift/origin/pull/15530 ?

stevekuznetsov:
These LGTM

stevekuznetsov:
@deads2k you need a pull to update prow and submitqueue configmaps too

deads2k:
> @deads2k you need a pull to update prow and submitqueue configmaps too

ok, but I can get these separately, right?

stevekuznetsov:
Not a huge fan of `just_integration` -- renaming it is going to be hairy if we want it to become `integration` later .... can we come up with something better?

deads2k:
> Not a huge fan of just_integration -- renaming it is going to be hairy if we want it to become integration later .... can we come up with something better?

I figured this was easiest path.  
 1. Create two new jobs, don't modify old.  
 2. Test new jobs to make sure they work
 3. activate new jobs.
 4. Remove old job
 5. clone the just_integration job
 6. wait for the next time you're going to reset the list of jobs and switch back.

Keeps the queue operating at each step without a big bang, right?

stevekuznetsov:
It's going to cause a thundering herd of triggers for the integration job when you clone it and add it back in but yeah it should not actually be an issue I don't think

PR #313, opened 2017-06-06
Cover case when target branch is pointing to release branch

jhadvig:
Fixing use case when either openshift-ansible or origin target branch was pointing to a release branch and we ended up installing/upgrading different versions of those two pkgs.
[Example](https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install_update/270/consoleFull): 
```
+ source ORIGIN_VARS
++ ORIGIN_INSTALL_VERSION=1.5.1-1.el7
++ ORIGIN_INSTALL_MINOR_VERSION=5
++ ORIGIN_UPGRADE_RELEASE_VERSION=3.6.0-0.alpha.1.1013.322171b
++ ORIGIN_UPGRADE_RELEASE_MINOR_VERSION=6
+ source OPENSHIFT_ANSIBLE_VARS
++ ATOMIC_OPENSHIFT_UTILS_INSTALL_VERSION=3.4.94-1.git.0.fe1fe27.el7
++ ATOMIC_OPENSHIFT_UTILS_INSTALL_MINOR_VERSION=4
++ ATOMIC_OPENSHIFT_UTILS_UPGRADE_RELEASE_VERSION=3.5.79-1.git.0.f5f10f5.el7
++ ATOMIC_OPENSHIFT_UTILS_UPGRADE_RELEASE_MINOR_VERSION=5
```

For that reason I've added an optional `--dependency_branch` flag which represents the target branch of the dependency repository. So for example if the PR is targeting  `OPENSHIFT_ANSIBLE_TARGET_BRANCH=release-1.5` we have to install `origin-1.4-x` and upgrade to `origin-1.5.x` and vice versa (if ORIGIN_TARGET_BRANCH=release-1.5 the install version of atomic-openshift-utils should be `1.4.x` and the upgrade version `1.5.x`) so the versions are matching.

Updated  `determine_install_upgrade_version.py` script && its TCs && install_udatte job

@stevekuznetsov PTAL

PR #482, opened 2017-08-01
Build origin release even if dependency branch points not to master

jhadvig:
Fixes https://github.com/openshift/origin/issues/15579

Here is a test job https://ci.openshift.redhat.com/jenkins/view/All/job/jhadvig_install_upgrade/1

@stevekuznetsov PTAL

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
@@ -74,17 +74,12 @@ extensions:
       repository: "origin"
       timeout: 5000
       script: |-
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
Why was this in the `if` in the first place?

jhadvig:
@stevekuznetsov so I've added it in https://github.com/openshift/aos-cd-jobs/pull/313 ... I probably thoughts the there is no ned to build the release version for origin if we are pointing to some older branch that already has built rpms, but now I see that I was wrong :cry:

stevekuznetsov:
@jhadvig ok yes always we need to build

PR #487, opened 2017-08-01
Add a test cross build

smarterclayton:
Only upload to GCS for the branch builds (as nightlies)

@stevekuznetsov


stevekuznetsov:
Are these the RPMs we want to gate for openshift-ansible?

smarterclayton:
No, these are the linux/darwin/windows release binaries that go up on GitHub.  Separate thing, this allows someone to test "does origin compile across all platforms" whenever someone starts doing windows / mac specific stuff in the future

PR #491, opened 2017-08-01
Updated message headers to use Java Properties format.

cjryan:


jupierce:
We fixed this another way.

PR #502, opened 2017-08-02
Move JUNIT_REPORT into the hack/env call

smarterclayton:
@stevekuznetsov @deads2k


deads2k:
according to what @smarterclayton tells me about `os::build::environment::run` this works on purpose.

PR #515, opened 2017-08-04
don't push as part of merge

bparees:


bparees:
@openshift/devex fyi, the jenkins image repo merge job will *not* push as part of the merging anymore.  This means your PR will merge and then shortly thereafter the push job will kick off and build/push the new image.  This is more like things used to be, but the latency should be lower (the push job runs every 5 mins, not every hour).

Advantages:

1) merges don't get blocked by push time
2) merges don't fail due to push failures

as long as it passes tests, it merges.

Disadvantages:

if you want to know when your changes have made it into the new docker image, you'll have to watch the push job.

bparees:
(note that previously there was no guarantee the merge pushed your images anyway because if you had passing tests and were the only new commit, the fast-merge process would merge the PR w/o running the merge job, meaning the images weren't pushed at that point anyway.  So we're just down to a single "push" mechanism to pay attention to instead of the previous 2 potential mechanisms)

PR #516, opened 2017-08-04
Stop disabling all excluders in the version script

stevekuznetsov:

### DIFF
--- a/sjb/hack/determine_install_upgrade_version.py
+++ b/sjb/hack/determine_install_upgrade_version.py
@@ -44,7 +44,7 @@ def get_minor_version(pkg_version_release):
 	return pkg_version_release.split('.', 2)[1]
 
 # Print install, upgrade and upgrade_release versions of desired package to STDOUT.
-# The upgrade version will be printed only in case there is more then one version of packages previous minor release. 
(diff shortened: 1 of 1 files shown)


sjb/hack/determine_install_upgrade_version.py
```diff
@@ -116,12 +116,9 @@ def sort_pkgs(available_pkgs):
 
 	yb = yum.YumBase()
 	# Have to redirect redirect the yum configuration to /dev/null cause we dont want to have any additional output
-	# that configuration prints. eg: `Loaded plugins: amazon-id, rhui-lb`  
+	# that configuration prints. eg: `Loaded plugins: amazon-id, rhui-lb`
 	old_stdout = sys.stdout
 	sys.stdout = open(os.devnull, 'w')
-	# Need make sure that all the packages are listed, cause if excluders are enabled for some reason,
-	# listing of origin or docker pkgs will be disabled, based on the enabled excluder.
-	yb.conf.disable_excludes = ["all"]
```
stevekuznetsov: @jhadvig I'm pretty much 100% sure that this was only necessary when the excluders were all sorts of broken. Both of the times we run this script before we run any ansible installation, so we should be fine to not need this now,

PR #518, opened 2017-08-04
Add a job for launching a cluster for two hours

smarterclayton:
Cluster can be used for debugging. Intended to be a new flag someone can use in prow

    /test launch-gce


smarterclayton:
@stevekuznetsov this is to make is so people don't even have to ask to get access to a cluster.

stevekuznetsov:
@smarterclayton awesome. Can you dovetail this with a PR to ci-config that explain the `gcloud compute describe-instances` and `gcloud compute ssh` commands that they would need to use to connect to the master?`

smarterclayton:
Yeah, I'm debating how much to print here.  I couldn't come up with a really safe way to send someone a secret, so the ssh will have to do for now.

PR #530, opened 2017-08-07
Run Oh the system container, Run!

ingvagabund:
Let's do it.

### DIFF
--- a/sjb/config/common/test_suites/openshift_ansible.yml
+++ b/sjb/config/common/test_suites/openshift_ansible.yml
@@ -4,3 +4,4 @@ children:
   - "test_pull_request_openshift_ansible_logging"
   - "test_pull_request_openshift_ansible_extended_conformance_install_with_status_check"
   - "test_pull_request_openshift_ansible_extended_conformance_install_update"
+  - "test_pull_request_openshift_ansible_extended_conformance_install_system_containers.yml"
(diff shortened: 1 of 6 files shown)


ingvagabund:
@sdodson @giuseppe @ashcrow @stevekuznetsov @jhadvig

ingvagabund:
@stevekuznetsov this needs to be run with OpenShift 3.6+. When the PR merges, will it affect 3.5 and less as well?

stevekuznetsov:
Tests for all branches will run the full suite of tests, yes. How common are 3.5 backports?

sdodson:
We won't be backporting system container support to 3.5.

ingvagabund:
> How common are 3.5 backports?

From time to time there are some. Well, we can always check if all tests passed up to the system containers. However, the [test]/[merge] will always fail. So we would have to run [test] and then merge manually.

stevekuznetsov:
@ingvagabund you can add a check to the job -- exit early with success:

```sh
if [[ "${OPENSHIFT_ANSIBLE_TARGET_BRANCH:-}" == "enterprise-3.5" ]]; then
    echo "[INFO] System containers are not supported, skipping."
    exit 0
fi
```

stevekuznetsov:
@ingvagabund you'll need to test if we need that in one stage or in all stages.

ingvagabund:
Not working currently, need to debug it before we merge it

ingvagabund:
@stevekuznetsov feel free to merge it

PR #536, opened 2017-08-07
sign rhaos RPMs before puddle build

adammhaile:
@jupierce I left off the `-candidate` as that didn't sign all the RPMs we needed with the openshift scripts build. This will it will definitely sign everything, though likely at the cost of some decent added time. We can likely just optimize and call for specific RPMs later if we need to speed it up.


jupierce:
@adammhaile How much time are we talking?

adammhaile:
When I ran this previously it took up to 25 minutes to sign everything with
rhaos-<version>-rhel-7  (no -candidate). But it really depended on if
anything had already been signed. If you ran it a second time it would only
sign packages that weren't. So testing `-candidate` vs regular was
difficult because most were already signed from the first run.

PR #542, opened 2017-08-08
call_puddle now adds sig to puddle confs

adammhaile:
@jupierce


build-scripts/rcm-guest/call_puddle.sh
```diff
@@ -2,24 +2,68 @@
...
 fi
 
+
 CONF_FILE=$(mktemp)
-curl -o "${CONF_FILE}" "${1}"
+curl -o "${CONF_FILE}" "${CONF_FILE_URL}"
 
-shift
+if [ -n "${SIGNED_KEYS}" ]; then
+  echo "$(grep "(keys[ \t]*=[ \t]*)\|(signed[ \t]*=[ \t]*)" --invert-match "${CONF_FILE}")" > "${CONF_FILE}"
```
jupierce: I don't think it is same to use the same file. 
https://stackoverflow.com/questions/6696842/bash-redirect-input-from-file-back-into-same-file

Maybe mktemp twice.. CONF_TEMPLATE and CONF_FILE

jobs/build/ocp/Jenkinsfile
```diff
@@ -378,14 +373,29 @@ node(TARGET_NODE) {
...
-                        PUDDLE_CONF,    // The puddle configuration file to use
-                        "-b",   // do not fail if we are missing dependencies
-                        "-d",   // print debug information
-                        "-n",   // do not send an email for this puddle
-                        "-s",   // do not create a "latest" link since this puddle is for building images
-                        "--label=building"   // create a symlink named "building" for the puddle
-                )
+                if ( SIGN_RPMS ) {
+                  buildlib.build_puddle(
+                          "--conf"
```
jupierce: Does not align with the parameters of buildlib.build_puddle. Happens to work, but we should not need to specify --conf, or --keys for each invocation. Keep things DRY as well.

```
keys = SIGN_RPMS?"b906ba72":null
buildlib.build_puddle( PUDDLE_CONF, keys, -b, .....)
```

This will require a signature change to build_puddle method. build_puddle( conf, keys, Object... args). Method implementation could include "--conf" and "--keys" if parameter is not null.

pipeline-scripts/buildlib.groovy
```diff
@@ -271,7 +271,7 @@ def build_puddle(conf_url, Object...args) {
     // Ideally, we would call invoke_on_rcm_guest, but jenkins makes it absurd to invoke with conf_url as one of the arguments because the spread operator is not enabled.
     def puddle_output = sh(
             returnStdout: true,
-            script: "ssh ocp-build@rcm-guest.app.eng.bos.redhat.com sh -s ${conf_url} ${this.args_to_string(args)} < ${env.WORKSPACE}/build-scripts/rcm-guest/call_puddle.sh",
+            script: "ssh ocp-build@rcm-guest.app.eng.bos.redhat.com sh -s ${this.args_to_string(args)} < ${env.WORKSPACE}/build-scripts/rcm-guest/call_puddle.sh",
```
jupierce: Encouraging the signature & implementation change here. In any case, it looks like the first parameter (conf_url) is always lost in this implementation.

build-scripts/rcm-guest/call_puddle.sh
adammhaile: Weird, unlike a lot of our stuff that's hard to test, this script was easy as I could just remove the puddle call. And it worked fine every time at least locally. But just in case, I'll add the second mktemp, just to be extra sure.

adammhaile:
@jupierce All changes have been made on this.

Issue #538, opened 2017-08-07
Clean up push-to-mirrors.sh


jupierce:
@adammhaile I recently updated push-openshift-online-to-mirrors.sh to improve variable naming and simplify it:

https://github.com/openshift/aos-cd-jobs/blob/master/build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh

Compare to what it would have looked like originally (before renaming): https://github.com/openshift/aos-cd-jobs/blob/70c3488790dde633c9b8c80b3667650d67872107/build-scripts/rcm-guest/libra-repo-to-mirrors.sh

Can you do this for https://github.com/openshift/aos-cd-jobs/blob/70c3488790dde633c9b8c80b3667650d67872107/build-scripts/rcm-guest/push-to-mirrors.sh ?

Using the EOF trick (https://github.com/openshift/aos-cd-jobs/blob/master/build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh#L56) can really make the script easier to read.

PR #550, opened 2017-08-09
Cleanup push-to-mirros.sh

adammhaile:
Issue #538 
@jupierce

### DIFF
--- a/build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh
+++ b/build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh
@@ -22,7 +22,7 @@ if [ "$BUILD_MODE" == "online:int" ]; then
 	REPO="online-openshift-scripts-int"
 elif [ "$BUILD_MODE" == "online:stg" ]; then
 	REPO="online-openshift-scripts-stg"
-elif [ "$BUILD_MODE" == "release" -o "$BUILD_MODE" == "pre-release" ]; then
+elif [ "$BUILD_MODE" == "release" ] || [ "$BUILD_MODE" == "pre-release" ]; then
 	REPO="online-openshift-scripts-${VERSION}"
 else
 	echo "Unknown build mode: $BUILD_MODE"
@@ -78,4 +78,3 @@ $MIRROR_SSH sh -s <<-EOF
 	ln -sfn "${LASTDIR}" "latest"  # replace the old "latest" symlink with one pointing to the newly copied puddle
 	/usr/local/bin/push.enterprise.sh "${REPO}" -v
 EOF
-
--- a/build-scripts/rcm-guest/push-to-mirrors.sh
+++ b/build-scripts/rcm-guest/push-to-mirrors.sh
@@ -4,6 +4,7 @@
 #
 
 set -o xtrace
+set -e
 
 ############
 # VARIABLES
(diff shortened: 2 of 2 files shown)


build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh
```diff
@@ -22,7 +22,7 @@ if [ "$BUILD_MODE" == "online:int" ]; then
 	REPO="online-openshift-scripts-int"
 elif [ "$BUILD_MODE" == "online:stg" ]; then
 	REPO="online-openshift-scripts-stg"
-elif [ "$BUILD_MODE" == "release" -o "$BUILD_MODE" == "pre-release" ]; then
+elif [ "$BUILD_MODE" == "release" ] || [ "$BUILD_MODE" == "pre-release" ]; then
```
stevekuznetsov: If you do not need to use `#!/bin/sh` or POSIX compliance... use `[[` as it's going to make this less annoying:

```sh
elif [[ "$BUILD_MODE" == "release" || "$BUILD_MODE" == "pre-release" ]]; then
```

stevekuznetsov: (and `[[` generally has fewer caveats and edge cases than `[`)

build-scripts/rcm-guest/push-to-mirrors.sh
```diff
@@ -68,28 +69,45 @@ LASTDIR=$(readlink ${PUDDLEDIR})
...
+MIRROR_SSH_SERVER="use-mirror-upload.ops.rhcloud.com"
+MIRROR_SSH_BASE="ssh ${BOT_USER} -o StrictHostKeychecking=no"
+MIRROR_SSH="${MIRROR_SSH_BASE} ${MIRROR_SSH_SERVER}"
+MIRROR_PATH="/srv/enterprise/${REPO}"
+ALL_DIR="/srv/enterprise/all/${MAJOR_MINOR}"
 
-# Replace latest link with new puddle content
-ssh ${BOT_USER} -o StrictHostKeychecking=no use-mirror-upload.ops.rhcloud.com "cd /srv/enterprise/${REPO} ; ln -sfn $LASTDIR latest"
+$MIRROR_SSH sh -s <<-EOF
+  set -e
```
stevekuznetsov: IMO here-docs with code in them are an anti-pattern. You don't get syntax highlighting in there, you have to worry about where and how variables expand, etc. If it's possible, maybe it would be cleaner to get all the remote logic in one place (forward params as necessary?) and use a separate file you can `scp` over and execute on the remote host.

build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh
adammhaile: Fair enough... was just doing what my linter told me to...

build-scripts/rcm-guest/push-to-mirrors.sh
adammhaile: @jupierce specifically asked for this.

stevekuznetsov: @jupierce how could you? :(

adammhaile:
@stevekuznetsov @jupierce 
Per IRC convo, `set -e` works perfectly fine inside of the here-doc. Tested with:

```
#!/bin/bash

MIRROR_SSH_SERVER="baby-build"
MIRROR_SSH="ssh ${MIRROR_SSH_SERVER}"

$MIRROR_SSH sh -s <<-EOF
  set -e
  set -o xtrace
  echo "Hello"
  achoo "Error"
  echo "World"
EOF
```
I tried this with and without `set -e` and with and without `bash -e` in the ssh call. As long as `-e` is in one or both of those places it reacts exact the same.

stevekuznetsov:
Ah ... maybe difference is `ssh sh <<EOF` versus just `ssh <<EOF`?

stevekuznetsov:
In any case -- who cares!! So many caveats that can bite you :(

adammhaile:
Probably so.

PR #555, opened 2017-08-10
Add documentation for cluster log gathering

codificat:
This adds some usage instructions around the log gathering script for online clusters.

### DIFF
--- a/docs/sop/gather-cluster-logs.md
+++ b/docs/sop/gather-cluster-logs.md
@@ -1,0 +1,136 @@
+# Collect logs from Online clusters
+
+## Synopsis
+
+````
+  ssh -i online/logs_access_key opsmedic@use-tower2.ops.rhcloud.com \
+      -- -u USER -c CLUSTER [-n node1 node2...] > logs.tar.gz
+````
+
+**NOTE**: the collected logs are dumped to the standard output as a compressed
+tarball, so **you must redirect the output of ssh** to a file of your choice.
+
+## SSH key to access logs
+
+Access to logs is done via ssh to a bastion host as the *opsmedic* user with a
+specific SSH key that is used for this purpose only. This SSH key is recreated
+on a weekly basis and stored in the *rotating_keys* branch of the
+[shared-secrets repo](https://github.com/openshift/shared-secrets/tree/rotating_keys).
+
+## Information collected
+
+The [log gathering script](../../tower-scripts/bin/gather-logs.sh) collects:
+
+* Master info:
+
+   - journal for these services: `atomic-openshift-master-{api,controllers}` and `etcd`
+   - The output of `oc get node`
+   - The output of the `/metrics` endpoint
+
+* Selected nodes info:
+
+   - journal for these services: docker, atomic-openshift-node, dnsmasq, openvswitch, ovs-vswitchd, ovsdb-server
+   - The output of `oc describe node` for that node
+   - The output of the `/metrics` endpoint for that node (via API proxy)
+
+## Command usage
+
+The log gathering command accepts 3 parameters:
+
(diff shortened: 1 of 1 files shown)


docs/sop/gather-cluster-logs.md
```diff
@@ -0,0 +1,136 @@
...
+    $ cd /path/to/your/cloned/shared-secrets
+    $ git checkout rotating_keys
+    $ git pull
+    $ chmod 600 online/logs_access_key
+    ````
+
+ 2. Collect logs from the cluster/nodes of choice. For example:
+
+    ````
+    $ ssh -i online/logs_access_key opsmedic@use-tower2.ops.rhcloud.com \
```
jupierce: @codificat This seems like something we could help the developer with. A shell script in the same directory as the key which wraps the ssh invocation. e.g.:

online/gather-logs.sh
```
#!/bin/bash
if [ ! -f log_access_key ]; then
    echo "....must run from correct directory..." >&2
fi
ssh -i log_access_key opsmedic@use-tower2.ops.rhcloud.com -- $@
```

That would enable us to change things around a bit (e.g. hosts/users and even parameters if we had to). 

I realize this is a change request because you've implemented precisely what I had originally requested :)

jupierce: The script could also set the permissions appropriately on the key file if it exists.

codificat: Yes, that's a good idea. The script could also help handle output redirection (e.g. accept an optional destination file, create a temp one if not, redirect ssh output in any case)

docs/sop/gather-cluster-logs.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Synopsis
+
+````
+  ssh -i online/logs_access_key opsmedic@use-tower2.ops.rhcloud.com \
+      -- -u USER -c CLUSTER [-n node1 node2...] > logs.tar.gz
+````
+
+**NOTE**: the collected logs are dumped to the standard output as a compressed
+tarball, so **you must redirect the output of ssh** to a file of your choice.
```
dbaker-rh: Correction: you *should* redirect ... to a file.

(technically, you might also want to pipe the output into tar to unpack the files locally)

docs/sop/gather-cluster-logs.md
```diff
@@ -0,0 +1,136 @@
...
+- `-u username` (**mandatory**): specify your kerberos id here.
+
+- `-c cluster` (**mandatory**): which cluster to collect logs from. Currently the list
+     of accepted clusters is:
+
+     - free-int
+     - free-stg
+     - starter-us-east-1
+     - starter-us-east-2
+     - starter-us-west-2
```
dbaker-rh: We have starter-us-west-1 as well now.  Suggest we enable that in the configuration, and update this text to include it, or update text to be descriptive and not literal.

docs/sop/gather-cluster-logs.md
```diff
@@ -0,0 +1,136 @@
...
+
+   - journal for these services: docker, atomic-openshift-node, dnsmasq, openvswitch, ovs-vswitchd, ovsdb-server
+   - The output of `oc describe node` for that node
+   - The output of the `/metrics` endpoint for that node (via API proxy)
+
+## Command usage
+
+The log gathering command accepts 3 parameters:
+
+- `-u username` (**mandatory**): specify your kerberos id here.
```
dbaker-rh: Just a comment on terminology.  "mandatory", or "required"?

They're synonymous in this context but I don't recall "mandatory" being used in man pages, etc for this meaning.

codificat:
Thanks for the feedback.

Comments addressed, and doc updated to use the helper script.

PR #556, opened 2017-08-10
added dist_git_copy option

adammhaile:
@jupierce

### DIFF
--- a/build-scripts/ose_images/ose_images.sh
+++ b/build-scripts/ose_images/ose_images.sh
@@ -650,6 +650,22 @@ merge_to_newest_dist_git() {
   popd >/dev/null
 }
 
+overwrite_dist_git_branch(){
+  pushd "${workingdir}/${container}" >/dev/null
+  echo "${branch}"
+  source_branch_files=$(mktemp -d)
+  rhpkg switch-branch "${branch}"
+  cp -r ./ "${source_branch_files}"
+  ls -al "${source_branch_files}"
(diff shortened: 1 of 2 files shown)


adammhaile:
Usage: ose_images.sh dist_git_copy --branch rhaos-3.7-rhel-7 --source_branch rhaos-3.6-rhel-7 --group base

jupierce:
@adammhaile This is a nit given how much I suspect we will use this, but since ose_images.sh determines groups based on the version in --branch, does it make sense to reverse these arguments?   `--branch rhaos-3.6-rhel-7   and --target_branch rhaos-3.7-rhel-7` ?

In the current version, I think if you did `dist_git_copy --branch rhaos-3.7-rhel-7 --source_branch=rhaos-3.3-rhel-7`  then you would find that it did not select the group specific to 3.3.

adammhaile:
@jupierce Yeah, I guess it just all depends if you want to resolve groups based on what we are pulling from or what we are pushing to. So yes, right now it would determine groups from --branch. Easy enough to flip. Will do.

adammhaile:
@jupierce Side note... this should have zero consequence for the updates I did yesterday because ose_images.sh doesn't have any specifics for 3.6 or 3.7... they are just handled in the group building code as `else` statements. So running it flipped would've been identical.

jupierce:
@adammhaile Agreed. I think 3.7 branches are ready.

adammhaile:
@jupierce Has been updated as suggested.

PR #560, opened 2017-08-11
stop pushing the jenkins v1 image

bparees:


bparees:
@openshift/devex fyi.

PR #562, opened 2017-08-12
use the ose image as the base image for jenkins images

bparees:


sjb/config/common/test_cases/rhel_base_images.yml
```diff
@@ -41,7 +41,7 @@ extensions:
 
         # create Dockerfile
         cat <<EOF > $contextdir/Dockerfile
-        FROM registry.access.redhat.com/rhel7.3:latest
+        FROM brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/openshift3/ose:latest
```
bparees: this might be overly optimistic that our CI machines have access to this, but i guess we'll find out.

PR #563, opened 2017-08-12
Revert "use the ose image as the base image for jenkins images"

bparees:
This reverts commit 2b757ee2894e430f85b8e6e242d83c2b37485225.


bparees:
brew isn't available from the CI environment.  will need to decide on a plan b.

PR #564, opened 2017-08-12
Gather origin-master-api and origin-master-controllers logs as well

smarterclayton:
We changed single node to use two processes instead of the all-in-one. Gather the old one as well for historical purposes


smarterclayton:
Need to get logs from masters for the current failing master branch.

stevekuznetsov:
Do we need this on the AWS boxes too?

PR #578, opened 2017-08-22
Blacklisting atomic-openshift-utils-3.6.173.0.7-1.git.0.f6f19ed.el7

jhadvig:
@sdodson @jupierce @stevekuznetsov here is the fix for the issue installing origin, caused by the buggy `atomic-openshift-utils-3.6.173.0.7-1.git.0.f6f19ed.el7`


jhadvig:
I've tested the change and the installation passed without any failures

stevekuznetsov:
Why?

stevekuznetsov:
Answer to why is that it would be prohibitively long to get the new one built.

PR #582, opened 2017-08-23
updater.py: long-lived branches, don't force-push

bbguimaraes:
@jupierce, @stevekuznetsov: PTAL.

### DIFF
--- a/aos_cd_jobs/updater.py
+++ b/aos_cd_jobs/updater.py
@@ -1,14 +1,14 @@
 #!/usr/bin/env python
-from os import listdir, rename, walk
-from os.path import join, relpath
+from os import listdir, remove, rename, walk
+from os.path import isdir, join, relpath
 from shutil import rmtree
 
 from aos_cd_jobs.common import JOBS_DIRECTORY, initialize_repo
 
 def update_branches(repo):
     for job in list_jobs(repo):
-        if job in repo.branches:
-            repo.branches[job].delete()
+        if job not in repo.branches:
+            repo.create_head(job, 'master')
         create_remote_branch(repo, job)
 
 def list_jobs(repo):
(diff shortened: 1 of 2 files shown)


aos_cd_jobs/updater.py
```diff
@@ -20,25 +18,42 @@ def list_jobs(repo):
...
 
-def initialize_orphan_branch(repo, name):
-    repo.heads.master.checkout(orphan=name)
-
-def populate_branch(repo, name):
-    directory = join(repo.working_dir, JOBS_DIRECTORY, name)
-    create_job_file_tree(repo.working_dir, directory)
+def checkout_branch(repo, name):
+    branch = getattr(repo.branches, name, None)
+    if branch is None:
```
stevekuznetsov: When does this happen?

aos_cd_jobs/updater_test.py
```diff
@@ -50,21 +46,27 @@ def test_update_branches(self, create_mock):
...
-        initialize_orphan_branch(repo, name)
-        repo.heads.master.checkout.assert_called_once_with(orphan=name)
+        repo.branches.job0 = Mock()
+        self.assertEqual(checkout_branch(repo, 'job0'), repo.branches.job0)
 
+    def test_checkout_branch_new(self):
+        repo = Mock()
+        repo.branches.job0 = None
+        branch = checkout_branch(repo, 'job0')
+        repo.create_head.assert_called_once_with('job0', 'master')
```
stevekuznetsov: Why does checking out `job0` expect to create head on master?

aos_cd_jobs/updater.py
bbguimaraes: When a new job is added (for which a branch didn't exist).

stevekuznetsov: I'd suggest to make the logic more explicit instead of having the fallback, create branch if it doesn't exist before calling into this one

bbguimaraes:
@stevekuznetsov: updated.

stevekuznetsov:
This LGTM. @jupierce when do we roll it out?

jupierce:
Thanks @bbguimaraes / @stevekuznetsov . LGTM. Let's give it a try!

PR #588, opened 2017-08-25
remove unnecessary docker volumes to free disk space

ingvagabund:
If a VM disk space is not big enough the docker volumes that were
used to build the origin images may lead to kubelet disk pressure.

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_install_system_containers.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_install_system_containers.yml
@@ -22,6 +22,8 @@ extensions:
         hack/build-images.sh
         sed -i 's|go/src|data/src|' _output/local/releases/rpms/origin-local-release.repo
         sudo cp _output/local/releases/rpms/origin-local-release.repo /etc/yum.repos.d/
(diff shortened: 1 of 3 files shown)


ingvagabund:
Testing right now

ingvagabund:
@stevekuznetsov @kargakis

ingvagabund:
The command succeeded. Feel free to merge it anytime.

sdodson:
These volume are used when building origin but unused after that?

ingvagabund:
That's right.

stevekuznetsov:
I would expect you to deliver this as a flag to havk/env and only set the flag in this repo.

ingvagabund:
That needs to be merged into origin first. We need to fix the CI first.

stevekuznetsov:
Is this causing breakage? I wasn't aware

sdodson:
It's suspected to be the root cause of why the system containers test is
taking 4-5hrs and failing more often than not.

stevekuznetsov:
Let's turn off that test and approach this the clean, non-hackey way.

ingvagabund:
Moving the command to origin hack scripts: https://github.com/openshift/origin/pull/16073

PR #597, opened 2017-08-25
Move openshift3/ose to push to the end

jupierce:
Closes the window between when we push the openshift3/ose image and when we have the latest puddle setup.


jupierce:
@sdodson fyi

sdodson:
thanks!

PR #603, opened 2017-08-30
logging - reduce es memory limit for testing

richm:
Use a 4Gi memory limit for logging testing.  This will allow us
to test with https://github.com/openshift/openshift-ansible/pull/5119
Once this PR merges, then we can merge 5119 and logging tests should
be able to resume.
@stevekuznetsov @portante @jcantrill @sdodson ptal

### DIFF
--- a/sjb/inventory/group_vars/OSEv3/logging.yml
+++ b/sjb/inventory/group_vars/OSEv3/logging.yml
@@ -1,8 +1,8 @@
 ---
 openshift_logging_use_mux: false
-openshift_logging_use_mux_client: false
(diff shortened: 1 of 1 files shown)


sdodson:
So the default in openshift-ansible would still be 8GiB but we'll override it to 4GiB here? Do we have a cluster size that we feel 4GiB is acceptable for or is this truly limited to the scope of a cluster sized to just run CI tests?

stevekuznetsov:
please place into `sjb/inventory/group_vars/OSEv3/logging.yml`

richm:
@stevekuznetsov done

PR #608, opened 2017-09-01
add new team members to email list for devex jobs

bparees:


bparees:
@legionus @miminar @dmage fyi, these are jobs the devex team needs to monitor for failure.  In general unless you are the test case babysitter for the sprint, you can ignore them, though.

PR #610, opened 2017-09-05
Add cluster smoke test job and run it during upgrades

codificat:
This adds a new job to run smoke tests on a cluster, and calls that job during the upgrade job to run a smoke test after a cluster has been upgraded.

The PR also contains some minor changes to the smoke test script itself - one of them is actually required: `oc get pod | grep Running` is prone to races resulting on a script failure, and instead of adding another `sleep` this uses `oc rollout status -w` instead which should be more reliable and appropriate.

Also taking the opportunity to update an unrelated job: just fixing the destination address for notifications of key rotation failures.

### DIFF
--- a/jobs/starter/smoke-test/Jenkinsfile
+++ b/jobs/starter/smoke-test/Jenkinsfile
@@ -1,0 +1,92 @@
+#!/usr/bin/env groovy
+
+// A Job to run a smoke test in a cluster
+// 
+// It invokes the 'smoketest' operation of 'cicd-control.sh' for the specified cluster.
+// Optionally emails test results.
+
+@Library('aos_cd_ops') _
+
+// Ask the shared library which clusters this job should act on
+cluster_choice = aos_cd_ops_data.getClusterList("${env.BRANCH_NAME}").join("\n")  // Jenkins expects choice parameter to be linefeed delimited
+
+properties([
+        [   $class: 'ParametersDefinitionProperty',
+            parameterDefinitions: [
+                [
+                    $class: 'hudson.model.ChoiceParameterDefinition',
+                    choices: "${cluster_choice}",
+                    name: 'CLUSTER_SPEC',
+                    description: 'Cluster to run tests on',
+                ],
+                [
+                    $class: 'hudson.model.BooleanParameterDefinition',
+                    defaultValue: false,
+                    description: 'Send email notifications',
+                    name: 'MAIL_RESULTS'
+                ],
+                [
+                    $class: 'hudson.model.StringParameterDefinition',
+                    defaultValue: 'aos-devel@redhat.com',
+                    description: 'Success Mailing List',
+                    name: 'MAIL_LIST_SUCCESS'
+                ],
+                [
+                    $class: 'hudson.model.StringParameterDefinition',
+                    defaultValue: 'aos-qe@redhat.com',
+                    description: 'Failure Mailing List',
+                    name: 'MAIL_LIST_FAILURE'
+                ],
+            ]
+        ],
+    ]
+)
+
+node('openshift-build-1') {
+
+    def deploylib = load( "pipeline-scripts/deploylib.groovy" )
+    deploylib.initialize(CLUSTER_SPEC)
+
+    try {
+
+        stage( 'smoketest' ) {
+            sshagent([CLUSTER_ENV]) {
+                smoketest = deploylib.run("smoketest", null, true)
+                echo smoketest
+            }
+        }
+
+        if ( MAIL_RESULTS.toBoolean() ) {
+            mail(
+                to: "${MAIL_LIST_SUCCESS}",
+                from: "aos-cd@redhat.com",
+                replyTo: 'jupierce@redhat.com',
+                subject: "Cluster smoke test succeeded: ${CLUSTER_NAME}",
+                body: """\
+Jenkins job: ${env.BUILD_URL}
+
+Smoke test output:
+${smoketest}
+""");
+
+        }
+
+    } catch ( err ) {
+        if ( MAIL_RESULTS.toBoolean() ) {
+            mail(to: "${MAIL_LIST_FAILURE}",
+                from: "aos-cd@redhat.com",
+                subject: "Error during smoke test on cluster ${CLUSTER_NAME}",
+                body: """Encountered an error: ${err}
+
+Jenkins job: ${env.BUILD_URL}
+
+Smoke test output:
+${smoketest}
+""");
+        }
+
+        // Re-throw the error in order to fail the job
+        throw err
+    }
+
+}
--- a/jobs/starter/upgrade/Jenkinsfile
+++ b/jobs/starter/upgrade/Jenkinsfile
@@ -1,16 +1,18 @@
 #!/usr/bin/env groovy
 
-def mail_success(list,detail) {
+def mail_success(list,detail,warn) {
+    body = "Cluster ${CLUSTER_NAME} upgrade details:\n"
+    if ( warn ) {
+        body += "\nWARNING: post-upgrade smoke test was not successful: ${warn}\n"
+    }
+    body += "\n${detail}\n\nJenkins job: ${env.BUILD_URL}\n"
     mail(
         to: "${list}",
         from: "aos-cd@redhat.com",
         replyTo: 'jupierce@redhat.com',
         subject: "[aos-devel] Cluster upgrade complete: ${CLUSTER_NAME}",
-        body: """\
-${detail}
-
-Jenkins job: ${env.BUILD_URL}
-""");
+        body: body
+    );
 }
 
 @Library('aos_cd_ops') _
@@ -61,7 +63,7 @@ node('openshift-build-1') {
                 echo "Cluster status BEFORE upgrade:"
                 deploylib.run("status")
             }
-            
+
             stage( "enable maintenance" ) {
                 // deploylib.run( "enable-statuspage" )
                 deploylib.run( "enable-zabbix-maint" )
@@ -80,7 +82,7 @@ node('openshift-build-1') {
                 deploylib.run( "enable-config-loop" )
                 deploylib.run( "run-config-loop" )
             }
-            
+
             stage( "disable maintenance" ) {
                 deploylib.run( "disable-zabbix-maint" )
                 //deploylib.run( "disable-statuspage" )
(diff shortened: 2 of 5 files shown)


jobs/cluster/rotate-log-access-key/Jenkinsfile
```diff
@@ -43,7 +43,7 @@ node('openshift-build-1') {
                     ],
                     [
                         $class: 'hudson.model.StringParameterDefinition',
-                        defaultValue: 'pep@redhat.com',
+                        defaultValue: 'jupierce@redhat.com,pep@redhat.com',
```
codificat: This change is unrelated to the topic of the PR, but thought I'd hijack the PR a bit to include just this minor fix...

jobs/cluster/smoke-test/Jenkinsfile
```diff
@@ -0,0 +1,86 @@
...
+                    [
+                        $class: 'hudson.model.StringParameterDefinition',
+                        name: 'CLUSTER_NAME',
+                        description: 'Cluster to run tests on',
+                    ],
+                    [
+                        $class: 'hudson.model.BooleanParameterDefinition',
+                        defaultValue: false,
+                        description: 'Send email notifications',
+                        name: 'MAIL_RESULTS'
```
codificat: So: this disables email notifications by default (i.e. no emails would be sent on success or failure). Should we have a default of `true` here? I set it to `false` because the nature of the job seems to me it might not be worth sending notifications for a single smoke test run...

Moreover: the upgrade job calls this one with email notifications explicitly disabled too, because it handles its own email notifications.

jobs/cluster/upgrade/Jenkinsfile
```diff
@@ -85,7 +85,13 @@ node('openshift-build-1') {
             }
 
             stage( "smoketest" ) {
-                // run smoketest job here, try{}catch to determine if it failed and  include details of job in success email
```
codificat: I did not include another try/catch specifically for the smoke test here - there's a surrounding try/catch block already.

So we could just set `propagate: true` if we wanted to fail the upgrade job on smoke test failure. But instead we probably want to check the result and throw a custom error...

About details of the smoke test itself: I believe that a URL pointing to the specific smoke test job run is enough here.

tower-scripts/bin/aos-cd-cluster-smoke-test.sh
```diff
@@ -63,24 +63,24 @@ fi
...
 # Using 'EOF' allows you to avoid escaping all uses of $
 ossh "root@${MASTER}" -c "/bin/bash" << 'EOF'
-    set -e
-    oc delete --ignore-not-found project aos-cd-smoke-test-proj
-    sleep 60 # Allow project to be deleted
-    oc new-project aos-cd-smoke-test-proj
+    set -ex
+    PROJ_RANDOM=aos-cd-smoketest-$(shuf -i 100000-999999 -n 1)
+    oc new-project ${PROJ_RANDOM}
     sleep 10  # Allow project to initialize
```
stevekuznetsov: This should try `oc get projects` until we see the project show up instead of a blind sleep

tower-scripts/bin/aos-cd-cluster-smoke-test.sh
```diff
@@ -29,7 +29,7 @@ if [ "$?" != "0" ]; then
 fi
 
 if oc get nodes -l type=infra | grep -e SchedulingDisabled -e NotReady; then
```
stevekuznetsov: This is as fragile as `oc get pods | grep Running`

tower-scripts/bin/aos-cd-cluster-smoke-test.sh
kargakis: Agreed.

tower-scripts/bin/aos-cd-cluster-smoke-test.sh
kargakis: Why don't we make the node schedulable if it's unschedulable? Also, I would retry for some time on NotReady.

tower-scripts/bin/aos-cd-cluster-smoke-test.sh
```diff
@@ -63,24 +63,24 @@ fi
...
-    oc new-project aos-cd-smoke-test-proj
+    set -ex
+    PROJ_RANDOM=aos-cd-smoketest-$(shuf -i 100000-999999 -n 1)
+    oc new-project ${PROJ_RANDOM}
     sleep 10  # Allow project to initialize
     oc new-app --image-stream=ruby --code=https://github.com/openshift/ruby-hello-world
     sleep 60  # Allow build to start
     timeout 10m oc logs -f bc/ruby-hello-world
-    oc get pods | grep "Running"
+    oc rollout status dc/ruby-hello-world -w
```
kargakis: -w is true by default, and I don't think we will ever break it

tower-scripts/bin/aos-cd-cluster-smoke-test.sh
codificat: I think that fixing stuff is not what the smoke test is for - if there's a problem, we want to highlight it, not try to fix it here as part of the test.

So, the question is if we should fail or not based on having *any* non-schedulable or NotReady node. Probably we don't... so we could just write down the warning; but then no one would read it probably (buried down in logs).

I'm thinking we should just remove this part of the smoke test (check for NotReady/unschedulable nodes). Note that the cluster status post-upgrade is already reported by the upgrade job, and this includes output of "oc get node".

jobs/cluster/smoke-test/Jenkinsfile
jupierce: Seems like a reasonable way to start.

jobs/cluster/upgrade/Jenkinsfile
jupierce: The smoketest should be informational. The upgrade has already happened by this point, so we want to finish the pipeline. However, the "success" email we sent out should include a highly visible warning that smoketests have failed along with a link to the failing smoketest job's output. This is why the original smoketest was counting relatively minor infractions like nodes being unschedulable. 

So, if result != SUCCESS, I suggest just setting a flag that can be interpreted by mail_success. mail_success can then tailor the output to include the warning and the URL.

codificat:
As mentioned I updated this so that the upgrade job doesn't fail due to smoke test failures but it adds a visible warning in the notification instead.

Also squashed and rebased (including the move from jobs/cluster to jobs/starter).

I still left out the check for non-schedulable/non-ready nodes because I believe that for large enough clusters is likely to have a tendency for false positives... also, the notification email, as part of cluster status info, already includes a list of all nodes and their statuses. Should I add that check back though?

jobs/starter/smoke-test/Jenkinsfile
```diff
@@ -0,0 +1,86 @@
...
+                        description: 'Failure Mailing List',
+                        name: 'MAIL_LIST_FAILURE'
+                    ],
+                ]
+            ],
+        ]
+    )
+
+    def deploylib = load( "pipeline-scripts/deploylib.groovy" )
+    deploylib.initialize()
```
jupierce: Cluster jobs have changed a bit while you have worked this card.

Populating cluster names:
https://github.com/openshift/aos-cd-jobs/blob/master/jobs/starter/perf/Jenkinsfile#L14-L17
https://github.com/openshift/aos-cd-jobs/blob/master/jobs/starter/perf/Jenkinsfile#L28

Initializing deploylib:
https://github.com/openshift/aos-cd-jobs/blob/master/jobs/starter/perf/Jenkinsfile#L39

Setting up ssh agent (CLUSTER_ENV instead of _NAME): 
https://github.com/openshift/aos-cd-jobs/blob/master/jobs/starter/perf/Jenkinsfile#L51

jobs/starter/upgrade/Jenkinsfile
```diff
@@ -1,16 +1,19 @@
 #!/usr/bin/env groovy
 
-def mail_success(list,detail) {
+def mail_success(list,detail,warn) {
+    body = "Cluster ${CLUSTER_NAME} upgrade details:\n"
+    if ( warn ) {
+        body += "\nWARNING: post-upgrade smoke test was not successful!\n"
+        body += "WARNING: smoke test job details: ${warn}\n"
```
jupierce: Redundant "WARNING" ?

jupierce:
@codificat Your point is well taken, but I don't consider a significant number of nodes being unschedulable/not-ready as being a false positive -- even in our large clusters. Experience has shown it to be an important signal for someone to check up on the health of the cluster. Perhaps we only fail the test if > 2 meet this condition. That gives us room for a little noise.

jobs/starter/upgrade/Jenkinsfile
codificat: This was intentional, for visual purposes... I thotught it makes it more obvious and easier to read/spot the smoke test URL vs the upgrade job URL which I also moved to the top for [hopefully] easier access.

A notification with smoke test warnings would look like this:

```
Cluster CLUSTER_NAME upgrade details:

WARNING: post-upgrade smoke test was not successful!
WARNING: smoke test job details: https://jenkins-jenkins.192.168.122.78.nip.io/job/smoketest/25/

Upgrade Jenkins job: https://jenkins-jenkins.192.168.122.78.nip.io/job/test/7/

<cluster status>

Smoke test results: FAILURE
```

I can leave the upgrade job URL at the bottom and supress the second WARNING line (leaving the smoke test URL only) if preferred

jobs/starter/smoke-test/Jenkinsfile
codificat: Sorry I missed these, will update...

jobs/starter/upgrade/Jenkinsfile
jupierce: Maybe combining the lines is sufficient. I just don't want it to look as scary :)
WARNING: Post-upgrade smoke test was not successful: URL

jobs/starter/smoke-test/Jenkinsfile
jupierce: You didn't miss them, they changed underneath you :)

codificat:
New rebase/squash with the recent suggested changes:

- Keep the check for non-ready nodes (I didn't add a threshold)
- Make the smoke test failure warning less scary
- Update to match new cluster selection and library handling

PR #616, opened 2017-09-08
Define osm_cluster_network_cidr and openshift_portal_net

sdodson:
These are now required values in order to ensure that we don't change
defaults during upgrades


sdodson:
Unblocks https://github.com/openshift/openshift-ansible/pull/5256

ashcrow:
I've updated the blocked PR to also require``osm_host_subnet_length`` be set. Add:

```
osm_host_subnet_length=9
```

and we should be good.

PR #617, opened 2017-09-08
Scan images

bbguimaraes:
@jupierce, @smunilla: PTAL.

### DIFF
--- a/build-scripts/ose_images/ose_images.sh
+++ b/build-scripts/ose_images/ose_images.sh
@@ -49,6 +49,7 @@ usage() {
   echo "Actions:" >&2
   echo "  build build_container :: Build containers in OSBS" >&2
   echo "  push push_images :: Push images to qe-registry" >&2
+  echo "  scan_images      :: Scan images with openscap" >&2
   echo "  compare_git      :: Compare dist-git Dockerfile and other files with those in git" >&2
   echo "  compare_auto     :: Auto compare dist-git files with those in git. Sends Dockerfile diff in email" >&2
   echo "  compare_nodocker :: Compare dist-git files with those in git.  Show but do not change Dockerfile changes" >&2
@@ -1177,6 +1178,14 @@ start_push_image() {
   popd >/dev/null
 }
 
+get_image_url() {
+  dockerfile=$1
+  name=$(grep " name=" "${dockerfile}" | cut -d'"' -f2)
+  version=$(grep version= "${dockerfile}" | cut -d'"' -f2)
+  release=$(grep release= "${dockerfile}" | cut -d'"' -f2)
+  echo "${PULL_REGISTRY}/${name}:${version}-${release}"
+}
+
 check_dependents() {
   if ! [ "${dependent_list_new}" == "" ] ; then
     dependent_list_working="${dependent_list_new}"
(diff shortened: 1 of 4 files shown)


jobs/build/scan-images/Jenkinsfile
```diff
@@ -0,0 +1,46 @@
+final MAIL_FROM = 'aos-cd@redhat.com'
+final MAILING_LIST_CVE = ['bbarcaro@redhat.com']
+final MAILING_LIST_ERR = ['bbarcaro@redhat.com']
```
bbguimaraes: Who should receive the notification email?

jobs/build/scan-images/Jenkinsfile
```diff
@@ -0,0 +1,46 @@
...
+"""
+      final ret = sh(
+        returnStatus: true,
+        script: 'grep -xq \'The following issues were found:\' scan.txt')
+      if(ret == 0) {
+        mail(
+          from: MAIL_FROM, to: MAILING_LIST_CVE,
+          subject: 'jobs/build/scan-images: cve(s) found',
+          body: readFile('scan.txt'))
+      }
```
bbguimaraes: Should the job fail at this point or is the email enough?

build-scripts/ose_images/ose_images.sh
```diff
@@ -1811,6 +1835,12 @@ case "$action" in
     echo "Fail Pushes: ${BUILD_FAIL}"
     cat ${workingdir}/logs/buildfailed | cut -d':' -f3-4
   ;;
+  scan_images )
+    popd > /dev/null
+    sudo atomic scan --scanner openscap \
+      $(< "${workingdir}/images_to_scan.txt") \
+      > scan.txt
```
bbguimaraes: What is the best way to get this to the machine that's going to execute the job?

`atomic scan` requires, obviously, the `atomic` command, and also the
`openscap` scanner image.  The steps I used to install on my fedora machine
were:

```
# dnf install atomic
# atomic install registry.access.redhat.com/rhel7/openscap

jobs/build/scan-images/Jenkinsfile
smunilla: I think email is good. We don't really want to fail on unpatched CVEs because a patch might not even exist yet.

bbguimaraes: After a discussion with @jupierce, we decided it would be better to have this
as a separate job (`build/scan-images`).  So "fail" here means only this
specific job, not the one which built the images.

jobs/build/scan-images/Jenkinsfile
jupierce: smunilla, jupierce, ahaile

jobs/build/scan-images/Jenkinsfile
jupierce: Right - It is ok for the job to fail, but the error must not propagate to the build job.

build-scripts/ose_images/ose_images.sh
bbguimaraes: The [docs][docs] say

> The OVAL definitions used by the CVE scan type are bundled in the container
> image during the build process, and as such are not always up-to-date. 

so we should keep the image up to date.  We could do a pull before each scan.

[docs]: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Security_Guide/sect-Using_OpenSCAP_with_Atomic.html

bbguimaraes: Also, can the jobs execute as the root user on this jenkins instance?  `atomic
scan` does a lot of black magic and has to execute as root.

jupierce: @bbguimaraes You should be able to use sudo in this environment to update the image.

jobs/build/scan-images/Jenkinsfile
bbguimaraes: Noted, I'll add them after we test the job.

build-scripts/ose_images/ose_images.sh
bbguimaraes: Command to update the image added.

bbguimaraes:
@jupierce: updated.

build-scripts/ose_images/ose_images.sh
adammhaile: Thanks @jupierce I have installed that on the new build vm.

PR #620, opened 2017-09-08
Moving around jobs to support dedicated upgrades

jupierce:


jupierce:
Apologies in advance @codificat  :) . Some renaming is necessary to take one some more upgrade jobs.

PR #624, opened 2017-09-09
cri-o: force image from docker.io for now

giuseppe:
https://github.com/openshift/openshift-ansible/pull/5310 was merged
before the image for RHEL was ready.  Until that it is done, force the
image from docker.io.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
@@ -58,6 +58,7 @@ extensions:
                          --inventory sjb/inventory/ \
                          -e deployment_type=origin  \
                          -e openshift_use_crio=True   \
+                         -e openshift_crio_systemcontainer_image_registry_override=docker.io/gscrivano \
                          -e etcd_data_dir="${ETCD_DATA_DIR}" \
                          -e openshift_master_default_subdomain="${local_ip}.nip.io"             \
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
--- a/sjb/generated/test_branch_origin_extended_conformance_crio.xml
+++ b/sjb/generated/test_branch_origin_extended_conformance_crio.xml
@@ -198,6 +198,7 @@ ansible-playbook -vv --become               \
                  --inventory sjb/inventory/ \
                  -e deployment_type=origin  \
                  -e openshift_use_crio=True   \
+                 -e openshift_crio_systemcontainer_image_registry_override=docker.io/gscrivano \
                  -e etcd_data_dir=&#34;\${ETCD_DATA_DIR}&#34; \
                  -e openshift_master_default_subdomain=&#34;\${local_ip}.nip.io&#34;             \
                  -e openshift_pkg_version=&#34;\$( cat ./ORIGIN_PKG_VERSION )&#34;               \


runcom:
@mrunalp this should fix the origin tests with cri-o

mrunalp:
👍

runcom:
@giuseppe we also released yesterday cri-o v1 RC1, could you also update your Docker image to use that? (and maybe also the RH registry image)

kargakis:
You also need to update the generated xml that will be pushed in Jenkins

./sjb/generate.sh

giuseppe:
@kargakis thanks for the review, I've pushed a new version right now.

@runcom sure, I am going to rebuild the system container image

sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
```diff
@@ -58,6 +58,7 @@ extensions:
                          --inventory sjb/inventory/ \
                          -e deployment_type=origin  \
                          -e openshift_use_crio=True   \
+                         -e openshift_crio_systemcontainer_image_registry_override=docker.io/gscrivano \
```
runcom: Should this also specify the name `cri-o-centos`|`cri-o-fedora`? This PR, afaict, is now pulling `docker.io/gscrivano/cri-o:latest` which doesn't exist. @giuseppe @ashcrow

ashcrow: @runcom Yes, you're correct. The other option which may be quicker and easier (as either would be temporary anyway) would be to tag a ``cri-o`` image in @giuseppe's docker.io.

giuseppe: @ashcrow do we need to change this in openshift-ansible?  I think it is reasonable to override the entire image not just the registry, what do you think if we drop the `openshift_crio_systemcontainer_image_registry_override` logic in favour of `openshift_crio_systemcontainer_image_override` and do the same for the docker system container?

ashcrow: @giuseppe I don't think we need to but we could. If you and @runcom both think it would be easier to have a full override I'll take care of modifying and getting a PR up for it today. I've had enough ☕️ already to be in code mode 😹

giuseppe: I've tagged 'docker.io/gscrivano/cri-o' for now.

ashcrow: I'll make the change but this can merge.

kargakis:
> I'll make the change but this can merge as it will work as is.

@giuseppe do you have access to push the update in Jenkins? Also, is there an issue somewhere to track when we can remove the override?

ashcrow:
@kargakis https://github.com/openshift/openshift-ansible/pull/5354

giuseppe:
@kargakis no, I have no access to push the update I think.  How can I check that?

kargakis:
> Also, is there an issue somewhere to track when we can remove the override?

If this didn't come out clear, I was asking for something like https://github.com/openshift/aos-cd-jobs/issues/627

> @kargakis no, I have no access to push the update I think. How can I check that?

You need credentials for https://ci.openshift.redhat.com/jenkins/

See https://mojo.redhat.com/docs/DOC-1081313#jive_content_id_Jenkins for more information.

In the meantime, I can push this in Jenkins. Is it ready to merge or do you need anything more?

giuseppe:
@kargakis it should be fine now

PR #626, opened 2017-09-10
Include [Feature:Builds] as part of extended builds run

smarterclayton:
Origin will be shifting over to this pattern. Brings us in line with sig-testing


bparees:
don't we need to go add that [Feature:Builds] tag to all our extended tests before we can do this?

bparees:
oops, sorry, missed the pipe

bparees:
lgtm.  @smarterclayton i assume you're going to run the job updater.

PR #636, opened 2017-09-13
json-file logging job - gating

richm:
This adds a gating logging job to run on every PR.  This configures
the operating system and OpenShift to use docker with
`--log-driver=json-file`.  The logging fluentd will automatically
detect this.
@stevekuznetsov @jcantrill ptal
[test]

### DIFF
--- a/sjb/config/common/test_suites/origin_aggregated_logging.yml
+++ b/sjb/config/common/test_suites/origin_aggregated_logging.yml
@@ -1,3 +1,4 @@
 ---
 children:
   - "test_pull_request_origin_aggregated_logging_ansible"
+  - "test_pull_request_origin_aggregated_logging_ansible_json_file"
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Seems reasonable to me -- if you want to push it up and test it before we merge, let me know when you're happy with it

richm:
I pushed it up and tested it
one question - I had to duplicate a lot of code in order to implement this == tech debt - is there some way to parameterize `config/test_cases/test_branch_openshift_ansible_logging` so that I could run it both with json-file and without?

stevekuznetsov:
Not in the current framework. To DRY this out, we'd want to install with a different inventory, right? In the future @csrwng and I will make sure that is easy to do. Unfortunately the `sjb` tool is very poor at handling that situation.

richm:
right - different inventory, same job

ok, then just please merge and push as is

stevekuznetsov:
Done and done :)

PR #639, opened 2017-09-14
CRI-O: remove broken docker images filter

runcom:
Turns out the "reference" filter isn't available in docker 1.12.x which
is used in the Origin CI. Replace that with a grep. For reference teh
error was:
```
++ docker images --filter 'reference=openshift/origin-*' --format
'{{.Repository}}:{{.Tag}}'
Error response from daemon: Invalid filter 'reference'
+ set +o xtrace
```


runcom:
@kargakis PTAL, this job failed because of that https://ci.openshift.redhat.com/jenkins/view/All/job/test_branch_origin_extended_conformance_crio/18

PR #641, opened 2017-09-14
WIP Remove aos-ci-test job dependencies

sdodson:
We've decided that these jobs and the coordination between them is not
providing any value so we're going to rely solely on the test and merge
jobs.

System container based tests have already been added to test and merge and we should invest in shoring those up rather than dealing with aos-ci-test.


sdodson:
When I say we've decided, I guess that's not yet decided. So if anyone objects to removing aos-ci-test from the requirements lets discuss it here.

ingvagabund:
Before we remove it completely, we need to extend the [merge]/[test] CI with the containerized deployment as well. If we disable the e2e conformance tests, we can already run the deployment. Still, we should debug and fix the failing e2e tests.

sosiouxme:
I'm not actually that aware of what's in aos-ci-tests vs test/merge so I don't have much useful to say here. Obviously the sooner we can go to a single set of tests the better.

ingvagabund:
The containerized CI job is here: https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_origin_extended_conformance_install_containerized.yml. I have already fix it at the same time I was working on the system container deployment.

stevekuznetsov:
@sosiouxme @rhcarvalho want to ensure that this doesn't regress on `tox` etc

sosiouxme:
> @sosiouxme <https://github.com/sosiouxme> @rhcarvalho
> <https://github.com/rhcarvalho> want to ensure that this doesn't regress
> on tox etc
>
>
That didn't run in aos-ci-test, just the rosie tests, don't know of
anything specific to checks in aos-ci-test so we're good.

sdodson:
WIP until Jan re-adds system container install minus e2e test. That provides parity with what aos-ci-test was doing.

ingvagabund:
https://github.com/openshift/aos-cd-jobs/pull/646

PR #644, opened 2017-09-15
allow use of mux, es external for testing

richm:
This will cause the mux dc and service to be created,
and the routes for es and es-ops.  However, fluentd will
not be configured to use mux by default, but mux tests can
use it.  The es and es-ops routes aren't used, but at least this
will exercise the ansible code that creates them.
@stevekuznetsov @jcantrill @nhosoi PTAL
[test]

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
@@ -133,6 +133,10 @@ extensions:
                          -e openshift_hosted_logging_hostname="kibana.127.0.0.1.nip.io"           \
                          -e openshift_logging_master_public_url="https://localhost:8443"          \
                          -e openshift_master_logging_public_url="https://kibana.127.0.0.1.nip.io" \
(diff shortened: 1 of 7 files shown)


stevekuznetsov:
Sounds good to me -- do you know the duration of these tasks? If they're long, maybe only enable them on the logging job instead of the global inventory

richm:
The tasks are very short - basically, create the mux dc, create the mux service, deploy a mux pod, create routes - they shouldn't add more than a few seconds to the ansible logging run

richm:
or are you saying that logging.yml is used by _all_ openshift ansible runs?  I'm not sure what you mean by "global"

stevekuznetsov:
That's used by every single job in the entire system that installs and/or upgrades Origin

richm:
ok - then I'll submit another PR which adds them to the logging test jobs

richm:
or, just update this one

richm:
Can this be merged?

stevekuznetsov:
Sorry, didn't see this got updated

/lgtm

PR #645, opened 2017-09-15
Add build stage for service-catalog user-broker binary

pmorie:
Part of openshift/origin#16220 - adds a stage to the extended_conformance_install_update job that builds the service catalog test broker binary

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
@@ -83,6 +83,12 @@ extensions:
         sed -i 's|go/src|data/src|' _output/local/releases/rpms/origin-local-release.repo
         sudo cp _output/local/releases/rpms/origin-local-release.repo /etc/yum.repos.d/
     - type: "script"
+      title: "build service-catalog user-broker image"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        REGISTRY=openshift/ NO_DOCKER=1 make -C cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog user-broker
+    - type: "script"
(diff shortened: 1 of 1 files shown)


pmorie:
I'm not sure if it works yet.

sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml
```diff
@@ -83,6 +83,12 @@ extensions:
         sed -i 's|go/src|data/src|' _output/local/releases/rpms/origin-local-release.repo
         sudo cp _output/local/releases/rpms/origin-local-release.repo /etc/yum.repos.d/
     - type: "script"
+      title: "build service-catalog user-broker image"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        REGISTRY=openshift/ NO_DOCKER=1 make -C cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog user-broker
```
jpeeler: Something to be mindful of (maybe it doesn't matter?) is that this is using the upstream build process rather than the origin tooling (build-go.sh).

stevekuznetsov: yeah we talked about this last night -- this doesn't need to be versioned etc so should be ok

jpeeler:
I think this can be closed - it's built here: https://github.com/openshift/aos-cd-jobs/blob/99684c80ea476b15581ce315ac044eec11c0d872/sjb/config/test_cases/test_branch_origin_service_catalog.yml#L25

And here: https://github.com/openshift/aos-cd-jobs/blob/99684c80ea476b15581ce315ac044eec11c0d872/sjb/config/test_cases/test_pull_request_origin_service_catalog.yml#L41

PR #646, opened 2017-09-15
extend openshift-ansible CI with conformance test-free containerized deployment

ingvagabund:
A must for termination of `aos-ci-test` based tests. Disabling e2e conformance tests for the same reason as I did for the system container based deployment. We still need to debug the failing tests. Currently, we don't have human power and capacity to do that.

### DIFF
--- a/sjb/config/common/test_suites/openshift_ansible.yml
+++ b/sjb/config/common/test_suites/openshift_ansible.yml
@@ -5,3 +5,4 @@ children:
   - "test_pull_request_openshift_ansible_extended_conformance_install_with_status_check"
   - "test_pull_request_openshift_ansible_extended_conformance_install_update"
   - "test_pull_request_openshift_ansible_extended_conformance_install_system_containers"
+  - "test_pull_request_openshift_ansible_extended_conformance_install_containerized"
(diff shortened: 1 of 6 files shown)


ingvagabund:
@stevekuznetsov @sdodson 
@jhadvig FYI

ingvagabund:
Tested here: https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install_containerized/18/

stevekuznetsov:
did we fix the installer to allow containerized to use the SHA hashes for the images? this job was borked before...

stevekuznetsov:
looks like no:
```
935f0d704319        openshift/node:v3.7.0-alpha.1                                                                          "/usr/local/bin/origi"   6 seconds ago        Up 5 seconds                                    origin-node
```

ingvagabund:
> did we fix the installer to allow containerized to use the SHA hashes for the images? this job was borked before...

@sdodson is that even possible? The `openshift-ansible` always expects the version, no hash.

stevekuznetsov:
Exactly... this was a conversation we had in an architecture call months ago -- we need a way to signal "I know what I'm doing, don't touch me" or other protections to ensure that if we tag it as a version we know that we are testing the correct images -- disable pulling? Idk.

ingvagabund:
Given the images are built locally, we always test the correct images. Is not that enough?

sdodson:
> did we fix the installer to allow containerized to use the SHA hashes for the images? this job was borked before...

I'd assumed that Jan had found some way around that when these tests were initially implemented. If not then we'll have to wait until we fix that.

stevekuznetsov:
This job is still valuable even with the published version of Origin on the Hub.

PR #653, opened 2017-09-18
Correct node port enablement

smarterclayton:
Space in the argument caused ansible to break the argument into two
parts and fail.


smarterclayton:
Ok, let's try this again

smarterclayton:
Pushed

PR #659, opened 2017-09-19
[BuildAuto] Automate OpenShift Jenkins uber-plugin rebuilds

codificat:
A job that invokes the `collect-jenkins-plugins.sh` and `update-dist-git.sh` scripts to create an updated `jenkins-x-plugins` RPM in dist-git and build in brew.

It uses buildlib to initialize kerberos credentials (needed by rhpkg). It doesn't need some of the steps that buildlib's `initialize` is doing, like login to registries, so suggesting to separate these init steps in the library.

@jupierce PTAL, I know some things much likely need updates, including:

- should we have a choice for the Jenkins Version parameter? If so, what choices and how are they maintained? Or just the plain string param might be enough here...
- should we move the choices for OCP release version into the library? (and update other jobs to use them)
- list for email notifications: more? less?

### DIFF
--- a/jobs/build/jenkins-plugins/Jenkinsfile
+++ b/jobs/build/jenkins-plugins/Jenkinsfile
@@ -7,8 +7,7 @@ properties(
                     name: 'JENKINS_VERSION',
                     description: 'Minimum required Jenkins version',
                     $class: 'hudson.model.StringParameterDefinition'
-                    // Probably better it this was a choice, but might not be
-                    // flexible enough
+                    defaultValue: '2.43'
                 ],
                 [
                     name: 'OCP_RELEASE',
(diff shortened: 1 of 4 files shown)


jobs/build/jenkins-plugins/Jenkinsfile
```diff
@@ -0,0 +1,123 @@
...
+    buildlib.path_setup()  // Put scripts in env.PATH
+
+    try {
+        stage ("prepare workspace") {
+
+            // Temporary file to write the plugin list to
+            tmpdir = pwd tmp:true
+            plugin_file = "${tmpdir}/jenkins-plugins.txt"
+            // Note that collect-jenkins-plugins.sh has a hardcoded output dir:
+            plugin_dir = "${env.WORKSPACE}/hacks/update-jenkins-plugins/working/hpis"
```
jupierce: We can change the output directory to CWD if it would simplify things (only CD uses this tool, so we won't break anyone). If we stay with absolute path, I think we might want to rm -rf the expected working directory. If we move to a temporary directory, let's try to clean up the directory if we succeed.

pipeline-scripts/buildlib.groovy
```diff
@@ -3,7 +3,21 @@ def commonlib = load("pipeline-scripts/commonlib.groovy")
 
 commonlib.initialize()
 
-def initialize() {
+def path_setup() {
```
jupierce: A more robust pattern in Jenkins pipelines is
```
withEnv(["PATH+SOMETHING=${env.WORKSPACE'}/somepath"]) {
      sh 'sometool'
}
```

I'd prefer to see this explicitly in each pipeline instead of changing the path globally in the library.

jobs/build/jenkins-plugins/Jenkinsfile
codificat: I'll go the quick/lazy way and leave the hardcoded path within the scripts if that's ok (adding a cleanup at the end)...

pipeline-scripts/buildlib.groovy
codificat: Ok!

I'll leave the entry to `build-scripts/ose_images` that was already there though - only moving PATH / GOPATH setup into a function to keep init a bit cleaner.

PR #660, opened 2017-09-19
scan_images: use oit

bbguimaraes:
Requires https://github.com/openshift/enterprise-images/pull/8.

@jupierce

### DIFF
--- a/jobs/build/scan-images/Jenkinsfile
+++ b/jobs/build/scan-images/Jenkinsfile
@@ -17,19 +17,33 @@ properties([[
 
 node('openshift-build-1') {
   try {
-    stage('clone') { dir('aos-cd-jobs') { checkout scm } }
+    def buildlib = null
+    stage('clone') {
+      dir('aos-cd-jobs') {
+        checkout scm
+        buildlib = load('pipeline-scripts/buildlib.groovy')
+      }
+      dir('oit') { git 'https://github.com/openshift/enterprise-images.git' }
+    }
+    stage('venv') {
+      if(!fileExists('env')) { sh 'virtualenv env' }
+      sh 'env/bin/pip install -r oit/oit/requirements.txt'
+    }
     stage('update') {
       // Pull scanner image to update CVE definitions.
       sh 'sudo docker pull registry.access.redhat.com/rhel7/openscap'
     }
     stage('scan') {
-      sh """\
-aos-cd-jobs/build-scripts/ose_images/ose_images.sh scan_images \
-  --user ocp-build \
-  --branch 'rhaos-${BUILD_VERSION}-rhel-7' \
-  --group base
-cat scan.txt
+      buildlib.with_virtualenv("${pwd()}/env") {
+        sh """\
+set -o pipefail
+sudo python oit/oit/oit.py \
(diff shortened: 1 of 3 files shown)


jobs/build/scan-images/Jenkinsfile
```diff
@@ -17,18 +17,26 @@ properties([[
 
 node('openshift-build-1') {
   try {
-    stage('clone') { dir('aos-cd-jobs') { checkout scm } }
+    stage('clone') {
+      dir('aos-cd-jobs') { checkout scm }
```
jupierce: What if these directories don't exist? Do we need to rm -rf them and mkdir them before we clone?

jobs/build/scan-images/Jenkinsfile
```diff
@@ -17,18 +17,26 @@ properties([[
...
+      dir('oit') { git 'https://github.com/openshift/enterprise-images.git' }
+    }
+    stage('venv') {
+      if(!fileExists('env')) { sh 'virtualenv env' }
```
jupierce: Any environment variables populated for the virtual environment here will be lost as soon as sh returns here. This is the pattern used in other Jenkins pipelines: https://github.com/adammhaile/aos-cd-jobs/commit/60eab300ea85d2582c1c540cab175f34e51e1454#diff-4afa940aacab03ebeba0c98c8cc67d64R194

jobs/build/scan-images/Jenkinsfile
bbguimaraes: I never thought about it, but it has worked on all my tests, with a repo that
I'm constantly updating.  I tested it again to be sure, the directory is
created if it doesn't exist and is reused if it does.  A `git fetch` is
executed to update the repository and, since we're not modifying any of them, I
don't think there will be a problem.

jobs/build/scan-images/Jenkinsfile
bbguimaraes: This command only creates the virtualenv the first time the job is executed.
All others should fail the `!fileExists('env')` test.  The virtualenv is used
on line 35, where `env/bin/python` is used instead of `python` or
`/bin/python`.

That has the same effect (again, it worked on all my tests), but I don't object
to a standard way of using virtualenvs.  Maybe something for the common libs?
I'll make the change to the job anyway.

jobs/build/scan-images/Jenkinsfile
```diff
@@ -17,19 +17,33 @@ properties([[
...
-  --group base
-cat scan.txt
+      buildlib.with_virtualenv("${pwd()}/env") {
+        sh """\
+set -o pipefail
+sudo python oit/oit/oit.py \
+  --metadata-dir oit/ \
+  --group 'openshift-${BUILD_VERSION}' \
+  distgits:scan-for-cves \
+  | tee scan.txt
```
bbguimaraes: @jupierce: this may need some tweaking depending on how `sudo` is configured.
On my machine, with the default sudo configuration, I need:

```
sudo --preserve-env PATH=$PATH python oit/oit/oit.py …

jupierce: Looks reasonable.

PR #666, opened 2017-09-21
Update new exclusion temporarily for router

smarterclayton:
The new labels will be added and the old labels removed. This test
should actually work, but we need to cut new images.


kargakis:
ok

PR #668, opened 2017-09-21
Add missing generated changes

kargakis:


kargakis:
@ingvagabund @stevekuznetsov ptal

kargakis:
@stevekuznetsov pushed but I would like a second review on @ingvagabund's changes

ingvagabund:
I skipped those intentionally as I did not want to affect the origin CI. I don't know if the same set of jobs is used in testing origin as well or not. Are they? @stevekuznetsov

PR #670, opened 2017-09-21
Set a recurring job for the conformance-k8s suite

smarterclayton:
This is the official upstream conformance suite. Not yet merged into
master but need PR testing to validate. Will not be required as part of
every PR


stevekuznetsov:
/lgtm

... lol

PR #671, opened 2017-09-21
Share origin_gce install code across conformance jobs

smarterclayton:


sjb/config/test_cases/test_pull_request_origin_extended_conformance_k8s.yml
```diff
@@ -1,15 +1,13 @@
 ---
-parent: 'common/test_cases/origin_release.yml'
+parent: 'test_cases/test_branch_origin_extended_conformance_gce.yml'
 overrides:
-  junit_analysis: False
```
stevekuznetsov: why? being pushed to gubernator, right?

smarterclayton: it's set in the base

smarterclayton: Or do I also have to override in a descendant?

stevekuznetsov: oh no sorry didn't see that it was in ancestor

PR #678, opened 2017-09-25
publish gcs artifacts regardless

smarterclayton:
@stevekuznetsov does this look right?


stevekuznetsov:
In the sense that it's not going to make a problem, sure. But we should really be putting the artifacts in the right place, and that will not do what you want -- we don't have logic in that upload to do the repo/org path segment like we should.

smarterclayton:
Updated

PR #686, opened 2017-09-27
started.json not uploaded

smarterclayton:


sjb/config/common/test_cases/origin.yml
```diff
@@ -57,7 +57,7 @@ actions:
         gsutil cp gcs/started.json "${path}/started.json"
         if [[ -n "${PULL_NUMBER:-}" ]]; then
           echo "${path}" > "${BUILD_NUMBER}.txt"
-          if [[ "${REPO_NAME}" != "origin" ]]; then
+          if [[ -n "${REPO_OWNER:-}" || "${REPO_NAME:-}" != "origin" ]]; then
             segment="${REPO_OWNER}_${REPO_NAME}/"
```
kargakis: You can still break here if one of the parameters is unset.

smarterclayton: there's no case where that is true

smarterclayton: we don't accept one or the other.

stevekuznetsov: This is wrong -- it will set whenever `$REPO_OWNER` is set, which should be always

smarterclayton:
Fixed.

PR #690, opened 2017-09-27
enable the eventrouter always for logging

richm:
This allows us to test those code paths in the ansible
logging roles and CI tests.
@wozniakjan @jcantrill PTAL
@stevekuznetsov I've already successfully tested this with
logging CI


stevekuznetsov:
Pushed

PR #691, opened 2017-09-27
Support updates to forks of this repository

bbguimaraes:
When executed against a fork:
- the pruner script will not be executed, to avoid removing branches
  accidentaly (e.g. PR branches)
- a user-specific key will be used for pushes

@jupierce

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,14 +22,25 @@ node('openshift-build-1') {
           sh 'git checkout master'
         }
         stage('run') {
-          sshagent(['openshift-bot']) {
-            sh '''\
+          final url = sh(
+            returnStdout: true,
+            script: 'git config remote.origin.url')
+          if(!(url =~ /^[-\w]+@[-\w]+(\.[-\w]+)*:/)) {
+            error('This job uses ssh keys for auth, please use an ssh url')
+          }
+          def prune = true, key = 'openshift-bot'
+          if(url.trim() != 'git@github.com:openshift/aos-cd-jobs.git') {
+            prune = false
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -16,22 +16,29 @@ node('openshift-build-1') {
...
-          sh '''\
-virtualenv env/
-. env/bin/activate
+        stage('run') {
+          final url = sh(
+            returnStdout: true,
+            script: 'git config remote.origin.url')
+          def prune = 'true', key = 'openshift-bot'
+          if(url.trim() != 'git@github.com:openshift/aos-cd-jobs.git') {
+            prune = ''
```
jupierce: 'false'? Seems odd to mix 'true' and empty string.

Jenkinsfile
```diff
@@ -16,22 +16,29 @@ node('openshift-build-1') {
...
+            key = "${(url =~ /.*:([^\/]+)/)[0][1]}-aos-cd-bot"
```
jupierce: Can we add a check at the top of this job.. if (url startsWith "git" == false ) then error(please use an ssh url)?
I know the job won't work without it, but a clean error will save time diagnosing.

Jenkinsfile
bbguimaraes: That's an old, shell style boolean.  But I agree it's not necessary here.
Changed to something better.

Jenkinsfile
bbguimaraes: Changed, but I could not find a way to extract the git remote url from jenkins,
so the check has to be done after cloning.

PR #699, opened 2017-10-02
Improve stg->prod promotion pipeline

codificat:
This is for [the corresponding trello card](https://trello.com/c/mo3Vr0TF/341-2-buildauto-improve-stg-prod-promotion-pipeline-add-online):

- Tidy up the `mirrors-stage-to-prod.sh` script, remove the push of `/srv/libra/*`, and add an option to specify the repo to push (i.e. which dir under `/srv/enterprise`).
- Use buildlib in `Jenkinsfile` to stream de script to rcm-guest
- Enable promotion of online-openshift-scripts as well as OCP
- Request confirmation before proeceeding

@jupierce PTAL.

PR #701, opened 2017-10-03
Fix typo in mirrors-stage-to-prod.sh

codificat:
Introduced in #699, thanks @stevekuznetsov for pointing it out.


stevekuznetsov:
while you are at it i suggest you run `shellcheck` on the file and submit all the fixes it identifies

codificat:
Done, thanks

PR #703, opened 2017-10-03
Record systemd-journald.service journal as an artifact

jim-minter:
it is the place that indicates whether journald has suppresssed messages from other journals.

@stevekuznetsov @gabemontero @smarterclayton @bparees


jim-minter:
Messages are of the form `Oct 03 12:42:20 t540p systemd-journald[789]: Suppressed 5140 messages from /...`; these do not appear in generated/pid1.journal.

bparees:
lgtm but will wait for @stevekuznetsov

PR #707, opened 2017-10-04
Remove use of OS_BUILD_ENV_TMP_VOLUME

detiber:
https://github.com/openshift/origin/pull/16686 removes the tmp volume from hack/env, this PR removes references to the no longer used OS_BUILD_ENV_TMP_VOLUME variable.


detiber:
@stevekuznetsov ptal

PR #706, opened 2017-10-04
Add Chrome for console integration tests

benjaminapetersen:
@spadgett @stevekuznetsov 

RE [Fixing e2e tests](https://github.com/openshift/origin-web-console/pull/2083) in web console.

PR #709, opened 2017-10-05
Add flag to assume yes when installing chrome

benjaminapetersen:
Follow to #706, adding the `-y` flag to install chrome.

Tested the command in a local VM first.

@stevekuznetsov @spadgett


stevekuznetsov:
Pushed up

PR #711, opened 2017-10-05
Remove path sanity check on rcm-guest

codificat:
The mirror script can't check for path existence on rcm-guest, as the target paths are on the mirrors. Removing the sanity check here... it shouldn't be much of a problem, specially when the script is invoked from the jenkins job, as the arguments are picked from a set list of valid choices...

@jupierce thanks for the patience merging fixes...


jupierce:
@codificat  No problem! This are very difficult to test beforehand.

PR #718, opened 2017-10-09
enable service catalog and tsb by default in CI

jim-minter:


jim-minter:
@stevekuznetsov ptal

stevekuznetsov:
This will not impact GCE, but should be sufficient to globally install the TSB on all Origin installs. Is this heavyweight and/or does it take time? If so, you may want to add it to only  the Origin conformance e2e job (or whatever jobs are running your tests) so you don't burden e.g. the Installer tests with it.

jim-minter:
@stevekuznetsov thanks for the feedback, how about this?

sjb/config/common/test_cases/origin_release_install_gce.yml
```diff
@@ -121,7 +121,7 @@ extensions:
           docker pull openshift/origin-gce:latest
         fi
 
-        ../../bin/local.sh ansible-playbook -e "provision_gce_docker_storage_driver=${DOCKER_STORAGE_DRIVER}" -e "openshift_test_repo=${location_url}" playbooks/launch.yaml
+        ../../bin/local.sh ansible-playbook -e "provision_gce_docker_storage_driver=${DOCKER_STORAGE_DRIVER}" -e "openshift_test_repo=${location_url}" -e "openshift_enable_service_catalog=true" -e "template_service_broker_install=true" playbooks/launch.yaml
```
stevekuznetsov: @smarterclayton PTAL

smarterclayton: Wrong place, add it to openshift/release cluster/test-deploy/data/vars.yaml

jim-minter:
@stevekuznetsov repushed, removed gce-specific bits, ptal

smarterclayton:
This broke install_update completely (since 8:30am on Friday).  

```
TASK [ansible_service_broker : set ansible-service-broker image facts using set prefix and tag] ***
task path: /usr/share/ansible/openshift-ansible/roles/ansible_service_broker/tasks/install.yml:25
ok: [localhost] => {
    "ansible_facts": {
        "ansible_service_broker_etcd_image": "quay.io/coreos/etcd:latest", 
        "ansible_service_broker_image": "ansibleplaybookbundle/ansible-service-broker:latest"
    }, 
    "changed": false, 
    "generated_timestamp": "2017-10-13 13:53:25.826243"
}

TASK [ansible_service_broker : validate Dockerhub registry settings] ***********
task path: /usr/share/ansible/openshift-ansible/roles/ansible_service_broker/tasks/validate_facts.yml:2
fatal: [localhost]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "generated_timestamp": "2017-10-13 13:53:25.880053", 
    "msg": "To use the dockerhub registry, you must provide the ansible_service_broker_registry_user. ansible_service_broker_registry_password, and ansible_service_broker_registry_organization parameters"
}
```

I'm inclined to revert to fix.

PR #720, opened 2017-10-10
retrieve systemd-journald artifact on gce

jim-minter:
follow-up to #703


stevekuznetsov:
@jim-minter nice catch, thanks

PR #724, opened 2017-10-12
new openshift-ansible install update containerized job

ingvagabund:
So we can test containerized upgrade as well

- no conformance tests run so far, will be done later once all failing tests are fixed
- upgrading to the same version (not an actual upgrade, just to test the upgrade execution path), once pre-built OCP images are available, we will perform a real upgrade
- health-checks are disabled: @sosiouxme 
- built on top of `test_pull_request_openshift_ansible_extended_conformance_install_containerized`

I have not added the new job to the openshift-ansible test-suite yet. I will do it once we agree the job is ready to go.

It takes ~1h to run the job, tested in the CI: https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install_update_containerized/4/

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update_containerized.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update_containerized.yml
@@ -1,0 +1,160 @@
+---
+parent: 'common/test_cases/origin.yml'
+overrides:
+  junit_analysis: False
+extensions:
+  sync_repos:
+    - name: "openshift-ansible"
+    - name: "aos-cd-jobs"
+  actions:
+    - type: "script"
+      title: "build an origin release"
+      repository: "origin"
+      script: |-
+        if [[ -n "${PULL_REFS:-}" ]]; then
+          export OPENSHIFT_ANSIBLE_TARGET_BRANCH="${PULL_REFS%%:*}"
+        fi
+        git checkout ${OPENSHIFT_ANSIBLE_TARGET_BRANCH}
+        hack/build-base-images.sh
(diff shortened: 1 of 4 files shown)


ingvagabund:
@stevekuznetsov @kargakis @sdodson

sjb/config/test_cases/test_branch_origin_extended_conformance_install_update_containerized.yml
```diff
@@ -0,0 +1,160 @@
...
+                         --connection local     \
+                         --inventory sjb/inventory/ \
+                         -e containerized=true      \
+                         -e deployment_type=origin                  \
+                         -e openshift_image_tag="$( cat ./ORIGIN_TAG )"                      \
+                         -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )" \
+                         -e oreg_url='openshift/origin-${component}:'"$( cat ./ORIGIN_COMMIT )" \
+                         -e openshift_node_port_range='30000-32000'                    \
+                         -e 'osm_controller_args={"enable-hostpath-provisioner":["true"]}' \
+                         -e openshift_disable_check="disk_availability,memory_availability,package_availability,docker_image_availability,docker_storage,package_version" \
```
sosiouxme: only health checks in place currently are

      - disk_availability      
      - memory_availability    
      - docker_image_availability     

Of these, seems like it would still be a good idea to run the last one.

If the intent is to disable them all so as not to deal with them ever, there's:

`-e openshift_disable_check="*"`

ingvagabund: I would like to run as many as we can. Though, it takes same time to reach the checks and determine which passed and which failed. Actually, it is the last one that failed.

PR #726, opened 2017-10-12
Set openshift_release explicitly

ingvagabund:
Make sure the `openshift_release` is always set.

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
@@ -66,6 +66,7 @@ extensions:
         if [[ "${curbranch}" == master ]] ; then
            git log -1 --pretty=%h >> "${jobs_repo}/ORIGIN_COMMIT"
            ( source hack/lib/init.sh; os::build::rpm::get_nvra_vars; echo "-${OS_RPM_VERSION}-${OS_RPM_RELEASE}" ) >> "${jobs_repo}/ORIGIN_PKG_VERSION"
+           ( source hack/lib/init.sh; os::build::rpm::get_nvra_vars; echo "${OS_GIT_MAJOR}.${OS_GIT_MINOR}" | sed "s/+//" ) >> "${jobs_repo}/ORIGIN_RELEASE"
         elif [[ "${curbranch}" =~ ^release-* ]] ; then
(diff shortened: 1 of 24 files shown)


sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -108,6 +110,7 @@ extensions:
                          -e deployment_type=origin  \
                          -e etcd_data_dir="${ETCD_DATA_DIR}" \
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
+                         -e openshift_release="$( cat ./ORIGIN_SHORT_VERSION )"               \
```
stevekuznetsov: Is this an API change for the installer? Seems weird to need the version and release at the same time

stevekuznetsov: Also a little upsetting that this "release" is not the RPM release field

ingvagabund: Please see https://trello.com/c/Z83qZeDi/530-1-require-openshiftrelease

ingvagabund: It's a good point. We could check if the `openshift_pkg_version` is set and if so, detect the `openshift_release` from it. @sdodson WDYT?

sdodson: Yeah, that seems fine. If they've specified pkg_version we can set release to the first two members of that.

ingvagabund:
@stevekuznetsov @sdodson PTAL again. Based on a discussion we had on a meeting and comment https://github.com/openshift/openshift-ansible/pull/5860#issuecomment-339985704, the `openshift-release` is unconditionally required.

stevekuznetsov:
Mostly LGTM. the bash bits to get the release look really fragile though -- I wonder if you could write utilities in the repos under test that would expose the correct value, like we did with `os::rpm`

ingvagabund:
> Mostly LGTM. the bash bits to get the release look really fragile though -- I wonder if you could write utilities in the repos under test that would expose the correct value, like we did with ``os::rpm``

https://github.com/openshift/origin/pull/17355

ingvagabund:
@stevekuznetsov updated PTAL

PR #727, opened 2017-10-13
`rhel7next` scheduled job

bbguimaraes:
@jupierce

### DIFF
--- a/build-scripts/rcm-guest/rhel7next.sh
+++ b/build-scripts/rcm-guest/rhel7next.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+set -eu
+
+BASE_DIR=/mnt/redhat
+REMOTE_DIR=/srv/enterprise/rhel
+LATEST_RHEL_COMPOSE=${BASE_DIR}/nightly/latest-RHEL-7
+LATEST_EXTRAS=${BASE_DIR}/nightly/EXTRAS-RHEL-7.4/latest-EXTRAS-7-RHEL-7/compose/Server/x86_64/os
+REPO_FILE=rhel7next.repo
+MIRROR=use-mirror-upload.ops.rhcloud.com
+rsync='rsync
+--archive --hard-links --verbose --delete-after --progress --no-p --no-g
+--omit-dir-times --chmod=Dug=rwX'
+
+_reposync() {
+    local repoid
+    repoid=$1
+    reposync \
+        --config "${REPO_FILE}" \
+        --download_path "${repoid}/" --repoid "${repoid}" \
+        --arch x86_64 --download-metadata --downloadcomps --delete --norepopath
+}
+
+_createrepo_comp() {
+    local item
+    item=$1
+    createrepo \
+        --update --outputdir "${item}/" --group "${item}/comps.xml" "${item}/"
+}
+
+_createrepo() {
+    local item
+    item=$1
+    createrepo --update --outputdir "${item}/" "${item}/"
+}
+
+mkdir -p /mnt/rcm-guest/puddles/RHAOS/rhel7next/
+cd /mnt/rcm-guest/puddles/RHAOS/rhel7next/
+cat > rhel7next.repo <<-'EOF'
+	[rhel-7-fast-datapath-rpms]
+	name = Red Hat Enterprise Linux 7 Fast Datapath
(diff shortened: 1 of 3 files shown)


scheduled-jobs/build/rhel7next/Jenkinsfile
```diff
@@ -0,0 +1,39 @@
...
+node('openshift-build-1') {
+  try {
+    dir('aos-cd-jobs') {
+      def buildlib = null
+      stage('clone') {
+        checkout scm
+        buildlib = load('pipeline-scripts/buildlib.groovy')
+      }
+      stage('sync') {
+        sshagent(['openshift-bot']) {
```
bbguimaraes: Is this key necessary to ssh to `rcm-guest`?

bbguimaraes:
The script needs a directory on `rcm-guest` to sync the repositories, and it
needs to be persistent between executions to avoid downloading the same
packages again and again (or it will be really slow each time).

Do we have that?

bbguimaraes:
Also, I tested the script manually on `rcm-guest` earlier today using `rsync`'s
`--dry-run` option and everything worked fine until the first `reposync` (which
I interrupted because it will take a long time with an empty package cache).

build-scripts/rcm-guest/rhel7next.sh
```diff
@@ -0,0 +1,83 @@
...
+set -eu
+
+BASE_DIR=/mnt/redhat
+REMOTE_DIR=/srv/enterprise/rhel
+LATEST_RHEL_COMPOSE=${BASE_DIR}/nightly/latest-RHEL-7
+LATEST_EXTRAS=${BASE_DIR}/nightly/EXTRAS-RHEL-7.4/latest-EXTRAS-7-RHEL-7/compose/Server/x86_64/os
+REPO_FILE=rhel7next.repo
+MIRROR=use-mirror-upload.ops.rhcloud.com
+
+rsync=$(echo \
```
stevekuznetsov: No need to echo in a subshell -- just assign the var to a string

scheduled-jobs/build/rhel7next/Jenkinsfile
jupierce: @bbguimaraes No, the key for rcm-guest just sits on the buildvm filesystem.

build-scripts/rcm-guest/rhel7next.sh
```diff
@@ -0,0 +1,83 @@
...
+_createrepo_comp "${PWD}/rhel-7-fast-datapath-rpms"
+_createrepo_comp "${PWD}/rhel-7-fast-datapath-htb-rpms"
+# Rebuild repos without comp files
+_createrepo "${PWD}/rhel-7-server-ose-3.4-rpms"
+# rsync repos to os mirror
+${rsync} \
+    rhel-7-{fast-datapath{,-htb},server-ose-3.4}-rpms \
+    "${MIRROR}:${REMOTE_DIR}/"
+# push to all geos
+ssh "${MIRROR}" /usr/local/bin/push.enterprise.sh -v
```
jupierce: Fought with a subtle error here before: when you stream a scriptA to ssh1 and then invoke ssh2 within scriptA, ssh2 will start eating stdin and read the rest of scriptA. If someone puts anything after this line in rhel7next.sh, it might not be executed because of this idiosyncrasy. Fix is here: https://github.com/openshift/aos-cd-jobs/blob/c3ae10a82e1fbbd74d1c384d926c8a799e9ede66/jobs/starter/package-overrides/scripts/update-cluster-overrides.sh#L75-L82  . Fixing it in invoke_rcm_guest by wrapping everything in { } < /dev/null should fix it for everything.

stevekuznetsov: oh my

jupierce:
@bbguimaraes  We have plenty of persistent locations on rcm-guest. We use directories under /mnt/rcm-guest/puddles/RHAOS .

bbguimaraes:
@jupierce, @stevekuznetsov: updated.

build-scripts/rcm-guest/rhel7next.sh
bbguimaraes: Makes sense.  Fixed (but it's not pretty).

stevekuznetsov:
🔥 🎆

PR #731, opened 2017-10-13
Remove quick installer

sdodson:
Depends on https://github.com/openshift/openshift-ansible/pull/5760 https://github.com/openshift/openshift-ansible/pull/5757


stevekuznetsov:
This needs conditional logic to handle the transitional caes

PR #732, opened 2017-10-13
Prevent oit exception from stopping critical cleanup

jupierce:


jobs/build/buildvm-maint/Jenkinsfile
```diff
@@ -25,20 +25,42 @@ node('openshift-build-1') {
...
+                    sh 'git clone git@github.com:openshift/enterprise-images.git'
+                }
+
+                stage("push images") {
+                    dir ( "enterprise-images" ) {
+                        sh './oit/oit.py --user=ocp-build --group sync-3.7 distgits:push-images --to-defaults'
+                    }
+                }
+            } catch ( ex1 ) {
+                e1 = ex1
```
stevekuznetsov: is there not a `finally`?

jupierce: There is. But maintaining something like .. try { X } finally { try { Y } finally { ..Z } } .. as more steps are added isn't a great pattern. Certainly cleaner ways than what I've done here though.

PR #743, opened 2017-10-18
refactor ocp for OIT

adammhaile:
refactor refresh-images for OIT


adammhaile:
@jupierce There was an errant } that the linter didn't seem to pick up. I think all the fully left justified multi-line strings are throwing it off.

PR #746, opened 2017-10-18
Add build version to puddle directory names

bbguimaraes:
@jupierce

### DIFF
--- a/build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh
+++ b/build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh
@@ -8,15 +8,16 @@ set -e
 
 if [ "$#" != "2" ]; then
 	echo "Invalid arguments"
-	echo "Syntax: $0 <major.minor> <build_mode>"
+	echo "Syntax: $0 <version> <build_mode>"
 	echo "  Build Modes: online:int, online:stg, pre-release, release"
 	exit 1
 fi
 
 
-VERSION="${1}"
+FULL_VERSION="${1}"
 BUILD_MODE="${2}"
 BASEDIR="/mnt/rcm-guest/puddles/RHAOS"
+VERSION=$(echo "${FULL_VERSION}" | cut -d . -f 1-2)
 
 if [ "$BUILD_MODE" == "online:int" ]; then
 	REPO="online-openshift-scripts-int"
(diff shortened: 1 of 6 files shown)


bbguimaraes:
This will have a minor conflict with #743.  But it looks like the relevant part
of the pipeline wasn't changed in that PR, so it should be easily fixable.

jupierce:
@bbguimaraes Needs a rebase.

jupierce:
@bbguimaraes Any clever way we could carry the symlink with us for stage-to-prod? https://github.com/openshift/aos-cd-jobs/blob/master/jobs/sprint/stage-to-prod/Jenkinsfile

bbguimaraes:
@jupierce: the symlink is [relative](symlink), so we can be clever by being
very dumb and copying it directly from one directory to the other.  Since we
are copying [the directory the symlink points to](copy), I don't see why it
shouldn't work.

[symlink]: https://github.com/openshift/aos-cd-jobs/pull/746/files#diff-189a1d69bc25339a86bcdab4eb7b3061R109
[copy]: https://github.com/openshift/aos-cd-jobs/blob/master/build-scripts/rcm-guest/mirrors-stage-to-prod.sh#L54

bbguimaraes:
Looking at the `mirrors-to-stage-prod.sh` script, we don't have the latest
version number at that point (maybe that's what you meant in the first place
and I misunderstood =).  Can we get that information somehow from the contents
of the directory?

We could also make the build number symlink point to the date symlink, i.e.

    latest -> 3.6.0.0-0 -> 2017-10-20.1

but that does not look very good.

jupierce:
If you create the symlink during an online:stg build, then the version symlink will be there by the time I run stage-to-prod. 

lastest symlink ---->  puddle dir <--- version symlink

You would need logic to say "copy all symlinks that point to this directory"

bbguimaraes:
That's the problem, a symlink is a one-way relation, there's no reverse
operation.  We can always iterate over all of them and test each, but that's not
a pretty solution either.

jupierce:
@bbguimaraes Paradigm shift -- instead of creating a symlink, let's rename the puddle directory. After it is built (e.g. "2017-10-17.2/"), let's rename it and include the build:  "2017-10-17.2_3.7.0-0.158.0"  ; i.e. date_version-release .

That way, existing scripts should work without change. Apologies in advance for the rework on the symlink approach.

bbguimaraes:
@jupierce: updated.

build-scripts/rcm-guest/push-to-mirrors.sh
```diff
@@ -73,7 +74,7 @@ else
 fi
 
 # dereference the symlink to the actual directory basename: e.g. "2017-06-09.4"
-LASTDIR=$(readlink ${PUDDLEDIR})
+LASTDIR=$(readlink ${PUDDLEDIR})_${FULL_VERSION}
```
bbguimaraes: Here (and similarly in `push-openshift-online-to-mirrors.sh`) I'm assuming we
want to keep the puddle directory in `rcm-guest` unchanged (i.e. without the
build version).

jupierce: I would prefer we change it everywhere.

build-scripts/rcm-guest/push-openshift-online-to-mirrors.sh
```diff
@@ -46,7 +47,7 @@ if [ ! -e "${PUDDLEDIR}" ]; then
 fi
 
 # dereference the symlink to the actual directory basename: e.g. "2017-06-09.4"
-LASTDIR=$(readlink ${PUDDLEDIR})
+LASTDIR=$(readlink ${PUDDLEDIR})_${FULL_VERSION}
```
jupierce: Comment is no longer correct. Let's call out exactly what is being done.

build-scripts/rcm-guest/push-to-mirrors.sh
jupierce: Can we also prefix the version with a 'v'? date_vVERSION.

bbguimaraes:
@jupierce: updated.

PR #753, opened 2017-10-18
Run Heapster in standalone mode

DirectXMan12:
This sets up Heapster to run in standalone mode (without Hawkular), in
order to run the Kubernetes HPA e2e tests.


DirectXMan12:
Once we're running Heapster when we run CI jobs, we should be able to remove the line in Origin which disables the Kube HPA e2e tests, so that we don't break HPA policy again.

stevekuznetsov:
It's been pointed out that we would be adding a significant runtime increase with this. In the inventory right now are arguments we add to every test we ever run with Ansible. You probably only want it added to Origin e2e, so please make PRs that mirror:

https://github.com/openshift/aos-cd-jobs/pull/757
https://github.com/openshift/release/pull/275

PR #757, opened 2017-10-19
enable service catalog and tsb in CI for conformance runs

jim-minter:
re-try of https://github.com/openshift/aos-cd-jobs/pull/718 now that https://github.com/openshift/openshift-ansible/pull/5764 and https://github.com/openshift/openshift-ansible/pull/5746 are in.


stevekuznetsov:
:shipit:

PR #777, opened 2017-10-20
Add cpu_request values now that they are available

portante:


portante:
@sdodson, what do you think?

portante:
@sdodson, I think we need this to get the jobs passing, as it ensure various components get a minimum amount of CPU needed to complete their required functions.

sdodson:
Makes sense.

PR #779, opened 2017-10-23
run the registry tests nightly

bparees:


bparees:
@stevekuznetsov @miminar ptal

PR #786, opened 2017-10-24
Decrease master and node log level for logging tests

sdodson:
The working theory is that debug_level of 5 is overwhelming the logging
stack in an already overloaded host


sdodson:
It does.

stevekuznetsov:
@sdodson is this going to be overwritten by the `debug_level` var in the inventory? I can't remember the precedence rules right now

sdodson:
Cmd line should be highest.

stevekuznetsov:
@sdodson awesome!

PR #788, opened 2017-10-25
add cluster up test job

bparees:


sjb/config/test_cases/test_branch_origin_extended_clusterup.yml
```diff
@@ -18,4 +18,13 @@ extensions:
       title: "run extended tests"
       repository: "origin"
       script: |-
+        sudo systemctl stop origin-master
```
stevekuznetsov: If it's from the built release none of these should be present

bparees: I think i was using built_installed release earlier, i'll try to remove these and see if things still work now.

bparees:
these jobs have been installed on jenkins.

PR #789, opened 2017-10-26
use proper parents for extended test jobs

bparees:


bparees:
@stevekuznetsov please sanity check this...I think we were using the wrong parents for a bunch of these jobs.

For the branch tests i assume we just want to use a pre-built release for the branch.

For the PR tests, I assume we want to build an origin release (based on the PR)

Does that seem right to you?  Are these the right jobs to inherit now?

bparees:
(and i'm not sure when we'd use built release vs built_installed release)

stevekuznetsov:
> For the branch tests i assume we just want to use a pre-built release for the branch.

I would not think this is correct -- they are postsubmit tests that should check the version that was just submitted.

> For the PR tests, I assume we want to build an origin release (based on the PR)

Yes.

That being said, it still looks like your changes are right because the branch tests based off of `origin_installed_release` were being based off of a pre-built AMI and those on `origin_built_installed_release` start from scratch.

PR #798, opened 2017-10-27
push-release should use PUSH_LOCAL

smarterclayton:
Otherwise we could accidentally retag what's in the hub

@stevekuznetsov


stevekuznetsov:
Updated

PR #824, opened 2017-11-08
origin_aggregated_logging: set oauth proxy image for tests

wozniakjan:
The oauth-proxy doesn't belong to origin aggregated logging project and does not get built during the `build_images.sh` phase of our tests.

It does not share image prefix with our other images, and therefore should be referenced fully. When setting `openshift_logging_image_prefix`, we should make sure the `openshift_logging_elasticsearch_proxy_image_prefix` also exists.

These should be upstream images working defaults:
```
docker.io/openshift/oauth-proxy:v1.0.0
docker.io/openshift/origin-logging-auth-proxy:latest
docker.io/openshift/origin-logging-curator:latest
docker.io/openshift/origin-logging-elasticsearch:latest
docker.io/openshift/origin-logging-fluentd:latest
docker.io/openshift/origin-logging-kibana:latest
```

when setting `openshift_logging_image_prefix=openshift/origin-`, it overwrites these defaults for all images and breaks the oauth-proxy
```
  openshift/origin-oauth-proxy:v1.0.0     <-- doesn't exist
( docker.io/openshift/oauth-proxy:v1.0.0  <-- exists )
```
and it requires setting `openshift_logging_elasticsearch_proxy_image_prefix=docker.io/openshift/`

This is not very elegant solution, it would be better if the upstream image prefixes were consistent just like our downstream prefixes are, but currently I don't see any other fix. We could also take oauth-proxy under logging team's wings and build it along the other logging images, but I was informed this is not desired.

https://hub.docker.com/r/openshift/oauth-proxy/

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
@@ -140,6 +140,7 @@ extensions:
                          -e debug_level=2           \
                          -e openshift_logging_install_logging=True \
                          -e openshift_logging_image_prefix="openshift/origin-" \
+                         -e openshift_logging_elasticsearch_proxy_image_prefix="docker.io/openshift/" \
                          -e openshift_hosted_logging_hostname="kibana.127.0.0.1.nip.io"           \
                          -e openshift_logging_master_public_url="https://localhost:8443"          \
                          -e openshift_master_logging_public_url="https://kibana.127.0.0.1.nip.io" \
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml
@@ -136,6 +136,7 @@ extensions:
                          -e openshift_logging_install_logging=True \
                          -e debug_level=2           \
                          -e openshift_logging_image_prefix="openshift/origin-" \
+                         -e openshift_logging_elasticsearch_proxy_image_prefix="docker.io/openshift/" \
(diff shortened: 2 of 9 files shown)


stevekuznetsov:
This seems fine to me. Alternatively we could think about locally re-tagging if we can't get the names to match. @wozniakjan have you tested this?

wozniakjan:
currently running test here: https://ci.openshift.redhat.com/jenkins/job/test_branch_openshift_ansible_logging_testing_wip/1/console

it will fail for fluentd because we have broken fluentd image
this broke it: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/89
this is going to fix it: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/94

But I can check event.log to see if it got elasticsearch proxy running

stevekuznetsov:
OK! Let me know when you'd like this to merge. The approach is
/lgtn

wozniakjan:
safe to merge, the events show that images got pulled correctly

https://ci.openshift.redhat.com/jenkins/job/test_branch_openshift_ansible_logging_testing_wip/1/s3/download/gathered/scripts/entrypoint/artifacts/events.txt

Is it ok if I delete this job https://ci.openshift.redhat.com/jenkins/job/test_branch_openshift_ansible_logging_testing_wip ? Do I do it by clicking 'disable project' or 'delete multijob project'? It was created only for this test

stevekuznetsov:
Click `delete multijob project`

/lgtm

stevekuznetsov:
@wozniakjan you can push this up to the normal job whenever you are good

PR #845, opened 2017-11-14
Re-include the commit hash when vendoring in web-console

smunilla:


jobs/build/ocp/Jenkinsfile
```diff
@@ -317,7 +318,7 @@ node(TARGET_NODE) {
                                 grunt build
 
                                 git add dist
-                                git commit -m "Merge master into enterprise-${BUILD_VERSION}" --allow-empty
+                                git commit -m "Merge master into enterprise-${BUILD_VERSION}\nWeb console commit: ${WEB_CONSOLE_COMMIT}" --allow-empty
                             """
```
jwforres: i dont think this is the right commit, if i understand this right this is the commit that is going to the enterprise branch of the web console repo.  What we need updated is the commit that gets generated from hack/vendor-console.sh that commits the bindata.go file into ose

smunilla: Ah! That explains some things. I'll update to call that instead.

smunilla:
Shot down a weird path here. Fixing.

jwforres:
i dont know tito, but this seems better

xingxingxia:
Thanks for the PR. Subscribing it and looking forward to its merge

jobs/build/ocp/Jenkinsfile
```diff
@@ -376,7 +376,7 @@ node(TARGET_NODE) {
                 buildlib.set_rpm_spec_version( "origin.spec", NEW_VERSION )
                 buildlib.set_rpm_spec_release_prefix( "origin.spec", NEW_RELEASE )
                 // Note that I did not use --use-release because it did not maintain variables like %{?dist}
-                sh "tito tag --accept-auto-changelog --keep-version --debug"
+                sh 'tito tag --accept-auto-changelog --keep-version --debug --changelog="Automatic commit of package [atomic-openshift] release [${NEW_VERSION}-${NEW_RELEASE}]; bump origin-web-console ${VC_COMMIT}"'
```
jupierce: ' in groovy means don't do any string substitution for ${vars}, so your ${NEW_VERSION} etc are literal values here. Change to:
```
sh "tito tag --accept-auto-changelog --keep-version --debug --changelog='Automatic commit of package [atomic-openshift] release [${NEW_VERSION}-${NEW_RELEASE}]; bump origin-web-console ${VC_COMMIT}' "
```

Groovy will do the string substitution and ultimately pass bash a fully formed string in single quotes for the --changelog parameter.

xingxingxia:
It is appreciated if it could pass checks and merge as early as possible :) so easy way https://url.corp.redhat.com/559353b can be used in feature testing.
Without this, today the way to check whether https://trello.com/c/zHWEtHgj/ can be testable is, launching Origin env (v3.8.0-alpha.0+fe4f498-286 in latest AMI, the AMI is built today), but found again the PR of https://trello.com/c/zHWEtHgj/ does not come in. (Why it still does not come in to latest AMI Origin, is another confusion :) )

PR #851, opened 2017-11-17
Sync the image registry whenever we are building it

stevekuznetsov:

/cc @bparees @deads2k @smarterclayton

### DIFF
--- a/sjb/config/test_cases/test_pull_request_openshift_ansible_extended_conformance_install_containerized.yml
+++ b/sjb/config/test_cases/test_pull_request_openshift_ansible_extended_conformance_install_containerized.yml
@@ -7,3 +7,4 @@ overrides:
     - name: "openshift-ansible"
       type: "pull_request"
     - name: "origin"
+    - name: "image-registry"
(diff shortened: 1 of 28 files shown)


sjb/config/test_cases/test_pull_request_openshift_ansible_extended_conformance_install_containerized.yml
```diff
@@ -7,3 +7,4 @@ overrides:
     - name: "openshift-ansible"
       type: "pull_request"
     - name: "origin"
+    - name: "image-registry"
```
smarterclayton: This doesn't scale.  Let's talk on monday

stevekuznetsov: The fact that we only allow for "extension" or "override" behavior in sjb framework makes this suck. I would not expect us to have this issue in the future with the provision+install pipeline. I think we can create an abstraction where you can provide whatever explicit sources/steps you want and let us default the rest, so that e.g. tests for OpenShift Ansible sub in their PR code for the installer image, image-registry tests sub in their container for the deployment, etc. /cc @csrwng

PR #854, opened 2017-11-18
use yum to find pkg verrel if tag doesn't work

richm:
use yum to find pkg verrel if tag doesn't work
This is a hack, but I don't know what else to do until the package
verrel for the 3.7 branch packages are a little more reasonable

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
@@ -90,6 +90,14 @@ extensions:
            echo "${closest_tag:1}" | cut -d'.' -f1,2 > ${jobs_repo}/ORIGIN_RELEASE
            # disable local origin repo
            sudo yum-config-manager --disable origin-local-release
+           # hackety hackety hack hack - pre-releases use a *wack[y]* versioning scheme
+           if ( sudo yum install --assumeno origin${pkgver} 2>&1 || : ) | grep -q 'No package .* available' ; then
+              # just ask yum what the heck the version is
+              pkgver=$( ( sudo yum install --assumeno origin 2>&1 || : ) | awk '$1 == "x86_64" {print $2}' )
+              echo "-${pkgver}" > ${jobs_repo}/ORIGIN_PKG_VERSION
+           else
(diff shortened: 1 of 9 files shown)


richm:
@stevekuznetsov @jcantrill logging CI for 3.7 branch is broken without this patch.  I've successfully tested with a cherrypick for 3.7

stevekuznetsov:
Can you explain the issue a bit more here?

richm:
We use this method to determine the correct rpm package verrel and image tag to use for branch releases: https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml#L67

specifically, this bit:
```
           # pkg ver is closest_tag with leading "-" instead of "v"
           pkgver=$( echo "${closest_tag}" | sed 's/^v/-/' )
```
e.g. for logging branch `release-3.7` it assumes there is a tag like `v3.7.0-rc.0`, which translates into an rpm package verrel like `-3.7.0-rc.0`.  Except this is not working for 3.7 pre-release(?) packages which have a verrel like this: `-3.7.0-0.rc.0.0.$commit_hash` where `$commit_hash` is the commit hash from origin that was used to produce the rpms.

Maybe this will all go away once 3.7 is officially released, but for now it takes this gross hack to figure out what the rpm verrel is.

stevekuznetsov:
Hmm. SJB really sucks for this. Why do you need to resolve the closest tag? If we're on `release-3.6` you should be OK to use `v3.6` or `-3.6` for the version, won't that give you latest 3.6?

richm:
> won't that give you latest 3.6?

Yes, works fine for 3.6, but fails for 3.7.

stevekuznetsov:
D'oh. Ew ew ew ew. Sorry about this mess. Are you happy with these changes? I can merge now.

richm:
I can't really say that I'm "happy" with these changes but this will have to do for now so we can have CI working for release-3.7.

stevekuznetsov:
@richm Yep. Can you publish these or should I?

richm:
@stevekuznetsov I will publish - thanks!

richm:
pushed

PR #858, opened 2017-11-20
Add prerequisite play before install origin

michaelgugino:
This commit adds the prerequisites play to all jobs.

This play should be run before 'install origin' is run
to ensure that all host prerequisites are run before
installing origin.


michaelgugino:
This depends on: https://github.com/openshift/openshift-ansible/pull/6188

PR #860, opened 2017-11-21
Heavy-handed hack to stop CentOS PaaS SIG repos from breaking jobs

stevekuznetsov:

/assign @sdodson


stevekuznetsov:
here goes nothing...

PR #866, opened 2017-11-22
Amend catalog jobs to specify user broker image

jpeeler:
Depends on openshift/origin#17378, currently deployed in Jenkins. Would like for the rebase PR to land and then double check the daily job works correctly as expected.


jpeeler:
I just tested the CI run here since the rebase referenced above has landed and it passed:
https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_service_catalog/

So this should be ready for merge @stevekuznetsov.

PR #868, opened 2017-11-23
Fixes for aws-ami

bbguimaraes:
@kwoodson, @jupierce: I'm opening this PR to start the discussion.

It contains only minor fixes to the pipeline, no major change in functionality.
I was able to reproduce the problem we had in prod on my machine and I believe
it happens because some of the things referenced in it do not exist (or my
account and the account we use in prod don't have access to it).

With those replaced with the ones we use for devenvs (see last commit) the job
is able to advance, provision the instance and start the node setup.  It now
fails when applying the `openshift_excluder` role (but that may be because of
the different base ami, I'll keep investigating).

### DIFF
--- a/jobs/build/aws-ami/Jenkinsfile
+++ b/jobs/build/aws-ami/Jenkinsfile
@@ -42,30 +42,19 @@ if ( MOCK.toBoolean() ) {
 }
 
 node(TARGET_NODE) {
-
-    set_workspace()
-
-    stage('build') {
-        try {
-
-            sh(script: '''if [ -d "${WORKSPACE}/openshift-ansible" ]; then 
-                cd ${WORKSPACE}/openshift-ansible ; 
-                git pull; 
-            else
-                git clone https://github.com/openshift/openshift-ansible
-            fi
-
-            cd ${WORKSPACE}
-
-            ''')
-
+    try {
+        set_workspace()
+        stage('clone') {
+            dir('openshift-ansible') {
+                git 'https://github.com/openshift/openshift-ansible.git'
+            }
+        }
+        stage('build') {
             // create the provisioning_vars.yml file to use as inventory
-            withEnv(["AWS_PROFILE=${WORKSPACE}/.aws/credentials", "PATH=${pwd()}/${PATH}",
-                     "RELEASE_VERSION=${RELEASE_VERSION}", "DEPLOYMENT_TYPE=${DEPLOYMENT_TYPE}",
-                     "BASE_AMI_ID=${BASE_AMI_ID}", "CLUSTERID=${CLUSTERID}",
-                     "AWS_SSH_KEY_USER=${AWS_SSH_KEY_USER}","YUM_BASE_URL=${YUM_BASE_URL}","BASE_AMI_NAME=${BASE_AMI_NAME}"]) {
-                withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'cloud-provider-creds']]) {
-                    writeFile file:"${WORKSPACE}/provisioning_vars.yml", text:"""---
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'cloud-provider-creds']]) {
+                writeFile(
+                    file: 'provisioning_vars.yml',
+                    text: """---
 openshift_aws_create_vpc: False
 openshift_deployment_type: ${DEPLOYMENT_TYPE}
 openshift_aws_clusterid: ${CLUSTERID}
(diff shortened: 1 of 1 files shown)


jobs/build/aws-ami/Jenkinsfile
```diff
@@ -42,34 +42,38 @@ if ( MOCK.toBoolean() ) {
...
-            ''')
-
+    try {
+        set_workspace()
+        def buildlib = null
+        stage('clone') {
+            checkout scm
+            buildlib = load('pipeline-scripts/buildlib.groovy')
+            dir('openshift-ansible') {
+                git 'https://github.com/openshift/openshift-ansible.git'
```
kwoodson: What does this do?  I'm assuming it performs a `git clone`.  

The original intention was to check it out if it wasn't there or git pull if it is.

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -81,30 +85,30 @@ openshift_cloudprovider_aws_access_key: ${env.AWS_ACCESS_KEY_ID}
...
-"""
-
-                    sshagent(["${AWS_SSH_KEY_USER}"]) {
-                        // need to load aws credentials
-                        withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'ami-build-creds']]) {
-                        sh("cat ${WORKSPACE}/provisioning_vars.yml")
-                        sh(script: "ANSIBLE_HOST_KEY_CHECKING=False AWS_ACCESS_KEY_ID=${env.AWS_ACCESS_KEY_ID} AWS_SECRET_ACCESS_KEY=${env.AWS_SECRET_ACCESS_KEY} ansible-playbook ${WORKSPACE}/openshift-ansible/playbooks/aws/openshift-cluster/build_ami.yml -e @${WORKSPACE}/provisioning_vars.yml -vvv")
+""")
+            }
+            sh 'cat provisioning_vars.yml'
```
kwoodson: This was for debug.  I'm fine keeping it or dropping it.  Kind of nice to see what settings are being used.

Thoughts?

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -42,34 +42,38 @@ if ( MOCK.toBoolean() ) {
...
             // create the provisioning_vars.yml file to use as inventory
-            withEnv(["AWS_PROFILE=${WORKSPACE}/.aws/credentials", "PATH=${pwd()}/${PATH}",
-                     "RELEASE_VERSION=${RELEASE_VERSION}", "DEPLOYMENT_TYPE=${DEPLOYMENT_TYPE}",
-                     "BASE_AMI_ID=${BASE_AMI_ID}", "CLUSTERID=${CLUSTERID}",
-                     "AWS_SSH_KEY_USER=${AWS_SSH_KEY_USER}","YUM_BASE_URL=${YUM_BASE_URL}","BASE_AMI_NAME=${BASE_AMI_NAME}"]) {
-                withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'cloud-provider-creds']]) {
-                    writeFile file:"${WORKSPACE}/provisioning_vars.yml", text:"""---
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'cloud-provider-creds']]) {
+                env.CLUSTERID = 'devenv'
+                env.BASE_AMI_ID = 'ami-10aed306'
```
kwoodson: I don't think we want this hard coded.  This should be a parameter so we build on whichever AMI is passed.

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -42,34 +42,38 @@ if ( MOCK.toBoolean() ) {
...
+                env.AWS_SSH_KEY_USER = 'libra'
```
kwoodson: This should be passed in as well I think.  These should be the same as the user that is inside of AWS

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -81,30 +85,30 @@ openshift_cloudprovider_aws_access_key: ${env.AWS_ACCESS_KEY_ID}
...
-            // Replace flow control with: https://jenkins.io/blog/2016/12/19/declarative-pipeline-beta/ when available
-            mail(to: "${MAIL_LIST_FAILURE}",
-                    from: "kwoodson@redhat.com",
-                    subject: "Error building aws ami",
-                    body: """Encoutered an error while running build_ami.yml: ${err}
+        }
+    } catch ( err ) {
+        // Replace flow control with: https://jenkins.io/blog/2016/12/19/declarative-pipeline-beta/ when available
+        mail(to: "${MAIL_LIST_FAILURE}",
+                from: "kwoodson@redhat.com",
```
kwoodson: Would love to change this from address :)

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -42,34 +42,38 @@ if ( MOCK.toBoolean() ) {
...
+                env.AWS_SSH_KEY_USER = 'libra'
+                writeFile(
+                    file: 'provisioning_vars.yml',
+                    text: """---
 openshift_aws_create_vpc: False
+openshift_aws_subnet_name: devenv-subnet-1
 openshift_deployment_type: ${DEPLOYMENT_TYPE}
 openshift_aws_clusterid: ${CLUSTERID}
 openshift_clusterid: ${CLUSTERID}
+openshift_aws_build_ami_group: default
```
kwoodson: This references the security group inside of the openshift.  I'd exclude this variable or variable-ize it to default to `${CLUSTERID}`

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -42,34 +42,38 @@ if ( MOCK.toBoolean() ) {
...
-                    writeFile file:"${WORKSPACE}/provisioning_vars.yml", text:"""---
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'cloud-provider-creds']]) {
+                env.CLUSTERID = 'devenv'
+                env.BASE_AMI_ID = 'ami-10aed306'
+                env.AWS_SSH_KEY_USER = 'libra'
+                writeFile(
+                    file: 'provisioning_vars.yml',
+                    text: """---
 openshift_aws_create_vpc: False
+openshift_aws_subnet_name: devenv-subnet-1
```
kwoodson: Shouldn't have to specify subnet.  These should be set before the provisioning happens or variable-ize this param.

jobs/build/aws-ami/Jenkinsfile
bbguimaraes: Thanks for clarifying.  I imagined that was the original purpose so I changed
it to use the more concise `git` command, which does exactly that.

jobs/build/aws-ami/Jenkinsfile
bbguimaraes: I'd leave it, it is useful information in the build log.

jobs/build/aws-ami/Jenkinsfile
bbguimaraes: Somehow I read that as 'to', not 'from'.  Changed it to the address we use on
other jobs.

bbguimaraes:
kwoodson: thanks for the review.  I've addressed some of the comments and
replied to them.

As for the others, I evidently didn't make myself clear (sorry about that).
The last commit on this PR is only meant to point out the variables I had to
change when I executed the job on my machine, using my aws account, in order to
get past the errors we saw in prod.

For example: the error we got was when querying for the `opstest` vpc.  That
apparently doesn't exist or can't be accessed by the account (you probably know
better than I), so I changed `CLUSTERID` to `devenv` to use that vpc (which the
account *can* access) instead, just to get past the error.  Then the same
happened with the subnet, so I added `openshift_aws_subnet_name`, and so on.

I hope it's clearer now and it points us to the things we need to change to fix
the issue.

kwoodson:
@bbguimaraes, 

Thanks for your explanation I understand now.  I'm not a jenkins expert but it was easier to fit my values in as the defaults for my testing.

Were there any action items here?  I think as long as we keep all the necessary variables as parameters we should be ok.

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -25,7 +25,7 @@ properties(
                                                     openshift-enterprise      Openshift Enterpriseonline <br>
                                     ''',
                                     name: 'DEPLOYMENT_TYPE'],
-                                 [$class: 'hudson.model.StringParameterDefinition', defaultValue: 'opstest', description: 'Clusterid to set for ami build. Also used for VPC and other settings', name: 'CLUSTERID'],
+                                 [$class: 'hudson.model.StringParameterDefinition', defaultValue: 'default', description: 'Clusterid to set for ami build. Also used for VPC and other settings', name: 'CLUSTERID'],
```
bbguimaraes: I changed this to use the `default` cluster and vpc (which the account has).
Is that a reasonable value for this variable?

jobs/build/aws-ami/Jenkinsfile
```diff
@@ -25,7 +25,7 @@ properties(
...
                                  [$class: 'hudson.model.StringParameterDefinition', defaultValue: '3.7.0', description: 'Release version (matches version in branch name for release builds)', name: 'RELEASE_VERSION'],
                                  [$class: 'BooleanParameterDefinition', defaultValue: false, description: 'Mock run to pickup new Jenkins parameters?.', name: 'MOCK'],
                                  [$class: 'hudson.model.StringParameterDefinition', defaultValue: 'ami-068a787c', description: 'Base AMI id to build from.', name: 'BASE_AMI_ID'],
```
bbguimaraes: If the job is supposed to be part of the build process, how do we determine
which AMI to use as a base?

bbguimaraes:
@kwoodson: a few updates after our discussion.

jobs/build/aws-ami/Jenkinsfile
kwoodson: This is a good question.  I hate to hard code it because if someone comes along and updates the base AMI to have the latest errata or updates we need to build on top of it.

Is there a way to force someone to pass in the AMI id? I defaulted in my test but I'm not sure we want a default.  If a security vulnerability is found in our default AMI or a mistake happens then we never want to build off this default.  We probably need to be told from which AMI we need to build on top of.  For this reason I placed it as a parameter.

The new current AMI should be updated as I built a recent, clean RHEL 7.4 image.

jobs/build/aws-ami/Jenkinsfile
kwoodson: I think its safe to change this to something else but what that something else is I'm not sure.  There are discussions ongoing about where this work will land.  When it lands and where it lands (which AWS account) will then help us better determine what this value should be.  

I think the push will be to get CICD team their own AWS account so that they can build a space for jobs like these that require some known set of environment and infrastructure settings.  Until then feel free to change this to whatever makes it work.

jobs/build/aws-ami/Jenkinsfile
bbguimaraes: We can make it a required parameter, for sure.  But if we want this jobs to be
part of the build process, that only pushes the problem up the stack: the job
that will call this one will have to know which AMI to pass as a parameter.

If we don't have a way to programmatically find the name of the image, then
making it a parameter of the (main) build job might be a solution.

@jupierce: would that be a problem?

kwoodson:
I think we want to test with these latest changes.

/lgtm

PR #871, opened 2017-11-23
Removing use of --include ; rely on unmanaged dirs in oit

jupierce:
fyi @adammhaile . Most image config.yml files in oit will be moved to an "unmanaged" directory so that we don't have to keep all of these --include / --optional-include parameters up to date as images are transitioned. We just move group/openshiftX.Y/unmanaged/distgit_name -> ../images/distgit_name in order to move an image over to oit management.


adammhaile:
ACK. Good idea.

PR #877, opened 2017-11-27
Updating location of logging playbook for master branch

ewolinetz:
Corresponding change for https://github.com/openshift/openshift-ansible/pull/6154
This should only be for master branch.

@stevekuznetsov 

CC @mtnbikenc @richm

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
@@ -144,6 +144,12 @@ extensions:
       title: "install origin-aggregated-logging"
       repository: "aos-cd-jobs"
       script: |-
+        playbook_base='/usr/share/ansible/openshift-ansible/playbooks/'
+        if [[ -s "${playbook_base}openshift-logging/config.yml" ]]; then
+            playbook="${playbook_base}openshift-logging/config.yml"
+        else
+            playbook="${playbook_base}byo/openshift-cluster/openshift-logging.yml"
+        fi
         ansible-playbook -vv --become               \
                          --become-user root         \
                          --connection local         \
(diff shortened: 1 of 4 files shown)


richm:
I'm not sure how aos-cd-jobs branches work, but afaik, the same sjb config runs against _all_ branches.  So if you want to use a different playbook location, you'll need to add some logic to the shell part to see if the new file exists, use it if it exists, or fall back to the old file if the new file doesn't exist.

stevekuznetsov:
Yes, that is unfortunately the (sad) state of affairs @richm

ewolinetz:
@mtnbikenc is this PR no longer needed then for https://github.com/openshift/openshift-ansible/pull/6154 ?

mtnbikenc:
@ewolinetz This PR or a similar one is still needed for updating the location of the playbooks.  I updated https://github.com/openshift/openshift-ansible/pull/6154 so that it wouldn't break logging CI, and we could get this updated while having both new and old files in place.

ewolinetz:
Updated to use the new logging playbook if its available or fall back to the prior location
PTAL

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -144,6 +144,8 @@ extensions:
       title: "install origin-aggregated-logging"
       repository: "aos-cd-jobs"
       script: |-
+        logging_playbook='/usr/share/ansible/openshift-ansible/playbooks/openshift-logging/config.yml'
+        [[ ! -f ${logging_playbook} ]] && logging_playbook='/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/openshift-logging.yml'
```
stevekuznetsov:
```sh
playbook_base='/usr/share/ansible/openshift-ansible/playbooks/'
if [[ -s "${playbook_base}openshift-logging/config.yml" ]]; then
    playbook="${playbook_base}openshift-logging/config.yml"
else
    playbook="${playbook_base}byo/openshift-cluster/openshift-logging.yml"
fi
```

ewolinetz:
@stevekuznetsov addressed your nit

stevekuznetsov:
/lgtm

@ewolinetz or @richm could you do the push?

PR #879, opened 2017-11-28
Add fedora/27/atomic to master blocking tests

sdodson:


sdodson:
The test has been green for a week now. though I'm not 100% certain this is still the correct way to make the test blocking now that we're using submit queue. Do you know @bbguimaraes

stevekuznetsov:
@sdodson you want to add it here https://github.com/openshift/release/blob/master/cluster/ci/config/submit-queue/submit_queue_openshift_ansible_config.yaml#L28

/close

PR #895, opened 2017-12-01
get list of repos using repolist all, and parse out the exclamation point

richm:
There is a job that runs before logging starts that disables
all of the centos-paas repos, and doing a yum repolist all
has an output like this:

    !centos-paas-sig-openshift-origin13-rpms                    C disabled
    !centos-paas-sig-openshift-origin14-rpms                    C disabled
    !centos-paas-sig-openshift-origin15-rpms                    C disabled
    !centos-paas-sig-openshift-origin36-rpms                    C disabled
    !centos-paas-sig-openshift-origin37-rpms                    C disabled

So have to add a bit of extra logic to look for `!centos-paas-...` and
strip off the exclamation point.
This makes logging CI work again.
@stevekuznetsov @jcantrill

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
@@ -79,7 +79,7 @@ extensions:
            # disable all of the centos repos except for the one for the
            # version being tested - this assumes a devenv environment where
            # all of the repos are installed
-           for repo in $( sudo yum repolist | awk '/^centos-paas-sig-openshift-origin/ {print $1}' ) ; do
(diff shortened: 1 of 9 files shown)


stevekuznetsov:
Are you seeing the wrong repo be enabled? If my logic is correct you should be able to just remove yours -- sorry for not noticing it would conflict!

richm:
I'm seeing the right repo not be enabled (because a. I was using `yum repolist` instead of `yum repolist all` and b. because the disabled repos were being listed with a leading `!`) - hence my hack.

I hope that all of this hackery will go away soon if I can get logging CI working with oc cluster up . . .

stevekuznetsov:
OK! LGTM

stevekuznetsov:
@richm can you push out?

richm:
ok

richm:
pushed

PR #896, opened 2017-12-01
Add openshift ansible crio pr test

sdodson:


sdodson:
@stevekuznetsov is this necessary or can we just add the prow config to call the existing crio job?
CC: @runcom @ashcrow

stevekuznetsov:
No, we need this or we won't be syncing an O-A PR

sdodson:
So LGT @stevekuznetsov  ?

stevekuznetsov:
Sure -- you will need to add the config to run it, too -- it should not be blocking today (the origin version is at like <10% success rate ATM)

PR #897, opened 2017-12-01
push a version tag for jenkins centos images

bparees:


stevekuznetsov:
is this the point where we make a for loop?

bparees:
obviously not. :)

sjb/config/test_cases/push_jenkins_images.yml
```diff
@@ -32,22 +32,32 @@ extensions:
         sudo chmod a+rw /tmp/.docker/config.json
         export DOCKER_CONFIG=/tmp/.docker
 
+        version_tag=`cat release.version`
```
stevekuznetsov: use `$()`

bparees: thanks, that worked.

PR #901, opened 2017-12-02
CRI-O: match version against openshift

runcom:
@mrunalp @kargakis @stevekuznetsov PTAL this will fix https://github.com/openshift/origin/pull/17546 by pulling the correct image for the CRI-O system container

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
@@ -93,6 +93,21 @@ extensions:
       title: "origin prerequisites"
       repository: "aos-cd-jobs"
       script: |-
+        crio_tag="$( cat ./ORIGIN_PKG_VERSION | grep -Po '\d.\d' )"
+        case $crio_tag in
+        "3.7")
+          crio_image="docker.io/runcom/cri-o-system-container:v3.7"
+          ;;
+        "3.8")
+          crio_image="docker.io/runcom/cri-o-system-container:v3.8"
+          ;;
+        "3.9")
+          crio_image="docker.io/runcom/cri-o-system-container:v3.9"
+          ;;
+        ""|*)
+          crio_image="docker.io/runcom/cri-o-system-container:latest"
+          ;;
+        esac
         local_ip="$( curl http://169.254.169.254/latest/meta-data/local-ipv4 )"
         ansible-playbook -vv --become               \
                          --become-user root         \
@@ -100,7 +115,7 @@ extensions:
                          --inventory sjb/inventory/ \
                          -e deployment_type=origin  \
                          -e openshift_use_crio=True   \
-                         -e openshift_crio_systemcontainer_image_override=docker.io/gscrivano/cri-o-centos \
+                         -e openshift_crio_systemcontainer_image_override="${crio_image}" \
                          -e etcd_data_dir="${ETCD_DATA_DIR}" \
                          -e openshift_master_default_subdomain="${local_ip}.nip.io"             \
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
@@ -119,6 +134,21 @@ extensions:
         else
             playbook="${playbook_base}byo/openshift-node/network_manager.yml"
         fi
+        crio_tag="$( cat ./ORIGIN_PKG_VERSION | grep -Po '\d.\d' )"
+        case $crio_tag in
+        "3.7")
+          crio_image="docker.io/runcom/cri-o-system-container:v3.7"
+          ;;
+        "3.8")
+          crio_image="docker.io/runcom/cri-o-system-container:v3.8"
+          ;;
+        "3.9")
+          crio_image="docker.io/runcom/cri-o-system-container:v3.9"
+          ;;
+        ""|*)
+          crio_image="docker.io/runcom/cri-o-system-container:latest"
+          ;;
+        esac
         local_ip="$( curl http://169.254.169.254/latest/meta-data/local-ipv4 )"
         ansible-playbook -vv --become               \
                          --become-user root         \
@@ -137,7 +167,7 @@ extensions:
                          --inventory sjb/inventory/ \
                          -e deployment_type=origin  \
                          -e openshift_use_crio=True   \
-                         -e openshift_crio_systemcontainer_image_override=docker.io/gscrivano/cri-o-centos \
+                         -e openshift_crio_systemcontainer_image_override="${crio_image}" \
                          -e etcd_data_dir="${ETCD_DATA_DIR}" \
                          -e openshift_master_default_subdomain="${local_ip}.nip.io"             \
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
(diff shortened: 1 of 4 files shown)


runcom:
Tag mapping lives here for now: https://hub.docker.com/r/runcom/cri-o/tags/

stevekuznetsov:
Historically logic like this has a habit of breaking whenever something crazy happens with versioning. I think it may be better to have a switch-case and very explicit requirements?

runcom:
You mean to not concatenate strings to build up the image string for the installer?  We can avoid that, but that requires us to update this repo every time a new release come up no? I think I'm missing something

stevekuznetsov:
Yes, it would require it to be updated. If you think releases are very frequent then maybe it's a bad idea. If it's once per release of Origin that's usually once every couple of months and totally OK. I would rather heave a static list/mapping we have to update than logic that fails in some weird edge cases or ends up installing something unexpected. Switch/case can have a well defined `default`

sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
```diff
@@ -118,7 +119,7 @@ extensions:
                          --inventory sjb/inventory/ \
                          -e deployment_type=origin  \
                          -e openshift_use_crio=True   \
-                         -e openshift_crio_systemcontainer_image_override=docker.io/gscrivano/cri-o-centos \
+                         -e openshift_crio_systemcontainer_image_override="docker.io/runcom/cri-o-system-container:v${crio_tag}" \
```
ashcrow: Is ``crio_tag`` being set? In output that @mrunalp passed over to @sdodson and I the end result looked like:

```docker.io/runcom/cri-o-system-container:v```

Comment (giuseppe) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: true.  If this is the version currently used by the CI then `crio_tag` looks empty

Comment (ashcrow) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: This *seems* to be merged .... but not "merged". 😕

Comment (mrunalp) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: I just talked with @runcom He deployed it to test to see if it works. We need to debug why the tag isn't getting set.

Comment (runcom) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: yup, I'll debug it asap

Comment (ashcrow) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: @runcom were you able to find the issue?

Comment (runcom) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: I'm debugging here just right now, I'll update this once I find something

Comment (runcom) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: should be fixed now, I'll work on Steve's switch case for images

Comment (runcom) [reply sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml]: I've deployed this to the origin CI already so we can test if it fixes the empty crio_tag issue

**Comment by runcom** [comment]
@stevekuznetsov added the case statement for the crio image PTAL (deployed to the CI as well for the crio jobs)

File: sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml [hunk 2]
```diff
@@ -105,6 +105,21 @@ extensions:
       timeout: 7200
       repository: "aos-cd-jobs"
       script: |-
+        crio_tag="$( cat ./ORIGIN_PKG_VERSION | cut -d'-' -f2 | cut -d'.' -f1,2 )"
```
stevekuznetsov: I was also interested in _this_ being fragile. Can you use regex in the `case` statements?

runcom: You mean extracting something like "3.9" with a regexp from that file? I could do that yeah

stevekuznetsov: Right

runcom:
@ashcrow so I've fixed the image tag but now we have another issue https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_crio/185/consoleFull#54402166758b6e51eb7608a5981914356

which is not related to CRI-O

stevekuznetsov:
Are you starting the job manually? Use `/test`. The job expects many parameters to be set that were not provided in that run.

runcom:
> Are you starting the job manually? Use /test. The job expects many parameters to be set that were not provided in that run.

oh, I see, I'll just /test crio then

runcom:
So it's failing again https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_extended_conformance_crio/427/consoleFull#-190934373858b6e51eb7608a5981914356

but this time it tries to restart CRI-O but it hasn't been installed, has something changed in openshift-ansible that now forgets to "atomic install" CRI-O?

@giuseppe @ashcrow

ashcrow:
@runcom not that I'm aware of ... but looking at the code it doesn't look correct. It literally installed node items but didn't install the system container 😕. 

I see that the ``docker`` role has been renamed ``container_runtime`` (which makes sense) ... I wonder if something slipped through the cracks.

ashcrow:
I **think** [this](https://github.com/openshift/openshift-ansible/commit/5120f8e90c0178ac7f6d911159ceb278dd87b4c9#diff-0d092b0eb8108379abb4511b6bd1c47dL20) may be it but I'm not sure. The role used to require the ``docker`` role, but it no longer has a dependency.

@michaelgugino ^^ Thoughts?

ashcrow:
Looks like it merged yesterday https://github.com/openshift/openshift-ansible/pull/6297

runcom:
So that's another reason to test cri-o as well in openshift ansible ha

michaelgugino:
Docker has been renamed to container_runtime to more accurately reflect its purpose.

This must now be called from prerequisites.yml before calling byo/config.yml

ashcrow:
@michaelgugino thanks. Was there a reason this wasn't called already or is it a bug from refactoring?

giuseppe:
I have some patches here, I will push them shortly.

Another thing, we should enable container-engine in the system-containers tests.

sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
runcom: fixed, now using grep

sdodson:
> @michaelgugino thanks. Was there a reason this wasn't called already or is it a bug from refactoring?

It's intentional. So we need to ensure those vars are set when calling prerequisites too before this PR will produce the desired outcome.

michaelgugino:
Any and all variables used with any and all of the plays must be used with prerequisites.yml.

In the case of container_runtime, that is all configured during prerequisites.yml.  prerequisites.yml configures all of the host-level settings for each container runtime.  In the case of system containers, this is when the runtime images are pulled and configured.

I highly recommend refactoring some of the tests to use a extra_vars file instead of passing most things via cli in the script.  This will help alleviate some of the issues in the future.

mrunalp:
@michaelgugino Is that something which can be done in a follow-on PR?

ashcrow:
@sdodson++

mrunalp:
@sdodson Thanks!

mrunalp:
@michaelgugino Does it look okay now with @sdodson commit?

ashcrow:
I opened up https://github.com/runcom/cri-o-system-container/issues/1 to ensure that this is the image that is in @runcom's namespace. Pulling and looking at it I'm pretty sure it is.

PR #915, opened 2017-12-05
logging use oc cluster up to deploy openshift

richm:
Use oc cluster up to deploy openshift.  Uses whatever
origin/oc is on the machine.  Assumes that origin 3.x can be used
to deploy logging images for 3.y, and run CI tests for logging 3.y.
Do not build and install openshift-ansible - this should help avoid
strange problems with tito tagging.  Instead, just use the already
cloned ansible source code in place to install logging.
Note: there is some hackery around having to restart docker and
oc cluster up, due to the fact that inter-pod networking for logging
does not work otherwise (e.g. fluentd cannot talk to elasticsearch).
@jcantrill @stevekuznetsov @ewolinetz @nhosoi PTAL

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
@@ -19,33 +19,14 @@ extensions:
         fi
         hack/build-images.sh
     - type: "script"
-      title: "build an openshift-ansible release"
-      repository: "openshift-ansible"
+      title: "install Ansible and other packages needed for OpenShift Ansible tasks"
+      repository: "origin-aggregated-logging"
       script: |-
-        tito_tmp_dir="tito"
-        rm -rf "${tito_tmp_dir}"
-        mkdir -p "${tito_tmp_dir}"
-        tito tag --offline --accept-auto-changelog
-        tito build --output="${tito_tmp_dir}" --rpm --test --offline --quiet
-        createrepo "${tito_tmp_dir}/noarch"
-        cat << EOR > ./openshift-ansible-local-release.repo
-        [openshift-ansible-local-release]
-        baseurl = file://$( pwd )/${tito_tmp_dir}/noarch
-        gpgcheck = 0
-        name = OpenShift Ansible Release from Local Source
-        EOR
-        sudo cp ./openshift-ansible-local-release.repo /etc/yum.repos.d
-    - type: "script"
-      title: "install the openshift-ansible release"
-      repository: "openshift-ansible"
-      script: |-
-        last_tag="$( git describe --tags --abbrev=0 --exact-match HEAD )"
-        last_commit="$( git log -n 1 --pretty=%h )"
-        sudo yum install -y "atomic-openshift-utils${last_tag/openshift-ansible/}.git.0.${last_commit}.el7"
-        rpm -V "atomic-openshift-utils${last_tag/openshift-ansible/}.git.0.${last_commit}.el7"
+        sudo yum install -y ansible python2 python-six tar java-1.8.0-openjdk-headless httpd-tools \
+           libselinux-python python-passlib python-pip
     - type: "script"
       title: "install Ansible plugins"
-      repository: "origin"
+      repository: "origin-aggregated-logging"
       script: |-
         sudo chmod o+rw /etc/environment
         echo "ANSIBLE_JUNIT_DIR=$( pwd )/_output/scripts/ansible_junit" >> /etc/environment
@@ -56,20 +37,17 @@ extensions:
         done
         sudo sed -r -i -e 's/^#?stdout_callback.*/stdout_callback = default_with_output_lists/' -e 's/^#?callback_whitelist.*/callback_whitelist = generate_junit/' /etc/ansible/ansible.cfg
     - type: "script"
-      title: "determine the release commit for origin images and version for rpms"
-      repository: "origin"
+      title: "ensure the correct package repositories are used for the correct release"
+      repository: "origin-aggregated-logging"
       script: |-
         # is logging using master or a release branch?
-        pushd "/data/src/github.com/openshift/origin-aggregated-logging/"
         curbranch=$( git rev-parse --abbrev-ref HEAD )
-        popd
-        jobs_repo="/data/src/github.com/openshift/aos-cd-jobs/"
         if [[ "${curbranch}" == master ]] ; then
-           git log -1 --pretty=%h >> "${jobs_repo}/ORIGIN_COMMIT"
-           ( source hack/lib/init.sh; os::build::rpm::get_nvra_vars; echo "-${OS_RPM_VERSION}-${OS_RPM_RELEASE}" ) >> "${jobs_repo}/ORIGIN_PKG_VERSION"
-           ( source hack/lib/init.sh; os::build::rpm::get_nvra_vars; echo "${OS_GIT_MAJOR}.${OS_GIT_MINOR}" | sed "s/+//" ) >> "${jobs_repo}/ORIGIN_RELEASE"
+           # enable centos-openshift-origin and centos-openshift-origin-testing
+           # assume everything else has been disabled
+           sudo yum-config-manager --enable centos-openshift-origin > /dev/null
+           sudo yum-config-manager --enable centos-openshift-origin-testing > /dev/null
         elif [[ "${curbranch}" =~ ^release-* ]] ; then
-           pushd "/data/src/github.com/openshift/origin-aggregated-logging/"
            # get repo ver from branch name
            repover=$( echo "${curbranch}" | sed -e 's/release-//' -e 's/[.]//' )
            # get version from tag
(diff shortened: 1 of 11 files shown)


jcantrill:
it LGTM

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -19,33 +19,14 @@ extensions:
         fi
         hack/build-images.sh
     - type: "script"
-      title: "build an openshift-ansible release"
```
stevekuznetsov: So in general whenever possible we need to be building RPMs and using built releases for CI testing. The issues you saw with the `tito tag` command were entirely user error on the part of whoever branched Origin and OpenShift Ansible for the new release. We have doc for whoever is doing that branching that, if followed, will cause no errors on that end. Let's not remove this step.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -19,33 +19,14 @@ extensions:
...
-      title: "install the openshift-ansible release"
-      repository: "openshift-ansible"
+      title: "install Ansible and other packages needed for OpenShift Ansible tasks"
+      repository: "origin-aggregated-logging"
       script: |-
-        last_tag="$( git describe --tags --abbrev=0 --exact-match HEAD )"
-        last_commit="$( git log -n 1 --pretty=%h )"
-        sudo yum install -y "atomic-openshift-utils${last_tag/openshift-ansible/}.git.0.${last_commit}.el7"
-        rpm -V "atomic-openshift-utils${last_tag/openshift-ansible/}.git.0.${last_commit}.el7"
+        sudo yum install -y ansible python2 python-six tar java-1.8.0-openjdk-headless httpd-tools \
```
stevekuznetsov: These sorts of tasks should be pushed into the base AMI: https://github.com/openshift/origin-ci-tool/blob/master/oct/ansible/oct/roles/dependencies/tasks/install_dependencies.yml#L5

`yum` tasks are notoriously slow and flaky.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -85,107 +63,141 @@ extensions:
...
+          mkdir $HOME/.kube
+        fi
+        sudo cp /var/lib/origin/openshift.local.config/master/admin.kubeconfig $HOME/.kube/config
+        sudo chown $USER $HOME/.kube/config
+        sudo mkdir -p /etc/origin/master
+        sudo cp /var/lib/origin/openshift.local.config/master/admin.kubeconfig /etc/origin/master/admin.kubeconfig
         sudo chmod a+x /etc/ /etc/origin/ /etc/origin/master/
         sudo chmod a+rw /etc/origin/master/admin.kubeconfig
-    - type: "script"
-      title: "ensure built version of origin is installed"
```
stevekuznetsov: If you're installing `origin-clients` you still 100% need this step to ensure you're installing the correct version. It's runtime is <1s but the peace of mind is huge.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -85,107 +63,141 @@ extensions:
...
-        fi
     - type: "script"
       title: "install origin-aggregated-logging"
-      repository: "aos-cd-jobs"
+      repository: "openshift-ansible"
       script: |-
-        playbook_base='/usr/share/ansible/openshift-ansible/playbooks/'
+        OS_A_C_J_DIR=/data/src/github.com/openshift/aos-cd-jobs
+        pushd $OS_A_C_J_DIR > /dev/null
+        # cannot use the networking related parameters with cluster up deployment
```
stevekuznetsov: Anything that cannot be used for all deployments should not be in this file. Move those options out of the file and into `-e` statements for the jobs that need them.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -85,107 +63,141 @@ extensions:
...
-                         --become-user root         \
-                         --connection local         \
-                         --inventory sjb/inventory/ \
-                         -e containerized=true      \
-                         -e deployment_type=origin  \
-                         /usr/share/ansible/openshift-ansible/playbooks/prerequisites.yml
-    - type: "script"
-      title: "install origin"
-      repository: "aos-cd-jobs"
+      title: "configure system to run oc cluster up and run it"
```
stevekuznetsov: There's a lot of setup steps here that I worry will be fragile.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: But is it absolutely necessary to do that here?  Because isn't building and installing ansible being done elsewhere, in other tasks?

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: These packages were being installed as a "side effect" of doing the ansible package installation.  So this job will be doing `yum install` here anyway, no matter what.  I will note that almost all of them are already installed - I think just the java, httpd, and selinux packages are not already installed.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: How will adding more steps make it less fragile?

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
stevekuznetsov: I would say yes, for the simple reason that the reason you took this out -- `tito` issues -- was an operator error and should not, in a perfect world, happen again. You incur no overhead with the RPM approach if we do our due diligence.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
stevekuznetsov: Sure, in any case -- whether `yum` is resolving them or they're explicitly listed -- add them to the base AMI so we can no-op on them as much as possible, please

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: not sure what you mean

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: Then it should be removed from the logging.yaml inventory

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
stevekuznetsov: Don't remove this step

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: That wasn't the only reason I took it out.  The primary reason is that it isn't necessary, and the secondary reason is that it has been fragile and caused problems in the past, the latest tito issue being just one of them.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: So the base ami builder tracks the Requires and BuildRequires in the openshift-ansible.spec?

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
stevekuznetsov: 100% -- the intent of those files was that they contain super basic options _everyone_ was going to use. If that's not the case, then we should make sure it is.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
stevekuznetsov: Not steps as in `steps:[]` but just like -- to run `oc cluster up` correctly here we had to change a bunch of firewall rules, do all sorts of stuff, all this is logic that will need to be maintained, is all.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
stevekuznetsov: No, it's just a static list, but we should aim to get all of our runtime RPM dependencies in there so we can lean up the jobs

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
stevekuznetsov: I guess I would challenge that -- why is it not necessary? How many of our customers are running playbooks from source? If we're not testing the build --> install --> use flow for these using RPMs we're not testing end user experience. Before we started doing this in our CI we had a slew of issues coming back from QE about this sort of thing.

How else has it been fragile? Of all things packaging the Ansible playbooks into an RPM has been the least fragile of all of our builds steps as far as I'm aware. If there are other specific concerns we should address those.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: hmm - these parameters are in the general.yaml, which is used by _all_ jobs. If I remove those parameters from general.yaml, I'll have to update _every_ job.

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: I just want a way to run logging tests as simply as possible.

If we want to simulate the customer experience as much as possible, then we need a job that will install all of openshift + logging in one run of ansible-playbook.

Perhaps we need two sets of jobs - one for "lower" level logging integration testing, and one for "higher" level full installation path testing, which will install openshift + logging (and + metrics, and + whatever other options).

logging developers need a way to run the logging CI tests, locally or in an AWS devenv, in as close a way as possible to the way jenkins runs them.  Using `oct sync local` to sync local ansible and logging repos has caused a lot of grief related to building/installing ansible and tagging.  This is simply not necessary for 95% of developer and "lower" level logging CI workflow.

PR #916, opened 2017-12-05
logging - workaround ansible eventrouter recursion bug

richm:
Enabling eventrouter causes ansible to crash with a
recursion bug - disable to unblock CI until we can fix it


richm:
@jcantrill @ewolinetz @nhosoi @wozniakjan

PR #920, opened 2017-12-06
have to explicitly set log-driver=journald now

richm:
Looks like the default install for origin is to omit the docker
log-driver setting, which by default means docker will use json-file
That means we need to explicitly tell origin to configure docker
to use the journald log driver (which we probably should have done
anyway . . .)


richm:
@jcantrill so the moral of the story is that we have not been testing with journald for container logs since 3.8/3.9 probably

stevekuznetsov:
@richm pushed out

PR #921, opened 2017-12-06
increase fluentd memory to avoid oom kill

richm:
seeing fluentd being oom killed during the mux tests - interesting
that it is fluentd being killed and not mux
Raise the memory request/limit (they are controlled by the same
ansible parameter) for fluentd to avoid this


richm:
@jcantrill @portante Not sure if this is related to k8s 1.8 upgrade or origin 3.8/3.9, or possibly moving to jemalloc for fluentd, which maybe increases the memory usage in the short term, but prevents long term unbounded memory growth?  At any rate, I've seen several oom kills of fluentd during the tests, primarily the mux tests

richm:
btw - how do I tell if a pod has been oom killed?  Since we disabled the supervisor on fluentd, I don't see oom-killer messages in /var/log/messages

stevekuznetsov:
You should be able to see the reason for termination in the pod status. For instance:

```
oc get pod $podname -o jsonpath={.status.containerStatuses[0].state.terminated.reason}
```

richm:
not jemalloc - it is off by default - trying a test with jemalloc on by default

PR #924, opened 2017-12-07
conformance_crio: Force prereq and install to use same options

ashcrow:
It was noted that the prereq and install playbook calls require the same
set of options. This change copies the options from the install up to
the prereq call.

/cc @giuseppe @runcom @sdodson @michaelgugino


runcom:
Is this fixing any issue we're having at the moment with the CI?

sdodson:
Yes, hopefully.

sdodson:
We need to revisit and start defining some vars files for various permutations of jobs and refactor them to use those vars files, but lets make the pain go away now.

PR #925, opened 2017-12-07
moving tito tmp dir

adammhaile:
@jupierce


jupierce:
@adammhaile Thanks. Does tito_tmp get created automatically if it does not exist? Has the change on buildvm been tested to ensure it has the intended effect?

jupierce:
@adammhaile Tested myself. Seems fine.

adammhaile:
Just saw this. Yeah, did as expected when I tried. tito_tmp was manually
created by me. When I ran without it, tito created it anyways. But wanted
directory to be there and there's nothing that should delete it.

PR #929, opened 2017-12-08
Fix missing variable local_ip for crio prerequisite install

michaelgugino:
This commit adds a missing variable to the call to ansible.

Without this, the call fails with 'unbound variable: local_ip'


michaelgugino:
@ashcrow @sdodson @stevekuznetsov

stevekuznetsov:
@michaelgugino pushed up

PR #930, opened 2017-12-08
Run go bits inside a container

kargakis:
/cc @deads2k @stevekuznetsov 

Depends on https://github.com/openshift/origin/pull/17681

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_end_to_end.yml
+++ b/sjb/config/test_cases/test_branch_origin_end_to_end.yml
@@ -24,4 +24,5 @@ extensions:
       script: |-
         OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=_output/local/bin/linux/amd64/end-to-end.test hack/env make build-router-e2e-test
         OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=_output/local/bin/linux/amd64/etcdhelper hack/env make build WHAT=tools/etcdhelper
-        OPENSHIFT_SKIP_BUILD='true' JUNIT_REPORT='true' make test-end-to-end -o build
+        OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=???? hack/env JUNIT_REPORT='true' make test-end-to-end-unit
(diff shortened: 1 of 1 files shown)


sjb/config/test_cases/test_branch_origin_end_to_end.yml
```diff
@@ -24,4 +24,5 @@ extensions:
       script: |-
         OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=_output/local/bin/linux/amd64/end-to-end.test hack/env make build-router-e2e-test
         OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=_output/local/bin/linux/amd64/etcdhelper hack/env make build WHAT=tools/etcdhelper
-        OPENSHIFT_SKIP_BUILD='true' JUNIT_REPORT='true' make test-end-to-end -o build
\ No newline at end of file
+        OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=???? hack/env JUNIT_REPORT='true' make test-end-to-end-unit
```
kargakis: Not sure what artifacts do you need back from the container

stevekuznetsov:
That test requires direct access to the daemon... we've been asking for it to be refactored for what feels like more than a year now. 

/cc @bparees @miminar @dmage

bparees:
> That test requires direct access to the daemon... we've been asking for it to be refactored for what feels like more than a year now.

what test does?

bparees:
(and why the heck do we have a target named `test-end-to-end-unit`? I can't think of anything much more contradictory).

stevekuznetsov:
Hmm I may have tagged the wrong set of people -- I was certain registry tests fell into the bucket (but maybe they just moved out of Origin? I definitely remember them being in here). @knobunc it's currently router tests that expect direct Docker daemon interaction https://github.com/openshift/origin/blob/master/test/end-to-end/router_test.go

bparees:
Yes presumably they moved when the registry was removed. (of course the
same problem may exist for them in their new location)

Ben Parees | OpenShift

stevekuznetsov:
I believe it remains. @mfojtik can we get prioritization for that?

PR #946, opened 2017-12-14
CRI-O: move junit results to gcs artifacts root

runcom:
Needed by the k8s testgrid as per
https://github.com/kubernetes/test-infra/pull/5943#issuecomment-351584590

@stevekuznetsov PTAL


runcom:
I've also noticed we've not been publishing results to https://openshift-gce-devel.appspot.com/ since 11/16 for the crio jobs :confused:

stevekuznetsov:
I would rather see TestGrid move to be like Gubernator and allow for the jUnit everywhere. We are using the federated bucket now, as we discussed when we set this up -- `gs://origin-federated-results/`

PR #960, opened 2017-12-18
Add kubernetes-metrics-server builds

smarterclayton:
Like image-registry, the new metrics server image is separate from the repo.

This is almost a direct copy of image-registry, except the exclusion rules only build it for master and release-3.9. Running this may break older release branches for push, possibly.

I think the next image we add should be forced to follow the centralized model - see openshift/release#501 for an exploration of how that might work.

@stevekuznetsov @DirectXMan12


smarterclayton:
Will try tomorrow to merge this and test

PR #968, opened 2017-12-28
use origin local if release specific repo not found

richm:
If there is no release specific repo e.g. currently the repo
centos-paas-sig-openshift-origin38-rpms does not exist, then
just use the latest origin built on the devenv machine.


stevekuznetsov:
We're so close on my end to not need any of this nonsense and write jobs explicitly to use specific versions ... but until then -- let's do this

richm:
@stevekuznetsov ok, good to know - I've already pushed these, and they are working

stevekuznetsov:
Nice

PR #972, opened 2018-01-02
Add in-tree CI script

michaelgugino:
This commit adds functionality for calling a CI
script from inside the openshift-ansible source.

This will allow for better control of testing between
release branches, as well as simplify changing of variables
between the various tests.


michaelgugino:
This is currently WIP.

michaelgugino:
I think we're in a good state for now.  I would like to get this merged @stevekuznetsov @sdodson

Issue #970, opened 2018-01-02
build/ocp job should accept a list of images to exclude


jupierce:
The ocp build job presently builds all images and fails if any images don't build. Occasionally, we just need to get through a build and would like to exclude a failing image from the build process. oit will soon support this with the --exclude flag (https://github.com/openshift/enterprise-images/issues/112). 

The task for this issue is to expose a new Jenkins pipeline _job parameter_ which allows the builder to exclude images from the jenkins job.

e.g. 
a parameter with `cri-o-docker logging-fluentd-docker`
would translate into additional command line arguments for the oit invocations: -x cri-o-docker -x logging-fluentd-docker --ignore-missing-base`  would be included as arguments in the oit invocations. 

When an image is excluded, the "build complete" email should clearly state that it is a partial build in the subject line and list what images were not included. 

e.g.
Subject: New PARTIAL build for OpenShift (Release Candidate): 3.8.18-1  [excluded images: cri-o-docker logging-fluentd-docker]

Finally, the UMB (sendCIMessage) message content should be changed to indicate it is a partial build. Otherwise, SVT testing would be triggered against an incomplete build.

adammhaile:
All of this should be in #973

PR #973, opened 2018-01-03
Updates to address #970

adammhaile:
exlusion job parameter
option to retry/contintue/abort on build failure
@jupierce

### DIFF
--- a/jobs/build/ocp/Jenkinsfile
+++ b/jobs/build/ocp/Jenkinsfile
@@ -25,6 +25,7 @@ online:stg                {origin,origin-web-console,openshift-ansible}/stage ->
                           [$class: 'hudson.model.BooleanParameterDefinition', defaultValue: false, description: 'Mock run to pickup new Jenkins parameters?', name: 'MOCK'],
                           [$class: 'hudson.model.BooleanParameterDefinition', defaultValue: false, description: 'Run as much code as possible without pushing / building?', name: 'TEST'],
                           [$class: 'hudson.model.TextParameterDefinition', defaultValue: "", description: 'Include special notes in the build email?', name: 'SPECIAL_NOTES'],
+                          [$class: 'hudson.model.StringParameterDefinition', defaultValue: "", description: 'Exclude these images from builds. Comma or space separated list. (i.e cri-o-docker,aos3-installation-docker)', name: 'BUILD_EXCLUSIONS'],
                   ]
             ],
             disableConcurrentBuilds()
@@ -35,6 +36,12 @@ IS_TEST_MODE = TEST.toBoolean()
 BUILD_VERSION_MAJOR = BUILD_VERSION.tokenize('.')[0].toInteger() // Store the "X" in X.Y
 BUILD_VERSION_MINOR = BUILD_VERSION.tokenize('.')[1].toInteger() // Store the "Y" in X.Y
 SIGN_RPMS = SIGN.toBoolean()
+
+if ( BUILD_EXCLUSIONS != "" ) {
+  // clean up string delimiting
+  BUILD_EXCLUSIONS = BUILD_EXCLUSIONS.replaceAll(',', ' ')
+  BUILD_EXCLUSIONS = BUILD_EXCLUSIONS.split().join(',')
+}
 
 def mail_success( version ) {
 
@@ -56,11 +63,17 @@ def mail_success( version ) {
         inject_notes = "\n***Special notes associated with this build****\n${SPECIAL_NOTES.trim()}\n***********************************************\n"
     }
 
+    PARTIAL = " "
+    exclude_subject = ""
+    if ( BUILD_EXCLUSIONS != "" ) {      
+      PARTIAL = " PARTIAL "
+      exclude_subject = " [excluded images: ${BUILD_EXCLUSIONS}]"
+    }
     mail(
             to: "${MAIL_LIST_SUCCESS}",
             from: "aos-cd@redhat.com",
             replyTo: 'smunilla@redhat.com',
-            subject: "[aos-cicd] New build for OpenShift ${target}: ${version}",
+            subject: "[aos-cicd] New${PARTIAL}build for OpenShift ${target}: ${version}${exclude_subject}",
             body: """\
 OpenShift Version: v${version}
 ${inject_notes}
@@ -92,19 +105,21 @@ ${OA_CHANGELOG}
 """);
 
     try {
-            timeout(3) {
-                sendCIMessage( messageContent: "New build for OpenShift ${target}: ${version}",
-                        messageProperties: """build_mode=${BUILD_MODE}
-                        puddle_url=${mirrorURL}/${OCP_PUDDLE}
-                        image_registry_root=registry.reg-aws.openshift.com:443
-                        brew_task_url_openshift=${OSE_BREW_URL}
-                        brew_task_url_openshift_ansible=${OA_BREW_URL}
-                        product=OpenShift Container Platform
-                        """,
-                        messageType: 'ProductBuildDone',
-                        overrides: [topic: 'VirtualTopic.qe.ci.jenkins'],
-                        providerName: 'Red Hat UMB'
-                )
+            if ( BUILD_EXCLUSIONS == "" ) {
+              timeout(3) {
+                  sendCIMessage( messageContent: "New build for OpenShift ${target}: ${version}",
+                          messageProperties: """build_mode=${BUILD_MODE}
+                          puddle_url=${mirrorURL}/${OCP_PUDDLE}
+                          image_registry_root=registry.reg-aws.openshift.com:443
+                          brew_task_url_openshift=${OSE_BREW_URL}
+                          brew_task_url_openshift_ansible=${OA_BREW_URL}
+                          product=OpenShift Container Platform
+                          """,
+                          messageType: 'ProductBuildDone',
+                          overrides: [topic: 'VirtualTopic.qe.ci.jenkins'],
+                          providerName: 'Red Hat UMB'
+                  )
+              }
             }
     } catch ( mex ) {
             echo "Error while sending CI message: ${mex}"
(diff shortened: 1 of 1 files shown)


jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +655,68 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
                 sh "ose_images.sh --user ocp-build build_container --branch rhaos-${BUILD_VERSION}-rhel-7 --group base --repo https://raw.githubusercontent.com/openshift/aos-cd-jobs/master/build-scripts/repo-conf/aos-unsigned-building.repo"
             }
 
-            buildlib.oit """
+            waitUntil {
+              try {
+                exclude = ""
+                if ( BUILD_EXCLUSIONS != "" ) {
+                  exclude = "-x ${BUILD_EXCLUSIONS} --ignore-missing-base"
```
jupierce: This won't work for space delimited. The second exclude will be treated like an argument unless you wrap it in quotes. I would preprocess BUILD_EXCLUSIONS somewhere near the top of the pipeline to clean it up:
- replace commas with whitespace
- split on whitespace
- join back with comma

This will turn things like "cri-o-docker,    xyz-docker" into parameter safe "cri-o-docker,xyz-docker".

jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +655,68 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
...
 --working-dir ${OIT_WORKING} --group openshift-${BUILD_VERSION}
+${exclude}
 images:build
 --push-to-defaults --repo-type unsigned
 """
+                return true // finish waitUntil
+              }
+              catch ( err ) {
+                try {
+                  record_log = buildlib.parse_record_log( OIT_WORKING )
```
jupierce: Each invocation of oit _appends_ to the record log. The list you are building, therefore, will include images from the last retry even if the image was subsequently successful.

jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +655,68 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
...
+Jenkins job: ${env.BUILD_URL}
+
+FAILED BUILDS:
+${failed_builds}
+""");
+                } catch ( mail_exception ) {
+                    echo "One or more errors sending email: ${mail_exception}"
+                }
+                
+                def resp = input    message: "Error during Image Build for OCP v${BUILD_VERSION} with args: ${opts}",
```
jupierce: opts?

jobs/build/ocp/Jenkinsfile
```diff
@@ -25,6 +25,7 @@ online:stg                {origin,origin-web-console,openshift-ansible}/stage ->
                           [$class: 'hudson.model.BooleanParameterDefinition', defaultValue: false, description: 'Mock run to pickup new Jenkins parameters?', name: 'MOCK'],
                           [$class: 'hudson.model.BooleanParameterDefinition', defaultValue: false, description: 'Run as much code as possible without pushing / building?', name: 'TEST'],
                           [$class: 'hudson.model.TextParameterDefinition', defaultValue: "", description: 'Include special notes in the build email?', name: 'SPECIAL_NOTES'],
+                          [$class: 'hudson.model.TextParameterDefinition', defaultValue: "", description: 'Exclude these images from builds. Comma or space separated list', name: 'BUILD_EXCLUSIONS'],
```
jupierce: If you don't permit newlines, use StringParameterDefinition to ensure people don't get tempted.

jupierce: Include an example string: "cri-o-docker, ...." so we remember what this should look like when we need it.

jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +655,68 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
...
+                        from: "aos-cd@redhat.com",
+                        subject: "RESUMABLE Error during Image Build for OCP v${BUILD_VERSION}",
+                        body: """Encountered an error: ${err}
+Input URL: ${env.BUILD_URL}input
+Jenkins job: ${env.BUILD_URL}
+
+FAILED BUILDS:
+${failed_builds}
+""");
+                } catch ( mail_exception ) {
```
jupierce: If the email can't be sent, let the exception stop the job. No one is going to be checking for input if they don't get an email and I don't want automated builds to start backing up because of it. This is different from the cluster upgrade pipeline you were modeling - there, people were normally paying a great deal of attention to each upgrade.

jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +655,68 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
...
+                                                  [$class: 'hudson.model.ChoiceParameterDefinition',
+                                                   choices: "RETRY\nCONTINUE\nABORT",
+                                                   description: 'Retry (try the operation again). Continue (fails are OK, continue pipeline). Abort (terminate the pipeline).',
+                                                   name: 'action']
+                                  ]
+
+                if ( resp == "RETRY" ) {
+                    return false  // cause waitUntil to loop again
+                } else if ( resp == "CONTINUE" ) {
+                    echo "User chose to build fails are OK."
```
jupierce: This codepath somehow needs to affect the PARTIAL build email log. We know images failed and so should consumers of that build.

adammhaile:
@jupierce all have been addressed.

jobs/build/ocp/Jenkinsfile
```diff
@@ -92,14 +105,20 @@ ${OA_CHANGELOG}
...
             timeout(3) {
-                sendCIMessage( messageContent: "New build for OpenShift ${target}: ${version}",
+                sendCIMessage( messageContent: "New${PARTIAL}build for OpenShift ${target}: ${version}",
                         messageProperties: """build_mode=${BUILD_MODE}
                         puddle_url=${mirrorURL}/${OCP_PUDDLE}
                         image_registry_root=registry.reg-aws.openshift.com:443
                         brew_task_url_openshift=${OSE_BREW_URL}
                         brew_task_url_openshift_ansible=${OA_BREW_URL}
-                        product=OpenShift Container Platform
+                        product=OpenShift Container Platform${product_suffix}
```
jupierce: @cjryan please note that we will be sending a "partial" build notice out on the UMB. Partial builds indicate that some images might not have been built.

jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +661,81 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
...
+                  }
+                  else {
+                    // build may have succeeded later. If so, remove.
+                    if ( failed_map.containsKey(distgit) ) {
+                      failed_map.remove(distgit)
+                    }
+                  }
+                }
+                
+                failed_list = buildlib.mapToList(failed_map)
```
jupierce: Starting on this line, I'd like to remove as much logic as we can from the pipeline. As we know, loc == pain  when it comes to Jenkins pipelines. To that end, can't we:
- just initialize exclusion_list with failed_map.keys()
- use "${failed_map}"  in the email. Groovy does a sufficient job of printing the map. It will be JSON-y, but this is an internal email and I'm not concerned with it being pretty.

jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +661,81 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
...
+                  if ( builds[i]['status'] == '-1' ){
+                    task_url = "No Brew task available, see jenkins log."
+                    if ( build[i].containsKey('task_url') ) {
+                      task_url = build[i]['task_url']
+                    }
+                    failed_map[distgit] = task_url
+                  }
+                  else {
+                    // build may have succeeded later. If so, remove.
+                    if ( failed_map.containsKey(distgit) ) {
```
jupierce: No need to check. Groovy doesn't throw an exception if it is not present.

jobs/build/ocp/Jenkinsfile
```diff
@@ -642,11 +661,81 @@ Please direct any questsions to the Continuous Delivery team (#aos-cd-team on IR
...
+              }
+              catch ( err ) {
+                record_log = buildlib.parse_record_log( OIT_WORKING )
+                builds = record_log['build']
+                failed_map = [:]
+                for(i = 0; i < builds.size(); i++) {
+                  distgit = build[i]['distgit']
+                  if ( builds[i]['status'] == '-1' ){
+                    task_url = "No Brew task available, see jenkins log."
+                    if ( build[i].containsKey('task_url') ) {
```
jupierce: Why would this ever be missing? In other oit records, I just set the field to "?" so that it will always be there and the pipeline code can stay tight.

jobs/build/ocp/Jenkinsfile
adammhaile: Good to know.

jobs/build/ocp/Jenkinsfile
adammhaile: It won't exist if the brew task is never created: https://github.com/openshift/enterprise-images/blob/2cf0d60b5e66620de8abc8145bc8f483baee8206/oit/packages/distgit.py#L539
Down on L556 is the only place task_url is added to the record. And if the task is never created, it's never added, so it wouldn't exist.

jobs/build/ocp/Jenkinsfile
adammhaile: Sure, that gets rid of 10 lines.

adammhaile:
@jupierce All have been addressed.

jobs/build/ocp/Jenkinsfile
cjryan: @jupierce understood, thanks for the heads-up!

akostadinov: @jupierce , you mean that we will send only a partial notice or a partial notice in addition to the final notice? We need a way to know when everything is ready.

akostadinov: Ok, I understand now what you mean. We need to have this indicated in headers to allow filtering of messages. I think changing message type. Perhaps use message type 'partial_build_done'. Hope such custom message will be allowed by the plugin. I see a `Custom` message in the list. So maybe we can just use `custom` as value? Presently there is an issue with our jenkins and I can't check how it works.

jupierce: @akostadinov Can't you filter on "product"? If not, I will just disable the message on partial builds for now.

jobs/build/ocp/Jenkinsfile
jupierce: I've added: https://github.com/openshift/enterprise-images/commit/78f4ea3c775de2fb3141863731dee56a692ebc38

PR #974, opened 2018-01-03
Ansible Service Broker Test

eriknelson:
Initial pass at a config for testing ASB nightly against upstream catalog. Feedback much appreciated.

### DIFF
--- a/sjb/config/test_cases/test_branch_ansible_service_broker.yml
+++ b/sjb/config/test_cases/test_branch_ansible_service_broker.yml
@@ -5,10 +5,9 @@ overrides:
   timer: 'H H * * *'
   email:
     - ernelson@redhat.com
-  sync:
-    - "openshift,ansible-service-broker=master"
-    - "rthallisey,service-broker-ci=master"
 extensions:
+  sync_repos:
+    - name: "ansible-service-broker"
   actions:
     - type: "script"
       title: "install asb to cluster"
@@ -30,7 +29,7 @@ extensions:
       # repository: "origin" # TODO: Doesn't appear to be a way to cd into non-openshift org?
       timeout: 1800
       script: |-
-        cd $GOPATH/src/github.com/rthallisey/service-broker-ci && make install
+        go get github.com/rthallisey/service-broker-ci
     - type: "script"
       title: "run ASB ci"
       repository: "ansible-service-broker"
--- a/sjb/config/test_cases/test_branch_ansible_service_broker.yml
+++ b/sjb/config/test_cases/test_branch_ansible_service_broker.yml
@@ -1,0 +1,48 @@
+---
+parent: 'test_cases/test_branch_origin_service_catalog.yml'
+overrides:
+  junit_analysis: False
+  timer: 'H H * * *'
+  email:
+    - ernelson@redhat.com
+  sync:
+    - "openshift,ansible-service-broker=master"
+    - "rthallisey,service-broker-ci=master"
+extensions:
+  actions:
+    - type: "script"
+      title: "install asb to cluster"
+      repository: "ansible-service-broker"
+      timeout: 1800
+      script: |-
+        export PATH=$(pwd)/_output/local/bin/linux/amd64:$PATH
+        oc create project ansible-service-broker
+        oc process ./templates/deploy-ansible-service-broker.template.yaml -p BROKER_CA_CERT=$(oc get secret -n kube-service-catalog -o go-template=\'{{ range .items }}{{ if eq .type \"kubernetes.io/service-account-token\" }}{{ index .data \"service-ca.crt\" }}{{end}}{{\"\\n\"}}{{end}}\' | tail -n 1) | oc create -n ansible-service-broker -f -'
+    - type: "script"
+      title: "relist Catalog to pickup new clusterserviceclasses"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        # relistRequets is intially zero, and to trigger relist, need to increment. Patch it to force 1.
+        ./_output/local/bin/linux/amd64/oc patch clusterservicebroker ansible-service-broker -p '{"spec":{"relistRequests":1}}'
+    - type: "script"
+      title: "install ASB ci"
+      # repository: "origin" # TODO: Doesn't appear to be a way to cd into non-openshift org?
+      timeout: 1800
+      script: |-
+        cd $GOPATH/src/github.com/rthallisey/service-broker-ci && make install
+    - type: "script"
+      title: "run ASB ci"
+      repository: "ansible-service-broker"
+      timeout: 1800
+      script: |-
+        $GOPATH/bin/ci
+  system_journals:
+    - origin-master.service
+    - origin-master-api.service
+    - origin-master-controllers.service
+    - origin-node.service
+    - openvswitch.service
+    - ovs-vswitchd.service
+    - ovsdb-server.service
+    - etcd.service
--- a/sjb/generated/test_branch_ansible_service_broker.xml
+++ b/sjb/generated/test_branch_ansible_service_broker.xml
@@ -1,0 +1,805 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<com.tikal.jenkins.plugins.multijob.MultiJobProject plugin="jenkins-multijob-plugin@1.21">
+  <actions/>
+  <description>&lt;div style=&#34;font-size: 32px; line-height: 1.5em; background-color: yellow; padding: 5px;&#34;&gt;WARNING: THIS IS AN AUTO-GENERATED JOB DEFINITION. CHANGES MADE USING THE ONLINE EDITOR WILL NOT BE HONORED. MAKE CHANGES TO THE JOB DEFINITIONS IN THE &lt;a href=&#34;https://github.com/openshift/aos-cd-jobs/tree/master/sjb&#34;&gt;openshift/aos-cd-jobs&lt;/a&gt; REPOSITORY INSTEAD.&lt;/div&gt;</description>
+  <keepDependencies>false</keepDependencies>
+  <properties>
+    <com.sonyericsson.rebuild.RebuildSettings plugin="rebuild@1.25">
+      <autoRebuild>false</autoRebuild>
+      <rebuildDisabled>false</rebuildDisabled>
+    </com.sonyericsson.rebuild.RebuildSettings>
+    <jenkins.model.BuildDiscarderProperty>
+      <strategy class="hudson.tasks.LogRotator">
+        <daysToKeep>7</daysToKeep>
+        <numToKeep>-1</numToKeep>
+        <artifactDaysToKeep>-1</artifactDaysToKeep>
+        <artifactNumToKeep>-1</artifactNumToKeep>
+      </strategy>
+    </jenkins.model.BuildDiscarderProperty>
+    <hudson.model.ParametersDefinitionProperty>
+      <parameterDefinitions>
+        <hudson.model.StringParameterDefinition>
+          <name>JOB_SPEC</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>buildId</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>BUILD_ID</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>REPO_OWNER</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>REPO_NAME</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>PULL_BASE_REF</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>PULL_BASE_SHA</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>PULL_REFS</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
+        <hudson.model.StringParameterDefinition>
+          <name>PULL_NUMBER</name>
+          <description></description>
+          <defaultValue></defaultValue>
+        </hudson.model.StringParameterDefinition>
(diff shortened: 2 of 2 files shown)


sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
+---
+parent: 'common/test_cases/origin_built_release.yml'
+overrides:
+  junit_analysis: False
+  timer: 'H H * * *'
+  email:
+    - jpeeler@redhat.com
```
jpeeler: change to your email so you'll know when there are problems

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+        git log -1 --pretty=%h > "${jobs_repo}/ORIGIN_COMMIT"
+        source hack/lib/init.sh
+        os::build::rpm::format_nvra > "${jobs_repo}/ORIGIN_BUILT_VERSION"
+    - type: "script"
+      title: "build junitreport"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        OS_BUILD_ENV_PULL_IMAGE=true OS_BUILD_ENV_PRESERVE=_output/local/bin/linux/amd64/junitreport hack/env make build WHAT=tools/junitreport
+    - type: "script"
```
jpeeler: you can delete this section since you won't need the user broker

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+      timeout: 1800
+      script: |-
+        REGISTRY=openshift/ NO_DOCKER=1 make -C cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog user-broker-image
+    - type: "script"
+      title: "build Service Catalog binary"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        unset GOPATH; cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog/hack/build-go.sh
+    - type: "script"
```
jpeeler: you can delete this section since you won't need to run the catalog unit tests again

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+      title: "run Service Catalog unit tests"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        NO_DOCKER=1 make -C cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog test-unit
+    - type: "script"
```
jpeeler: you can delete this section since you won't need to run the integration tests again

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+        # https://github.com/kubernetes-incubator/service-catalog/pull/1566 (and vendored in rebase)
+        sed -i 's/\(go test.*\)\( -i \)\(.*\)/\1 \3/' cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog/test/integration.sh
+        NO_DOCKER=1 make -C cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog scBuildImageTarget=.scBuildImage test-integration
+    - type: "script"
+      title: "build Service Catalog image"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        hack/build-local-images.py service-catalog
+    - type: "script"
```
jpeeler: you can delete this section, same reason

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+      timeout: 1800
+      script: |-
+        NO_DOCKER=1 make -C cmd/service-catalog/go/src/github.com/kubernetes-incubator/service-catalog  bin/e2e.test
+    - type: "script"
+      title: "start cluster with Service Catalog"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        ./_output/local/bin/linux/amd64/oc cluster up --version=latest --service-catalog
+    - type: "script"
```
jpeeler: you can delete this section, same reason

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+      title: "Run Service Catalog E2E"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        sudo chmod a+rw /var/lib/origin/openshift.local.config/master/admin.kubeconfig
+        export KUBECONFIG=/var/lib/origin/openshift.local.config/master/admin.kubeconfig
+        ./_output/local/bin/linux/amd64/oc login -u system:admin
```
jpeeler: Though you may need the above configuration for oc commands.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+################################################################################
+# ASB CI ADDITIONS
+################################################################################
+    - type: "script"
+      title: "Install ASB to cluster"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        ./_output/local/bin/linux/amd64/oc create project ansible-service-broker
+        curl "https://raw.githubusercontent.com/openshift/ansible-service-broker/master/templates/simple-broker-template.yaml" | ./_output/local/bin/linux/amd64/oc process -n "ansible-service-broker" -f - | ./_output/local/bin/linux/amd64/oc create -f -
```
jpeeler: (This is where my expertise ends.)
You may consider adding a retry to curl download attempts.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,137 @@
...
+      timeout: 1800
+      script: |-
+        # relistRequets is intially zero, and to trigger relist, need to increment. Patch it to force 1.
+        ./_output/local/bin/linux/amd64/oc patch clusterservicebroker ansible-service-broker -p '{"spec":{"relistRequests":1}}'
+    - type: "script"
+      title: "Configure ASB_CI_ROOT and go build environment"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        # TODO: Sanity check on configuring environment this way? Go tools? Is this the right location?
```
jpeeler: It might be worth checking what GOPATH is set to before being unset on line 35 and using that.

eriknelson:
@jpeeler Thanks! Updated based on feedback. Can you elaborate on the GOPATH item? Not sure what you mean.

jpeeler:
I was only pointing out that earlier in the file the original GOPATH is
unset, but I'll just retract what I said. I don't think anything is wrong
with manually specifying for such a distinct usage.

eriknelson:
Great, think this should be all set.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
jboyd01: FYI, I just had to do a change in service catalog tests around our use of kubeconfig.  See https://github.com/openshift/origin/issues/18722.  The work around is simple (use oc config view --raw > tmp and use the tmp file for kubeconfig).

jpeeler: Unrelated to the above, soon I plan on updating the way origin is built for catalog. And you'll probably want to copy such that origin changes won't be able to break catalog (and therefore cause your job to fail).

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,109 @@
...
+        export GOBIN=$ASB_CI_ROOT/bin
+        export PATH="$PATH:GOBIN"
+    - type: "script"
+      title: "Install ASB CI tool"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        go get github.com/rthallisey/service-broker-ci/cmd/ci
+    - type: "script"
+      title: "Install ASB source"
```
stevekuznetsov: If this is the code under test you don't need to clone it manually. If it's not, let's use `sync:`

eriknelson: It is the code under test; can you point me towards an example that shows how I should be cloning it?

jpeeler: Like https://github.com/openshift/aos-cd-jobs/blob/6e5814f8a2fbdab27b95c5bd02ddd0b9dcc691ea/sjb/config/test_cases/test_pull_request_origin_extended_conformance_azure_39.yml#L4-L10

stevekuznetsov: Err as in the above link -- each explicitly listed item under `sync:` is code that is not directly under test -- the test is written for Origin PRs. In this job, therefore, if you are checking out ASB source in a job on the ASB repo, you can just write `sync: []`. If ASB source is being checked out in an Origin job, explicitly list it

eriknelson: Sorry, I'm still not understanding what the ask is and I'm unfamiliar with any of the aos-ci stuff. What is the difference between adding to the `sync` list and "explicitly listing it"? The intention of this job is to nightly run the latest asb code with the new catalog source so we can hopefully be notified of breaking changes in the catalog sooner.

eriknelson:
@jpeeler @stevekuznetsov Pushed https://github.com/openshift/aos-cd-jobs/pull/974/commits/795356a24ddfe58a342a784134500bbc60a458a8, I think this is what you guys were looking for? Please let me know if that's good or if I should be doing something else.

Also, does sync just pull code into the `GOPATH`?

jpeeler:
Basically, but you want to put it in the extensions section. I just realized that this whole thing can be shorted by using service catalog as a parent, will start a review now.
sync pulls code into /data/src/..., which I assume go has been configured to use.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,93 @@
+---
+parent: 'common/test_cases/origin_built_release.yml'
```
jpeeler: You should consider making the parent test_cases/test_branch_origin_service_catalog.yml and remove all the origin/service-catalog building steps. The alternative is that you have to maintain the catalog build steps yourself, but it's not a clear decision since you'd be able to skip running catalog tests with the way things currently are.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,93 @@
...
+overrides:
+  sync:
+    - "openshift,ansible-service-broker=master"
```
jpeeler: This should be in the extensions section so the repos synced is appended, not replaced.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,93 @@
...
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        if [[ "${PULL_BASE_REF:-master}" == "release-3.9" ]]; then
+          ./_output/local/bin/linux/amd64/oc cluster up --loglevel=5 --version=latest --service-catalog
+        else
+          ./_output/local/bin/linux/amd64/oc cluster up --loglevel=5 --tag=latest --enable=service-catalog
+        fi
+        ./_output/local/bin/linux/amd64/oc login -u system:admin
+        ./_output/local/bin/linux/amd64/oc describe po --all-namespaces
```
jpeeler: Everything up to here in the extensions section can go away with catalog set as the parent.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,93 @@
...
+    - type: "script"
+      title: "Install ASB to cluster"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        ./_output/local/bin/linux/amd64/oc create project ansible-service-broker
+        ./_output/local/bin/linux/amd64/oc process 'https://raw.githubusercontent.com/openshift/ansible-service-broker/master/templates/deploy-ansible-service-broker.template.yaml' -p BROKER_CA_CERT=$(oc get secret -n kube-service-catalog -o go-template=\'{{ range .items }}{{ if eq .type \"kubernetes.io/service-account-token\" }}{{ index .data \"service-ca.crt\" }}{{end}}{{\"\\n\"}}{{end}}\' | tail -n 1) | oc create -n ansible-service-broker -f -'
```
jpeeler: I think instead of processing a template (especially this late in the process - it could fail due to github being down or whatever), you should change the repository line to be "ansible-service-broker" and run oc process on the file that's been synced.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,93 @@
...
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        go get github.com/rthallisey/service-broker-ci/cmd/ci
+    - type: "script"
+      title: "Run ASB ci"
+      repository: "origin"
+      timeout: 1800
+      script: |-
+        cd $GOPATH/src/openshift/ansible-service-broker && ci
```
jpeeler: I think if you change the repo line on this one and above (probably to "ansible-service-broker"), you don't need to worry about using cd. ./ci instead of just ci though.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,48 @@
+---
+parent: 'test_cases/test_branch_origin_service_catalog.yml'
+overrides:
+  sync:
+    - "openshift,ansible-service-broker=master"
```
jpeeler: I don't think you want any overrides of sync. The extensions part should be all you need.

sjb/config/test_cases/test_branch_ansible_service_broker.yml
```diff
@@ -0,0 +1,48 @@
...
+  sync_repos:
+    - "ansible-service-broker"
+  actions:
+    - type: "script"
+      title: "install asb to cluster"
+      repository: "ansible-service-broker"
+      timeout: 1800
+      script: |-
+        $GOPATH/src/github.com/origin/_output/local/bin/linux/amd64/oc create project ansible-service-broker
+        $GOPATH/src/github.com/origin/_output/local/bin/linux/amd64/oc process ./templates/deploy-ansible-service-broker.template.yaml -p BROKER_CA_CERT=$(oc get secret -n kube-service-catalog -o go-template=\'{{ range .items }}{{ if eq .type \"kubernetes.io/service-account-token\" }}{{ index .data \"service-ca.crt\" }}{{end}}{{\"\\n\"}}{{end}}\' | tail -n 1) | oc create -n ansible-service-broker -f -'
```
jpeeler: If oc is indeed not in the path, you'll need to specify the full path throughout the command.

jpeeler:
I'll think you'll have to do like this: https://github.com/openshift/origin-ci-tool/commit/ebaec23624fac48e6876462417908cb8ab4de98c.

PR #978, opened 2018-01-08
Remove PR-specific checks from base AMI

kargakis:


kargakis:
https://ci.openshift.redhat.com/jenkins/view/AMIs/job/ami_build_origin_int_rhel_base/524

stevekuznetsov:
AMI is only valid for master so this seems reasonable

kargakis:
Sorry for not mentioning it but previously the job was broken because `PULL_BASE_REF` does not exist in its env. Trying to get it to pass tests now.

PR #981, opened 2018-01-08
Using new push_status build record from OIT

adammhaile:
For use with openshift/enterprise-images#124
@jupierce

### DIFF
--- a/jobs/build/ocp/Jenkinsfile
+++ b/jobs/build/ocp/Jenkinsfile
@@ -678,8 +678,11 @@ images:build
                 for(i = 0; i < builds.size(); i++) {
                   bld = builds[i]
                   distgit = bld['distgit']
-                  if ( bld['status'] == '-1' ){
+                  if ( bld['status'] != '0' ){
                     failed_map[distgit] = bld['task_url']
+                  }
+                  else if ( bld['push_status'] != '0' ){
+                    failed_map[distgit] = 'Failed to push built image. See debug.log'
                   }
                   else {
                     // build may have succeeded later. If so, remove.
@@ -694,7 +697,7 @@ images:build
 Input URL: ${env.BUILD_URL}input
 Jenkins job: ${env.BUILD_URL}
 
-FAILED BUILDS:
+BUILD / PUSH FAILURES:
 ${failed_map}
 """);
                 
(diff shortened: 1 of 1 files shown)


jobs/build/ocp/Jenkinsfile
```diff
@@ -674,16 +674,25 @@ images:build
...
-                    failed_map[distgit] = bld['task_url']
+                    failed_build_map[distgit] = bld['task_url']
                   }
                   else {
                     // build may have succeeded later. If so, remove.
-                    failed_map.remove(distgit)
+                    failed_build_map.remove(distgit)
+                  }
+                  
+                  if ( bld['push_status'] == '-1' ){
```
jupierce: Just noticed on this read through, but these status checks should be `if  ... != 0` instead of ==-1 . We may want useful non-zeros at some point.

jobs/build/ocp/Jenkinsfile
```diff
@@ -710,7 +722,7 @@ ${failed_map}
                     return false  // cause waitUntil to loop again
                 } else if ( resp == "CONTINUE" ) {
                     echo "User chose to build fails are OK."
```
jupierce: wording

jobs/build/ocp/Jenkinsfile
```diff
@@ -674,16 +674,25 @@ images:build
...
-                    failed_map.remove(distgit)
+                    failed_build_map.remove(distgit)
+                  }
+                  
+                  if ( bld['push_status'] == '-1' ){
+                    failed_push_list += distgit
+                  }
+                  else {
+                    // push may have succeeded later. If so, remove.
+                    failed_push_list.removeAll{ it == distgit }
```
jupierce: I would just put everything in the same failed_map, otherwise your build exclusion logic breaks down. The key:value can be `<dist-git-name>:"failed to push..."` . For simplicity, we don't need to discriminate between build & push failures in the email.

jobs/build/ocp/Jenkinsfile
adammhaile: Fair enough. Fixed.

jobs/build/ocp/Jenkinsfile
adammhaile: What I was going for was an image push was basically non-fatal. But that's fine. Will simplify.

jobs/build/ocp/Jenkinsfile
adammhaile: Wow... I'm gonna blame autocorrect there. Sure. That's it...
Updated

adammhaile:
@jupierce all requests updated

PR #995, opened 2018-01-15
Fail jobs with correct message

smarterclayton:
conformance_k8s ran directly failed with "undefined buildId", not the
appropriate error.

Also, branch jobs are sometimes invoked directly.  Having this check is annoying, not sure it's correct in all cases.

@stevekuznetsov


smarterclayton:
@kargakis

PR #996, opened 2018-01-16
fix ansible logging tests for release branches

richm:
the ansible logging json-file and journald jobs are not in sync
this should get them more in sync
@sdodson @jcantrill


sdodson:
CI jobs pushed.

PR #1006, opened 2018-01-21
Unify GCE jobs, use existing RPMs where possible

smarterclayton:
Add a launch job for Ansible. Continues work from previous PR


smarterclayton:
Babysitting these manually so we can get the ansible gce job up (can't wait for new stuff anymore)

PR #1014, opened 2018-01-23
publish-oc-binary: minor fixes

bbguimaraes:
@dobbymoodge @jupierce @stevekuznetsov

### DIFF
--- a/build-scripts/rcm-guest/publish-oc-binary.sh
+++ b/build-scripts/rcm-guest/publish-oc-binary.sh
@@ -1,9 +1,9 @@
-#!/bin/sh
+#!/bin/bash
 set -eux
 
 rpm_name() {
     printf "${RPM}" "${arch}"
-    [ "${arch}" == x86_64 ] && printf %s -redistributable
+    [[ "${arch}" == x86_64 ]] && printf %s -redistributable
     printf %s "-${VERSION}"
 }
 
@@ -12,13 +12,13 @@ extract() {
     mkdir macosx windows
     for arch in x86_64 ${ARCH}; do
         rpm=$(echo "$(rpm_name "${arch}")"*)
-        [ "${arch}" != x86_64 -a ! -e "${rpm}" ] && continue
+        if [[ "${arch}" != x86_64 && ! -e "${rpm}" ]]; then continue; fi
         mkdir "${arch}"
-        if [ "${arch}" != x86_64 ]; then
-            rpm2cpio ${rpm}* | cpio -idm --quiet ./usr/bin/oc
+        if [[ "${arch}" != x86_64 ]]; then
+            rpm2cpio "${rpm}" | cpio -idm --quiet ./usr/bin/oc
             mv usr/bin/oc "${arch}"
         else
-            rpm2cpio ${rpm}* \
+            rpm2cpio "${rpm}" \
                 | cpio -idm --quiet \
                     ./usr/share/atomic-openshift/{linux,macosx}/oc \
                     ./usr/share/atomic-openshift/windows/oc.exe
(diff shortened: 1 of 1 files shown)


build-scripts/rcm-guest/publish-oc-binary.sh
```diff
@@ -12,13 +12,13 @@ extract() {
     mkdir macosx windows
     for arch in x86_64 ${ARCH}; do
         rpm=$(echo "$(rpm_name "${arch}")"*)
-        [ "${arch}" != x86_64 -a ! -e "${rpm}" ] && continue
+        if [[ "${arch}" != x86_64 ]] && [[ ! -e "${rpm}" ]]; then continue; fi
```
dobbymoodge: `if [[ "${arch}" != x86_64 `**&&**` ! -e "${rpm}" ]]` should work

build-scripts/rcm-guest/publish-oc-binary.sh
```diff
@@ -12,13 +12,13 @@ extract() {
...
         mkdir "${arch}"
-        if [ "${arch}" != x86_64 ]; then
-            rpm2cpio ${rpm}* | cpio -idm --quiet ./usr/bin/oc
+        if [[ "${arch}" != x86_64 ]]; then
+            rpm2cpio "${rpm}" | cpio -idm --quiet ./usr/bin/oc
```
dobbymoodge: @bbguimaraes not sure why the wildcard was removed here.

build-scripts/rcm-guest/publish-oc-binary.sh
```diff
@@ -12,13 +12,13 @@ extract() {
...
             mv usr/bin/oc "${arch}"
         else
-            rpm2cpio ${rpm}* \
+            rpm2cpio "${rpm}" \
```
dobbymoodge: and here

build-scripts/rcm-guest/publish-oc-binary.sh
bbguimaraes: Updated.

build-scripts/rcm-guest/publish-oc-binary.sh
bbguimaraes: It should have been moved to line 14, but was never removed from these two
lines.  This second wildcard is redundant, it can be seen in the logs that the
final file name is already contained in the variable.

build-scripts/rcm-guest/publish-oc-binary.sh
bbguimaraes: Same as above.

build-scripts/rcm-guest/publish-oc-binary.sh
```diff
@@ -1,9 +1,9 @@
-#!/bin/sh
+#!/bin/bash
 set -eux
 
 rpm_name() {
     printf "${RPM}" "${arch}"
-    [ "${arch}" == x86_64 ] && printf %s -redistributable
+    [[ "${arch}" == x86_64 ]] && printf %s -redistributable
```
stevekuznetsov: nit: writing this as 

```sh
if [[ "${arch}" == x86_64 ]]; then
    printf %s -redistributable
fi
```

Makes this look like "normal" code you would see in another language (except Ruby, but whatever) and IMHO improves readability/lowers the context switching overhead for a dev to read this script.

Also of note, explicit `if` and `else` will ensure devs are not using the `test && someting || else` construct, which many thing will run like a ternary operator but in reality has a different control flow.

Anyway, just a nit. Up to you @bbguimaraes @jupierce if you want to bother with that.

PR #1018, opened 2018-01-24
Revert "Removing emails from puddle build"

jupierce:
Reverts openshift/aos-cd-jobs#1015


jupierce:
Puddle has this option required. Need to solve this another way.

PR #1028, opened 2018-01-24
RPM artifacts were being stomped by branch job publishes

smarterclayton:
Postsubmits are supposed to be published, but branch jobs were. Switch how we publish postsubmits to be in /releases/ORG/REPO/BRANCH_OR_TAG/.latest-*.

Requires a manual move of GCS, which I will be doing later.  @stevekuznetsov if you can give it a side eye


smarterclayton:
Hold on to your butts

PR #1032, opened 2018-01-25
The oc binary must be built for extended tests to run

smarterclayton:
Future jobs will be less hacky, but this is necessary for k8s to run.


smarterclayton:
Broke on the 22nd, maybe we stopped copying oc out of the built artifact implicitly.

PR #1046, opened 2018-01-30
sources.yml write needed to happen sooner

adammhaile:
@jupierce - Found a timing bug while trying to fix the 3.9 build this morning. It was masked by the fact that we don't cleanup everything in the workspace between runs and sources.yml basically never changes. But I had a clean slate this time, so it was causing an issue.


jupierce:
Nice catch!

PR #1051, opened 2018-01-30
specify a more specific registry

jcantrill:
Set a specific image registry so we know we are getting the desired logging images


smarterclayton:
Yeah, this won't work with local images unless you tag them in, right?

PR #1052, opened 2018-01-31
make sure docker is using journald for journald tests

richm:
Not sure which playbook enables docker log-driver, so just use
all of them.
With this, I can confirm:

    $ sudo docker info |grep -i log
    Logging Driver: journald


richm:
@jcantrill @ewolinetz

PR #1054, opened 2018-01-31
Adding buildvm snapshot email to buildvm-maint

adammhaile:
@jupierce - We can certainly do something other than / in addition to email. This was just a very quick first pass to layout what I was thinking of capturing.

### DIFF
--- a/jobs/build/buildvm-maint/scripts/gen_snapshot.sh
+++ b/jobs/build/buildvm-maint/scripts/gen_snapshot.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+echo "INSTALLED PYTHON PACKAGES"
+echo "========================="
+pip freeze
+
+echo ""
+echo "INSTALLED YUM PACKAGES"
+echo "======================"
+yum list installed
+
+echo ""
+echo "INSTALLED RUBY GEMS"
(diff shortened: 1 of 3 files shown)


adammhaile:
Haha, good idea. Doubt we'll really get emails that often... but would also
be a good sanity check, letting us know that something has changed we
weren't expecting.
Maybe even store to /home/jenkins the actual diff (still send the whole
thing in email) so that we could login and see what the diff was, should we
care.

adammhaile:
@jupierce updated to include diff suggestion

adammhaile:
@jupierce The requested changes have already been addressed here

jobs/build/buildvm-maint/scripts/gen_snapshot.sh
```diff
@@ -0,0 +1,25 @@
...
+
+echo ""
+echo "INSTALLED RUBY GEMS"
+echo "==================="
+gem list --local
+
+echo ""
+echo "DOCKER INFO"
+echo "==========="
+docker info
```
jupierce: docker info will change constantly since it lists freespace. Incompatible with a diff model.

adammhaile:
@jupierce Removed docker info

PR #1060, opened 2018-02-01
Build AMI after an OCP build / image refresh

bbguimaraes:
@jupierce @mwoodson

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -75,10 +75,12 @@ def initialize_enterprise_images_dir() {
     echo "Initialized env.ENTERPRISE_IMAGES_DIR: ${env.ENTERPRISE_IMAGES_DIR}"
 }
 
-def oit( cmd ){
+def oit(cmd, opts=[:]){
     cmd = cmd.replaceAll( '\n', ' ' ) // Allow newlines in command for readability, but don't let them flow into the sh
     cmd = cmd.replaceAll( ' \\ ', ' ' ) // If caller included line continuation characters, remove them
-    sh "${env.ENTERPRISE_IMAGES_DIR}/oit/oit.py --user=ocp-build --metadata-dir ${env.ENTERPRISE_IMAGES_DIR} ${cmd.trim()}"
+    return sh(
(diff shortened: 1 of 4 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -78,7 +78,9 @@ def initialize_enterprise_images_dir() {
 def oit( cmd ){
     cmd = cmd.replaceAll( '\n', ' ' ) // Allow newlines in command for readability, but don't let them flow into the sh
     cmd = cmd.replaceAll( ' \\ ', ' ' ) // If caller included line continuation characters, remove them
-    sh "${env.ENTERPRISE_IMAGES_DIR}/oit/oit.py --user=ocp-build --metadata-dir ${env.ENTERPRISE_IMAGES_DIR} ${cmd.trim()}"
+    return sh(
```
jupierce: suggesting capture=false as a default argument. The output of oit is usually useful in the Jenkins log, so I'd like it to continue streaming there by default. 
`return sh( returnStdout: capture, "......")`

bbguimaraes: Updated.

pipeline-scripts/buildlib.groovy
```diff
@@ -354,6 +356,53 @@ def build_puddle(conf_url, keys, Object...args) {
...
+    waitUntil {
+        try {
+            build(job: 'build%2Faws-ami', parameters: [
+                param('String', 'OPENSHIFT_VERSION', version),
+                param('String', 'OPENSHIFT_RELEASE', release),
+                param('Boolean', 'USE_CRIO', true),
+                param(
+                    'String', 'CRIO_SYSTEM_CONTAINER_IMAGE_OVERRIDE',
+                    'registry.reg-aws.openshift.com:443/openshift3/cri-o:v'
+                        + full_version)])
```
bbguimaraes: @jupierce, @mwoodson: I believe these are the only parameters that need to be
passed at the moment, and they should not be affected (complemented, if
anything) by future changes to keep the base ami updated.

Let me know if that's not the case.

PR #1064, opened 2018-02-02
debug_level = 5 for gce jobs

sdodson:
Normally this is imported from the groupvars in sjb/inventory but the GCE jobs don't pick those up. Need a better long term fix for ensuring that they're picked up but this should ensure that GCE jobs log at loglevel 5.


stevekuznetsov:
@smarterclayton any objections?

sdodson:
sounds like no

PR #1074, opened 2018-02-06
aws-ami: fix boolean parameter

bbguimaraes:
Correct the incorrect correction in 6ecf674: use an actual
`BooleanParameter`, convert it to the string representation of a python
bool when passing it to ansible.

@jupierce @mwoodson


jupierce:
@mwoodson FYI

PR #1077, opened 2018-02-06
Fix crio moar space

cevich:
Because these commands operate inside a shell with ``set -u`` on it, any
undefined variable will throw an error instead of being treated as if it
were empty (default behavior).  Fix generation of the oct commands
WRT ``$ROOT_VOLUME_SIZE`` such that it's explicitly treated as empty
when undefined.

Also wrap the oct configure command in a complete 'if' statement, using
the bash conditional syntax (double-bracket) rather than shelling out to
"test" (single bracket).

### DIFF
--- a/sjb/actions/oct_install.py
+++ b/sjb/actions/oct_install.py
@@ -4,6 +4,7 @@ from actions.named_shell_task import render_task
 from .interface import Action
 
 _OCT_INSTALL_TITLE = "INSTALL THE ORIGIN-CI-TOOL"
+# N/B: Running inside bash with 'set -u'!
 _OCT_INSTALL_ACTION = """latest="$( readlink "${HOME}/origin-ci-tool/latest" )"
 touch "${latest}"
 cp "${latest}/bin/activate" "${WORKSPACE}/activate"
@@ -16,7 +17,10 @@ mkdir -p "${OCT_CONFIG_HOME}"
 rm -rf "${OCT_CONFIG_HOME}/origin-ci-tool"
 oct configure ansible-client verbosity 2
 oct configure aws-client 'keypair_name' 'libra'
-oct configure aws-client 'private_key_path' '/var/lib/jenkins/.ssh/devenv.pem'"""
+oct configure aws-client 'private_key_path' '/var/lib/jenkins/.ssh/devenv.pem'
+if [[ -n "${ROOT_VOLUME_SIZE:-}" ]] ; then
+    oct configure aws-defaults master_root_volume_size $ROOT_VOLUME_SIZE
+fi"""
 
 
 class OCTInstallAction(Action):
--- a/sjb/config/common/test_cases/crio.yml
+++ b/sjb/config/common/test_cases/crio.yml
@@ -16,6 +16,9 @@ parameters:
     description: "Pull request number."
   - name: PULL_PULL_SHA
     description: "Pull request head SHA."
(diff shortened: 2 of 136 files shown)


cevich:
Failure observed in [this job's console (unbound variable).](https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_cmd/9730/consoleFull#-133342521558b6e51eb7608a5981914356).

To validate the fix, I copied the [initial parts of sjb for the failing job](https://github.com/openshift/aos-cd-jobs/pull/1077/commits/7e904a2b7570e9bdf5a7fe5443ec23118e08e441), and will run it manually.

I will then modify the same job so it defines ``$ROOT_VOLUME_SIZE`` and confirm the new size is realized.

cevich:
Previously failing job, w/ [fix from this PR shows success when variable is not defined](https://ci.openshift.redhat.com/jenkins/job/cevich_testing_pr_1077/1/console)(last line is most significant):

```
Option keypair_name updated to be libra.
+ oct configure aws-client private_key_path /var/lib/jenkins/.ssh/devenv.pem
Option private_key_path updated to be /var/lib/jenkins/.ssh/devenv.pem.
+ [[ -n '' ]]
```

[Modified testing job ](https://github.com/openshift/aos-cd-jobs/pull/1077/commits/45779cf3960ba49e500e31f806d9e81d6962b1e2)to define ``$ROOT_VOLUME_SIZE`` and [manually executed after ``sjb_push-update.sh`` also shows success](https://ci.openshift.redhat.com/jenkins/job/cevich_testing_pr_1077/2/console):

```
Option keypair_name updated to be libra.
+ oct configure aws-client private_key_path /var/lib/jenkins/.ssh/devenv.pem
Option private_key_path updated to be /var/lib/jenkins/.ssh/devenv.pem.
+ [[ -n 42 ]]
+ oct configure aws-defaults master_root_volume_size 42
Option master_root_volume_size updated to be 42.
```

So this fixes the problem in both the positive and negative cases.

cevich:
@dobbymoodge @stevekuznetsov @kargakis Apologies for not adequately testing this previously.    I believe I've done so now (details above).  If there's another way to confirm, or if someone could tripple-check before merging, I'd sleep better tonight 😀 

@mrunalp Just FYI - The "disk space problem" may re-appear until this fix-of-the-fix is merged.  Original attempt was fine for crio, but apparently broke everything else 😲

stevekuznetsov:
https://github.com/openshift/origin-ci-tool/commit/7083cf2802334da01afa461ca2314e6eae38a563 should fix the issue at the root

/close

cevich:
> Hmmm @cevich can you show what the size is set to if you don't manually change the default? I was expecting the default to be 75GB right now.

@stevekuznetsov You're the expert, but my understanding is in AWS, VMs are lazy-deleted, possibly hours later.  So consider this:  Changing the default for every single VM by +40 GB, multiplied by the total number of VMs, is potentially almost a 33% increase in storage-allocation pressure.  I don't know if/how well AWS can handle that, or what the limit is, but are you sure we should push it?   Such a large change could cause other, much harder-to-diagnose problems, vs only increasing space where it's needed.

cevich:
@stevekuznetsov Clarification:  It's the quantity increase AND the rate of increase I was referring to.  Throwing more quota at the problem is a valid solution just so long as the rubber-stamps have ink.  No need to respond, maybe it's not even a problem, just sharing my thoughts.

PR #1084, opened 2018-02-07
build_ami: add a return statement to waitUntil

bbguimaraes:
The `waitUntil` step requires a `true` return value, even if the end of
the block is reached without any errors.

> Job: I was successful.
> Jenkins: SUCCESS IS NOT ENOUGH!


bbguimaraes:
https://buildvm.openshift.eng.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Focp/249/console

PR #1105, opened 2018-02-09
Changing variable used to checking out branch for openshift-ansible

ewolinetz:
@stevekuznetsov does this do what I think it does?
For the origin-aggregated-logging ci tests `es5.x` and `master` branches we should be using openshift-ansible `master` branch but we want to use the appropriate branch if we're doing a `release-x.y` pr/test

### DIFF
--- a/sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml
+++ b/sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml
@@ -17,7 +17,11 @@ extensions:
       title: "build an openshift-ansible release"
       repository: "openshift-ansible"
       script: |-
-        git checkout "${PULL_REFS%%:*}"
+        if [[ "${PULL_REFS%%:*}" == es5.x ]]; then
+          git checkout "${OPENSHIFT_ANSIBLE_TARGET_BRANCH}"
+        else
+          git checkout "${PULL_REFS%%:*}"
+        fi
         tito_tmp_dir="tito"
         rm -rf "${tito_tmp_dir}"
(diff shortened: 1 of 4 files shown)


sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml
```diff
@@ -17,7 +17,7 @@ extensions:
       title: "build an openshift-ansible release"
       repository: "openshift-ansible"
       script: |-
-        git checkout "${PULL_REFS%%:*}"
```
ewolinetz: We should do a check here similar to the other `if` statements for evaluating if its master/es5.x ....

ewolinetz:
cc @richm

sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml
```diff
@@ -17,7 +17,11 @@ extensions:
...
+        if [[ "${PULL_REFS%%:*}" == es5.x ]]; then
+          git checkout "${OPENSHIFT_ANSIBLE_TARGET_BRANCH}"
```
stevekuznetsov: I don't think this will be set to anything other than `master`.

ewolinetz: that's fine, before I wasn't using an if else block, so it was set to `OPENSHIFT_ANSIBLE_TARGET_BRANCH` all the time. If we're running ci for the es5.x branch we should use the ansible target branch (which is fine to be `master` always)

stevekuznetsov:
Pushed out

PR #1122, opened 2018-02-16
Get all container logs from builds

smarterclayton:
I need this to debug some failures. Seemed to work locally but haven't added one of these before.

@kargakis @stevekuznetsov


stevekuznetsov:
Should not be able to crash jobs

/shrug
/lgtm

stevekuznetsov:
pushed

PR #1082, opened 2018-02-06
Deploy branch specific vars from sjb/inventory/${branch}.cfg

sdodson:
Necessary because master branch needs to override several image tags

PR #1147, opened 2018-02-26
Re-Add branch specific evars

sdodson:


sdodson:
@stevekuznetsov PTAL, previous impelmentation was appending "" to the command line and causing errors on non master branches. https://github.com/openshift/aos-cd-jobs/pull/1082 Just work around this by providing a default with meaningless variable

stevekuznetsov:
Can't give meaningful review here, but sure

PR #1152, opened 2018-02-27
Record latest build number for cri-o jobs

stevekuznetsov:

/cc @kargakis


sjb/config/common/test_cases/crio.yml
```diff
@@ -73,6 +73,8 @@ actions:
         echo "${path%\/}" > "${BUILD}.txt"
         gsutil cp "${BUILD}.txt" "gs://origin-federated-results/pr-logs/directory/${JOB_NAME}/${BUILD_NUMBER}.txt"
       fi
+      echo ${BUILD} > latest-build.txt
+      gsutil cp latest-build.txt "gs://origin-ci-test/pr-logs/directory/${JOB_NAME}/latest-build.txt"
```
kargakis: Different bucket from the one used by the test but yeah we can only give a single bucket to tot now or potentially update it to map jobs to buckets?

/shrug

PR #1162, opened 2018-02-28
Upload latest-build.txt for kubernetes conformance jobs

kargakis:
Last set of jobs that seems to upload starting metadata in GCS

/cc stevekuznetsov


kargakis:
@stevekuznetsov the kubernetes and crio jobs updated here seem to only run inside Jenkins. We should move them to run as periodics in prow. Opened https://github.com/openshift/aos-cd-jobs/issues/1165.

PR #1172, opened 2018-03-01
Fall back to vX.Y for image tag during install

smarterclayton:
Have system containers use the latest major/minor tag. Have other jobs explicitly set openshift_image_tag to the current commit.

### DIFF
--- a/sjb/config/common/test_cases/origin_installed_release.yml
+++ b/sjb/config/common/test_cases/origin_installed_release.yml
@@ -87,6 +87,7 @@ extensions:
                          -e etcd_data_dir="${ETCD_DATA_DIR}" \
                          -e openshift_master_default_subdomain="${local_ip}.nip.io"             \
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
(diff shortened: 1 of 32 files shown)


smarterclayton:
@stevekuznetsov not urgent but if you can take a look

sjb/config/common/test_cases/origin_installed_release.yml
```diff
@@ -87,6 +87,7 @@ extensions:
                          -e etcd_data_dir="${ETCD_DATA_DIR}" \
                          -e openshift_master_default_subdomain="${local_ip}.nip.io"             \
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
+                         -e openshift_image_tag="$( cat ./ORIGIN_COMMIT )"                      \
```
stevekuznetsov: Should this be the tag not the commit?

smarterclayton: In this case we want to use the images that got built by the job.  Once every job is building images the same way, this value would be set to whatever the tag is for the combined job.

PR #1180, opened 2018-03-02
DRY out the origin conformance install parents

stevekuznetsov:
Consolidate passed evars in base install template

---

DRY out the origin conformance install parents

---

/cc @smarterclayton 
/assign @michaelgugino @sdodson


stevekuznetsov:
Really unclear why the Origin conformance test was not using the common parent or why there was a huge amount of drift between the two in how they did installs/passed `-e` vars. I synced the parent to have the union of the `-e` vars passed in both, then moved the conformance install to inherit.

sjb/config/common/test_cases/origin_built_installed_release.yml
```diff
@@ -105,7 +107,10 @@ extensions:
                          -e etcd_data_dir="${ETCD_DATA_DIR}" \
                          -e openshift_master_default_subdomain="${local_ip}.nip.io"             \
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
+                         -e openshift_release="$( cat ./ORIGIN_RELEASE )"                       \
```
michaelgugino: Setting openshift_release with extra_vars is bad.

We should be removing extra_vars usage, not expanding it.

stevekuznetsov: This is how the current Origin jobs do it -- what is better? Can you post a snippet of how you'd rather have this info passed in?

stevekuznetsov: I don't have strong feelings about this, but our jobs are wildly inconsistent about this right now and I would love for them not to be.

michaelgugino: export these values as environment variables, then use lookups:

https://github.com/openshift/openshift-ansible/blob/master/test/ci/inventory/group_vars/all.yml#L13

stevekuznetsov: How are we pulling in that file _right now_?

sdodson: Even if we should stop using extra_vars, this seems like a step in the right direction by consolidating them all in one place, or at the very least fewer places.

sjb/config/test_cases/test_pull_request_openshift_ansible_extended_conformance_install.yml
```diff
@@ -2,10 +2,9 @@
...
 overrides:
   junit_analysis: True
-  sync_repos:
-    - name: "aos-cd-jobs"
-    - name: "openshift-ansible"
-      type: "pull_request"
-    - name: "origin"
-    - name: "image-registry"
-    - name: "kubernetes-metrics-server"
+  sync:
```
sdodson: This seems unrelated to the PR, fine with it if you are, just don't know if it was intentional.

stevekuznetsov: These have to change since this PR conflates a) the migration to a parent and b) the migration from `sync_repos` to `sync`

PR #1181, opened 2018-03-02
For non-master (read - backporting) prs disable all health checks for origin logging tests

ewolinetz:
We only want to disable them for the origin aggregated logging repo and only if they are `release-*` branches.

PTAL @sosiouxme @richm @stevekuznetsov

### DIFF
--- a/sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml
+++ b/sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml
@@ -119,6 +119,13 @@ extensions:
       title: "origin prerequisites"
       repository: "aos-cd-jobs"
       script: |-
+        curbranch=$( git rev-parse --abbrev-ref HEAD )
+        logging_extras=""
+        if [[ "${curbranch}" == master ]] ; then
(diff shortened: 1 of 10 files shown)


sosiouxme:
Logging runs into enough random other stuff that reducing what can go wrong seems like a worthy goal. I'm a little skeptical that this helps though... I don't know that this logging test is running into check failures that aren't showing up in every other test too. Backport tests are running into more version failures because the CI env results in more dependency quirks, and since logging does a lot of backports... that's a frequent pain.

But see [this test today](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-ansible/7347/test_pull_request_openshift_ansible_extended_conformance_install_system_containers/5643/) for instance. Four other tests on the PR had checks fail on the same thing, but in this one the failure happened outside the check anyway. The check's just usually the first thing to fail... something else is likely to fail if the check is disabled. (Notably, not true in the case of the docker-1.13 issue.)

I'm fine with this if you want. The only real harm is in adding another logical branch. Just not sure it's going to help either.

ewolinetz:
This is no longer needed, the problem has been resolved on 3.7 by https://github.com/openshift/openshift-ansible/pull/7347

PR #1184, opened 2018-03-02
DRY Out the system containers jobs

stevekuznetsov:

/assign @ingvagabund 

I am trying to DRY out these configs -- need help understanding the atomic pull stuff in there.


ingvagabund:
I will check it out Steve. Looks like all the DRY code is now included by all jobs. Originally, I took the containerized job configuration and just updated it so it works for the system containers. So it should be just simple tweaking of the original template.

ingvagabund:
I don't see anything building the origin images? Where did that go?

The key difference from the containerized install was to run:

```
# install atomic rpm
sudo yum install atomic -y
# only origin, openvswitch and node docker images are run as system containers
sudo atomic pull --storage ostree docker:openshift/origin:$( cat ./ORIGIN_TAG )
sudo atomic pull --storage ostree docker:openshift/openvswitch:$( cat ./ORIGIN_TAG )
sudo atomic pull --storage ostree docker:openshift/node:$( cat ./ORIGIN_TAG )
```

calling the ansible-playbook with proper ansible variables:
```
-e openshift_use_system_containers=true \
-e openshift_docker_systemcontainer_image_override=brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/rhel7 \
-e system_images_registry=docker \
-e containerized=true      \
```

and verifying the system containers are running at the end:

```
docker_image_tag="$( cat ./ORIGIN_TAG )"
docker_repository="${OS_IMAGE_PREFIX:-"openshift/origin"}"
[[ "$(sudo atomic containers list -n --no-trunc -f runtime=runc -f image=${docker_repository}:${docker_image_tag} --json | jq '.[0].running')" == "true" ]]
```

The rest was just a copy-paste of the containerized installation.

ingvagabund:
Did you tried to run the DRY-out JJ if it still works?

stevekuznetsov:
This is the only part that is hard with the current DRY out:

```
# install atomic rpm
sudo yum install atomic -y
# only origin, openvswitch and node docker images are run as system containers
sudo atomic pull --storage ostree docker:openshift/origin:$( cat ./ORIGIN_TAG )
sudo atomic pull --storage ostree docker:openshift/openvswitch:$( cat ./ORIGIN_TAG )
sudo atomic pull --storage ostree docker:openshift/node:$( cat ./ORIGIN_TAG )
```

Could you help me understand what that is doing? Why is the installer not doing this?

ingvagabund:
The atomic does not know about the docker images unless they are uploaded to its storage. @giuseppe knows more about that.

ingvagabund:
IIUC, the atomic storage is like the docker registry for the system containers.

stevekuznetsov:
@giuseppe @ingvagabund can I ask two more questions:

 - how long does the above step take?
 - are there any downsides to loading the images into atomic if the system containers will not be used?

Trying to figure out if it just makes sense to always pull them in or not

giuseppe:
"atomic pull --storage ostree" pulls the image from the registry to the ostree storage  (it internally just uses skopeo copy).

There is no downside to pull the images if they are not used.

IIRC, we are doing that in the aos-cd-jobs so we can build the images with "docker build" and then pull them in the ostree storage so that they can be used as a system container, there is no network involved as it translates to `skopeo copy docker-daemon:openshift/node:$( cat ./ORIGIN_TAG ) ostree:openshift/node:$( cat ./ORIGIN_TAG )@/ostree/repo`

sjb/config/test_cases/test_pull_request_openshift_ansible_extended_conformance_install_system_containers.yml
```diff
@@ -1,10 +1,9 @@
 ---
```
stevekuznetsov: @ingvagabund @giuseppe what branches do we care about for this job against openshift-ansible?

ingvagabund: IIRC 3.6+. Given we do the last three, in all branches now? @giuseppe can you verify?

stevekuznetsov: Done

PR #1186, opened 2018-03-04
impl envvars for logging jobs

jcantrill:
@stevekuznetsov  I think this would superceded #1181.  It utilizes the pattern of reusing the jobs as you described with overriding the repos.  It removes the parts I think @richm added to disable repos but that I believe is moot given the changes identified by @ewolinetz  in #1181.

Please have a look.

### DIFF
--- a/build-scripts/rcm-guest/rhel7next.sh
+++ b/build-scripts/rcm-guest/rhel7next.sh
@@ -41,14 +41,14 @@ _createrepo() {
 mkdir -p /mnt/rcm-guest/puddles/RHAOS/rhel7next/
 cd /mnt/rcm-guest/puddles/RHAOS/rhel7next/
 cat > rhel7next.repo <<-'EOF'
-	[rhel-7-fast-datapath-rpms]
+	[rhel-7-fast-datapath-rpms-2]
 	name = Red Hat Enterprise Linux 7 Fast Datapath
 	baseurl = http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/7Server/x86_64/fast-datapath/os/
 	# baseurl = http://pulp.dist.stage.ext.phx2.redhat.com/content/dist/rhel/server/7/7Server/x86_64/fast-datapath/os/
 	enabled = 0
 	gpgcheck = 0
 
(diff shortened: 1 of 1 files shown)


sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml
```diff
@@ -18,169 +18,6 @@ extensions:
...
-        else
-            playbook="${playbook_base}byo/config.yml"
-        fi
-        ansible-playbook -vv --become               \
-                         --become-user root         \
-                         --connection local         \
-                         --inventory sjb/inventory/ \
-                         -e openshift_deployment_type=origin  \
-                         -e debug_level=2           \
-                         -e openshift_logging_install_logging=False \
```
stevekuznetsov: You need to be sure to pass these still?

sjb/config/test_cases/test_pull_request_openshift_ansible_logging_36.yml
```diff
@@ -4,3 +4,5 @@ overrides:
   sync_repos: []
   sync:
     - "openshift,origin-aggregated-logging=release-3.6"
+    - "openshift,origin=release-3.6"
+  evars: "-e openshift_disable_check=* -e openshift_install_examples=false"
```
stevekuznetsov: Do you expect this in all branches? Push it up a level to the parent then.

sjb/config/test_cases/test_pull_request_origin_aggregated_logging_ansible.yml
```diff
@@ -13,174 +13,7 @@ extensions:
...
-            playbook="${playbook_base}byo/config.yml"
-        fi
-        ansible-playbook -vv --become               \
-                         --become-user root         \
-                         --connection local         \
-                         --inventory sjb/inventory/ \
-                         -e openshift_logging_install_logging=False \
-                         -e openshift_deployment_type=origin  \
-                         -e debug_level=2           \
-                         -e openshift_docker_log_driver=journald \
```
stevekuznetsov: need to pass these

PR #1192, opened 2018-03-07
Collect entire journal since boot for crio tests

cevich:
Otherwise debugging ugly problems, like kernel-tracebacks is a PITA.
Use the --boot flag to limit data to only the most recent (current)
boot.

### DIFF
--- a/sjb/config/common/test_cases/crio.yml
+++ b/sjb/config/common/test_cases/crio.yml
@@ -131,6 +131,7 @@ artifacts:
   - /tmp/kube-proxy.log
   - /tmp/kube-proxy.yaml
   - /tmp/kube-scheduler.log
+  - /etc/crio/crio.conf
 generated_artifacts:
   installed_packages.log: 'sudo yum list installed'
   avc_denials.log: 'sudo ausearch -m AVC -m SELINUX_ERR -m USER_AVC'
--- a/sjb/config/common/test_cases/crio.yml
+++ b/sjb/config/common/test_cases/crio.yml
@@ -136,6 +136,7 @@ generated_artifacts:
   avc_denials.log: 'sudo ausearch -m AVC -m SELINUX_ERR -m USER_AVC'
   filesystem.info: 'sudo df -h && sudo pvs && sudo vgs && sudo lvs'
   pid1.journal: 'sudo journalctl _PID=1 --no-pager --all --lines=all'
+  system.journal: 'sudo journalctl --no-pager --no-hostname --boot'
 system_journals:
   - crio.service
   - customcluster.service
--- a/sjb/generated/ami_build_origin_int_fedora_crio.xml
+++ b/sjb/generated/ami_build_origin_int_fedora_crio.xml
@@ -301,6 +301,10 @@ if ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel sudo sta
     ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel sudo chmod -R o+rX /tmp/kube-scheduler.log
     scp -r -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel:/tmp/kube-scheduler.log &#34;${ARTIFACT_DIR}&#34;
 fi
+if ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel sudo stat /etc/crio/crio.conf; then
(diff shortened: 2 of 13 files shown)


cevich:
Ref: https://github.com/kubernetes-incubator/cri-o/issues/1424

cevich:
@stevekuznetsov @mrunalp I know you're both busy, but when you get a chance PTAL.  Small change to collect the whole journal on every system.  Needed for diagnosing some classes of intermittent failures :frowning_face:

sjb/config/common/test_cases/crio.yml
```diff
@@ -136,6 +136,7 @@ generated_artifacts:
   avc_denials.log: 'sudo ausearch -m AVC -m SELINUX_ERR -m USER_AVC'
   filesystem.info: 'sudo df -h && sudo pvs && sudo vgs && sudo lvs'
   pid1.journal: 'sudo journalctl _PID=1 --no-pager --all --lines=all'
+  system.journal: 'sudo journalctl --no-pager --no-hostname --boot'
```
stevekuznetsov: Seems like a big hammer -- you already have the kernel logs, systemd logs, cri-o logs ... but if this is what you need, seems fine

wking: > …you already have the kernel logs…

Where are those?  I can't find anything that looks like `dmseg` in [here][1].

[1]: https://ci.openshift.redhat.com/jenkins/job/test_pull_request_crio_integration_rhel/621/

cevich: @wking you wanna look under ``S3 Artifacts`` (left side).  Still, even there I don't see any kernel or general logs...looking again...ya, for a crio-integration run, the only files I see are:

```
ansible_client_configuration.yml	￼ more details
ansible_variables.yml	￼ more details
aws_client_configuration.yml	￼ more details
aws_variables.yml	￼ more details
gathered/artifacts/testout.txt	￼ more details
generated/avc_denials.log	￼ more details
generated/filesystem.info	￼ more details
generated/installed_packages.log	￼ more details
generated/pid1.journal	￼ more details
journals/crio.service	￼ more details
journals/customcluster.service	￼ more details
journals/systemd-journald.service	￼ more details
logs/junit/HaULvfQPhQ.xml	￼ more details
logs/junit/rQMmUKvWmJ.xml	￼ more details
logs/log.txt	￼ more details
```

The pid1.journal has only systemd stuff, nothing about other services.  IMHO, this is an area where it's better to just collect the big log and have it when needed, rather that wish we had it.

@stevekuznetsov would it be better if I piped it into gzip maybe?

stevekuznetsov: Ah hmmm ok looks like we only added kernel logs for Origin [here](https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/common/test_cases/origin.yml#L215) but it would be simple to add them for y'all too. 

Either way let me know when you'd like me to push the merge button on this and get the changes pushed out to Jenkins and I can do it.

wking: Can someone with VM access compare the size of this (does it need a `-lines=all`) with the Origin dmesg log?  I suspect grabbing the full log is not going to be all that much bigger than individuall collecting a pid1, crio, dmesg, etc.  But I'm personally only interested in adding dmesg logs at the moment, in case the a la carte approach turns out to be significantly cheaper.

stevekuznetsov: Size is not a concern as far as storage is concerned -- I'm more worried about a dev looking through ten million unrelated messages they don't care about.

wking: I'm comfortable grepping, although 10-million lines might be a bandwidth problem for me ;).

cevich: Aye, better to have the data + grep than be in a pickle with only grep and no data.

>  Either way let me know when you'd like me to push the merge button on this and get the changes pushed out to Jenkins and I can do it.

@mrunalp @runcom Y'all know better than I when's a safe-ish time to push this.  PTAL and let us know what and when you think.

Thanks all.

cevich:
@mrunalp @rhatdan @runcom PTAL.  I think this change would help diagnose some of the CRIO CI failures where we simply don't have these extra details available.

rhatdan:
LGTM

cevich:
@rhatdan Okay, snuck in collection of ``/etc/crio/crio.conf`` as a separate commit.

cevich:
@stevekuznetsov this is good-to-go then, when you have a more-free moment than normal :open_mouth:

stevekuznetsov:
changes pushed out

PR #1194, opened 2018-03-07
Add rpm based CRIO CI job

sdodson:
Depends on https://github.com/openshift/openshift-ansible/pull/7425

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_crio_rpm.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_crio_rpm.yml
@@ -1,0 +1,241 @@
+---
+parent: 'common/test_cases/origin.yml'
+overrides:
+  junit_analysis: False
+  timer: 'H H * * *'
+  email:
+    - decarr@redhat.com
+    - gscrivan@redhat.com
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
How urgent is this? If not p0 can we try to use the new format?

sdodson:
I'd want to have this job working by end of week.

stevekuznetsov:
EOW gives us ages. Let's sync tomorrow about how we might be able to implement this job without 2k LOC copies

smarterclayton:
Any update on this?  The current crio job is not correct and we can't test with it anymore.

smarterclayton:
Specifically openshift/openshift-ansible#7486 can't test crio because the system container in 3.9 is busted (until we rebase we can't use the new crio image).

Also, we need to be careful to make sure we have specific cri-o install branches - 1.9 and 1.10 must be used appropriately for each branch (and we can't switch to 1.10 until we get the rebase landed).

ashcrow:
:+1: on getting this finished.

Is the system container defective or is it a matter of version matching for Openshift? If the system container is broken please make sure an issue is opened so folks can fix it.

PR #1208, opened 2018-03-08
Build more in logging jobs, explicitly set repos to use

stevekuznetsov:

/cc @jcantrill @smarterclayton


sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -1,38 +1,20 @@
 ---
-parent: 'common/test_cases/origin_built_release.yml'
-extensions:
+parent: 'common/test_cases/origin_release_with_ecosystem.yml'
```
stevekuznetsov: @jcantrill so we now depend on the parent that builds origin, etc

PR #1210, opened 2018-03-09
Use proper oc binary and openshift images when testing origin-web-console

jhadvig:
https://trello.com/c/eBFQn4Vh

Updating origin-web-console jobs so the web-console tests are running against proper version of `oc` binary and also with right image versions.
To clarify why there are two special statements for the enterprise branch:
1. On out [dockerhub](https://hub.docker.com/r/openshift/origin/tags/) we dont have a v3.6 tag that would point to the latest 3.6 release. 
2. Since there wasnt an official 3.8 release its better to use the 3.9 images. Talked to @mfojtik regarding the the 3.8 and since we are not shipping the 3.8 or using it in online, the backports are not crucial.

The `oc` binary will be copied from the right `openshift/origin` image and used to cluster up.  

@stevekuznetsov @spadgett PTAL

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_web_console.yml
+++ b/sjb/config/test_cases/test_branch_origin_web_console.yml
@@ -3,13 +3,46 @@ parent: 'common/test_cases/origin_built_release.yml'
 extensions:
   actions:
     - type: "script"
+      title: "determine binary and image version"
+      repository: "origin"
+      script: |-
+        if [[ "${PULL_BASE_REF}" = "enterprise-"* ]]; then
+          case $PULL_BASE_REF in
+          enterprise-3.[2-5])
+            echo "IMAGE_VERSION=${PULL_BASE_REF/enterprise-3/v1}.1" > IMAGE_VERSION_VAR
+            ;;
+          enterprise-3.1)
+            echo "IMAGE_VERSION=v1.1.6" > IMAGE_VERSION_VAR
+            ;;
+          *)
+            echo "IMAGE_VERSION=${PULL_BASE_REF/enterprise-/v}" > IMAGE_VERSION_VAR
+            ;;
+          esac
+        else
+          echo "IMAGE_VERSION=latest" > IMAGE_VERSION_VAR
+        fi
+        source IMAGE_VERSION_VAR
+        sudo docker pull openshift/origin:"${IMAGE_VERSION}"
+        sudo docker create --name temp-container openshift/origin:"${IMAGE_VERSION}"
+        sudo docker cp temp-container:/usr/bin/oc ./
+        sudo docker rm temp-container
+        sudo mv -f ./oc /bin/oc
+    - type: "script"
       title: "start openshift server"
       repository: "origin"
       script: |-
-        sudo yum --disablerepo=\* --enablerepo=origin-local-release,oso-rhui-rhel-server-releases install -y origin-clients
         wget https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
         sudo yum install ./google-chrome-stable_current_*.rpm -y
-        oc cluster up --version=latest --public-hostname=localhost --loglevel=5 --server-loglevel=5
+        source IMAGE_VERSION_VAR
+        oc version
+        # If the version is not specified for the 3.9 and 3.10(latest) images, the oc binary will use download will fail to start the cluster,
+        # since both of those image versions currently contain bug, so we have to specify the version. Once the bug is fixed and the image is
+        # rebuilt, we can remove this statement.
+        if [ "${IMAGE_VERSION}" == "v3.9" ] || [ "${IMAGE_VERSION}" == "latest" ]; then
+          oc cluster up --version="${IMAGE_VERSION}" --public-hostname=localhost --loglevel=5 --server-loglevel=5
+        else
+          oc cluster up --public-hostname=localhost --loglevel=5 --server-loglevel=5
+        fi
     - type: "script"
       title: "run web console tests"
       repository: "origin-web-console"
--- a/sjb/generated/test_branch_origin_web_console.xml
+++ b/sjb/generated/test_branch_origin_web_console.xml
@@ -264,16 +264,58 @@ ssh -F ./.config/origin-ci-tool/inventory/.ssh_config -t openshiftdevel &#34;bas
         </hudson.tasks.Shell>
         <hudson.tasks.Shell>
           <command>#!/bin/bash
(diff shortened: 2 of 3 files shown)


sjb/config/test_cases/test_branch_origin_web_console.yml
```diff
@@ -2,14 +2,40 @@
...
+          fi
+        else
+          echo "IMAGE_VERSION=latest" > IMAGE_VERSION_VAR
+        fi
+        source IMAGE_VERSION_VAR
+        sudo docker pull openshift/origin:"${IMAGE_VERSION}"
+        sudo docker create --name temp-container openshift/origin:"${IMAGE_VERSION}"
+        sudo docker cp temp-container:/usr/bin/oc ./
+        sudo docker rm temp-container
+        sudo mv -f ./oc /bin/oc
```
spadgett: I'd probably move it to a tmp folder instead of overwriting /bin/oc, then use the full path when invoking it

sjb/config/test_cases/test_branch_origin_web_console.yml
```diff
@@ -2,14 +2,40 @@
...
       repository: "origin"
       script: |-
-        sudo yum --disablerepo=\* --enablerepo=origin-local-release,oso-rhui-rhel-server-releases install -y origin-clients
         wget https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
         sudo yum install ./google-chrome-stable_current_*.rpm -y
-        oc cluster up --version=latest --public-hostname=localhost --loglevel=5 --server-loglevel=5
+        source IMAGE_VERSION_VAR
+        oc version
+        if [ "${IMAGE_VERSION}" == "v3.9" ]; then
+          oc cluster up --version=v3.9 --public-hostname=localhost --loglevel=5 --server-loglevel=5
```
spadgett: Any reason not to use `--version="${IMAGE_VERSION}"` always and skip this if/else?

spadgett:
Do we need to special case releases earlier than 3.6 which use tags like `v1.5.0`, `v1.4.0`, etc.?

sjb/config/test_cases/test_branch_origin_web_console.yml
spadgett: Or leave it where it is and call it with `./oc`

jhadvig:
@spadgett no, from the conversations that I had we should be only concerned about releases 3.6 and up.

sjb/config/test_cases/test_branch_origin_web_console.yml
jhadvig: `oc` will download the images based on its version. The `if` exception is because of an issue in our 3.9 images that prevents from starting the cluster.
I think that there was one other issue with one specific version, but cant recall what it was. Anyway will try it on my testing job for all the release versions and will update you

spadgett:
> no, from the conversations that I had we should be only concerned about releases 3.6 and up.

My concern is that this will definitely break our CI for earlier releases, and today it works. I'd at least like to fall back to something that might work.

We recently backported a fix to 3.3 (#2774), so sometimes we do need to backport fixes that far.

spadgett:
We might cluster up with --version=v3.6.0 for earlier releases. I'm not sure offhand how far back cluster up support goes anyway.

stevekuznetsov:
@smarterclayton how does this jive with your plans

jhadvig:
@spadgett comments addressed

spadgett:
@stevekuznetsov Do you mind reviewing the bash?

I'm still hesitant to knowingly break the CI for several branches that work today :/

sjb/config/test_cases/test_branch_origin_web_console.yml
```diff
@@ -2,14 +2,44 @@
...
+          echo "IMAGE_VERSION=v3.6.1" > IMAGE_VERSION_VAR
+          ;;
+        enterprise-3.[2-5])
+          echo "IMAGE_VERSION=${PULL_BASE_REF/enterprise-3/v1}.1" > IMAGE_VERSION_VAR
+          ;;
+        enterprise-3.1)
+          echo "IMAGE_VERSION=v1.1.6" > IMAGE_VERSION_VAR
+          ;;
+        *)
+          echo "IMAGE_VERSION=latest" > IMAGE_VERSION_VAR
```
spadgett: This looks like any newer branches like enterprise-3.10, enterprise-3.11, etc. will use latest. I'd rather write it in a way that will work for future versions without needing to update the job.

jhadvig:
@spadgett updated

sjb/config/test_cases/test_branch_origin_web_console.yml
stevekuznetsov: Should explicitly write down why we are doing this in the job or here in the comments.

sjb/config/test_cases/test_branch_origin_web_console.yml
stevekuznetsov: Agreed

jhadvig: The thing is that the hack/test-headless.sh test-integration is calling the oc binary, if we wont put it into PATH the job will fail

smarterclayton:
I can push a v3.6 tag and we will have actual origin v3.8 images.

smarterclayton:
otherwise this is no more terrible than anything else

sjb/config/test_cases/test_branch_origin_web_console.yml
stevekuznetsov: ACK

stevekuznetsov:
When do we aim to have 3.8 images? I thought we were not planning on it

sjb/config/test_cases/test_branch_origin_web_console.yml
spadgett: We should be remove to remove this special case when we have a `v3.9` git tag since 3.9 cluster up will work without `--version`.

spadgett:
@deads2k What's the status for cluster up in origin master? I think you mentioned there were some problems.

I'm worried this will break the console merge queue if 3.10 cluster up is not reliable.

/hold

deads2k:
> @deads2k What's the status for cluster up in origin master? I think you mentioned there were some problems.

We haven't completely broken anything.  We're looking at changing some flags, but https://github.com/openshift/origin/pull/18907 makes it work against the latest minor version tag.  Some flags are going away, but these are fine: `oc cluster up --public-hostname=localhost --loglevel=5 --server-loglevel=5` and if I'm reading right, that's what you're using.

jhadvig:
@spadgett I've hit the issue with master today. To cluster up I had to add the `--version=latest` flag to the oc command:
`oc cluster up --version=latest --public-hostname=localhost --loglevel=5 --server-loglevel=5`

otherwise I was hitting following issue:
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_web_console_jakub/66/consoleFull#187430740058b6e51eb7608a5981914356
since the oc binary is looking for `openshift/origin:v3.10` image, which ios weird cause the image with that tag is available at [dockerhub](https://hub.docker.com/r/openshift/origin/tags/)

@smarterclayton +1 for having both 3.6 and 3.8 images... will make all things less hacky on all fronts ..

@stevekuznetsov so I've create a sanity job https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_web_console_jakub/
where I'm testing several versions of 3.6 - latest. The thing is that I'm hitting unrelated errors like
PROVISION CLOUD RESOURCES or FORWARD GCS CREDENTIALS TO REMOTE HOST

jhadvig:
@stevekuznetsov the tests have passed at https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_web_console_jakub/ but there have been also couple of flakes. Tested 3.6+ versions.
If you are ok with the state of the PR we could merge it

spadgett:
@jhadvig I just want to make sure that cluster up in 3.10 starts reliably before cutting over, otherwise it can break the console merge queue. @deads2k has made major changes.

If we think it's reliable, then no objections to going ahead.

sjb/config/test_cases/test_branch_origin_web_console.yml
```diff
@@ -2,14 +2,53 @@
...
 extensions:
   actions:
+    - type: "script"
+      title: "determine binary and image version"
+      repository: "origin"
+      script: |-
+        if [[ "${PULL_BASE_REF}" = "enterprise-"* ]]; then
+          case $PULL_BASE_REF in
+          enterprise-3.8)
+            echo "IMAGE_VERSION=v3.9" > IMAGE_VERSION_VAR
```
spadgett: Looks like there's a `v3.8` image now, so we could remove this special case.

jhadvig:
@spadgett you were right. `3.8` and also `v3.6` is available. Re-run all the tests with updated job and everytihng is green. I thing we are good to merge now.
cc'ing @stevekuznetsov

PR #1211, opened 2018-03-09
fix logging jobs to use envvars

jcantrill:
@stevekuznetsov I think this would superceded #1181. It utilizes the pattern of reusing the jobs as you described with overriding the repos. It removes the parts I think @richm added to disable repos but that I believe is moot given the changes identified by @ewolinetz in #1181.

Please have a look.


jcantrill:
reopening #1186

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
```diff
@@ -192,11 +31,7 @@ extensions:
         popd
         logging_extras=""
         if [[ "${PULL_REFS%%:*}" == es5.x ]]; then
```
stevekuznetsov: you could also add `$EXTRA_EVARS` to your `ansible-playbook` call and pass these in only in the es5.x job, your call

sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml
```diff
@@ -1,5 +1,11 @@
 ---
-parent: 'common/test_cases/origin_built_release.yml'
+parent: 'common/test_cases/origin_built_installed_release.yml'
+overrides:
+  sync_repos:
```
stevekuznetsov: i would expect this to be the same as in the other one -- an override on `sync:`

sjb/config/test_cases/test_pull_request_openshift_ansible_logging_37.yml
```diff
@@ -5,3 +5,4 @@ overrides:
     - "openshift,origin=release-3.7"
     - "openshift,aos-cd-jobs=master"
     - "openshift,origin-aggregated-logging=release-3.7"
+  evars: "-e openshift_disable_check=* -e openshift_install_examples=false"
```
stevekuznetsov: if `test_cases/test_branch_openshift_ansible_logging.yml` exposes these, I don't think you will need them in each child?

sjb/config/test_cases/test_pull_request_origin_aggregated_logging_journald_37.yml
```diff
@@ -0,0 +1,8 @@
+---
```
stevekuznetsov: If these are re-names, maybe we can do that in a separate PR? this one is big enough

sjb/config/test_cases/test_branch_openshift_ansible_logging.yml
richm: We really want to merge es5.x into master (for 3.10) ASAP

sjb/config/test_cases/test_pull_request_origin_aggregated_logging_journald.yml
```diff
@@ -0,0 +1,71 @@
+---
+parent: 'common/test_cases/origin_built_installed_release.yml'
+overrides:
+  sync:
+    - "openshift,aos-cd-jobs=master"
+    - "openshift,openshift-ansible=master"
+  evars: "-e skip_sanity_checks=true"
```
richm: need to tell origin install to use journald - add `-e openshift_docker_log_driver=journald -e openshift_docker_options=--log-driver=journald`
likewise for the json-file jobs - add `-e openshift_docker_log_driver=json-file -e openshift_docker_options=--log-driver=json-file`

sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml
stevekuznetsov: outstanding

sjb/config/test_cases/test_pull_request_openshift_ansible_logging.yml
```diff
@@ -1,3 +1,4 @@
 ---
 parent: 'test_cases/test_branch_openshift_ansible_logging.yml'
-overrides: {}
\ No newline at end of file
+overrides:
+  evars: "-e skip_sanity_checks=true"
```
stevekuznetsov: So the parent has ` +  evars: "-e skip_sanity_checks=true -e openshift_disable_check=* -e openshift_install_examples=false" ` so this child will remove some of those? why?

jcantrill:
close in favor of #1220

PR #1212, opened 2018-03-09
Make image tag logic more consistent, stop passing tag explicitly

stevekuznetsov:

/cc @sdodson @deads2k @smarterclayton


smarterclayton:
This breaks openshift/sdn, which needs to use openshift/node, which needs to get the openshift_image_tag.  We still have to set openshift_image_tag no matter what.  We can't parse `oreg_url` because there is no guarantee that the tag is actually a tag.

michaelgugino:
@smarterclayton openshift_image_tag is always set by openshift_version role if it's not provided.  We set openshift_image_tag based on openshift_release (query registry for latest tag referred to by the v3.x tag).  If openshift_release and openshift_image_tag is not set, we query the registry for the latest available tag and use the numeric version of that tag.

PR #1214, opened 2018-03-09
remove repo sync override

bparees:


sjb/config/test_cases/test_pull_request_jenkins_images.yml
```diff
@@ -1,6 +1,4 @@
 ---
 parent: 'test_cases/test_branch_jenkins_images.yml'
-overrides:
-  sync_repos:
-    - name: "jenkins"
-      type: "pull_request"
```
stevekuznetsov: erm you still need to sync this as a PR so you want to add s2i to this

gabemontero:
The test_branch_jenkins_*_plugin jobs should sync the jenkins image repo as well

don't need the s2i repo for those

sjb/config/test_cases/test_pull_request_jenkins_images.yml
bparees: so does that mean https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_pull_request_wildfly_images.yml is totally wrong?

bparees:
> The test_branch_jenkins_*_plugin jobs should sync the jenkins image repo as well

@gabemontero let's discuss that elsewhere.  I don't think your plugin tests jobs are building a jenkins image from the local repo, they are just extending the jenkins image on docker hub to override the plugin.

bparees:
@stevekuznetsov updated, but now i'm wondering if https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_pull_request_wildfly_images.yml is totally wrong?

gabemontero:
On Fri, Mar 9, 2018 at 1:38 PM, Ben Parees <notifications@github.com> wrote:

> The test_branch_jenkins_*_plugin jobs should sync the jenkins image repo
> as well
>
> @gabemontero <https://github.com/gabemontero> let's discuss that
> elsewhere. I don't think your plugin tests jobs are building a jenkins
> image from the local repo, they are just extending the jenkins image on
> docker hub to override the plugin.
>
You are correct - I'm good

We'll worry about if / when we ever have to tweak the image in a way that
extension won't be sufficient for testing

stevekuznetsov:
Yeah it is totally wrong. I thought it was using the new sync and it is not. Ugh.

stevekuznetsov:
Let me just do the sync migration. ugh.

PR #1222, opened 2018-03-10
push out the new agent images

bparees:


bparees:
@gabemontero fyi, another necessary step I forgot about.  

so
1) i had to create the new repos (https://hub.docker.com/r/openshift/jenkins-agent-maven-35-centos7/ and https://hub.docker.com/r/openshift/jenkins-agent-nodejs-8-centos7/)
2) had to update the CI job to push out the new images

however the unbound variable issue is still going to be a problem until @stevekuznetsov can tell us what is up w/ that.

PR #1225, opened 2018-03-10
Sync master in push jobs so they can be periodic

stevekuznetsov:


smarterclayton:
Lgtm

PR #1236, opened 2018-03-13
always use latest logging image version

richm:
The images are always built and tagged with `latest` before running
logging install and CI, so just hard code the use of `latest` for
all of the logging images.


stevekuznetsov:
The worry with this approach is that if the images didn't get built etc they are just pulled and it's going to lead to super difficult debugging when you can't figure out which version you actually used. So a better approach is to tag them appropriately after the build. We use the commit SHA for Origin so we know the images don't exist anywhere else.

PR #1239, opened 2018-03-14
json-file: Use logging images with release_commit tag for security

richm:
The hack/build-images.sh script uses `os::build::image` which tags
each build with both `latest` and the commit hash.  When installing
the logging images, use the commit hash as the logging_image_version
to avoid any conflicts with latest or other extant image versions.


richm:
@jcantrill @ewolinetz

PR #1240, opened 2018-03-14
Drop extended_networking_minimal

danwinship:
With the merge of https://github.com/openshift/openshift-ansible/pull/7373 we now seem to be successfully using the NetworkPolicy in the GCE tests (eg https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/test_branch_origin_extended_conformance_gce/2669/), so extended_networking_minimal is pretty redundant, and since it's currently disabled anyway due to dind-related flaking, we should just kill it. (We still need the full extended_networking test to exist, so we can test all three networking plugins when needed, but we should eventually port that to use ansible rather than dind.)

I don't know the details of how CI works so I don't know if this PR is right at all...


stevekuznetsov:
@danwinship can you also delete references to it in `openshift/release/cluster/ci/config/prow/config.yaml`?

stevekuznetsov:
@danwinship can you rebase and I'll merge?

danwinship:
@stevekuznetsov done

PR #1242, opened 2018-03-14
peg logging 3.6 to image latest

jcantrill:


richm:
what about sjb/config/test_cases/test_pull_request_origin_aggregated_logging_journald.yml ?

sjb/config/test_cases/test_pull_request_origin_aggregated_logging_journald_36.yml
```diff
@@ -5,3 +5,4 @@ overrides:
     - "openshift,origin=release-3.6"
     - "openshift,aos-cd-jobs=master"
     - "openshift,openshift-ansible=release-3.6"
+  evars: "-e openshift_logging_image_version=latest"
```
stevekuznetsov: you are overriding whatever else was in this var from the parent job -- is that correct?

richm: no, it is not correct

This is the parent: https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_pull_request_origin_aggregated_logging_journald.yml
imo it should not set these evars here:
```
 evars: "-e skip_sanity_checks=true -e openshift_disable_check=* -e openshift_docker_log_driver=journald -e openshift_docker_options=--log-driver=journald -e openshift_install_examples=false"
```
because that does not allow the child to override.  Instead, these should be the default values in the parent job.  The parent job should check EXTRA_EVARS to see if any of these are set, and use the settings from there instead.

jcantrill:
This worked for me on 

https://github.com/openshift/origin-aggregated-logging/pull/992

@stevekuznetsov please merge and push out

PR #1249, opened 2018-03-16
Fix image registry jobs

bbguimaraes:
This may not be the best solution to the problem, but this issue is blocking
the image registry team:

https://github.com/kubernetes/test-infra/pull/7151#pullrequestreview-103766727

The problem here is that a parent job of `test_branch_image_registry_extended`,
`origin_release_with_ecosystem.yml`, adds `openshift,image-registry=master` to
`sync`.  But this is the repository that triggers the job, so it causes the
validation added on the PR mentioned above to fail.

This PR works around the issue by overriding `sync` and removing
`openshift,image-registry=master` from it.  Not the ideal solution, but it
unblocks the folks from image registry while we work on a better one.

@kargakis, @stevekuznetsov: can you verify this?

### DIFF
--- a/sjb/config/test_cases/test_branch_image_registry_extended.yml
+++ b/sjb/config/test_cases/test_branch_image_registry_extended.yml
@@ -1,5 +1,11 @@
 ---
 parent: 'common/test_cases/origin_built_installed_release.yml'
+overrides:
+  sync:
+    - "openshift,openshift-ansible=master"
+    - "openshift,kubernetes-metrics-server=master"
+    - "openshift,origin-web-console-server=master"
+    - "openshift,aos-cd-jobs=master"
 extensions:
   actions:
     - type: "script"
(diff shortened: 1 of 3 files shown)


kargakis:
Generate the diff so we can figure out whether the --repo call-out to image-registry is actually dropped?

bbguimaraes:
Done.

kargakis:
Ship it!

bbguimaraes:
@kargakis: added explanation to the commit message, removed WIP.

kargakis:
> (edit: I didn't know that removed the LGTM, sorry =)

Yeah, if you repush your changes, lgtm drops. Still

/lgtm

kargakis:
Push out to Jenkins, I am merging

sjb/config/test_cases/test_branch_image_registry_extended.yml
```diff
@@ -1,5 +1,11 @@
 ---
 parent: 'common/test_cases/origin_built_installed_release.yml'
+overrides:
+  sync:
+    - "openshift,openshift-ansible=master"
```
stevekuznetsov: This is the correct approach, but we also want to list `openshift,origin` in here

PR #1250, opened 2018-03-16
Remove option from journalctl generated artifact

cevich:
I must have been reading a Fedora man-page when I added the command
originally.  The ``--no-hostname`` option isn't supported by RHEL's
``journalctl``.


cevich:
***note***: This goof won't affect running jobs, only the contents of the ``system.journal`` log file.

cevich:
Thanks Steve :smiley:

PR #1251, opened 2018-03-16
Doubling kibana memory limit to help it start up more frequently

ewolinetz:
Currently Kibana starts up intermittently (even with additional `oc rollout latest` follow up commands).
This should help it start up more frequently as tested by @t0ffel 

PTAL @stevekuznetsov @richm


ewolinetz:
hopefully the ci machines aren't so constrained that 256 Mi causes issues... guess we'll see

PR #1265, opened 2018-03-21
Use clonerefs on all prow jobs

bbguimaraes:
@stevekuznetsov

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
+++ b/sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
@@ -1,16 +1,18 @@
 ---
-parent: 'common/test_cases/origin.yml'
+parent: 'common/test_cases/origin_minimal.yml'
 overrides:
   junit_analysis: False
-  timer: 'H H * * *'
   email:
     - decarr@redhat.com
     - gscrivan@redhat.com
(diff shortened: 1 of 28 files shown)


sjb/config/test_cases/test_branch_origin_extended_conformance_crio.yml
```diff
@@ -1,5 +1,5 @@
 ---
-parent: 'common/test_cases/origin.yml'
+parent: 'common/test_cases/origin_minimal.yml'
 overrides:
   junit_analysis: False
   timer: 'H H * * *'
```
bbguimaraes: Is this timer still needed?

stevekuznetsov: Do we have it as a postsubmit in `openshift/release`?

bbguimaraes: Yes, these jobs are all pre/post submits in their respective repositories.

stevekuznetsov: OK, then no timer needed here any longer

bbguimaraes: Removed.

PR #1270, opened 2018-03-23
Fix usage of cluster up to use --tag instead of --version

mfojtik:
We changed the flag in 3.10/master. I'm not sure if this job is being reused for <=3.9, if yes, we need to rethink this @soltysh  :-)


mfojtik:
@stevekuznetsov that was fast, how about 3.9 and 3.8 jobs?

mfojtik:
@pmorie @jhadvig and I think we should move the oc cluster up into bash script inside the git repo, instead of having it in job spec.

spadgett:
+1 to using a bash script in the repo, then we change what it does per branch if necessary

PR #1273, opened 2018-03-27
[WIP] Setup an Azure job

kargakis:
@pschiffe @kwoodson

### DIFF
--- a/sjb/config/common/test_cases/origin_release_install_azure.yml
+++ b/sjb/config/common/test_cases/origin_release_install_azure.yml
@@ -1,0 +1,190 @@
+---
+parent: 'common/test_cases/origin_release_with_ecosystem.yml'
+extensions:
+  sync:
+    - "openshift,release=master"
+  parameters:
+    - name: "CLUSTER_PROFILE"
+      description: "The subdirectory of openshift/release cluster/test-deploy to use for configuration of the cluster."
+      default_value: "gcp"
+    - name: "FOCUS"
+      description: "Literal string to pass to <code>--ginkgo.focus</code>"
+    - name: "OPENSHIFT_ANSIBLE_IMAGE"
+      description: >
+        The image to install the cluster with. If set, defaults to the value defined by the
+        <a href='https://github.com/openshift/release/blob/master/cluster/bin/local.sh'><code>cluster/bin/local.sh</code></a>
+        script.
+      default_value: "openshift/origin-ansible:latest"
+    - name: "ADDITIONAL_SKIP"
+      description: |-
+        Regular expression to filter additional tests from the conformance suite.
+        See also:
+        <div>
+        </div>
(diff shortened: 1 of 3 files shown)


sjb/config/common/test_cases/origin_release_install_azure.yml
```diff
@@ -0,0 +1,190 @@
+---
+parent: 'common/test_cases/origin_release_with_ecosystem.yml'
+extensions:
+  sync:
+    - "openshift,release=master"
+  parameters:
+    - name: "CLUSTER_PROFILE"
+      description: "The subdirectory of openshift/release cluster/test-deploy to use for configuration of the cluster."
+      default_value: "gcp"
```
kwoodson: is this shared? If this is the azure job then maybe this should be azure as its default. If this is shared then gcp might be ok

sjb/config/common/test_cases/origin_release_install_azure.yml
```diff
@@ -0,0 +1,190 @@
...
+        sudo systemctl restart docker
+
+        suffix='origin-${component}'
+        suffix="${suffix}:$( cat /data/src/github.com/openshift/aos-cd-jobs/ORIGIN_TAG )"
+        cd cluster/test-deploy/${CLUSTER_PROFILE}/
+        ../../bin/ansible.sh ansible-playbook \
+          -e "openshift_test_repo=${location_url}" \
+          -e debug_level=5 \
+          -e "oreg_url=${OS_PUSH_BASE_REGISTRY}${OS_PUSH_BASE_REPO}${suffix}" \
+          playbooks/gcp/openshift-cluster/launch.yml \
```
kwoodson: should this be azure?  This should probably be a call to acs-engine or our cluster create, right?

sjb/config/common/test_cases/origin_release_install_azure.yml
```diff
@@ -0,0 +1,190 @@
...
+        cp admin.kubeconfig /tmp/cluster-admin.kubeconfig
+  post_actions:
+    - type: "script"
+      title: "deprovision test cluster"
+      repository: "release"
+      timeout: 900
+      script: |-
+        trap 'exit 0' EXIT
+        cd cluster/test-deploy/${CLUSTER_PROFILE}
+        ../../bin/ansible.sh ansible-playbook playbooks/gcp/openshift-cluster/deprovision.yml
```
kwoodson: We do not have a deprovision yet.  This should probably be:
```
az group delete --resource-group <resource group name>
```

pschiffe: we do have `deprovision.yml` which deletes resource group, so I think this is fine, but it should be `make down`

kargakis:
@pschiffe let me know if you need any help

/close

PR #1276, opened 2018-03-28
Handle syntax differences for "oc cluster up"

jpeeler:
Used the PULL_BASE_REF environment variable to select a different syntax
based on the branch being tested.


jpeeler:
(Note: the new job that I'll soon be adding will use parent and avoid duplication.)

PR #1278, opened 2018-03-29
WIP: new catalog repo at openshift/service-catalog

jpeeler:
Do not merge - has not been tested and generated files not produced.


stevekuznetsov:
How is the pull request file different from the branch file?

jpeeler:
I assume you're talking about the other PR made in the release repo. I was following a list of things I thought was required to shift over to the new repo. After getting feedback on the other review, I see that really there's not much to do there.

stevekuznetsov:
No, I mean in this PR you are touching two files which look like copies of each other -- what's the diff?

stevekuznetsov:
Right, we have the `parent` mechanism to DRY jobs out in the framework, might be nice to see if we can use that here if the config is literally the same for the two jobs other than the trigger

jpeeler:
The PR job doesn't run the integration tests. Though I suppose adding them would be better? Then I wonder if a post-submit job is even needed at that point.

jpeeler:
Refer to my other PR. Want to start debugging the new job way in advance of the switch over.

PR #1279, opened 2018-03-29
[3.9] Only test with openshift_release

michaelgugino:
We don't need to test unreleased versions after we have
GA release.

This commit removes image tag and pkg versions and uses
only openshift_release.


michaelgugino:
(pushed)

michaelgugino:
Pushed reversion commit.  Closing.

PR #1283, opened 2018-04-04
Always fall back to the default GCE service account and project

ingvagabund:
Some jobs change the GCE credentials. However, given all artefacts are uploaded to the same GCS bucket, we need to flip back to the default credentials in case the job fails or the JJ creator forget to switch back.


ingvagabund:
@stevekuznetsov PTAL

stevekuznetsov:
@ingvagabund you can push this out

ingvagabund:
@stevekuznetsov all jobs under `sjb/generated` push-updated.

PR #1286, opened 2018-04-05
[WIP] acs-engine build.

kwoodson:


kwoodson:
After chatting with @stevekuznetsov we might take a different direction here.  

Steve suggested doing this work through a build and deployment in openshift.  I'm experimenting with that at the moment and will report back.

stevekuznetsov:
This is a silly thing but we check in cluster resources into `github.com/openshift/release/projects`, not this repo.

PR #1287, opened 2018-04-06
 Don't call out to gcloud in the test VMs 

stevekuznetsov:
Revert "Always fall back to the default GCE service account and project"

This reverts commit e83b9a21662db8721c494616646ae3e63afc4ce0.

---

Don't call out to gcloud in the test VMs

---

@ingvagabund didn't review carefully, didn't realize you were doing this in the test VM ... we don't even have `gcloud` in that VM. Why was this necessary? The upload tool should read and login using the auth config given, no extra steps necessary.

### DIFF
--- a/sjb/config/common/test_cases/origin_minimal.yml
+++ b/sjb/config/common/test_cases/origin_minimal.yml
@@ -94,11 +94,6 @@ post_actions:
     script: |-
       trap 'exit 0' EXIT
       if [[ -n "${JOB_SPEC:-}" ]]; then
-        # Set the original GCE SA and project to openshift-gce-devel that can upload results to GCS bucket
-        # in case different SA was used to provision/teardown GCE cluster
-        gcloud auth activate-service-account --key-file /data/credentials.json
-        gcloud config set project openshift-gce-devel
-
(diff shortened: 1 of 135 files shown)


ingvagabund:
@stevekuznetsov which test VM you have in mind? Or, what the `test_` prefix stands for then? In case I switch to a SA that has no write access to the GCE bucket, I can no longer upload results into the bucket.

stevekuznetsov:
I don't think any of our AWS AMIs have `gcloud` installed. Maybe you meant if you changed the SA on the Jenkins master (`host_script`?)

ingvagabund:
Ahhh, I see. I need to revisit the job again.

ingvagabund:
Yeah, I need to change the SA back to the default one on the testing machine. So maybe:
```sh
# Set the original GCE SA and project to openshift-gce-devel that can upload results to GCS bucket
# in case different SA was used to provision/teardown GCE cluster
if hash gdate 2>/dev/null; then
    gcloud auth activate-service-account --key-file /data/credentials.json
    gcloud config set project openshift-gce-devel
fi
```
? The `hash` is part of the `bash` package so it should be everywhere.

It's important to run the gcloud command no matter if the job succeeds or fails.

ingvagabund:
Cool. The data are uploaded into the GCS bucket via docker container so the SA is set to the default one.

PR #1293, opened 2018-04-09
Azure image build job

pschiffe:
@kwoodson @jim-minter


pschiffe:
Currently broken b/c:
```
TASK [openshift_azure : Create an image from VM] *******************************
Monday 09 April 2018  20:02:01 +0000 (0:00:02.804)       0:15:00.810 ********** 
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: list object has no element 0\n\nThe error appears to have been in '/usr/share/ansible/openshift-ansible/roles/openshift_azure/tasks/create_image_from_vm.yml': line 32, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Create an image from VM\n  ^ here\n"}
```

pschiffe:
This is working now.

sjb/config/test_cases/azure_build_node_image_rhel.yml
```diff
@@ -0,0 +1,78 @@
...
+      popd
+
+  - type: "script"
+    title: "clean up resources"
+    script: |-
+      pushd release/cluster/test-deploy/${CLUSTER_PROFILE}
+      TYPE=$CLUSTER_PROFILE ../../bin/ansible.sh ansible-playbook \
+        -e "openshift_azure_image=openshift-gi-${INSTANCE_PREFIX}" \
+        -e "openshift_azure_clusterid=magentaciimages" \
+        playbooks/${CLUSTER_PROFILE}/openshift-cluster/remove_instance.yml
```
kargakis: Is this cleaning up the image that is built? How are we going to GC old base images?

pschiffe: It's cleaning only the instance and it's resources (network interface, etc). The image stays (that's the point of the job). But we don't clean older images here, or in any other way.

kargakis: ack

PR #1299, opened 2018-04-11
Set openshift_image_tag explicitely

vrutkovs:
openshift-ansible uses `version` output to detect the image tag if its
not set. It can be 3.x.0, while CI tags only 3.x. CI should pass
openshift_image_tag to avoid this.

This should resolve https://github.com/openshift/openshift-ansible/pull/7919#issuecomment-380572925

PTAL @sdodson @stevekuznetsov

### DIFF
--- a/sjb/config/common/test_cases/origin_built_installed_release_39.yml
+++ b/sjb/config/common/test_cases/origin_built_installed_release_39.yml
@@ -58,6 +58,7 @@ extensions:
                          --connection local         \
                          --inventory sjb/inventory/ \
                          -e openshift_deployment_type=origin  \
+                         -e openshift_image_tag="$( cat ./ORIGIN_RELEASE )" \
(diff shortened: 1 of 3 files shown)


sdodson:
@michaelgugino can you take a look at this?

sjb/config/common/test_cases/origin_built_installed_release_39.yml
```diff
@@ -58,6 +58,7 @@ extensions:
                          --connection local         \
                          --inventory sjb/inventory/ \
                          -e openshift_deployment_type=origin  \
+                         -e openshift_image_tag="$( cat ./ORIGIN_RELEASE )" \
```
ingvagabund: Does it make sense to update `sjb/config/common/test_cases/origin_built_installed_release.yml` as well?

vrutkovs: We haven't seen this happening on master yet, so I'd rather not touch working parts.

PR #1311, opened 2018-04-18
Adds correct testing for catalog and origin 3.9

jpeeler:
Depends on https://github.com/openshift/release/pull/790.


jpeeler:
@stevekuznetsov good to go, thanks in advance

PR #1316, opened 2018-04-20
Fix 'generated' artifact order for all crio jobs

cevich:
Formally these commands were part of the 'generated:' hash/dictionary.
Unfortunately b/c hashes are un-ordered, some unknown problem on Fedora
sometimes causes the ``pvs, vgs, lvs`` command to render the VM
inaccessable.  If by chance that happened to occur early/first, it meant
all the subsequent commands would fail, and produce empty output.

However, because some of the commands are expected to produce
empty-output (e.g. ``ausearch``), order becomes significant in cases
of result ambeguity.  For example, were there actually no AVC denials,
or was the host inaccessable?  Determining that requres trudging through
the console output, which nobody enjoys.

Fix this situation for all crio jobs, by moving all the former
`generated:' commands and output files, into a ``script``.  Also,
move the problematic ``pvs, vgs, lvs`` command to the end of the list.
This removes ambiguity and will allow easier troubleshooting of the
Fedora problem.

### DIFF
--- a/sjb/config/common/test_cases/crio.yml
+++ b/sjb/config/common/test_cases/crio.yml
@@ -86,6 +86,23 @@ actions:
       fi
 post_actions:
   - type: "host_script"
+    title: "Produce and Retrieve artifacts from sequence of commands"
+    timeout: 300
+    script: |-
+      trap 'exit 0' EXIT
+      DESTDIR="${WORKSPACE}/artifacts/generated"
+      set +e
+      SSH="ssh -F ${WORKSPACE}/.config/origin-ci-tool/inventory/.ssh_config openshiftdevel"
+      mkdir -p "$DESTDIR"
+      $SSH 'sudo yum list installed' &> "$DESTDIR/installed_packages.log"
+      $SSH 'sudo ausearch -m AVC -m SELINUX_ERR -m USER_AVC' &> "$DESTDIR/avc_denials.log"
+      $SSH 'sudo journalctl _PID=1 --no-pager --all --lines=all' &> "$DESTDIR/pid1.journal"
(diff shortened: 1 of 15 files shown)


stevekuznetsov:
Lets see what this looks like intesting

cevich:
@stevekuznetsov Yep, that's the plan.  Holding off in case there were any change-suggestions.  This isn't urgent IMHO, but figuring out what's breaking on Fedora will help me sleep slightly better at night :grinning:

cevich:
Testing with https://ci.openshift.redhat.com/jenkins/job/testing_pr1316_ami_fedora/

cevich:
@stevekuznetsov @rhatdan @mrunalp PTAL, this is ready to go.

Test job (with CRI-O playbook disabled) is here: 
https://ci.openshift.redhat.com/jenkins/job/testing_pr1316_ami_fedora/13/s3

Fixes generating/collecting logs in a random-order, where a collection command (vgs, lvs, or pvs) sometimes blocks subsequent collection commands.

Also fixes the problem of a failed a command generating an empty file w/o any other accountability (Jenkins console also useless).  With these changes, the log file will contain any error messages produced during collection.  Failure of any collection command will not impact subsequent collection commands.

cevich:
@runcom @mrunalp PTAL, this is ready IMHO.

runcom:
lgtm but I'd love @stevekuznetsov to ack as well

cevich:
Recap:  This fixes the order of running artifact generating commands.  Currently the order is 'undefined' and In the past some of the commands result in hanging the test-host.  Preventing further collection of debugging data.

cevich:
@stevekuznetsov could you PTAL?  Test job w/ positive results: https://ci.openshift.redhat.com/jenkins/job/testing_pr1316_ami_fedora

PR #1322, opened 2018-04-24
rework azure jobs

jim-minter:


jim-minter:
@kargakis ptal

- fixes up the jobs to work with https://github.com/openshift/openshift-ansible/pull/8121 (now merged)
- condenses the 4 almost identical image build jobs into 1
- removes the master pr & branch tests because they just don't work yet.  we can readd them when we're closer
- other small fixes

cf https://github.com/openshift/release/pull/818

This PR doesn't include the definition of https://ci.openshift.redhat.com/jenkins/job/azure_build_images, how should we deal with that?

kargakis:
> This PR doesn't include the definition of https://ci.openshift.redhat.com/jenkins/job/azure_build_images, how should we deal with that?

Unclear what that job is doing. How is it different from the generic build job you introduce here? 

/cc pschiffe

kargakis:
Also there is no way to pass environment variables from the prow config to Jenkins so you still need 4 distinct Jenkins jobs. cc @stevekuznetsov

jim-minter:
@kargakis ptal again

PR #1325, opened 2018-04-25
release-3.9 playbook: set openshift_image_tag to latest to copy it to

vrutkovs:
This should resolve failing system-container tests for 3.9

PTAL @sdodson @michaelgugino


vrutkovs:
Nope, that doesn't work

PR #1327, opened 2018-04-26
openshift-ansible PRs don't need to build Origin

vrutkovs:
This commit adds a common testcase for openshift-ansible. Instead of built Origin it would use branch-specific RPMs to test PRs.

Only test_pull_request_openshift_ansible_extended_conformance_install.yml  for master is 
updated for now

### DIFF
--- a/sjb/config/test_cases/test_pull_request_openshift_ansible.yml
+++ b/sjb/config/test_cases/test_pull_request_openshift_ansible.yml
@@ -1,0 +1,140 @@
+---
+parent: 'common/test_cases/origin_minimal.yml'
+overrides:
+  junit_analysis: False
+extensions:
+  sync:
+    - "openshift,origin=master"
+    - "openshift,aos-cd-jobs=master"
+  actions:
+    - type: "script"
+      title: "build an openshift-ansible release"
+      repository: "openshift-ansible"
+      script: |-
+        tito_tmp_dir="tito"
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
No, we need Origin and OpenShift Ansible to run against latest HEAD of each other. This will become much less painful in the next few weeks as we start using `ci-operator` and caching builds on our cluster.

/close

vrutkovs:
>No, we need Origin and OpenShift Ansible to run against latest HEAD of each other.

Isn't that repo actually contains the latest built versions of Origin? Also, there is no benefit in the CI if the issues there are not reproducible - and the only way we can reproduce it is using this (or any other) RPM repo.

>This will become much less painful in the next few weeks as we start using ci-operator and caching builds on our cluster.

Its awesome, but its safe to revert this PR in a few weeks, as meanwhile 3.7 PRs (and master yesterday) are blocked by Origin builds failing

stevekuznetsov:
> as meanwhile 3.7 PRs (and master yesterday) are blocked by Origin builds failing

And similarly every time openshift-ansible `HEAD` breaks, Origin merge queue falls over. This is what we want until we have a sensible ratcheting system for testing against each other. We've dealt with this for a little over a year now and we should be able to deal with it for a little while longer. If you have bandwidth to spare on trying to fix this I'm happy to explain our plans and show how you can help, but I do not think that churn on jobs like this is the appropriate solution at this time.

PR #1329, opened 2018-04-27
Add cluster_operator e2e test

csrwng:
Adds e2e tests for Cluster Operator

### DIFF
--- a/sjb/config/test_cases/test_branch_cluster_operator_e2e.yml
+++ b/sjb/config/test_cases/test_branch_cluster_operator_e2e.yml
@@ -1,0 +1,115 @@
+---
+parent: 'common/test_cases/origin_built_release.yml'
+overrides:
+  sync: []
+extensions:
+
+  actions:
+    - type: "script"
+      title: "setup prerequisites"
+      script: |-
+        sudo pip install openshift
+        sudo pip install --upgrade awscli
+        cd /tmp
+        wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+        sudo yum -y install epel-release-latest-7.noarch.rpm
+        sudo yum --disablerepo="*" --enablerepo=epel -y update ansible
+        sudo yum --disablerepo="*" --enablerepo=epel -y install jq
+        sudo env GOPATH=/data go get -u github.com/cloudflare/cfssl/cmd/...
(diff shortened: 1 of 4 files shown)


csrwng:
@dgoodwin fyi
The test is working but failing for right now because of the cert mismatch on the external ELB
It's also pending proper credentials from @dobbymoodge 
Currently it simply uses the ones from jenkins

sjb/config/test_cases/test_branch_cluster_operator_e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+        sudo yum -y install epel-release-latest-7.noarch.rpm
+        sudo yum --disablerepo="*" --enablerepo=epel -y update ansible
+        sudo env GOPATH=/data go get -u github.com/cloudflare/cfssl/cmd/...
+        mkdir -p /tmp/.aws
+        mkdir -p /tmp/.ssh
+    - type: "host_script"
+      title: "transfer aws credentials to remote host"
+      script: |-
+        scp -F ./.config/origin-ci-tool/inventory/.ssh_config ~/.aws/clusteroperator-credentials openshiftdevel:/tmp/.aws/credentials
+        scp -F ./.config/origin-ci-tool/inventory/.ssh_config ~/.ssh/devenv.pem openshiftdevel:/tmp/.ssh/libra.pem
```
dobbymoodge: Isn't the SSH pubkey added to the instance by the cloud provider?

dobbymoodge: Also, maybe for these instances we should use an ad-hoc or less widely shared keypair. Thoughts?

csrwng: @dobbymoodge the public key is added to the instance, but not the private one.
We need the private key in the test vm so *it* can ssh into the target cluster machines.

If you want to create a new key just for this, that's great. I'd only ask that you let me know the name and to place the private key somewhere on the jenkins filesystem so the job can get to it.

csrwng:
@dobbymoodge I've now updated the job to create a keypair and use that for the target cluster's machines.

csrwng:
@dobbymoodge updated to remove bash calls with sudo, got one passing test. It's ready to ship! :)

PR #1336, opened 2018-05-02
system-containers 3.9: run additiona docker tag for 3.9 branch only

vrutkovs:
Fixes https://github.com/openshift/origin/issues/19582

PTAL @stevekuznetsov @smarterclayton

### DIFF
--- a/sjb/config/common/test_cases/origin_built_installed_release.yml
+++ b/sjb/config/common/test_cases/origin_built_installed_release.yml
@@ -45,11 +45,6 @@ extensions:
           sudo atomic pull --storage ostree docker:openshift/node:$( cat ./ORIGIN_TAG )
         fi
     - type: "script"
-      title: "tag openshift/origin for system containers"
-      repository: "aos-cd-jobs"
-      script: |-
-        sudo docker tag openshift/origin:$( cat ./ORIGIN_TAG ) openshift/origin:$( cat ./ORIGIN_TAG ).$(cat ORIGIN_PKG_VERSION | grep -Po "^\-\w\.\w\.(\w)" | cut -d '.' -f 3) || true
-    - type: "script"
       title: "origin prerequisites"
       repository: "aos-cd-jobs"
       script: |-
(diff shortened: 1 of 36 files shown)


sjb/config/common/test_cases/origin_built_installed_release.yml
```diff
@@ -48,7 +48,9 @@ extensions:
       title: "tag openshift/origin for system containers"
       repository: "aos-cd-jobs"
       script: |-
-        sudo docker tag openshift/origin:$( cat ./ORIGIN_TAG ) openshift/origin:$( cat ./ORIGIN_TAG ).$(cat ORIGIN_PKG_VERSION | grep -Po "^\-\w\.\w\.(\w)" | cut -d '.' -f 3) || true
+        if [[ "${PULL_BASE_REF}" == "release-3.9" ]]; then
+          sudo docker tag openshift/origin:$( cat ./ORIGIN_TAG ) openshift/origin:$( cat ./ORIGIN_TAG ).$(cat ORIGIN_PKG_VERSION | grep -Po "^\-\w\.\w\.(\w)" | cut -d '.' -f 3) || true
```
smarterclayton: We already have a 39 fork of this job.  Don’t add this here.

vrutkovs: Right, updated `_release_39` instead

sjb/config/common/test_cases/origin_built_installed_release.yml
```diff
@@ -44,11 +44,6 @@ extensions:
           sudo atomic pull --storage ostree docker:openshift/openvswitch:$( cat ./ORIGIN_TAG )
           sudo atomic pull --storage ostree docker:openshift/node:$( cat ./ORIGIN_TAG )
         fi
-    - type: "script"
-      title: "tag openshift/origin for system containers"
-      repository: "aos-cd-jobs"
-      script: |-
-        sudo docker tag openshift/origin:$( cat ./ORIGIN_TAG ) openshift/origin:$( cat ./ORIGIN_TAG ).$(cat ORIGIN_PKG_VERSION | grep -Po "^\-\w\.\w\.(\w)" | cut -d '.' -f 3) || true
```
ingvagabund: What about to keep it here? We will have to same problem once `origin_built_installed_release_310.yml` is forked. It would make more sense to copy-paste the `origin_built_installed_release.yml` -> `origin_built_installed_release_310.yml` and just remove the `|| true` part in the process. Plus adding comment here saying "Remove the `|| true` once forked into `origin_built_installed_release_3N.yml`".

@smarterclayton WDYT?

PR #1339, opened 2018-05-03
Bump k8s CI jobs to go1.10.1

ingvagabund:
The minimal required version of Go in k8s was bump to 1.10.1. Which caused the CI to fail for the past 9 days.


ingvagabund:
@stevekuznetsov PTAL

ingvagabund:
@sjenning FYI

PR #1347, opened 2018-05-04
Fix crio rpm installs

sdodson:


sdodson:
Testing the new job here https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-ansible/8262/test_pull_request_openshift_ansible_extended_conformance_install_crio_rpm/1409/

sjb/config/test_cases/test_branch_origin_extended_conformance_crio_rpm.yml
```diff
@@ -38,6 +38,35 @@ extensions:
...
+            major="$( echo "${branch}" | grep -Po "(?<=release-)[0-9](?=\.[0-9])" )"
+            minor="$( echo "${branch}" | grep -Po "(?<=release-[0-9]\.)[0-9]" )"
+            ;;
+          enterprise-* )
+            major="$( echo "${branch}" | grep -Po "(?<=release-)[0-9](?=\.[0-9])" )"
+            minor="$( echo "${branch}" | grep -Po "(?<=release-[0-9]\.)[0-9]" )"
+            ;;
+        esac
+        previous="$(( minor -1))"
+        sudo yum-config-manager --enable "rhel-7-server-ose-${major}.${minor}"
```
cevich: @sdodson you know more than I: Are all the above complications necessary?  Couldn't we simply rely on yum/rpm to always pick the latest version available.  In other words, just:  ``yum-config-manager --enable "rhel-7-server-*"`` or does that cause problems?

sdodson: We use the same jobs to install 3.9, 3.10, and future releases so we need to make sure we enable only the correct repos.  AFAIK cri-o 1.10 is not guaranteed to be compatible with Origin 3.9

cevich: OIC...thx for the explanation.  It seems like a lot of extra maintenance work, given the total number of jobs here (not just CRI-O).  Is this section at all common with other jobs?  (Just trying to think if I could help out writing some shared script or table or something)

cevich:
So it seems the issue here is simply the lack of a 3.10 repository file.  Is there something I can do/look at to help move a solution for that forward?

sdodson:
I think the commit I just pushed here and the change in https://github.com/openshift/origin-ci-tool/pull/170 should make this work

sjb/config/test_cases/test_branch_origin_extended_conformance_crio_rpm.yml
```diff
@@ -38,6 +38,36 @@ extensions:
...
+        sudo yum-config-manager --disable origin-deps-rhel7\* || true
+        sudo yum-config-manager --disable rhel-7-server-ose\* || true
+
+        branch="${PULL_BASE_REF:-"master"}"
+        case "${branch}" in
+          master )
+            # 3.10
+            major=3
+            minor=10
+            ;;
```
cevich: This is the case I'm most worried about, maintenance wise.  It means when 3.11 comes around, some human needs to remember to edit this.  Is there no, like, ``VERSION`` file or something that could be used?  How about pulling it from ``git describe`` output?

cevich: idea:

```bash
    tag_version=$(git describe --abbrev | sed -r 's/^openshift-ansible-([0-9\.]+)-.+/\1/')
    major=$(echo $tag_version | cut -d . -f 1)
    minor=$(echo $tag_version | cut -d . -f 2)
```

sdodson:
The AMI still lacks the repo definitions from https://github.com/openshift/origin-ci-tool/pull/170 checking on when we'll get a new AMI

cevich:
Oh, I thought it would have happened by now, good to know.  Thx for following up Scott.

sdodson:
Tagged a new image, this works now as evidenced by https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-ansible/8618/test_pull_request_openshift_ansible_extended_conformance_install_crio_rpm/2359/

PR #1350, opened 2018-05-07
Bump k8s CI jobs to go1.10.2

ingvagabund:
Because https://github.com/kubernetes/kubernetes/pull/63412

Wondering if it would be better to detect the version automatically, e.g. by running `grep "minimum_go_version=go" hack/lib/golang.sh | cut -d'=' -f2`. Still, even the `hack/lib/golang.sh` and the `minimum_go_version` variable can change over time so it's not bulletproof.


ingvagabund:
@stevekuznetsov PTAL

ingvagabund:
Jobs updated, finished successfully.

stevekuznetsov:
I think explicitly having to bump this is good, implicit bumps may make changes sneak in that are not expected.

ingvagabund:
Thanks

PR #1353, opened 2018-05-07
Skip cluster operator builds when deploying.

dgoodwin:
This is handled separately for the CI jobs due to path issues.

Will be needed before https://github.com/openshift/cluster-operator/pull/195 can merge. Should be a no-op in the meantime.

CC @csrwng


csrwng:
@dobbymoodge I tagged you because you have the power to merge :)

dobbymoodge:
@csrwng This passes tests? :)

csrwng:
Yes

PR #1358, opened 2018-05-07
catalog: add support for other branches

jpeeler:
This will allow reuse of this configuration for other branches.
Currently only master is being tested for now.

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_service_catalog.yml
+++ b/sjb/config/test_cases/test_branch_origin_service_catalog.yml
@@ -11,6 +11,13 @@ extensions:
     - name: "service-catalog"
   actions:
     - type: "script"
+      title: "checkout the correct branch of service-catalog"
+      repository: "service-catalog"
+      script: |-
+        if [[ "${PULL_BASE_REF:-master}" =! "master" ]]; then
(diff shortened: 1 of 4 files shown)


sjb/config/test_cases/test_branch_origin_service_catalog.yml
```diff
@@ -10,6 +10,13 @@ extensions:
     - name: "origin"
     - name: "service-catalog"
   actions:
+    - type: "script"
```
stevekuznetsov: You should migrate from `sync_repos:` to `sync:` and this will happen for you automatically

jpeeler: @stevekuznetsov will this work for both PRs and post submits?
```
sync:
- "openshift,service-catalog=${PULL_BASE_REF}"
```
If so, I can fix that up.

stevekuznetsov: That should work, yes

jpeeler: Actually, the syntax document says sync and sync_repos can't be mixed. That would mean that common/test_cases/origin.yml would need migrating first, right?

PR #1370, opened 2018-05-10
Fix failing install of openshift client

csrwng:
Fixes issue with latest ami


csrwng:
@stevekuznetsov ptal, will do a follow-up to switch to venv

PR #1378, opened 2018-05-14
Run "delete-yum-repos" during stage "upgrade: misc"

mbarnes:
Removes the `reposdir` override in `/etc/yum.conf`.


mbarnes:
@jupierce, for context the "delete-yum-repos" operation now cleans up after some `/etc/yum.conf` hacks that give us tighter control over yum repos during an upgrade (see https://github.com/openshift/openshift-ansible-ops/pull/4174).

PR #1380, opened 2018-05-15
switch email notifications for cluster up tests

bparees:


bparees:
@openshift/sig-developer-experience fyi
@mfojtik @deads2k ptal

PR #1394, opened 2018-05-21
Set system_images_registry=docker to enable prepull in openshift-ansible

vrutkovs:
Prerequisite for https://github.com/openshift/openshift-ansible/pull/8172

In order to enable prepull of control plane and node images CI should set 
system_images_registry flag to avoid pulling images from docker.io and use 
local images


vrutkovs:
Not required anymore, openshift-ansible would use that as a default

PR #1398, opened 2018-05-22
cluster-operator: increase timeouts for cluster creation and teardown

csrwng:


csrwng:
@dobbymoodge ptal

sjb/config/test_cases/test_branch_cluster_operator_e2e.yml
```diff
@@ -58,7 +58,7 @@ extensions:
     - type: "script"
       title: "Create Cluster"
       repository: "cluster-operator"
-      timeout: 2400
+      timeout: 5400
       script: |-
         # Allow some time for the API server to be registered
         sleep 20
```
dobbymoodge: LGTM, except this sleep smells bad to me. There's no way we can replace this with a poll + backoff?

If there's no better solution - or - if I'm being too nitpicky, I'm fine with the rest of the change.

dobbymoodge: referring to the `sleep 20`, that is

csrwng: I can come up with something better. Is it ok to fix in a separate PR? This one is just for bumping the timeouts on the create and deprovision

dobbymoodge: @csrwng yeah, however you prefer.

PR #1400, opened 2018-05-23
reformatted parameters in refresh-images

markllama:
Reformat the parameter section of the refresh-images Jenksinfile:
https://trello.com/c/CXTLjfpu

### DIFF
--- a/jobs/build/refresh-images/Jenkinsfile
+++ b/jobs/build/refresh-images/Jenkinsfile
@@ -52,24 +52,92 @@ node('openshift-build-1') {
 
     // Expose properties for a parameterized build
     properties(
-            [buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '720')),
-             [$class              : 'ParametersDefinitionProperty',
-              parameterDefinitions:
-                      [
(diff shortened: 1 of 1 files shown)


jobs/build/refresh-images/Jenkinsfile
```diff
@@ -52,24 +52,92 @@ node('openshift-build-1') {
...
+                        description: 'OSE Major Version',
+                        $class: 'hudson.model.ChoiceParameterDefinition',
+                        choices: "3",
+                        defaultValue: '3'
+                    ],
+                    [
+                        name: 'OSE_MINOR',
+                        description: 'OSE Minor Version',
+                        $class: 'hudson.model.ChoiceParameterDefinition',
+                        choices: "1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15",
```
jupierce: Using join('\n') model for choices parameters is fine here. Parameters are added/modified infrequently and testing them is relatively fast. Up to you whether you want to take that on in this refactor.

markllama: I looked at joining the numbers or generating a range. IN the case of "1..13" or what ever it ended up being less clear than the string list. If it gets much longer it might make more sense.

jupierce: Ultimately, the versions themselves will be in a pipeline library (sourced from a shared location). We can dial in the extent of the refactor to make here:
1. Leave it as-is.
2. Use join as a quick improvement.
3. Refactor to use BUILD_VERSION (combo of major.minor like build/ocp job) and join('\n') to reduce potential for human error. 

Generating a sequence will be problematic if we ever move to a 4.0, so hoping to be at shared library at that point.

markllama: I'm for leaving this alone right now.  It's out of scope and not hurting anything.
Re: if ever... deal with it then.  It's just a string.
Join isn't an improvement in this case, the string reads cleaner.

PR #1403, opened 2018-05-23
Disable cluster-operator e2e tests

staebler:
Disable cluster-operator e2e tests while transitioning from clust-op API to upstream cluster-api API.


csrwng:
These need to be disabled in openshift/release. Will open a PR there.
/close

PR #1420, opened 2018-05-28
Add alderaan pipeline scripts

chaitanyaenr:
This commit:
- Adds pipeline-script for each stage in the alderaan pipeline including osp, ocp installations,
  performance and scale tests. All the scripts assume that they are being executed in 
  openshift-qe jenkins.
- Adds sample property files used by the scripts.
- Adds a Jenkinsfile which ties up all the jobs.


jeremyeder:
Are you planning to merge into master?

PR #1429, opened 2018-05-30
Updating logging component images used

ewolinetz:
@stevekuznetsov  -- this inventory is only used for the origin-aggregated-logging jobs, correct?

This is to address the change in image name specification from https://github.com/openshift/openshift-ansible/pull/8489

Image prefix and image version are no longer honored for logging components

CC @jcantrill  @sdodson

### DIFF
--- a/sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml
+++ b/sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml
@@ -64,6 +64,8 @@ extensions:
                          -e openshift_logging_mux_allow_external=True                             \
                          -e openshift_logging_es_allow_external=True                              \
                          -e openshift_logging_es_ops_allow_external=True                          \
+                         -e oreg_url='openshift/origin-${component}:'"${release_commit:-latest}"                 \
+                         -e openshift_logging_elasticsearch_proxy_image="docker.io/openshift/oauth-proxy:v1.0.0" \
                          ${EXTRA_EVARS:-} \
                          ${playbook} \
                          ${logging_extras} \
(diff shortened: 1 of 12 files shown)


stevekuznetsov:
This is the global inventory used by everyone. Also, as noted before, we should not be using `:latest` as if there was some issue in building the images, they will get pulled from the hub, which is pretty explicitly _not_ the point of the jobs originally, as we want to integrate the projects at `master`, not from published images.

jcantrill:
@stevekuznetsov Is this where we should be making these changes? https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_openshift_ansible_logging_json_file.yml#L50

stevekuznetsov:
I'm not sure the changes are appropriate

ewolinetz:
@stevekuznetsov ack, we only want these changes to occur for the logging jobs -- so we can move them into `-e` options for our logging playbook executions.

Currently, we are seeing that the images in the DC/DS during the jobs where we build our images are pointing to dockerhub due to the oreg_url changes.

sdodson:
We could set oreg_url='openshift/origin-${component}:latest' as an alternative.

ewolinetz:
@sdodson that could work too... that, coupled with setting our elasticsearch proxy to still point to dockerhub

stevekuznetsov:
`-e` vars changes 

/lgtm

ewolinetz:
Do we need to push the generated files for this? I am prompted for a password, I don't remember that being the case in the past...

sdodson:
Yeah, they need to be pushed, I just pushed the latest though no way to tell if they required changes or not. README.md explains how to configure your credentials.

retry your test?

PR #1434, opened 2018-06-04
catalog: specify version to execute

jpeeler:
There's really no good way currently to quickly build images for
non-HEAD branches as hack/build-local-images.py doesn't support anything
other than using a latest tag. So given that this branch is highly
unlikely to have breaking catalog changes, just use oc cluster up
without the version tag and pull the latest images.

(This fixed the build for openshift/origin/pull/19880.)


jpeeler:
Actually, do not merge. I think this needs a better fix.

jpeeler:
This PR adds support to build the correct tags in origin: openshift/origin/pull/19918. Once it merges this can merged and CI should be well again.

jpeeler:
@stevekuznetsov this is good to go!

Issue #1449, opened 2018-06-08
build_ami job still requires openshift-ansible git information


markllama:
The *ami* stage of the ocp Jenkins job still wants to pass an OPENSHIFT_ANSIBLE_SOURCE_BRANCH value that is no longer present.

The build/aws_ami job must be updated to use the RPM built by the ocp and ose jobs.

markllama:
Proposed fix: https://github.com/openshift/aos-cd-jobs/pull/1451

PR #1404, opened 2018-05-24
Openshift ansible rpm

markllama:
Remove custom code for building the _openshift-ansible_ package.  The package has been added to the `oit` group RPM list.

This should be accepted and tested in sync with https://github.com/openshift/enterprise-images/pull/298

PR #1451, opened 2018-06-08
add definition for OPENSHIFT_ANSIBLE_SOURCE_BRANCH back to ocp call t…

markllama:
This change replaces a previously removed calculation of the OPENSHIFT_ANSIBLE_SOURCE_BRANCH value.  It was removed when the _openshift-ansible_ RPM build was moved to the oit  rpms:build step, but it is still needed for the build_ami step.


markllama:
Addresses issue https://github.com/openshift/aos-cd-jobs/issues/1449

jupierce:
This has a subtle but important issue without https://github.com/openshift/aos-cd-jobs/pull/1404#discussion_r190674022 . This PR's implementation requires openshift-ansible to cut the release branch simultaneously with ose. If openshift-ansible cuts before ocp first, you will build the RPM with the 3.X branch, but the AMI will build with openshift-ansible/master.

PR #1453, opened 2018-06-08
Avoid unnecessary builds in azure jobs

kargakis:
These images are built and published as part of separate processes anyway

@openshift/sig-azure


kwoodson:
@kargakis, I'll take your word for it.  If this is true, +:100: !

/lgtm

kargakis:
This should be fine to merge, will monitor our jobs but 99% we don't need these pushes.

PR #1455, opened 2018-06-11
Fixing parameter reorg from #1414

jupierce:


tbielawa:
More evidence we need to prioritize an automatic linter in these PRs

PR #1459, opened 2018-06-13
Pass buildlib into mail_success and get_rpm_changelog

adammhaile:
@tbielawa @markllama we missed that buildlib was still needed in get_rpm_changelog
Honestly I feel like we should make buildlib a global but there's probably a reason that I'm unaware of.

### DIFF
--- a/jobs/build/ocp/Jenkinsfile
+++ b/jobs/build/ocp/Jenkinsfile
@@ -159,7 +159,7 @@ def get_mirror_url(build_mode, version) {
     return "https://mirror.openshift.com/enterprise/enterprise-${version}"
 }
 
-def mail_success(version, mirrorURL, record_log) {
+def mail_success(buildlib, version, mirrorURL, record_log) {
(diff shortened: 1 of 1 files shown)


jupierce:
@adammhaile - buildlib loads a file from the filesystem that is not available until `checkout scm` runs which cannot be run until a `node(...)` block. So it can't be truly global. There are alternatives for sharing business logic which Jenkins provides without these pre-conditions. One of those I've detailed at the bottom of this card: https://trello.com/c/biPEXBBF

markllama:
The only required function is `buildlib.read_changelog(specfile_path)` A cleaner solution would be to have `get_rpm_changelog` just generate the _specfile_path_ and pass it up to the function that already has buidlib for execution.

For now this fix works.  I suggest +! accept and merge and create a card to re-factor.

adammhaile:
@jupierce I figured. Thanks for confirming.

Issue #1458, opened 2018-06-12
CI failure due to tito exception - test_pull_request_origin_extended_conformance_gce_39


jpeeler:
(test_pull_request_origin_extended_conformance_install_39 also fails in the same way)
Feel free to do any necessary retests on https://github.com/openshift/origin/pull/19965.
```
########## STARTING STAGE: BUILD AN OPENSHIFT-ANSIBLE RELEASE ##########
+ [[ -s /var/lib/jenkins/jobs/test_pull_request_origin_extended_conformance_gce_39/workspace/activate ]]
+ source /var/lib/jenkins/jobs/test_pull_request_origin_extended_conformance_gce_39/workspace/activate
++ export VIRTUAL_ENV=/var/lib/jenkins/origin-ci-tool/4b405957477ba1b70cfacd1cf43c6d41a605fc8e
++ VIRTUAL_ENV=/var/lib/jenkins/origin-ci-tool/4b405957477ba1b70cfacd1cf43c6d41a605fc8e
++ export PATH=/var/lib/jenkins/origin-ci-tool/4b405957477ba1b70cfacd1cf43c6d41a605fc8e/bin:/sbin:/usr/sbin:/bin:/usr/bin
++ PATH=/var/lib/jenkins/origin-ci-tool/4b405957477ba1b70cfacd1cf43c6d41a605fc8e/bin:/sbin:/usr/sbin:/bin:/usr/bin
++ unset PYTHON_HOME
++ export OCT_CONFIG_HOME=/var/lib/jenkins/jobs/test_pull_request_origin_extended_conformance_gce_39/workspace/.config
++ OCT_CONFIG_HOME=/var/lib/jenkins/jobs/test_pull_request_origin_extended_conformance_gce_39/workspace/.config
++ mktemp
+ script=/tmp/tmp.wgjWpOjkCM
+ cat
+ chmod +x /tmp/tmp.wgjWpOjkCM
+ scp -F /var/lib/jenkins/jobs/test_pull_request_origin_extended_conformance_gce_39/workspace/.config/origin-ci-tool/inventory/.ssh_config /tmp/tmp.wgjWpOjkCM openshiftdevel:/tmp/tmp.wgjWpOjkCM
[...]

PR #1461, opened 2018-06-13
force auto tag with release 9999

markllama:
Force testing auto-tags to be unique.
This addresses https://github.com/openshift/aos-cd-jobs/issues/1458 per @jupierce and @stevekuznetsov


stevekuznetsov:
Run `sjb/generate.sh`

stevekuznetsov:
Here goes nothing...

PR #1462, opened 2018-06-13
Tito tag 9999 - Added %dist back to tag string

markllama:
The tag string for openshift-ansible is derived from the release number and some arch information contained in %{dist}.  Restored the %{dist} substring to the tag.


jupierce:
@stevekuznetsov forgive me.

PR #1471, opened 2018-06-18
Increase timeout for cri-o e2e testing

cevich:
The timeout for the cri-o e2e tests is set at 4-hours.  However, the
default timeout for the shell process running Ansible (remotely) is
shorter.  On occasion, depending on many factors, this has caused
intermittant run failures.  Increase the shell timeout for all jobs
that execute the e2e test playbook.  This provides a much larger window
where the tests may fail more naturally.

### DIFF
--- a/sjb/config/test_cases/test_branch_crio_e2e_rhel.yml
+++ b/sjb/config/test_cases/test_branch_crio_e2e_rhel.yml
@@ -18,6 +18,7 @@ extensions:
       git checkout "${PULL_BASE_SHA}"
   - type: "script"
     title: "run the cri-o e2e tests"
+    timeout: "21600"  # 6 hours.  Playbook has shorter timeout for the test-task.
     script: |-
       ansible-playbook -vv --become  \
                        -i localhost, \
--- a/sjb/config/test_cases/test_pull_request_crio_ami_rhel.yml
+++ b/sjb/config/test_cases/test_pull_request_crio_ami_rhel.yml
@@ -55,6 +55,7 @@ extensions:
         oct package ami --stage=crio
     - type: "script"
       title: "run the cri-o tests"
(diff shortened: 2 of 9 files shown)


cevich:
@mrunalp @rhatdan PTAL.  This should resolve the intermittent e2e failures for CRI-O CI.  Note however, it brings along a side-effect:

In the (hopefully extremely rare) case the e2e test or something else is hung or exceptionally slow, the job will not conclude in failure until the new (6-hour) job-level timeout expires.

That said, the most likely situation is the 4-hour timeout set in the CRI-O Ansible playbook will take effect first.

rhatdan:
LGTM

cevich:
@rhatdan there are no tests for job-definition changes :face_with_head_bandage:   This is what I did to test and verify the fix:

1. Duplicate a crio job, but replace the call to ``ansible-playbook`` with a ``sleep 3h``.
2. Run the test job, it timed out after about 2-hours killing the ``sleep`` process.
3. Update the ``timeout`` key in the job definition to be 4-hours.
4. Run the test job, it completed successfully w/o killing ``sleep``.

@stevekuznetsov this is ready to merge, assuming it looks good to you also.

cevich:
@runcom PTAL

runcom:
looks ok but the issue we're seeing is, weirdly, just on release-1.10, so not sure what changed in master crio, it's not failing as often as release-1.0

BTW, LGTM

cevich:
@runcom thx.  Could easily be an edge-case, where 1.10 is just barely longer than the shell timeout.  In all cases though, for debugging purposes, it's much better if the ansible playbook has control of the specific-testing timeout.  Once the shell timeout (fixed here) hits, it kills off the ansible process, resulting in debugging data-loss :confused:

PR #1473, opened 2018-06-18
made name of oa_specfile variable consistent

markllama:
This change makes the name of the variable that holds the location of the RPM spec file for openshift ansible consistent for all uses.


markllama:
//reviewer tbielawa

PR #1493, opened 2018-06-23
Return `openshift-ansible` changelog to ocp success email

markllama:
This change retrieves the changelog for the `openshift-ansible` package, previously built, to to the success email for the build job.

It finds the brew task id from the build_rpm record, retrieves the task report web page, parses the package build ID
Finally it calls `brew buildinfo --changelog` to retrieve the changelog for the package and adds that to the
success email.

### DIFF
--- a/jobs/build/ocp/Jenkinsfile
+++ b/jobs/build/ocp/Jenkinsfile
@@ -1046,7 +1046,14 @@ Jenkins job: ${env.BUILD_URL}
 
             record_log = buildlib.parse_record_log(OIT_WORKING)
             oa_specfile = get_rpm_specfile_path(record_log, "openshift-ansible")
-            oa_changelog = get_changelog("openshift-ansible", record_log)
+	    try {
+		retry(3) {
+		    oa_changelog = get_changelog("openshift-ansible", record_log)
+		}
+	    } catch (changelog_err) {
+		echo "WARNING: unable to retrieve changelog for openshift-ansible: ${changelog_err}"
+		oa_changelog = "WARNING: unable to retrieve changelog for openshift-ansible.  See console output"
+	    }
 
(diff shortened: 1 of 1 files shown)


jobs/build/ocp/Jenkinsfile
```diff
@@ -162,6 +162,58 @@ def get_mirror_url(build_mode, version) {
...
+    }
+    build_record = rpm_builds[build_record_index]
+
+    // then get the task_id and task_url out of it
+    // task_id = build_record['task_id']
+    task_url = build_record['task_url']
+
+    // get the build ID from the web page
+    // there must be an API way to do this MAL 20180622
+    build_id = sh(
```
adammhaile: This should really silently fail or retry. I give 100% odds that a build will "Fail" because this curl fails at some point.
Honestly the entire maill_success stuff should be wrapped in a giant try/catch/pass because failing to send out the success email should never, ever give us red.

jobs/build/ocp/Jenkinsfile
```diff
@@ -162,6 +162,58 @@ def get_mirror_url(build_mode, version) {
...
+    // the Changelog: line
+    try {
+        changelog = sh(
+            returnStdout: true,
+            script: [
+                "brew buildinfo ${build_id} --changelog",
+                "sed -n '/Changelog/,\$p'"
+            ].join(' | ')
+        ).trim()
+    } catch (err) {
```
adammhaile: Above should be like this in terms of silently failing.

jobs/build/ocp/Jenkinsfile
markllama: Hrrm.  I think that might be the next little refactor.

jobs/build/ocp/Jenkinsfile
markllama: Will address.

PR #1495, opened 2018-06-25
Fix openshift-ansible repo sync for azure jobs

kargakis:
@openshift/sig-azure


kwoodson:
@kargakis, thanks for this!

PR #1498, opened 2018-06-26
Persist azure launch artifacts.

kwoodson:
Follow up to https://github.com/openshift/openshift-ansible/pull/8972.  This is required for e2e testing to occur.

The previous version was throwing an exception in the openshift-ansible code when attempting to deploy acs-engine cluster.

@kargakis @jim-minter @pweil-

### DIFF
--- a/sjb/config/common/test_cases/origin_release_install_azure.yml
+++ b/sjb/config/common/test_cases/origin_release_install_azure.yml
@@ -169,8 +169,9 @@ extensions:
         set +a -o xtrace
 
         cd /data/src/github.com/Azure/acs-engine
-        mkdir -p _output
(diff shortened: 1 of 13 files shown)


sjb/config/common/test_cases/origin_release_install_azure.yml
```diff
@@ -170,7 +170,7 @@ extensions:
 
         cd /data/src/github.com/Azure/acs-engine
         mkdir -p _output
-        cp /var/tmp/${RESOURCE_GROUP}.json _output/
+        cp /var/tmp/deploy/apimodel.json _output/${RESOURCE_GROUP}.json
```
kargakis: Are you sure you don't want to be copying the whole directory?

PR #1499, opened 2018-06-26
sjb/children.sh: lookup parents only once

bbguimaraes:
Checked with:

```sh
$ for x in sjb/config/test_cases/*.yml sjb/config/common/test_cases/*.yml; do
> diff <(sjb/children.sh "$x") <(sjb/children.sh.old "$x")
> done
$ TIMEFORMAT='%R %U %S'
$ time sjb/children.sh.old sjb/config/common/test_cases/origin_minimal.yml
125.569 60.926 71.239
$ time sjb/children.sh sjb/config/common/test_cases/origin_minimal.yml
1.204 1.000 0.214
```

@stevekuznetsov: it had to be done.

/me awaits the impending bash code review.


sjb/children.sh
```diff
@@ -4,21 +4,24 @@ set -o errexit
...
 
-function children() {
-	local config=$1
-
-	for child_config in sjb/config/test_cases/*.yml sjb/config/common/test_cases/*.yml; do
-		if ! grep -q 'parent' "${child_config}"; then
-			continue
-		fi
+CONFIGS='sjb/config/test_cases/*.yml sjb/config/common/test_cases/*.yml'
+declare -A PARENTS
```
stevekuznetsov: This is not portable and won't work on latest MacOS AFAIR

sjb/children.sh
```diff
@@ -4,21 +4,24 @@ set -o errexit
...
+            echo "${child_config}"
+            children "${child_config}"
+        fi
+    done
 }
 
-children "$1"
\ No newline at end of file
+for config in $CONFIGS; do
+    if parent=$(grep -Po "(?<=parent: ')[^']+(?=')" "${config}"); then
```
stevekuznetsov: We could use Python and `yaml.load()` :)

stevekuznetsov:
I am generally fine with this but worry someone using a mac will break on it

sjb/children.sh
bbguimaraes: Who uses mac anyw…

sjb/children.sh
bbguimaraes: See below.

stevekuznetsov:
Cool, let's use the regex then since it's not built-in. Thanks for looking at this

bbguimaraes:
@stevekuznetsov: updated.

You can thank my performance OCD =)

PR #1500, opened 2018-06-28
Add functions to compare and sort version strings, and to select "build mode" automatically

markllama:
This change adds seven (7) functions to the `pipeline-scripts/buildlib.groovy` library:

*`get_branches()` - get the list of branches from a github remote repository
* `get_releases()` - get the list of release numbers from a github remote repository branch list
* `get_single_file()` - get a single file from a github repo (given an OAuth key)
* `sort_versions()` - sort version strings
* `cmp_version()` - compare version strings
* `eq_version()` - determine if two version strings are equal
* `auto_mode()` - select the _build mode_ for a build when given only the desired version

These functions together will make it possible to allow users of the Jenkins CD build service of OCP to let the _build mode_ default when they give the version they wish to build.

Additional work is needed to integrate this functionality into the build jobs.

### DIFF
--- a/pipeline-scripts/Jenkinsfile
+++ b/pipeline-scripts/Jenkinsfile
@@ -1,0 +1,91 @@
+#!/usr/bin/groovy
+
+// Test the functions in the buildlib for OCP builds
+
+properties(
+    [
+        buildDiscarder(
+            logRotator(
+                artifactDaysToKeepStr: '',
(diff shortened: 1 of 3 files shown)


markllama:
This PR also adds unit tests for the 4 functions that do not have side effects or depend on outside git repositories for input.  It includes a Jenkinsfile that can be used to run the unittests.

PR #1504, opened 2018-07-03
Modify browbeat pipeline script to load up all the params

chaitanyaenr:
This commit modifies the browbeat script to load and pass all the
params defined in the property file so that we need not modify the
script when a new param is added.


chaitanyaenr:
@jeremyeder @akrzos PTAL.

mbruzek:
This change looks good to me. +1 

It will avoid us having to update the groovy code each time a property or parameter changes!

akrzos:
Should this also modify the browbeat.properties file located in https://github.com/openshift/aos-cd-jobs/tree/master/alderaan/property_files 

The OpenStack specifically looks out of date as well.  I assume those are just example files and not used elsewhere.

Also does this specifically set the properties file name to "browbeat.properties"?  Such that we can not use other names?

chaitanyaenr:
@akrzos the property files are just examples but we need to keep them updated like you said.
About the property file name, it can have a different name. We get the file and rename it to browbeat.properties before reading the params.

I have the updated the PR to address the comments. PTAL.

akrzos:
@chaitanyaenr I see it now.  This looks good as the pipeline stands right now.  When we get a staging cluster back I have several more vars for OSP in order to reduce staging cluster build time and we will have OSP13 specific vars but we can fix that when we have the hardware available to test.

chaitanyaenr:
@akrzos @mbruzek @jeremyeder is the PR ready for merge?

akrzos:
LGTM, Thanks @chaitanyaenr

PR #1506, opened 2018-07-05
update devex team emails

bparees:


bparees:
@wozniakjan fyi because i was in there anyway and i don't want to do it twice in a month :/

PR #1512, opened 2018-07-09
Remove the hostname from property file

chaitanyaenr:
This also updates the scaleup properties file with the sample params.


chaitanyaenr:
@jeremyeder PTAL.

PR #1513, opened 2018-07-09
fix typo

jpeeler:
I'm going to rip this out and replace the repo syncing with "sync" soon, but this needs to go in for now.

@stevekuznetsov


jpeeler:
This was pushed to fix the error, but as originally said is going to be replaced by another PR.

PR #1518, opened 2018-07-11
Add pre-release testing for OSA

mjudeikis:
Add a variable to replace multiple versions in the scripts. 
Add rhel pre-release secret passing to ansible code. This will need to be only for pre-release builds of rhel. 

cc: @openshift/sig-azure

### DIFF
--- a/sjb/config/test_cases/azure_build_node_image_centos_39.yml
+++ b/sjb/config/test_cases/azure_build_node_image_centos_39.yml
@@ -10,17 +10,20 @@ sync:
 parameters:
 - name: "AZURE_VM_SIZE"
   default_value: "Standard_D2s_v3"
+- name: "OPENSHIFT_RELEASE"
+  default_value: "3.9"
 actions:
   - type: "forward_parameters"
     parameters:
       - JOB_SPEC
       - AZURE_VM_SIZE
+      - OPENSHIFT_RELEASE
 
   - type: "host_script"
     title: "determine the resource group"
     script: |-
       echo "RESOURCE_GROUP=ci-$JOB_NAME-$BUILD_NUMBER" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
-      echo "IMAGE_NAME=centos7-3.9-$(TZ=Etc/UTC date +%Y%m%d%H%M)" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
+      echo "IMAGE_NAME=centos7-$OPENSHIFT_RELEASE-$(TZ=Etc/UTC date +%Y%m%d%H%M)" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
 
   - type: "script"
     title: "build openshift installer image"
@@ -69,7 +72,7 @@ actions:
       docker run --rm -v ${PWD}:/go/src/github.com/Azure/acs-engine:z \
         -w /go/src/github.com/Azure/acs-engine \
         -e ORCHESTRATOR=openshift \
-        -e ORCHESTRATOR_VERSION=3.9.0 \
+        -e ORCHESTRATOR_VERSION=${OPENSHIFT_RELEASE}.0 \
         -e IMAGE_RESOURCE_GROUP=images \
         -e IMAGE_NAME=$IMAGE_NAME \
         -e CLUSTER_DEFINITION=examples/openshift.json \
@@ -80,7 +83,7 @@ actions:
         -e CLIENT_SECRET=$AZURE_SECRET \
         -e TENANT_ID=$AZURE_TENANT \
         -e SUBSCRIPTION_ID=$AZURE_SUBSCRIPTION_ID \
-        registry.svc.ci.openshift.org/azure/acs-engine-tests:v3.9 make test-openshift
+        registry.svc.ci.openshift.org/azure/acs-engine-tests:v$OPENSHIFT_RELEASE make test-openshift
       set -o xtrace
 
   - type: "script"
--- a/sjb/config/test_cases/azure_build_node_image_rhel_310.yml
+++ b/sjb/config/test_cases/azure_build_node_image_rhel_310.yml
@@ -34,11 +34,12 @@ actions:
   - type: "host_script"
     title: "move secrets to remote host"
     script: |-
+      # TODO: Remove token inject logic after GA
       set +x
       export KUBECONFIG=/var/lib/jenkins/.azure/kubeconfig
-      NEW_TOKEN=$(oc get secret aws-reg-token --template={{.data.token}} | base64 -d)
-      if [[ -n $NEW_TOKEN ]]; then
-        echo $NEW_TOKEN > /var/lib/jenkins/.azure/registry
+      AWS_REGISTRY_TOKEN=$(oc get secret aws-reg-token -n azure --template={{.data.token}} | base64 -d)
+      if [[ -n $AWS_REGISTRY_TOKEN ]]; then
+        echo "AWS_REGISTRY_TOKEN=$AWS_REGISTRY_TOKEN" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
       fi
       set -x
       rsync --copy-links --omit-dir-times --archive --rsh "ssh -F ./.config/origin-ci-tool/inventory/.ssh_config" /var/lib/jenkins/.azure/* openshiftdevel:/data/src/github.com/openshift/release/cluster/test-deploy/azure/
@@ -49,11 +50,13 @@ actions:
     title: "build node image"
     repository: "release"
     script: |-
-      cd cluster/test-deploy/azure
+     # TODO: Remove token injection after GA (oreg_* vars)
+     cd cluster/test-deploy/azure
       echo "Build parameters:"
       echo "openshift_azure_resource_group_name=$RESOURCE_GROUP"
       echo "openshift_azure_output_image_name=$IMAGE_NAME"
       set +x
+      export OREG_URL='registry.reg-aws.openshift.com/openshift3/ose-${component}:${version}'
       TYPE=azure INSTANCE_PREFIX=unused ../../bin/ansible.sh ansible-playbook \
         -e "openshift_azure_resource_group_name=$RESOURCE_GROUP" \
         -e "openshift_azure_resource_location=eastus" \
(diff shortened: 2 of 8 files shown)


sjb/config/test_cases/azure_build_node_image_rhel_310.yml
```diff
@@ -10,17 +10,23 @@ sync:
 parameters:
 - name: "AZURE_VM_SIZE"
   default_value: "Standard_D2s_v3"
+- name: "OPENSHIFT_RELEASE"
+  default_value: "3.10"
```
kargakis: The job is already versioned so we don't need these variables.

sjb/config/test_cases/azure_build_node_image_rhel_310.yml
```diff
@@ -10,17 +10,23 @@ sync:
...
+- name: "OPENSHIFT_RELEASE_LATEST"
```
kargakis: I am not sure we need this parameter. Playbooks are already versioned based on the branch (we know that playbooks run from release branches are not latest).

mjudeikis: as discusses on irc will move to maintain this in the ansible code via release branches.

mjudeikis:
@kargakis  I know we discussed this on chat, but its big pita for rhel builds to run without knowing which version they are running. Idea is that our ansible playbooks is version agnostic, so they are capable to build centos and rhel for all releases. 

If we stick with "branch code versioning" we will soon end-up maintaining  versions of code (even if its just 3 places in 2 repos where we need to change few lines of version per branches), it will be poor man solution as we will need to make 2 repos 3 releases = 6 PR if we want to change anything... Not the place I want to be in.

As much as I dont like this I would better improve our openshift-ansible to build based on the variable we pass from here, and do maintenance in one place and not over 2-3 separate repositories :/ 

Long story short problem is our RPM channels and how we treat some of the logic in openshift-ansible. To be safe we need these in our node build scripts
```
    openshift_image_tag: "v3.10"
    openshift_version: "3.10"
    openshift_release: "3.10"
```

sjb/config/test_cases/azure_build_node_image_centos_311.yml
```diff
@@ -0,0 +1,112 @@
+---
```
kargakis: This shouldn't be `*centos_311.yml` but `*centos.yml` like all other master jobs.

sjb/config/test_cases/azure_build_node_image_centos_311.yml
```diff
@@ -0,0 +1,112 @@
...
+provision:
+  os: "rhel"
+  stage: "base"
+  provider: "aws"
+sync:
+  - openshift,openshift-ansible=release-3.11
```
kargakis: There is no such branch (yet). Use master.

sjb/config/test_cases/azure_build_node_image_centos_311.yml
```diff
@@ -0,0 +1,112 @@
...
+        -e IMAGE_NAME=$IMAGE_NAME \
+        -e CLUSTER_DEFINITION=examples/openshift.json \
+        -e TIMEOUT=30m \
+        -e CLEANUP_ON_EXIT=true \
+        -e REGIONS=eastus \
+        -e CLIENT_ID=$AZURE_CLIENT_ID \
+        -e CLIENT_SECRET=$AZURE_SECRET \
+        -e TENANT_ID=$AZURE_TENANT \
+        -e SUBSCRIPTION_ID=$AZURE_SUBSCRIPTION_ID \
+        registry.svc.ci.openshift.org/azure/acs-engine-tests:v$OPENSHIFT_RELEASE make test-openshift
```
kargakis: Use 3.10 here

sjb/config/test_cases/azure_build_node_image_centos_311.yml
mjudeikis: tempted to delete all these for now. I was not expting to get config for jenkis so fast :D

sjb/config/test_cases/azure_build_node_image_centos_311.yml
```diff
@@ -0,0 +1,112 @@
...
+    parameters:
+      - JOB_SPEC
+      - AZURE_VM_SIZE
+      - OPENSHIFT_RELEASE
+
+  - type: "host_script"
+    title: "determine the resource group"
+    script: |-
+      echo "RESOURCE_GROUP=ci-$JOB_NAME-$BUILD_NUMBER" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
+      echo "IMAGE_NAME=centos7-$OPENSHIFT_RELEASE-$(TZ=Etc/UTC date +%Y%m%d%H%M)" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
```
kargakis: Use `master` instead of OPENSHIFT_RELEASE

sjb/config/test_cases/azure_build_node_image_rhel_311.yml
```diff
@@ -0,0 +1,121 @@
+---
+provision:
+  os: "rhel"
+  stage: "base"
+  provider: "aws"
+sync:
+  - openshift,openshift-ansible=release-3.11
```
kargakis: master

sjb/config/test_cases/azure_build_node_image_rhel_311.yml
```diff
@@ -0,0 +1,121 @@
+---
```
kargakis: Need to drop the 311 suffix

sjb/config/test_cases/azure_build_node_image_rhel_311.yml
```diff
@@ -0,0 +1,121 @@
...
+    parameters:
+      - JOB_SPEC
+      - AZURE_VM_SIZE
+      - OPENSHIFT_RELEASE
+
+  - type: "host_script"
+    title: "determine the resource group"
+    script: |-
+      echo "RESOURCE_GROUP=ci-$JOB_NAME-$BUILD_NUMBER" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
+      echo "IMAGE_NAME=rhel7-$OPENSHIFT_RELEASE-$(TZ=Etc/UTC date +%Y%m%d%H%M)" | ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "cat >>/etc/environment"
```
kargakis: master

sjb/config/test_cases/azure_build_node_image_rhel_311.yml
```diff
@@ -0,0 +1,121 @@
...
+        -e IMAGE_NAME=$IMAGE_NAME \
+        -e CLUSTER_DEFINITION=examples/openshift.json \
+        -e TIMEOUT=30m \
+        -e CLEANUP_ON_EXIT=true \
+        -e REGIONS=eastus \
+        -e CLIENT_ID=$AZURE_CLIENT_ID \
+        -e CLIENT_SECRET=$AZURE_SECRET \
+        -e TENANT_ID=$AZURE_TENANT \
+        -e SUBSCRIPTION_ID=$AZURE_SUBSCRIPTION_ID \
+        registry.svc.ci.openshift.org/azure/acs-engine-tests:v$OPENSHIFT_RELEASE make test-openshift
```
kargakis: 3.10

mjudeikis:
/hold cancel
ptal @kwoodson @kargakis @jim-minter   This passes token to be backed into the image for RHEL build which needs to be built from aws registry. This particular code needs to be removed when going GA and moving jobs.

kwoodson:
@mjudeikis, I noticed you have the ...311.xml files but not the corresponding  ....yml files.  Are they missing or did I miss something?

sjb/config/test_cases/azure_build_node_image_centos_39.yml
```diff
@@ -51,6 +54,7 @@ actions:
         -e "openshift_azure_storage_account_ns=images" \
         -e "openshift_azure_container=images" \
         -e "openshift_azure_vm_size=$AZURE_VM_SIZE" \
+        -e "openshift_azure_release=$OPENSHIFT_RELEASE" \
```
kargakis: Is this needed?

mjudeikis: ack.

sjb/config/test_cases/azure_build_node_image_rhel_310.yml
```diff
@@ -31,6 +34,10 @@ actions:
   - type: "host_script"
     title: "move secrets to remote host"
     script: |-
+      set +x
+      export KUBECONFIG=/var/lib/jenkins/.azure/kubeconfig
+      echo $(oc get secret aws-reg-token --template={{.data.token}} |base64 -d) > /var/lib/jenkins/.azure/registry
```
kargakis: I assume we will use this eventually for running PR tests. If `oc get secret` fails then the token will get overwritten and jobs running in parallel beyond this step will not have a token to use?

mjudeikis: Whats true. How unreliable is our CI cluster for this? Do we need to build some contingency in this now? I would like to assume perfect world for now and if we see a lot of flakes we can add some checks.

kargakis: The CI cluster is mostly reliable but the minimal fix for avoiding this flake comes at no cost.
```bash
NEW_TOKEN=$(oc get secret aws-reg-token --template={{.data.token}} | base64 -d)
if [[ -n $NEW_TOKEN ]]; then
  echo $NEW_TOKEN > /var/lib/jenkins/.azure/registry
fi
```

sjb/config/test_cases/azure_build_node_image_rhel_310.yml
```diff
@@ -31,6 +34,13 @@ actions:
   - type: "host_script"
     title: "move secrets to remote host"
     script: |-
+      set +x
+      export KUBECONFIG=/var/lib/jenkins/.azure/kubeconfig
+      NEW_TOKEN=$(oc get secret aws-reg-token --template={{.data.token}} | base64 -d)
+      if [[ -n $NEW_TOKEN ]]; then
+        echo $NEW_TOKEN > /var/lib/jenkins/.azure/registry
```
jim-minter: don't write to the jenkins filesystem, it introduces a (tiny) race - echo the token up to the box via ssh echo.  Think about the filename you choose - make it descriptive

sjb/config/test_cases/azure_build_node_image_rhel_310.yml
```diff
@@ -31,6 +34,13 @@ actions:
   - type: "host_script"
     title: "move secrets to remote host"
     script: |-
+      set +x
```
mjudeikis: Move to file-less way to avoid race conditions:
 ssh -F ./.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "echo 'location_url=${location_url:-}' >> /etc/environment"
or 
>/data/src/github.com/openshift/release/cluster/test-deploy/azure/registry

sjb/config/test_cases/azure_build_node_image_rhel_310.yml
```diff
@@ -51,7 +65,9 @@ actions:
         -e "openshift_azure_storage_account_ns=images" \
         -e "openshift_azure_container=images" \
         -e "openshift_azure_vm_size=$AZURE_VM_SIZE" \
-        playbooks/azure/openshift-cluster/build_node_image.yml
+        -e "openshift_azure_dev_registry_token=$(cat /data/src/github.com/openshift/release/cluster/test-deploy/azure/registry)" \
```
jim-minter: add a comment that this line should be removed at GA

jim-minter:
2 nits then will merge

jim-minter:
@mjudeikis merged - please will you update the jenkins jobs?

mjudeikis:
@jim-minter done and running

PR #1519, opened 2018-07-11
Adding elasticsearch node selectors since they are required as of 3.11

ewolinetz:
Required as of https://github.com/openshift/openshift-ansible/pull/9144

PTAL @stevekuznetsov @richm


richm:
These new parameters should be added to https://github.com/openshift/aos-cd-jobs/blob/master/sjb/inventory/group_vars/OSEv3/logging.yml instead of trying to pass them in as -e params

PR #1521, opened 2018-07-11
move logging nodeselectors to inventory logging.yml

richm:
move logging nodeselectors to inventory logging.yml
this should fix current logging CI issues


ewolinetz:
/lgtm

thanks @richm I was going to try a few different permutations to fix this, I think this is the cleanest approach

PR #1525, opened 2018-07-13
update kibana proxy image

jcantrill:
Update proxy image to 1.1.0


ewolinetz:
/lgtm

we should also update the default value here https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_defaults/defaults/main.yml#L16

jcantrill:
@ewolinetz ref https://github.com/openshift/openshift-ansible/pull/9210

jcantrill:
closing per @richm alternate fix.

PR #1528, opened 2018-07-17
Add helper functions for updating RPM version numbers and getting versions from .spec files on github

markllama:
This PR adds functions that can be used in the future to generate version number updates and to retrieve
version number information from RPM spec files on github without cloning the entire repo.

It makes no change to existing jobs or functions.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -293,18 +293,33 @@ def get_releases(repo_url) {
 }
 
 /**
+ * Read an OAuth token from a file on the jenkins server.
+ * Because groovy/jenkins sandbox won't let you read it without sh()
+ * @param token_file - a file containing a single OAuth token string
+ * @return - a string containing the OAuth token
+ */
+def read_oath_token(token_file) {
+    token_string = sh (
+        returnStdout: true,
+        script: "cat ${token_file}"
+    ).trim()
+    return token_string
+}
+
+/**
  * Retrieve a single file from a Github repository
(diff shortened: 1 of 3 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -437,6 +452,105 @@ invalid mode build != master and no release branch
     return mode
 }
 
+//
+// Create a new version string based on the build mode
+//
+//@NonCPS
+def new_version(mode, version_string, release_string) {
+
```
tbielawa: Can you document some valid input/outputs with those fancy `@param` comments you use so well

pipeline-scripts/buildlib.groovy
```diff
@@ -437,6 +452,105 @@ invalid mode build != master and no release branch
     return mode
 }
 
+//
+// Create a new version string based on the build mode
+//
+//@NonCPS
```
tbielawa: You meant to leave the NonCPS commented out, right?

pipeline-scripts/buildlib.groovy
markllama: You silver-tongued dude you.  Yep will do.

pipeline-scripts/buildlib.groovy
markllama: I think I tested it with and without.  Groovy alone and Jenkins are different.  I'll verify.

markllama:
Updated comment strings.  Unit tests passed.

PR #1529, opened 2018-07-18
Enable ansible-2.6 repo on master branches

sdodson:


sdodson:
/cc @vrutkovs 
this should work now, will merge after install job goes green on the pr to require ansible 2.6

PR #1534, opened 2018-07-24
Switch to using the AWS cluster operator account for co e2e

csrwng:
The initial vm will still be created in the regular account but all test artifacts will now be created in the cluster operator AWS account, making it easier to manage and cleanup and reducing the risk to the regular account. After this is in place, the credentials to the regular account may be removed from Jenkins.


csrwng:
@dobbymoodge ptal

dobbymoodge:
@csrwng you've tested this, right?

csrwng:
@dobbymoodge Yes the test is successful :)

PR #1549, opened 2018-08-08
cluster-operator e2e: switch to using coutil for cluster verification

csrwng:
Use our consolidated command instead of the separate utility


csrwng:
@stevekuznetsov or @dobbymoodge ptal

dobbymoodge:
lgtm

PR #1554, opened 2018-08-16
Provide aws machine controller ci job

ingvagabund:
For testing https://github.com/openshift/cluster-api-provider-aws.
Successful run: https://ci.openshift.redhat.com/jenkins/job/ci-kubernetes-aws-actuator/17/console

@stevekuznetsov PTAL

### DIFF
--- a/sjb/config/test_cases/ci-kubernetes-aws-actuator.yml
+++ b/sjb/config/test_cases/ci-kubernetes-aws-actuator.yml
@@ -1,0 +1,236 @@
+---
+parent: 'common/test_cases/minimal.yml'
+extensions:
+  actions:
+    - type: "host_script"
+      title: "Upload the default AWS credentiasl"
+      script: |-
+         ssh -F ${WORKSPACE}/.config/origin-ci-tool/inventory/.ssh_config openshiftdevel "mkdir ~/.aws"
+         scp -F ${WORKSPACE}/.config/origin-ci-tool/inventory/.ssh_config ~/.aws/credentials openshiftdevel:~/.aws
+    - type: "script"
+      title: "Install minikube"
+      script: |-
+        # Install minikube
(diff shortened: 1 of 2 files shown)


sjb/config/test_cases/ci-kubernetes-aws-actuator.yml
```diff
@@ -0,0 +1,236 @@
...
+          name: test
+        ---
+        apiVersion: v1
+        kind: Secret
+        metadata:
+          name: aws-credentials-secret
+          namespace: test
+        type: Opaque
+        data:
+          awsAccessKeyId: $(echo -n $(aws configure get aws_access_key_id) | base64)
```
ingvagabund: This two lines has the verbose mode disabled so the aws credentials are not printed into the logs.

ingvagabund:
@enxebre @paulfantom FYI

sjb/config/test_cases/ci-kubernetes-aws-actuator.yml
enxebre: is this happening on an ec2 machine? if so we could also apply a know iam role and let the aws client omitting the credentials

enxebre:
great!

sjb/config/test_cases/ci-kubernetes-aws-actuator.yml
ingvagabund: I did not check this use case. Though, in future we may need more permissions which we might not want to share with other ec2 instances (since all ec2 instance are provisioned by the oct). So I still think it's better to use the explicit aws credentials. Later, we can switch to a different AWS account.

stevekuznetsov:
@ingvagabund push whenver

ingvagabund:
@stevekuznetsov thanks!!!

PR #1558, opened 2018-08-21
Add all required env to azure jobs

stevekuznetsov:

@kwoodson you were missing the `$PROW_JOB_ID` var since you don't use a common parent. Fixed the jobs now.


kwoodson:
@stevekuznetsov, thanks!

PR #1561, opened 2018-08-24
cluster-operator e2e: Install python-devel tools

csrwng:
This patches up the e2e job for cluster-operator by installing the python-devel package. The real long-term fix is to use the ansible image to run any playbooks and not rely on software installed on the ami. That will be coming in a follow-up fix.


csrwng:
@dobbymoodge @stevekuznetsov ptal

stevekuznetsov:
@csrwng you push this out

csrwng:
done

PR #1565, opened 2018-08-28
CRI-O: disable df -h collection

runcom:

### DIFF
--- a/sjb/config/common/test_cases/crio.yml
+++ b/sjb/config/common/test_cases/crio.yml
@@ -133,7 +133,9 @@ artifacts:
 generated_artifacts:
   installed_packages.log: 'sudo yum list installed'
   avc_denials.log: 'sudo ausearch -m AVC -m SELINUX_ERR -m USER_AVC'
-  filesystem.info: 'sudo df -h && sudo pvs && sudo vgs && sudo lvs'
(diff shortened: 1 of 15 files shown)


sjb/config/common/test_cases/crio.yml
```diff
@@ -133,7 +133,9 @@ artifacts:
 generated_artifacts:
   installed_packages.log: 'sudo yum list installed'
   avc_denials.log: 'sudo ausearch -m AVC -m SELINUX_ERR -m USER_AVC'
-  filesystem.info: 'sudo df -h && sudo pvs && sudo vgs && sudo lvs'
+  #filesystem.info: 'sudo df -h && sudo pvs && sudo vgs && sudo lvs'
```
rhatdan: Would something like
 df -h /var/lib/containers/
Filesystem              Size  Used Avail Use% Mounted on
/dev/mapper/fedora-var  200G   36G  165G  18% /var

Be better?

runcom: so, this is collecting the whole df situation when a crio job finishes. For some reason, after e2e, some nfs mount from some e2e test is still hanging around causing df -h to hang. I'm not sure about the root cause but for now, I'd disable it entirely to keep working on the upstream test-grid. I'll re-enable it once we figure the nfs leak out.

PR #1570, opened 2018-08-31
Ensure cockpit image is set correctly

michaelgugino:
Due to how we use oreg_url in CI, we need to set
cockpit-ui image specifically.

Needed-by: https://github.com/openshift/openshift-ansible/pull/9827


stevekuznetsov:
Jobs grab this dir by cloning `master` of `aos-cd-job` so without a merge you haven't pushed anything anywhere FYI

michaelgugino:
@stevekuznetsov I just realized that.  I pushed up a temporary change reflected here.

PR #1574, opened 2018-09-04
Remove sjb extra_var openshift_cockpit_deployer_image

michaelgugino:
This commit removes extra_var openshift_cockpit_deployer_image
as it is no longer needed because it is defined in sjb/inventory.


michaelgugino:
This is pushed.

PR #1575, opened 2018-09-05
only specify to sync openshift/origin for openshift/service-catalog prs

jboyd01:
We were getting errors from Non Service Catalog origin & openshift prs complaining that a sync for origin was specified more then once.  This has been corrected, we only specify the origin sync now for the Service Catalog PRs.

Also cleaned up the reporting of commits for origin & service-catalog and fixed the commands used to pull the Catalog pod logs (it was failing).


stevekuznetsov:
Pushed out

PR #1580, opened 2018-09-10
Cluster operator: Use 3.10 oc binary to start cluster up

csrwng:
Latest version of `oc cluster up` is no longer compatible with the cluster operator test. This ensures that we use the last known good version (3.10) until we can migrate to the newer version.


csrwng:
@dobbymoodge @stevekuznetsov ptal

stevekuznetsov:
@csrwng you push

PR #1589, opened 2018-09-19
use mailing lists instead of individuals

markllama:
This change sends email outputs from the ocp/Jenkinsfile jobs to the aos-team-art mailing list instead of to individuals.


sosiouxme:
thpbpbt I don't have approve here :)

PR #1595, opened 2018-09-27
Allow user to choose the jenkins slave to schedule the job on

chaitanyaenr:
This commit allows users to run their workload on a particular slave
using labels.


chaitanyaenr:
@akrzos PTAL.
/cc @jeremyeder

PR #1596, opened 2018-09-28
ci-kubernetes-aws-actuator: run e2e from within a go binary

ingvagabund:
Replace the bash craziness with a simple `make test-e2e` command.


ingvagabund:
@stevekuznetsov PTAL

PR #1600, opened 2018-10-02
Add support to convert xml templates to jjb

chaitanyaenr:
This commit:
- Will allow users to submit a xml template and alderaan-watcher
  will take care of converting it to jjb and creates the jenkins job.
- Paramerizes github token used for the jobs so that we can expose
  the templates in the public repo.
- Also adds support to add users public key to the cluster.


chaitanyaenr:
@jeremyeder PTAL.

PR #1620, opened 2018-10-12
add extended test jobs for 311

bparees:


bparees:
@smarterclayton @stevekuznetsov ptal.

bparees:
@stevekuznetsov @smarterclayton sjb/generate is not generating the new pull_request jobs I have defined, but it is generating the branch jobs.  (you can see what was generated in this PR).  Any theories on why not?

bparees:
sigh. yml vs yaml.

PR #1630, opened 2018-10-26
hardcode openshift_cli_image to not pull from docker.io

richm:
in a devenv ami, the origin-node image should already be there
so no need to pull it from docker.io

### DIFF
--- a/sjb/config/common/test_cases/origin_built_installed_release.yml
+++ b/sjb/config/common/test_cases/origin_built_installed_release.yml
@@ -128,6 +128,7 @@ extensions:
                          -e openshift_pkg_version="$( cat ./ORIGIN_PKG_VERSION )"               \
                          -e openshift_release="$( cat ./ORIGIN_RELEASE )"                       \
                          -e oreg_url='openshift/origin-${component}:'"$( cat ./ORIGIN_COMMIT )" \
+                         -e openshift_cli_image='openshift/origin-node:'"$( cat ./ORIGIN_COMMIT )" \
                          -e openshift_node_port_range='30000-32000'                             \
                          -e 'osm_controller_args={"enable-hostpath-provisioner":["true"]}'      \
                          "${evars:-}" \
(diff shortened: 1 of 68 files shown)


richm:
There are some assumptions in this PR
- these jobs only run against master
- these jobs only run against aws devenv ami

smarterclayton:
I think this was a bug in ansible and we should answer whether it will be fixed there first.

gabemontero:
@richm @smarterclayton is there an item opened against https://github.com/openshift/openshift-ansible on somewhere else to track what was noted in https://github.com/openshift/aos-cd-jobs/pull/1630#issuecomment-433520957

this is currently blocking any of the jenkins plugin PRs

gabemontero:
@vrutkovs responded in slack at https://coreos.slack.com/archives/CBN38N3MW/p1540928638142500?thread_ts=1540925477.140600&cid=CBN38N3MW that if we have any jobs that use openshift-ansible they need to be reconfigured to release-3.11

That basically that the time has come as part of the move from prow that at least for our jenkins runs, openshift-ansible master is broken and won't work in Jenkins based tests at all very soon.

Now, this might just pertain to my https://github.com/openshift/aos-cd-jobs/issues/1631 vs. what @richm is dealing with ... if so, we can move the subsequent discussion there.

But if no fixes are coming in openshift-ansible to address the current install origin stage in these aos-cd-jobs, this would imply
a) short term, we change the aos-cd-jobs for the jenkins plugin to use 3.11 so the PR tests can pass .. in the very short term interim, validating plugin changes against 3.11 is tolerable
b) we have to figure out how to port the current injection of the plugin changes from a PR into the jenkins image used in the ci-operator/prow e2e tests, then invoke the tests as we do today ... most likely some form of docker build on the jenkins image that injects the new plugin file, using the extension capabilities of our jenkins image 

@richm @smarterclayton @bparees @stevekuznetsov @vrutkovs .... fyi / that sound right?

bparees:
(a) sounds fine to me
(b) needs a little more thought imho.  I think there are a few approaches we could take but in general i'd like to see a test architecture that allows us to "build an image based on an existing image, but adding content from another repo/pr, and then be able to run extended tests against that image on cluster".

One way to get there might involve adding Dockerfiles to the various plugin repos that effectively do:

FROM jenkins
ADD locally-built-plugin.war

If those images were then published to api.ci, the extended tests could consume them (when the extended tests were being run for a PR from that repo).  

Something like that anyway.  Like I said, needs more discussion.  But getting away from one-off processes that build the image via docker would be good.

This is at least somewhat related to https://jira.coreos.com/browse/DPTP-208 (jira issue for getting the jenkins images to be built/published to api.ci)

PR #1650, opened 2018-11-15
Replace OIT with Doozer

adammhaile:
@tbielawa @sosiouxme @smunilla @joeldavis84 
NOT TO BE MERGED UNTIL AFTER THANKSGIVING!

This is merely a brutal replacement of OIT calls with Doozer.
Oh, and I deleted ose_images.sh and some other no longer used jobs.
Feel free to take a look and let me know if I did something dumb, but definitely do not merge until I'm back after the holiday.


tbielawa:
Decided together to merge this because it should only be a matter of reverting this if it breaks.

PR #1663, opened 2018-12-05
Copy the properties file to the jump host to generate configmap

chaitanyaenr:
This commit:
- adds a stage to clean up the workspace at the end of the run.
- adds support to run containerized tooling by copying the properties
  files for the selected jobs in the pipeline onto the jump host. The
  properties file gets converted to a configmap and fed into the
  controller pod.
- adds a byo scale test to allow users to run any script as long as
  it's part of the openshift/svt repo.
- modifies the properties files with the new vars.


chaitanyaenr:
@jeremyeder PTAL.

chaitanyaenr:
@akrzos  PTAL.

chaitanyaenr:
Updated the PR with support to run BYO scale test. 
@jeremyeder @akrzos PTAL.

PR #1669, opened 2018-12-11
Adding automatic merge job

adammhaile:
@tbielawa

### DIFF
--- a/jobs/build/merge_ocp/Jenkinsfile
+++ b/jobs/build/merge_ocp/Jenkinsfile
@@ -1,0 +1,127 @@
+
+// Expose properties for a parameterized build
+properties(
+    [
+        buildDiscarder(
+            logRotator(
+                artifactDaysToKeepStr: '',
+                artifactNumToKeepStr: '',
+                daysToKeepStr: '',
+                numToKeepStr: '1000')),
+        [
+            $class: 'ParametersDefinitionProperty',
+            parameterDefinitions: [
+                [
+                    name: 'VERSIONS',
+                    description: 'CSV list of merges to run merge on.',
+                    $class: 'hudson.model.StringParameterDefinition',
+                    defaultValue: "3.9,3.10,3.11,4.0"
+                ],
+                [
+                    name: 'MAIL_LIST_SUCCESS',
+                    description: 'Success Mailing List',
+                    $class: 'hudson.model.StringParameterDefinition',
+                    defaultValue: [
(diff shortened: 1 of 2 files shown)


jobs/build/merge_ocp/Jenkinsfile
```diff
@@ -0,0 +1,126 @@
...
+                daysToKeepStr: '',
+                numToKeepStr: '1000')),
+        [
+            $class: 'ParametersDefinitionProperty',
+            parameterDefinitions: [
+                [
+                    name: 'VERSIONS',
+                    description: 'Version string for build without leading "v" (i.e. 4.0.0)',
+                    $class: 'hudson.model.StringParameterDefinition',
+                    defaultValue: "3.9,3.10,3.11,master"
```
tbielawa: Just curious why `master` is listed, but `4.0.0` is in the example. Is this supposed to to be multiple choice, or is the `defaultValue` just there to give you something to erase as an example?

tbielawa: Oh, `MERGE_VERSIONS` looks like `VERSIONS` is a comma separated input. So, maybe mention that in the `description` if that's correct?

jobs/build/merge_ocp/merge_ocp.sh
```diff
@@ -0,0 +1,30 @@
+# usage: ./merge_ocp.sh ./working/ enterprise-3.11 release-3.11
+set -o xtrace
```
tbielawa:
```suggestion
set -e
set -x
```

`-e` for exit on error and `-o` is just another way to print out commands before executing (like how you did `xtrace` before).

tbielawa: Also probably enable the one for exiting on undefined variables.

richm: `set -euxo pipefail` is sort of like `gcc -Wall` . . .

jobs/build/merge_ocp/Jenkinsfile
adammhaile: Oops. Copy error from the job I based this on. 
Fixed and removed master from the list in lieu of checking if a given version is the current master and then handling it specially.

jobs/build/merge_ocp/merge_ocp.sh
adammhaile: Ack. Will change.

PR #1676, opened 2018-12-19
[CNV] add scale job

ILpinto:
Job for CNV scale.


chaitanyaenr:
@ILpinto Thanks for the PR. It looks good at the first glance, I will test it and add the review.

PR #1680, opened 2019-01-08
Use origin-cli for 3.10+ to get oc for console tests

jhadvig:
For 3.10 and also 3.11 we need to get the `oc` binary from the `origin-cli` image.
This should be a quick fix, so we free our merge queue. Will refactor je jobs as follow-up.

@spadgett FYI

@stevekuznetsov PTAL

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_web_console_310.yml
+++ b/sjb/config/test_cases/test_branch_origin_web_console_310.yml
@@ -22,8 +22,8 @@ extensions:
           echo "IMAGE_VERSION=latest" > IMAGE_VERSION_VAR
         fi
         source IMAGE_VERSION_VAR
-        sudo docker pull openshift/origin:"${IMAGE_VERSION}"
-        sudo docker create --name temp-container openshift/origin:"${IMAGE_VERSION}"
+        sudo docker pull openshift/origin-cli:"${IMAGE_VERSION}"
+        sudo docker create --name temp-container openshift/origin-cli:"${IMAGE_VERSION}"
         sudo docker cp temp-container:/usr/bin/oc ./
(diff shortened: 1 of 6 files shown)


sjb/config/test_cases/test_branch_origin_web_console_310.yml
```diff
@@ -21,9 +21,20 @@ extensions:
...
+        # openshift/origin-cli in 3.10 and newer. Use openshift/origin for
+        # earlier releases.
+        case $PULL_BASE_REF in
+        enterprise-3.[3-9])
+          CLI_IMAGE="origin"
+          ;;
+        *)
+          CLI_IMAGE="origin-cli"
+          ;;
+        esac
```
spadgett: I'd remove the version check here and just hard-code the right value. We know this is 3.10. (Same for 3.11 below.)

spadgett: We should come back and remove some of the version sniffing from these files since it's no longer needed, but OK as a separate PR.

jhadvig:
@spadgett updated

stevekuznetsov:
@spadgett can you push this up?

spadgett:
I've never been able to successfully push. @jhadvig can you try?

https://github.com/openshift/aos-cd-jobs#custom-xml-generator-configuration-under-sjb

bbguimaraes:
```
[INFO] Updating test_pull_request_origin_web_console_310...
[INFO] Updating test_branch_origin_web_console_311...
[INFO] Updating test_pull_request_origin_web_console_311...
[INFO] Updating test_branch_origin_web_console_310...
```

PR #1685, opened 2019-01-16
Lock pulled oc binary to v3.11 version when testing against master branch

jhadvig:
Since post 3.11 we dropped the `oc cluster up` command we need to lock version of pulled image to v3.11 when sending PR against master branch. 

/assign @spadgett

@stevekuznetsov fyi

### DIFF
--- a/sjb/config/test_cases/test_branch_origin_web_console.yml
+++ b/sjb/config/test_cases/test_branch_origin_web_console.yml
@@ -6,35 +6,8 @@ extensions:
       title: "determine binary and image version"
       repository: "origin"
       script: |-
-        if [[ "${PULL_BASE_REF}" = "enterprise-"* ]]; then
-          case $PULL_BASE_REF in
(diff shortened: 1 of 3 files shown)


spadgett:
@jhadvig I'd remove the version testing from this script since we have separate jobs for 3.9/3.10/3.11. It's not needed anymore.

jhadvig:
@spadgett right... thought for some reason that we would want to keep it but since there are tests for each branch we support it shouldn't matter.

jhadvig:
@stevekuznetsov can you pls merge and push ? Thx <3

bbguimaraes:
```
[INFO] Checking for existence of test_branch_origin_web_console...
[INFO] Checking for existence of test_pull_request_origin_web_console...
[INFO] Updating test_branch_origin_web_console...
[INFO] Updating test_pull_request_origin_web_console...
```

PR #1686, opened 2019-01-24
Always continue image refresh jobs

smunilla:
There is almost never a circumstance when you want to abort or retry
an image refresh job. In 99% of cases you just want to push and attach
whatever you have.

### DIFF
--- a/jobs/build/refresh-images/Jenkinsfile
+++ b/jobs/build/refresh-images/Jenkinsfile
@@ -233,6 +233,7 @@ images:build
                         return true  // finish waitUntil
                     } catch (err) {
                         failed_builds = buildlib.get_failed_builds(DOOZER_WORKING)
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
FYI This repo doesn't have auto-merges

sosiouxme:
Can I suggest as an alternative https://github.com/openshift/aos-cd-jobs/pull/1690

smunilla:
@sosiouxme #1690 seems to leave the RETRY/CONTINUE/ABORT option still up to the user. Am I reading that wrong or did we change our mind about the value of that?

smunilla:
Discussed with @sosiouxme and I'm going to close this one.

sosiouxme:
@smunilla sorry i completely failed to see your comment!

PR #1698, opened 2019-02-14
build/custom: build signed when set

sosiouxme:
This fixes a bug where the build is always with unsigned repos, even if SIGNED is set.


sosiouxme:
@adammhaile @tbielawa PTAL

PR #1705, opened 2019-02-21
build/ocp: fix and expand safeArchiveArtifacts call

sosiouxme:
will also gather the new state.yaml file


sosiouxme:
apparently if I don't put the parentheses it treats it like a map lookup. oops.
@tbielawa @adammhaile

PR #1714, opened 2019-02-26
build/custom: wishlist items

sosiouxme:
Add some features to the custom job that I've really wanted.

Divided into commits for easier reviewing; you can also try it out at https://buildvm.openshift.eng.bos.redhat.com:8443/job/hack/job/lmeyer-aos-cd-builds/job/build%252Fcustom/

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -46,10 +46,11 @@ node {
                         defaultValue: ""
                     ],
                     [
-                        name: 'REBASE_IMAGES',
-                        description: 'Run images:rebase? Otherwise use images:update-dockerfile',
-                        $class: 'hudson.model.BooleanParameterDefinition',
-                        defaultValue: true
+                        name: 'IMAGE_MODE',
+                        description: 'How to update image dist-gits: with a source rebase, just dockerfile updates, or not at all (no version/release update)',
+                        $class: 'hudson.model.ChoiceParameterDefinition',
+                        choices: ['rebase', 'update-dockerfile', 'nothing'].join('\n'),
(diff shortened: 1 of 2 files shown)


sosiouxme:
@adammhaile @tbielawa

jobs/build/custom/Jenkinsfile
```diff
@@ -17,7 +17,7 @@ node {
                     commonlib.oseVersionParam('BUILD_VERSION'),
                     [
                         name: 'VERSION',
-                        description: 'Version string for build without leading "v" (i.e. 4.0.0)',
+                        description: 'Version string for build (e.g. 4.0.0)',
```
adammhaile: I liked that note. Because I did it too many times :P

pipeline-scripts/commonlib.groovy
```diff
@@ -16,12 +16,6 @@ ocpVersions = [
     "3.1",
 ]
 
-ocp4DefaultVersion = "4.0"
```
adammhaile: You gotta rebase! You just removed what I just added.

jobs/build/custom/Jenkinsfile
sosiouxme: The change here is that it works with or without the "v" so you don't have to remember

pipeline-scripts/commonlib.groovy
sosiouxme: Yeah! sorry 'bout that.

sosiouxme:
Updated for re-review

sosiouxme:
Heh, my bad. The changes aren't too big though. I saw a way to make the regex bit work the way I want, and I added the last commit so build name/description get updated

PR #1718, opened 2019-02-28
ocp/4 RPM rebuilds

sosiouxme:
Added retries around the ose and openshift-ansible RPM builds.

You can see an example run of ocp at https://buildvm.openshift.eng.bos.redhat.com:8443/job/hack/job/lmeyer-aos-cd-builds/job/build%252Focp/20/console

### DIFF
--- a/jobs/build/ocp/Jenkinsfile
+++ b/jobs/build/ocp/Jenkinsfile
@@ -792,37 +792,42 @@ node {
                 // Allow both brew builds to run at the same time
 
                 dir(OSE_DIR) {
-                    OSE_TASK_ID = sh(
+                    oseTaskId = sh(
                         returnStdout: true,
                         script: "tito release --debug --yes --test aos-${BUILD_VERSION} | grep 'Created task:' | awk '{print \$3}'"
                     )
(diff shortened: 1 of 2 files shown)


pipeline-scripts/commonlib.groovy
```diff
@@ -85,6 +85,24 @@ def cleanCommaList(str) {
     return str.replaceAll(',', ' ').split().join(',')
 }
 
+def retry(tries=3, Closure it) {
```
tbielawa: Are you using this custom `retry` function instead of the stdlib `retry` because of anything specific (the closure?)

sosiouxme:
@adammhaile @tbielawa this should address https://jira.coreos.com/browse/ART-344

pipeline-scripts/commonlib.groovy
adammhaile: @tbielawa the stdlib retry has a super annoying syntax IMHO... don't really blame him.

sosiouxme: No, I just wasn't aware it existed. Should have suspected :) All mine adds is a bit of output that's not too useful. I'll remove it.

sosiouxme: Updated...

PR #1720, opened 2019-03-01
commonlib: separate major versions

sosiouxme:
@adammhaile this is what I was thinking for doing different things for the different major versions. Fairly straightforward...


sosiouxme:
Before too long, it may make sense to rename `build/ocp` to `build/ocp3`. That will be clearer but we'll lose the build history, so I'm not super eager to do it.

adammhaile:
@sosiouxme agree on the rename, but I feel like after 4.0 GA is a more proper time.

PR #1721, opened 2019-03-02
use commonlib.email to archive and suppress email

sosiouxme:
Builds on https://github.com/openshift/aos-cd-jobs/pull/1720/ so first commit is from that.

This provides a wrapper method that archives emails with the other job artifacts. It also makes it easy to suppress sending the actual emails (and marks builds that do this so that we don't wonder why email didn't show up). Together, this makes it easy to test the jobs without having to consult email, or running the risk of emailing the team. It also means we can see what emails were actually sent even though we're not the recipients.

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -64,6 +64,7 @@ node {
                         $class: 'hudson.model.BooleanParameterDefinition',
                         defaultValue: false
                     ],
+                    commonlib.suppressEmailParam(),
                     [
                         name: 'MAIL_LIST_SUCCESS',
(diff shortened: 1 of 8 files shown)


tbielawa:
I can't quite tell from my cursory glance so I'm going to ask. Did you make sure this PR doesn't change any functions shared by other code in here? We have maintenance jobs and other tasks still running on our jenkins and I wouldn't want to interrupt them.

Thanks!

sosiouxme:
@tbielawa I do try to keep that in mind. I don't think it's a problem for this one -- all the changes are either net new, or to stuff I recently introduced for our own scripts.

PR #1725, opened 2019-03-04
ART-268 email owners when their image build fails

sosiouxme:
Builds on https://github.com/openshift/aos-cd-jobs/pull/1724 so only last three commits are new.

When image builds fail, collect the container logs and email the owners about what happened. Unless a whole lot of images fail; in that case probably something else is going on and we shouldn't spam them.

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -203,14 +203,29 @@ node {
                 if (!any_images_to_build) { return }
                 command = "--working-dir ${doozer_working} --group 'openshift-${params.BUILD_VERSION}' "
                 command += "${include_exclude} images:build --push-to-defaults --repo-type ${repo_type} "
-                buildlib.doozer command
+                try {
+                    buildlib.doozer command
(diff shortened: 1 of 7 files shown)


sosiouxme:
/assign @adammhaile @tbielawa 

I won't hide that I'm pretty pleased with this. See [an example of the email sent](https://buildvm.openshift.eng.bos.redhat.com:8443/job/hack/job/lmeyer-aos-cd-builds/job/build%252Fcustom/40/artifact/email/email001-aos_team_art@redhat.com-Failed_OCP_build_of_openshift3_ose_base_v3_8_44.txt).

You might wonder why the copy-pasta in the Jenkinsfiles determining whether to email or not. I expect we'll need to do a certain amount of tweaking, balancing how much we spam with trying to be informative. I expect this might vary per job, as well as how we want to respond. I could not think of a way of reconciling "DRY code" and "configurable per job" without it getting complicated or unintuitive. Perhaps something suitably simple will come to me later. For now, the duplicate logic is straightforward.

PR #1726, opened 2019-03-04
bypass extending s2i from push_jenkins_image (copy needed bits from s…

gabemontero:
…2i, but no longer build s2i)

/assign @bparees

### DIFF
--- a/sjb/config/test_cases/push_jenkins_images.yml
+++ b/sjb/config/test_cases/push_jenkins_images.yml
@@ -1,63 +1,114 @@
 ---
-parent: 'test_cases/test_branch_jenkins_images.yml'
-overrides:
-  email:
+parameters:
+  - name: BUILD_ID
+    description: "Unique build number for each run."
+  - name: REPO_OWNER
+    description: "GitHub org that triggered the job."
+  - name: REPO_NAME
+    description: "GitHub repo that triggered the job."
+  - name: PULL_BASE_REF
+    description: "Ref name of the base branch."
+  - name: PULL_BASE_SHA
+    description: "Git SHA of the base branch."
+  - name: PULL_REFS
+    description: "All refs to test."
+  - name: PULL_NUMBER
+    description: "Pull request number."
+  - name: PULL_PULL_SHA
+    description: "Pull request head SHA."
+  - name: JOB_SPEC
+    description: "JSON form of job specification."
+  - name: PROW_JOB_ID
+    description: "The ID that prow sets on a Jenkins job in order to correlate it with a ProwJob."
+provision:
+  os: "rhel"
+  stage: "base"
+  provider: "aws"
+email:
     - bparees@redhat.com
     - gmontero@redhat.com
     - cdaley@redhat.com
     - obulatov@redhat.com
     - agladkov@redhat.com
     - adkaplan@redhat.com
-extensions:
-  actions:
-    - type: "script"
-      title: "create directory for docker config"
-      script: |-
-        mkdir -p /tmp/.docker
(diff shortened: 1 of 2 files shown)


sjb/config/test_cases/push_jenkins_images.yml
```diff
@@ -1,63 +1,114 @@
 ---
-parent: 'test_cases/test_branch_jenkins_images.yml'
```
bparees: doesn't this mean test_branch_jenkins is still going to be broken, which means PR tests will still fail?

gabemontero: yep I need to move the s2i bits (minus building of s2i) into test_branch_jenkins

will reswizzle in a bit

gabemontero: well ... just realized a few things ... depends on which release branch I think
- 3.11 and 4.0 use ci-operator and not aos-cd-jobs
- we are generally only updating those two for security advisories ...
- ... but for say plugin/image fixes outside of that, we'd use test_branch_jenkins ..
- but we'd also use the unit tests and s2 there ... 

So probably just need to keep the s2i build in play, and update the golang

agree @bparees ?

sjb/generated/push_jenkins_images.xml
```diff
@@ -123,106 +113,6 @@ oct provision remote all-in-one --os &#34;rhel&#34; --stage &#34;base&#34; --pro
...
-SCRIPT
-chmod +x &#34;${script}&#34;
-scp -F ${WORKSPACE}/.config/origin-ci-tool/inventory/.ssh_config &#34;${script}&#34; openshiftdevel:&#34;${script}&#34;
-ssh -F ${WORKSPACE}/.config/origin-ci-tool/inventory/.ssh_config -t openshiftdevel &#34;bash -l -c \&#34;timeout 10800 ${script}\&#34;&#34; </command>
-        </hudson.tasks.Shell>
-        <hudson.tasks.Shell>
-          <command>#!/bin/bash
-SCRIPT_START_TIME=&#34;$( date +%s )&#34; &amp;&amp; export SCRIPT_START_TIME &amp;&amp; echo &#34;########## STARTING STAGE: TEST THE JENKINS IMAGES ##########&#34; &amp;&amp; trap &#39;export status=FAILURE&#39; ERR &amp;&amp; trap &#39;set +o xtrace; SCRIPT_END_TIME=&#34;$( date +%s )&#34;; ELAPSED_TIME=&#34;$(( SCRIPT_END_TIME - SCRIPT_START_TIME ))&#34;; echo &#34;########## FINISHED STAGE: ${status:-SUCCESS}: TEST THE JENKINS IMAGES [$( printf &#34;%02dh %02dm %02ds&#34; &#34;$(( ELAPSED_TIME/3600 ))&#34; &#34;$(( (ELAPSED_TIME%3600)/60 ))&#34; &#34;$(( ELAPSED_TIME%60 ))&#34; )] ##########&#34;&#39; EXIT &amp;&amp; set -o errexit -o nounset -o pipefail -o xtrace &amp;&amp; if [[ -s &#34;${WORKSPACE}/activate&#34; ]]; then source &#34;${WORKSPACE}/activate&#34;; fi
```
bparees: looks like we used to actually test the jobs prior to pushing them.

bparees:
@gabemontero given that this job used to actually test the images before pushing them, i'm now disinclined to merge it as is.  let's just fix the s2i build step to install a newer golang.

gabemontero:
10-4 @bparees

PR #1722, opened 2019-03-02
logging - use ansible vars.yaml files for branch specific params

richm:
This assumes the path for logging code under test is
/data/src/github.com/openshift/origin-aggregated-logging/

The file ci-ansible-common.yaml is for parameters common to
all logging jobs for this branch.
The file ci-ansible-journald.yaml is for `docker --log-driver=journald`
specific configuration.
The file ci-ansible-json-file.yaml is for `docker --log-driver=json-file`
specific configuration.

This requires https://github.com/openshift/origin-aggregated-logging/pull/1530

PR #1729, opened 2019-03-05
openshift-ansible - use ansible evars vars.yaml for branch specific params

patrickdillon:
This PR is based on #1722 

We are seeing release-3.9 jobs fail because they need infra nodes, which were removed in #1688. As all jobs in sjb use the same inventory, we can use this method of passing evars linked to a vars file in the target repo. 

This PR depends on openshift/openshift-ansible#11286


patrickdillon:
openshift-ansible pr has merged, please merge this asap

PR #1732, opened 2019-03-05
Add release-3.9 specific ansible config

vrutkovs:
This should fix 3.9 jobs in origin. It sets `openshift_node_labels` and a custom `openshift_dependencies_repo`

PTAL @stevekuznetsov @patrickdillon

### DIFF
--- a/sjb/inventory/release-3.9.cfg
+++ b/sjb/inventory/release-3.9.cfg
@@ -1,0 +1,5 @@
+openshift_node_labels:
+  region: infra
+  zone: default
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
How is this loaded conditionall?

patrickdillon:
@stevekuznetsov https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/common/test_cases/origin_built_installed_release.yml#L73-L77

vrutkovs:
Right, found it in the job logs. Its also the reason why passing custom evars before doesn't really work

PR #1741, opened 2019-03-07
Revert "Merge pull request #1733 from vrutkovs/3.9-disable-centos-repo"

patrickdillon:
This commit was intended to fix the gcp job, but had a side effect of breaking install job, so we would like to revert.

This reverts commit 7e1dfd65108dbbf034dc0fa2e26b2b101744733d, reversing
changes made to 46a12315e806d5954b738cb37504cf687aa9acb2.

/cc vrutkovs


vrutkovs:
The sideeffect which broke 3.9 install jobs was that a repo with openvswitch was not installed

PR #1748, opened 2019-03-08
custom only syncs for 4.x, build-sync updates description of progress

tbielawa:
custom only syncs for 4.x, build-sync updates description of progress

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -91,6 +91,8 @@ node {
     buildlib.initialize(false)
     GITHUB_BASE = "git@github.com:openshift" // buildlib uses this global var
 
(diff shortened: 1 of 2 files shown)


jobs/build/custom/Jenkinsfile
```diff
@@ -91,6 +91,8 @@ node {
     buildlib.initialize(false)
     GITHUB_BASE = "git@github.com:openshift" // buildlib uses this global var
 
+    majorVersion = params.BUILD_VERSION.split('.')[0]
+    minorVersion = params.BUILD_VERSION.split('.')[1]
```
sosiouxme: We should really standardize this in commonlib somehow. But for now, whatever.

tbielawa: I agree.

PR #1751, opened 2019-03-11
build/ocp(4) -> consolidate to buildlib

sosiouxme:
Moving some bits that were cloned, really are the same, and may be reusable into buildlib.


sosiouxme:
PTAL @tbielawa @adammhaile

tbielawa:
I like the way this looks.

tbielawa:
/lgtm
I assume you can lint this yourself :-)

sosiouxme:
I had to fix in ocp4 where I referred to `oseTaskId` vs `OSE_TASK_ID`.

I do not follow the lint comment? Just dense, sorry

adammhaile:
@sosiouxme I think @tbielawa is telling you to use the jenkins linter on it.

pipeline-scripts/buildlib.groovy
```diff
@@ -873,6 +873,66 @@ def get_distgit_notify( record_log ) {
...
+            github_url = github_url.replaceFirst(":", "/")
+            dockerfile_sub_path = val['source_dockerfile_subpath']
+            dockerfile_url = "Upstream source file: https://" + github_url + "/blob/" + sourceBranches[alias] + "/" + dockerfile_sub_path
+            // always mail success list, val.owners will be comma delimited or empty
+            commonlib.email(
+                to: "aos-team-art@redhat.com,${val.owners}",
+                from: "aos-cicd@redhat.com",
+                subject: "${val.image} Dockerfile reconciliation for OCP v${params.BUILD_VERSION}",
+                body: """
+Why am I receiving this?
```
adammhaile: Is there a way in jenkins groovy to have a text constants file? I mean... we could just have textconsts.groovy and import it inside buildlib.groovy I guess. Multi-line strings in groovy are terrible.
@sosiouxme I realize this is not your fault... it was like this before. Just pontificating.

sosiouxme: yes, but using a resource bundle introduces complications too. typically there are things you want to format in. then you have to move back and forth between the locations when you want to change anything. I feel that embedding the text in/near the code is the lesser evil. may improve things if we extract the text to a constant right before the method using it, and use formatting instead of interpolation.

PR #1752, opened 2019-03-12
Adding 4.1 and now pushes

adammhaile:
@sosiouxme @tbielawa 
Note: I checked and git merge as well as git push automatically no-op if nothing changed. So no extraneous pushes will occur.

### DIFF
--- a/jobs/build/merge_ocp/Jenkinsfile
+++ b/jobs/build/merge_ocp/Jenkinsfile
@@ -19,7 +19,7 @@ node {
                         name: 'VERSIONS',
                         description: 'CSV list of versions to run merge on.',
                         $class: 'hudson.model.StringParameterDefinition',
-                        defaultValue: "3.9,3.10,3.11,4.0"
+                        defaultValue: commonlib.ocpMergeVersions.join(',')
                     ],
                     commonlib.suppressEmailParam(),
                     [
@@ -49,7 +49,7 @@ node {
 
     SSH_KEY_ID = "openshift-bot"
     MERGE_VERSIONS = VERSIONS.split(',')
-    CURRENT_MASTER = "4.0"
+    CURRENT_MASTER = commonlib.ocp4MasterVersion
 
     try {
         sshagent([SSH_KEY_ID]) {
--- a/jobs/build/merge_ocp/merge_ocp.sh
+++ b/jobs/build/merge_ocp/merge_ocp.sh
@@ -29,6 +29,6 @@ echo 'pkg/assets/bindata.go merge=ours' >> .gitattributes
 echo 'pkg/assets/java/bindata.go merge=ours' >> .gitattributes
 git merge -m "Merge remote-tracking branch ${UPSTREAM_SOURCE_BRANCH}" "upstream/${UPSTREAM_SOURCE_BRANCH}"
 
-# git push # probabaly don't run this yet
+git push
 
 popd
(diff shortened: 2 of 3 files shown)


jobs/build/merge_ocp/Jenkinsfile
```diff
@@ -49,7 +49,7 @@ node {
 
     SSH_KEY_ID = "openshift-bot"
     MERGE_VERSIONS = VERSIONS.split(',')
-    CURRENT_MASTER = "4.0"
+    CURRENT_MASTER = "4.1"
```
tbielawa: I think we should look at moving this kind of value into `commonlib` or something equivalent.

jobs/build/merge_ocp/Jenkinsfile
```diff
@@ -19,7 +19,7 @@ node {
                         name: 'VERSIONS',
                         description: 'CSV list of versions to run merge on.',
                         $class: 'hudson.model.StringParameterDefinition',
-                        defaultValue: "3.9,3.10,3.11,4.0"
+                        defaultValue: "3.9,3.10,3.11,4.0,4.1"
```
tbielawa: These are the current supported building versions right? This feels like it should be in a `commonlib` type place.

adammhaile: Fair. I'll see what I can do.

sosiouxme: Do we actually need to merge 4.1, as long as we're building 4.1 from master? I would rather merge master->master and not merge 4.1 until master becomes 4.2

adammhaile: That's what setting `CURRENT_MASTER = "4.1"` does... 4.1 will be master -> master

jobs/build/merge_ocp/merge_ocp.sh
```diff
@@ -29,6 +29,6 @@ echo 'pkg/assets/bindata.go merge=ours' >> .gitattributes
 echo 'pkg/assets/java/bindata.go merge=ours' >> .gitattributes
 git merge -m "Merge remote-tracking branch ${UPSTREAM_SOURCE_BRANCH}" "upstream/${UPSTREAM_SOURCE_BRANCH}"
```
sosiouxme: This is already a no-op when there's nothing new. I was confused because of what the build job does -- creates and extra commit even if it's empty. So anyway :+1:

jobs/build/merge_ocp/Jenkinsfile
sosiouxme: yeah, nevermind me :)

sosiouxme: they're not the current supported building versions; they're the versions that still merge from origin. i don't think we will need to know that anywhere else any longer.

jobs/build/merge_ocp/merge_ocp.sh
adammhaile: Yep... should've clarified there. It's the build job that's adding that :)

adammhaile:
Alright @tbielawa & @sosiouxme 
Although I agree with @sosiouxme that the merge versions aren't really needed anywhere else, I agree more that it's the type of thing that is better in commonlib. If we ever need that info elsewhere then we're good now.
Also, current master version is also in commonlib now.

PR #1755, opened 2019-03-13
build/ocp4: unblock building 4.y with openshift package and new versioning

sosiouxme:
This is basically temporary (a good stopping point while I go about dismantling most of this and turning it into usually-incremental builds) with fixes to allow us to build at all, pretty similarly to before. Even that much turns out to be kind of involved, but keep in mind a lot of it probably goes away or transforms in the end.

Temporary home at https://buildvm.openshift.eng.bos.redhat.com:8443/job/hack/job/lmeyer-stage/job/build%252Focp4/ until merged.
Can also hack around at https://buildvm.openshift.eng.bos.redhat.com:8443/job/hack/job/lmeyer-dev/job/build%252Focp4/ while real builds are happening at the other.

### DIFF
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -8,61 +8,6 @@ def get_mirror_url(build_mode, version) {
     return "https://mirror.openshift.com/enterprise/enterprise-${version}"
 }
 
-def get_changelog(rpm_name, record_log) {
-    //
-    // INPUTS:
-    //   rpm_name - the name of an RPM build previously
-    //   record_log - an array of build records with | separated fields
-
-    rpm_builds = record_log['build_rpm']
-    if (rpm_builds == null || rpm_builds.size() == 0) {
-        return ""
-    }
-
-    // filter for the desired RPM using name
-    build_record_index = rpm_builds.findIndexOf {
-        it['rpm'] == rpm_name
-    }
-    if (build_record_index == -1) {
-        return ""
-    }
-    build_record = rpm_builds[build_record_index]
-
-    // then get the task_id and task_url out of it
-    // task_id = build_record['task_id']
-    task_url = build_record['task_url']
-
-    // get the build ID from the web page
-    // there must be an API way to do this MAL 20180622
-    try {
-        build_id = sh(
-            returnStdout: true,
-            script: [
-                "curl --silent --insecure ${task_url}",
-                "sed -n -e 's/.*buildID=\\([0-9]*\\).*/\\1/p'"
-            ].join(" | ")
-        ).trim()
-    } catch (err) {
-        error("failed to retrieve task page from brew: ${task_url}")
-    }
-
-    // buildinfo can return the changelog.  Return just the text after
-    // the Changelog: line
-    try {
-        changelog = sh(
-            returnStdout: true,
-            script: [
-                "brew buildinfo ${build_id} --changelog",
-                "sed -n '/Changelog/,\$p'"
-            ].join(' | ')
-        ).trim()
-    } catch (err) {
-        error "failed to get build info and changelog for build ${build_id}"
-    }
-
-    return changelog
-}
-
 def mail_success(version, mirrorURL, record_log, commonlib) {
 
     def target = "(Release Candidate)"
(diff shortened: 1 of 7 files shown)


sosiouxme:
@adammhaile @tbielawa PTAL
@smunilla not sure how interested you'll be in the job changes but heads up that ocp4 will be producing things that look different.

I suggest reviewing a commit at a time, ignoring whitespace changes (there was quite a lot of moving around and re-indenting).

pipeline-scripts/buildlib.groovy
```diff
@@ -576,6 +586,20 @@ def mock_merge_driver(repo_dir, files) {
     }
 }
 
+/**
+ * Clones ose (with origin as 'upstream')
+ * Sets OSE_DIR to ose repo directory
+ * @return Returns the directory where ose is cloned
+ */
+def initialize_ose_4() {
```
adammhaile: We should not need to do this anymore. If we add the openshift rpm to ocp-build-data this will all be automatically cloned by doozer.

jobs/build/ocp4/Jenkinsfile
```diff
@@ -477,65 +384,9 @@ node {
...
-                        ver_fields[ver_fields.size() - 1] = "${ver_fields[ver_fields.size() - 1].toInteger() + 1}"
-                        NEW_VERSION = ver_fields.join(".")
-                        NEW_RELEASE = "1"
-                        NEW_DOCKERFILE_RELEASE = NEW_RELEASE
-
-                    } else {
-                        error("Unknown BUILD_MODE: ${BUILD_MODE}")
-                    }
+                    NEW_VERSION = params.NEW_VERSION ?: spec.version
+                    NEW_RELEASE = new Date().format("yyyy.MM.dd.HH.mm")
```
adammhaile: Do we still have concerns about the release field having leading zeros in them? Or was that resolved?

jobs/build/ocp4/Jenkinsfile
```diff
@@ -546,38 +397,43 @@ node {
             }
 
             stage("merge origin") {
+                // note, this will go away once merge is handled in its own job
```
adammhaile: It already is. Was merged in yesterday :)

jobs/build/ocp4/Jenkinsfile
```diff
@@ -546,38 +397,43 @@ node {
...
             }
 
             stage("ose tag") {
                 dir(OSE_DIR) {
+                    // we no longer want to push tito's changes (and tags) to ose.
+                    // but push a tag for tito to generate the next changelog against and to match historical practice.
+                    def tag = "v${NEW_VERSION}-${NEW_RELEASE}"
+
+                    sh "git tag -am '' ${tag}"
+                    if (!IS_TEST_MODE) {
```
adammhaile: Still concerned about pushing anything back to OSE, I do not believe we should be. Especially if we build the RPM with doozer.

jobs/build/ocp4/Jenkinsfile
sosiouxme: Roger, I'll take that out

jobs/build/ocp4/Jenkinsfile
sosiouxme: Updated in later commit

sosiouxme:
Basically, I discovered ocp4 does more than just build things, and I wanted to make sure that kept working while determining what we don't need any more and working on the full transition to incremental.

As I put above, this is not the final solution. It's a stopgap.

jobs/build/ocp4/Jenkinsfile
sosiouxme: We have been pushing back. It doesn't look like we'll need to continue doing that. But this is not the final solution, this is reviving the ocp4 job for now.

pipeline-scripts/buildlib.groovy
sosiouxme: yes but we use it to populate the ose alias which I think is still a good idea for efficiency. or does doozer do only one ose checkout for multiple builds that use it? I think maybe you made it that smart but haven't checked lately :)

Of course not cloning the entire repo would improve efficiency too.

adammhaile: Yeah... just pull the one file you need directly and use that. Then let doozer do the full clone and build.

sosiouxme:
It was actually pretty easy to stop doing anything with the ose repo and just let doozer build it. So this is way simpler now.

PR #1757, opened 2019-03-14
build/incremental: fix and update

sosiouxme:
Fixes a scoping bug that crept in.
Now looks at openshift pkg for version if not given.
Uses a timestamp for default release.


sosiouxme:
PTAL @tbielawa @vikaslaad @adammhaile 
BTW it does sync images at the end. Probably a good idea to start running this for 4.0 regularly :)

sosiouxme:
Soon, yes. But for now this job is completely broken, this is a simple fix, and the converged ocp4/incremental job is not ready yet.

sosiouxme:
This is harmless and will fix a totally broken job. Can we merge this today and start running some of these?

PR #1758, opened 2019-03-15
build/custom: no need to check out ose repo at all

sosiouxme:
Let doozer do it as needed.

PTAL @adammhaile @tbielawa


sosiouxme:
3.x builds still use it :)

PR #1760, opened 2019-03-20
commonlib.shell wrapper for useful shell errors

sosiouxme:
"Script exited with result 1" isn't a very useful message for the end
user. Jenkins sh() step is woefully lacking in options for capturing
what happened. This fixes those problems by wrapping sh commands to
capture output and stderr and provide a meaningful error message when
there is an error.

Here be much devilry, but I'll take it for the signal boost in errors. See the build description and the shell/email artifacts at https://buildvm.openshift.eng.bos.redhat.com:8443/job/hack/job/lmeyer-stage/job/build%252Fcustom/2/ to see how much more context we can get.

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -122,7 +122,7 @@ node {
 
             stage("puddle: ose 'building'") {
                 if (rpms.toUpperCase() != "NONE") {
-                    AOS_CD_JOBS_COMMIT_SHA = sh(
+                    AOS_CD_JOBS_COMMIT_SHA = commonlib.shell(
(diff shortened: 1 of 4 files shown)


sosiouxme:
PTAL @adammhaile @tbielawa @smunilla

sosiouxme:
Once this is in places and expanded to wrap all shell commands, that will give us a fun next step: record all of our shell interactions with timings and results. Then upload them to elastic, or prometheus, or whatever, for analysis.

I didn't wrap all shell commands yet because it seemed wise to start with a few workhorses in case there are bugs when this sees real usage.

PR #1761, opened 2019-03-21
prevent silly errors on dockerfile reconciliation notices

sosiouxme:
build/ocp4 still had a reference to `OSE_SOURCE_BRANCH` though it no longer knows about source branches. Only shows up when there's a reconciliation notice to be sent. This torpedoed my build last night: https://buildvm.openshift.eng.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Focp4/33/console

Also, reconciliation notices break if they're not for repos that have an alias. So basically they're failing all the time, and the only notice we get is a message buried in the console logs, e.g. search for "Failure resolving alias for email" in https://buildvm.openshift.eng.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Focp/1055/consoleFull

So, this fixes that.


tbielawa:
I like bugfixes
/lgtm

PR #1762, opened 2019-03-26
build/ocp4: converged incremental or full build

sosiouxme:
For 4.x this builds anything from one component to changed components to all components. My plan is to schedule it as frequently as possible with the default, which is to build changed. Then schedule it once a week or so with a force build so everything is built (this might include a version bump for releases).

This also demos what I hope become standards for designing our jobs (in particular, the Jenkinsfile that is largely just structure calling out to a module; the reduction of global variables down to a few well-known Jenkins-related ones plus module state; adding info to build description/displayName as the build progresses...).

@tbielawa @adammhaile @smunilla @shiywang @joeldavis84 @vikaslaad

### DIFF
--- a/jobs/build/ocp4/build.groovy
+++ b/jobs/build/ocp4/build.groovy
@@ -1,0 +1,604 @@
+#!/usr/bin/env groovy
+
+buildlib = load("pipeline-scripts/buildlib.groovy")
+commonlib = buildlib.commonlib
+
+// Properties that should be initialized and not updated
+version = [
+    stream: "",     // "X.Y" e.g. "4.0"
+    full: "",       // e.g. "4.0.0"
+    release: "",    // e.g. "201901011200"
+    major: 0,       // X in X.Y, e.g. 4
+    minor: 0,       // Y in X.Y, e.g. 0
+]
+doozerWorking = "${env.WORKSPACE}/doozer_working" // must be in WORKSPACE to archive artifacts
+doozerOpts = "--working-dir ${doozerWorking}"
+
+// this plan is to be initialized but then adjusted for incremental builds
+buildPlan = [
+    dryRun: false, // report build plan without performing it
+    forceBuild: false, // build regardless of whether source has changed
+    buildRpms: false,
+    rpmsIncluded: "", // comma-separated list
+    rpmsExcluded: "", // comma-separated list
+    puddleConf: "",  // URL to download the puddle conf
+    buildImages: false,
+    imagesIncluded: "", // comma-separated list
+    imagesExcluded: "", // comma-separated list
+]
+// initialized but composeDir only filled in when puddle creates a compose
+rpmMirror = [       // where to mirror RPM compose
+    composeDir: "", // e.g. "YYYY-MM-DD.1"
+    url: "",
+]
+
+// some state to record and preserve between stages
+finalRpmVersionRelease = ""  // set before building a compose
+
+/**
+ * Initialize properties from Jenkins parameters.
+ * @return map which is the buildPlan property of this build.
+*/
+def initialize() {
+    GITHUB_BASE = "git@github.com:openshift"  // buildlib uses this :eyeroll:
+
+    currentBuild.displayName = "#${currentBuild.number} - ${params.BUILD_VERSION}.??"
+    echo "Initializing build: ${currentBuild.displayName}"
+
+    version.stream = params.BUILD_VERSION.trim()
+    version << determineBuildVersion(version.stream)
+    doozerOpts += " --group 'openshift-${version.stream}'"
+
+    buildPlan << [
+        dryRun: params.DRY_RUN,
+        forceBuild: params.FORCE_BUILD,
+        buildRpms: params.BUILD_RPMS != "none",
+        buildImages: params.BUILD_IMAGES != "none",
+    ]
+
+    // determine whether the user wanted to specify includes or excludes
+    rpmList = [only: "rpmsIncluded", except: "rpmsExcluded"][params.BUILD_RPMS]
+    if (rpmList) {
+        buildPlan[rpmList] = commonlib.cleanCommaList(params.RPM_LIST)
+    } else if (params.RPM_LIST.trim()) {
+        error("aborting because a list of RPMs was specified; you probably want to specify only/except.")
+    }
+
+    imageList = [only: "imagesIncluded", except: "imagesExcluded"][params.BUILD_IMAGES]
+    if (imageList) {
+        buildPlan[imageList] = commonlib.cleanCommaList(params.IMAGE_LIST)
+    } else if (params.IMAGE_LIST.trim()) {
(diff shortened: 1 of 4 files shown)


jobs/build/ocp4/build.groovy
```diff
@@ -0,0 +1,595 @@
...
+            return buildPlan
+        }
+        report "Found ${changed.images.size()} image(s) with changes:\n  " + changed.images.join("\n  ")
+
+        // also determine child images of changed
+        yamlData = readYaml text: buildlib.doozer(
+            """
+            ${doozerOpts}
+            ${includeExclude "images", buildPlan.imagesIncluded, buildPlan.imagesExcluded}
+            images:show-tree --yml
```
adammhaile: haha... I had no idea this was used for anything real. <3

jobs/build/ocp4/build.groovy
```diff
@@ -0,0 +1,595 @@
...
+ * @param stream: OCP minor version "X.Y"
+ * @return a map to merge into the "version" property representing the determined version
+ */
+def determineBuildVersion(stream) {
+    def full = "${stream}.0"  // default
+    def release = new Date().format("yyyyMMddHHmm")
+
+    def prevBuild = retry(3) {
+        commonlib.shell(
+            script: "brew latest-build --quiet rhaos-${stream}-rhel-7-candidate openshift | awk '{print \$1}'",
```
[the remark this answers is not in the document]
sosiouxme: Yes, I think that is a great idea. Since doozer already manages ocp-build-data it seems natural for the job to be able to query doozer for group data, rather than have the job read it directly. Especially since this is templated. Do you want to work on providing that from doozer?

This can remain hardcoded for now and be improved at our leisure.

sosiouxme:
Also I think that you're right that we can allow concurrency on this job and just prevent concurrency on the per-stream jobs.

It's worth noting that we used to run product builds in serial for all versions. Since splitting out ocp4 there's nothing keeping v3 and v4 jobs from both running at the same time. This means they'll be contending for the same OSBS capacity. I don't have a good solution in mind other than scheduling full builds far enough apart and minding current/scheduled activity when initiating builds. And increasing OSBS capacity so it's not an issue.

jobs/build/ocp4/build.groovy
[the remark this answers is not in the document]
adammhaile: Yep -> https://jira.coreos.com/browse/ART-535

jobs/build/ocp4/build.groovy
sosiouxme: :laughing: it's now super-critical!

jobs/build/ocp4/build.groovy
```diff
@@ -0,0 +1,606 @@
...
+
+    def changed = [:]
+    try {
+        def yamlData = readYaml text: buildlib.doozer(
+            """
+            ${doozerOpts}
+            ${includeExclude "rpms", buildPlan.rpmsIncluded, buildPlan.rpmsExcluded}
+            ${includeExclude "images", buildPlan.imagesIncluded, buildPlan.imagesExcluded}
+            config:scan-sources --yaml
+            """, [capture: true]
```
tbielawa: You're depending on doozer never accidentally spitting out cruft text to `stdout` here. I recall a recent conversation where this happened, doozer spit out some garbage into the yaml stream. I'd suggest to consider parsing from the output yaml file but it looks like doozer doesn't give the choice to `-o something.yaml` yet. I guess this is just something to keep in mind.

sosiouxme: Yes. If doozer is broken it might do that even when it's supposed to output yaml. I was thinking one of our automated doozer tests should try a few commands that have easily-defined output and ensure that's not happening.

sosiouxme: Although I suppose `-o foo.yaml` would be a more robust solution.

jobs/build/ocp4/build.groovy
```diff
@@ -0,0 +1,606 @@
...
+    buildPlan.puddleConf = "${puddleConfBase}/atomic_openshift-${version.stream}.conf"
+    // and where to mirror the compose when it's done
+    rpmMirror.url = "https://mirror.openshift.com/enterprise/enterprise-${version.stream}"
+
+    // adjust the build "title"
+    currentBuild.displayName = "#${currentBuild.number} - ${version.full}-${version.release}"
+    if (buildPlan.dryRun) { currentBuild.displayName += " [DRY RUN]" }
+    if (buildPlan.forceBuild) { currentBuild.displayName += " [force build]" }
+    if (!buildPlan.buildRpms) { currentBuild.displayName += " [no RPMs]" }
+    if (!buildPlan.buildImages) { currentBuild.displayName += " [no images]" }
```
tbielawa: This is nice

jobs/build/ocp4/build.groovy
```diff
@@ -0,0 +1,606 @@
...
+        ${includeExclude "images", buildPlan.imagesIncluded, buildPlan.imagesExcluded}
+        images:rebase --version v${version.full} --release ${version.release}
+        --message 'Updating Dockerfile version and release v${version.full}-${version.release}' --push
+        """
+    if(buildPlan.dryRun) {
+        echo "doozer ${cmd}"
+        return
+    }
+    buildlib.doozer(cmd)
+    // TODO: if rebase fails for required images, notify image owners, and still notify on other reconciliations
```
tbielawa: Don't want to forget about these 2 TODOs. That's referring to how we are notifying image owners presently right?

tbielawa:
Wow there's a lot to take in here.

* `-390` / `+70` on `ocp4/Jenkinsfile` is a big :+1: from me
* You moved a lot of stage logic into another groovy file, I guess there's advantages to that (I assume testing?) and it seems to have worked out well. Cool!
* All the new options seem really shiny and nice. 
* I'd like to see a `version dumps` stage in a later version, as is [done in build-sync](https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/build-sync/Jenkinsfile#L50). That can wait until later though. I should make that into a *lib function instead so we don't have to copy pasta all the time.
* Adding the dry run stuff is a HUGE win

This seems pretty legit to me. Lots of value here (not being facetious either). I like it!

/lgtm

jobs/build/ocp4/build.groovy
sosiouxme: Right. It requires a little doozer work that I haven't prioritized yet, because right now if `doozer images:rebase` fails for a required image, then we stop the build, and so notifications don't go out for what was successfully rebased.

I put it off because I don't think reconciliation notices are that critical.

sosiouxme:
Main advantage to having logic in separate `build.groovy` is that it allows freedom to keep state and define methods where it makes sense. You can't define methods inside a `node {}` block and so we'd have a bunch of helper methods stuck up at the top before you got to the job definition, which to me is annoying.

The other advantage is you can look at the Jenkinsfile and have a pretty concise summary of the execution flow...

PR #1765, opened 2019-03-29
buildlib: wait before re-submitting brew tasks

sosiouxme:
On immediate re-submit, brew seems to either do nothing and return success, or complain that the task is still open. Wait for it to get sorted out before re-submitting.


sosiouxme:
@adammhaile @tbielawa PTAL

PR #1771, opened 2019-04-01
build/ocp: watch/retry o-a/a-o builds in serial for now

sosiouxme:
commonlib.shell turns out not to be safe for concurrency.


sosiouxme:
@tbielawa @adammhaile @smunilla PTAL

PR #1772, opened 2019-04-02
build-sync: Parameterize selecting base image stream file

tbielawa:
Instead of using a hard-coded path to a base image stream, substitute
in the build version parameter for selection.

Now we'll be able to push to the 4.0 and 4.1 image streams separately.

On the buildvm there is now a new base image stream. In a future PR we should do something to remove that dependency and generate them from inputs.

### DIFF
--- a/jobs/build/build-sync/Jenkinsfile
+++ b/jobs/build/build-sync/Jenkinsfile
@@ -35,7 +35,7 @@ node {
     // Location of the image stream to apply
     def ocIsObject = "${mirrorWorking}/release-is.yaml"
     // Locally stored image stream stub
-    def baseImageStream = "/home/jenkins/base-art-latest-imagestream.yaml"
+    def baseImageStream = "/home/jenkins/base-art-latest-imagestream-${params.BUILD_VERSION}.yaml"
     // Kubeconfig allowing ART to interact with api.ci.openshift.org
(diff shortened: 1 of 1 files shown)


jobs/build/build-sync/Jenkinsfile
```diff
@@ -35,7 +35,7 @@ node {
     // Location of the image stream to apply
     def ocIsObject = "${mirrorWorking}/release-is.yaml"
     // Locally stored image stream stub
-    def baseImageStream = "/home/jenkins/base-art-latest-imagestream.yaml"
+    def baseImageStream = "/home/jenkins/base-art-latest-imagestream-${params.BUILD_VERSION}.yaml"
```
adammhaile: If you need a place to write to you should move this to `/mnt/workspace`. Create a dir in there if you want just remember to `sudo chown jenkins:jenkins` it as the base dir is root owned

tbielawa: I want to do that kind of thing in a followup PR.

tbielawa: (For reference, both the 4.0 and 4.1 files exist already in the path I modify in this PR)

adammhaile: ack... sounds good

PR #1773, opened 2019-04-02
Add 4.1 RHEL8 puddle conf

tbielawa:
Enables RHCOS to build using 4.1 content built in RHEL8 buildroots

### DIFF
--- a/build-scripts/puddle-conf/atomic_openshift-4.1.el8.conf
+++ b/build-scripts/puddle-conf/atomic_openshift-4.1.el8.conf
@@ -1,0 +1,17 @@
+[puddle]
+type = simple
+rootdir = /mnt/rcm-guest/puddles/RHAOS
+mashroot = /mnt/rcm-guest/puddles/RHAOS/mash
+brewroot = file:///mnt/redhat/brewroot
+topurl = http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/
+name = AtomicOpenShift.el8
+version = 4.1
+emails = aos-devel@redhat.com,aos-qe@redhat.com
+announcer = openshift-puddle <aos-devel@redhat.com>
+publish = no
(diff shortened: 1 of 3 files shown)


build-scripts/puddle-conf/atomic_openshift-4.1.el8.conf
```diff
@@ -0,0 +1,17 @@
+[puddle]
+type = simple
+rootdir = /mnt/rcm-guest/puddles/RHAOS
+mashroot = /mnt/rcm-guest/puddles/RHAOS/mash
+brewroot = file:///mnt/redhat/brewroot
+topurl = http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/
+name = AtomicOpenShift.el8
```
yuqi-zhang: So the full repo url will be something like... http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/AtomicOpenShift.el8/4.1/latest/ Feel like it might be slightly confusing given the base repo is http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/AtomicOpenShift/4.1 but no real objections whichever way the naming scheme ends up

tbielawa:
/hold

let me add the new job files first

build-scripts/puddle-conf/atomic_openshift-4.1.el8.conf
yuqi-zhang: (i.e. there is no `el7` in the other)

tbielawa: You're right, the other doesn't have a `.el7` part, but we can't change it right now because there are a lot of systems depending on that URL.

yuqi-zhang: Yeah its fine, either way if the job files look good I'm :+1:

cgwalters:
Thanks for doing this!  Looks sane to me.

PR #1775, opened 2019-04-05
update jobs that refer to ocp4 job

sosiouxme:
ose4.0 job removed because we don't need to build 4.0 any more.
ose4.1 job updated to run ocp4 with new parameters.
ose-master job removed because I don't think anyone needs to use it and it was just an extra hop for the scheduled builds.

The scheduled ose-master job I split into two; one incremental, one force (full) to run on different schedules, but both channeled through the ose4.1 job to prevent concurrent builds.


jobs/build/ocp4/Jenkinsfile
```diff
@@ -12,8 +12,8 @@ node {
                 logRotator(
                     artifactDaysToKeepStr: '',
                     artifactNumToKeepStr: '',
-                    daysToKeepStr: '',
-                    numToKeepStr: '1000')),
+                    daysToKeepStr: '30',
```
tbielawa: Good call. But 60 or 90 days might be better. Personally I have gone back as far as 3 months and ran analytics on build logs before

jobs/build/ose4.1/Jenkinsfile
```diff
@@ -1,14 +1,48 @@
-properties(
+#!/usr/bin/env groovy
+
+node { 
+    checkout scm
+    def commonlib = load("pipeline-scripts/commonlib.groovy")
+    properties([
+        buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '30', numToKeepStr: '')),
```
tbielawa: Same as above. Consider 60/90 days.

scheduled-jobs/build/ose-master/Jenkinsfile
```diff
@@ -1,15 +1,12 @@
 
-properties( [
-    buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '100')),
+properties([
+    buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '30', numToKeepStr: '')),
```
tbielawa: Same as above. Consider 60/90

tbielawa:
All in all it looks pretty good to me. I'll wait for some other eyes on this. Will review again on Monday.

sosiouxme:
Obviously it's easy to change 30 to 60 or 90. And fine with me.

I would argue though that retaining Jenkins builds for analytics isn't that useful. We should be entering what we need for metrics in a system that actually helps analyze metrics, and discarding the builds after we're unlikely to have a use for looking at logs/artifacts.

tbielawa:
You're totally right. And I know we already both agree that that addition is outside of the scope of this PR :smiley: 

The occasional reminder doesn't hurt though.

jobs/build/ocp4/Jenkinsfile
adammhaile: +1 - 60 should be fine

scheduled-jobs/build/ose-master-full/Jenkinsfile
```diff
@@ -0,0 +1,19 @@
+
+properties([
+    buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '30', numToKeepStr: '')),
+    disableConcurrentBuilds(),
+])
+
+currentBuild.displayName += "4.1 full build"
+def b = build(
+    job: '../aos-cd-builds/build%2Fose4.1',
```
adammhaile: Just talking out loud here - We should put some data in `ocp-build-data:master` that specifies what version master is and things like that. Single yaml file that jobs like this could curl and read. That way version are no longer hard coded in jobs and we don't have to update them anymore when version bumps happen.

sosiouxme: We'd still have to create e.g. the `ose4.2` job and repoint others when that time rolls around. I don't see updating this job at the same time as a problem worth the trouble of automating away.

The idea with `ocp-build-data:master` seems good though. It would be good for example to be able to have the commonlib options for which versions are available key off... which versions are available.

sosiouxme:
I updated the jobs to retain for 60 days. This should probably be merged after https://github.com/openshift/doozer/pull/76 so that the scheduled incremental builds get the right stuff.

scheduled-jobs/build/ose-master-full/Jenkinsfile
adammhaile: Well, to explain my full vision, eventually I would like "One Scheduled Job to Rule Them All" that just runs every ~15 min, looks at `ocp-build-data:master` and then fires off whatever jobs are required.

sosiouxme: Sounds good :)

PR #1776, opened 2019-04-05
Run bug sweeps on a schedule

tbielawa:
Two new jobs, one parameterized, one wrapper.

The parameterized job should be called by wrapper jobs. The included
4.1 wrapper job calls the 'sweep' job with a BUILD_VERSION parameter.

Uses the new --use-default-advisory behavior in elliott.

### DIFF
--- a/scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
+++ b/scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
@@ -15,7 +15,7 @@ properties(
 description = ""
 failed = false
 
-b = build job: 'sweep', propagate: false, parameters: [
+b = build job: '../../aos-cd-builds/sweep', propagate: false, parameters: [
     [
 	$class: 'StringParameterValue',
 	name: 'BUILD_VERSION',
@@ -23,8 +23,6 @@ b = build job: 'sweep', propagate: false, parameters: [
     ]
 ]
 
-description += "${b.displayName} - ${b.result}\n"
-currentBuild.description = description.trim()
-currentBuild.displayName = "${b.displayName}"
-failed |= (b.result != "SUCCESS")
-currentBuild.result = failed?"FAILURE":"SUCCESS"
+currentBuild.result = (b.result == "SUCCESS") ? "SUCCESS" : "FAILURE"
+currentBuild.displayName = "[${b.result}] ${b.displayName}"
+currentBuild.description = b.description
--- a/scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
+++ b/scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
@@ -1,0 +1,30 @@
+properties(
+    [
+        buildDiscarder(
+            logRotator(
+                artifactDaysToKeepStr: '',
+                artifactNumToKeepStr: '',
+                daysToKeepStr: '',
+                numToKeepStr: '90'
+            )
+        ),
+	disableConcurrentBuilds(),
+    ]
+)
+
+description = ""
+failed = false
+
+b = build job: 'sweep', propagate: false, parameters: [
+    [
+	$class: 'StringParameterValue',
+	name: 'BUILD_VERSION',
+	value: '4.1'
+    ]
+]
+
+description += "${b.displayName} - ${b.result}\n"
+currentBuild.description = description.trim()
+currentBuild.displayName = "${b.displayName}"
+failed |= (b.result != "SUCCESS")
+currentBuild.result = failed?"FAILURE":"SUCCESS"
--- a/scheduled-jobs/build/bug-sweeps/sweep/Jenkinsfile
+++ b/scheduled-jobs/build/bug-sweeps/sweep/Jenkinsfile
@@ -1,0 +1,49 @@
+#!/usr/bin/env groovy
+
+node {
+    checkout scm
+    def buildlib = load("pipeline-scripts/buildlib.groovy")
+    def commonlib = buildlib.commonlib
+
+    properties(
+        [
+            buildDiscarder(
+                logRotator(
+                    artifactDaysToKeepStr: '',
+                    artifactNumToKeepStr: '',
+                    daysToKeepStr: '',
+                    numToKeepStr: '90'
+                )
+            ),
+            [
+                $class : 'ParametersDefinitionProperty',
+                parameterDefinitions: [
+                    commonlib.ocpVersionParam('BUILD_VERSION'),
+                ]
+            ]
+        ]
+    )
+
+    echo "Initializing bug sweep for ${params.BUILD_VERSION}. Sync: #${currentBuild.number}"
+
+    currentBuild.displayName = "${params.BUILD_VERSION} bug sweep"
+
+    stage("Version dumps") {
+        buildlib.elliott "--version"
+        sh "which elliott"
+    }
+
+    stage("kinit") {
+	buildlib.kinit()
+    }
+
+    try {
+	stage("Sweep bugs") {
+	    currentBuild.description = "Searching for and attaching bugs"
+	    buildlib.elliott "--group=openshift-${params.BUILD_VERSION} find-bugs --auto --use-default-advisory rpm"
+	    currentBuild.description = "Ran bug attaching command without errors"
+	}
+    } catch (findBugsError) {
+	currentBuild.description = "Error sweeping:\n${findBugsError}"
+    }
+}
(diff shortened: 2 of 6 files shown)


scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+            )
+        ),
+	disableConcurrentBuilds(),
+    ]
+)
+
+description = ""
+failed = false
+
+b = build job: 'sweep', propagate: false, parameters: [
```
tbielawa: I am **not sure** if using the `build job` thing understands to make relative references. This `4.1` job is _supposed to_ call the `sweep` job in this same folder with the `BUILD_VERSION` parameter.

scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+	name: 'BUILD_VERSION',
+	value: '4.1'
+    ]
+]
+
+description += "${b.displayName} - ${b.result}\n"
+currentBuild.description = description.trim()
+currentBuild.displayName = "${b.displayName}"
+failed |= (b.result != "SUCCESS")
+currentBuild.result = failed?"FAILURE":"SUCCESS"
```
tbielawa: This weirdness I just copied from another scheduled wrapper job (ose3.9 I think). It's not very pretty but I'm open to suggestions for improvements.

scheduled-jobs/build/bug-sweeps/sweep/Jenkinsfile
```diff
@@ -0,0 +1,48 @@
...
+                    artifactDaysToKeepStr: '',
+                    artifactNumToKeepStr: '',
+                    daysToKeepStr: '',
+                    numToKeepStr: '90'
+                )
+            ),
+            [
+                $class : 'ParametersDefinitionProperty',
+                parameterDefinitions: [
+                    commonlib.ocpVersionParam('BUILD_VERSION'),
```
tbielawa: Not specifying a `majorVersion` parameter here so we will default to `all` which leaves this open to easily add more per-version sweep jobs later.

smunilla: Not sure I follow here. Don't we restrict to the version we just built?

tbielawa: There are 2 jobs in this PR

* `sweep` which is the parameterized job. It's like running `ocp` directly (which is the `Jenkinsfile` you commented on)
* `4.1` - Calls the `sweep` job with `BUILD_VERSION` set to `4.1`

> Don't we restrict to the version we just built?

Since we still do `n-2` advisories I could see us adding `3.11` and `3.10` and `3.9` jobs in here that would call the `sweep` job w/ the `BUILD_VERSION` set appropriately.

scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
sosiouxme: It's relative, yeah.

But why not put the parameterized job in the aos-cd-builds folder? I thought we had this dynamic going where the manual jobs went there and the scheduled jobs went here.

scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
sosiouxme: I'm a bit partial to copying up the result from the build https://github.com/openshift/aos-cd-jobs/blob/master/scheduled-jobs/build/ose-master/Jenkinsfile#L10-L12
```
currentBuild.result = (b.result == "SUCCESS") ? "SUCCESS" : "FAILURE"
currentBuild.displayName = "[${b.result}] ${b.displayName}"
currentBuild.description = b.description
```
though it should probably have its own build id in there.

scheduled-jobs/build/bug-sweeps/sweep/Jenkinsfile
```diff
@@ -0,0 +1,49 @@
...
+    echo "Initializing bug sweep for ${params.BUILD_VERSION}. Sync: #${currentBuild.number}"
+
+    currentBuild.displayName = "${params.BUILD_VERSION} bug sweep"
+
+    stage("Version dumps") {
+        buildlib.elliott "--version"
+        sh "which elliott"
+    }
+
+    stage("kinit") {
```
sosiouxme: Could these be a single "init" stage?

Really this job is so simple I'm not sure you need stages at all...

scheduled-jobs/build/bug-sweeps/sweep/Jenkinsfile
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    try {
+	stage("Sweep bugs") {
+	    currentBuild.description = "Searching for and attaching bugs"
+	    buildlib.elliott "--group=openshift-${params.BUILD_VERSION} find-bugs --auto --use-default-advisory rpm"
+	    currentBuild.description = "Ran bug attaching command without errors"
+	}
+    } catch (findBugsError) {
+	currentBuild.description = "Error sweeping:\n${findBugsError}"
```
sosiouxme: How about re-throwing the exception?

Might we want to notify someone this is failing?

scheduled-jobs/build/bug-sweeps/sweep/README.md
```diff
@@ -0,0 +1,4 @@
+Parameterized job for running bugzilla bug sweeps.
+
+Preferrably a wrapper job will run on a schedule which calls out to
```
sosiouxme:
```suggestion
Preferably a wrapper job will run on a schedule which calls out to
```

scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
tbielawa: I have no strong feelings about where it lives.

scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile
tbielawa: OK.

scheduled-jobs/build/bug-sweeps/sweep/Jenkinsfile
tbielawa: Good point.

scheduled-jobs/build/bug-sweeps/sweep/Jenkinsfile
tbielawa: I wish we had a way to buffer notice emails so we don't spam blast them over and over again. Emailing it is a good idea though.

sosiouxme:
Only thing that stands out is `scheduled-jobs/build/bug-sweeps/4.1/Jenkinsfile` being one dir down from where the other scheduled jobs are. Since scheduled jobs are all manually configured anyway instead of having branches generated, I suppose that doesn't matter.
/lgtm

tbielawa:
That's weird. I wasn't expecting this

from https://buildvm.openshift.eng.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Fsweep/2/console

```
[build%2Fsweep] Running shell script
+ set +x
/mnt/workspace/jenkins/working/aos-cd-builds/build%2Fsweep@tmp/durable-e795ec40/script.sh: line 9: null/tools/bin/elliott: No such file or directory
```

So, naturally, let's grep around for some stuff here.

```$ grep -r 'tools/bin/elliott'      
pipeline-scripts/buildlib.groovy:        script: "${env.ENTERPRISE_IMAGES_DIR}/tools/bin/elliott --user=ocp-build ${cmd.trim()}")
```

And that command is part of `def elliott`. Which is weird because I don't see `ENTERPRISE_IMAGES_DIR` anywhere else. I tried to go through the refresh images logs where `buildlib.elliott` is actually used but the logs don't include any jobs where the actual attach images step was ran. It was skipped in all the ones we have logs for. https://buildvm.openshift.eng.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Frefresh-images/ 

Weird.

tbielawa:
I'm going to try a dirty hack for this and change it to just call `elliott` the same way that `def doozer` calls doozer.

sosiouxme:
Yeah, that's been broken. Thought Sam had fixed it but haven't run it in some time...

Would ship that solution :)

PR #1778, opened 2019-04-06
build/*: background cleaning workdir

sosiouxme:
Removing the workdir before beginning can take time, and I'm impatient. So this does it in the background.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -1087,4 +1087,17 @@ def getGroupBranch(doozerOpts) {
     return branch
 }
 
+def cleanWorkdir(workdir) {
+    // get a fresh workdir; removing the old one is left in the background.
+    // NOTE: if wrapped in commonlib.shell, this would wait for the background process;
+    // this is designed to run instantly and never fail, so just run it in a normal shell.
+    sh """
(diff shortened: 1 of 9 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -1087,4 +1087,17 @@ def getGroupBranch(doozerOpts) {
...
+def cleanWorkdir(workdir) {
+    // get a fresh workdir; removing the old one is left in the background.
+    // NOTE: if wrapped in commonlib.shell, this would wait for the background process;
+    // this is designed to run instantly and never fail, so just run it in a normal shell.
+    sh """
+        mkdir -p ${workdir}
+        mv ${workdir} ${workdir}.rm.${currentBuild.number}
+        mkdir -p ${workdir}
+        # see discussion at https://stackoverflow.com/a/37161006 re:
+        JENKINS_NODE_COOKIE=dontKill BUILD_ID=dontKill nohup bash -c 'rm -rf ${workdir}.rm.*' &
```
tbielawa: The referenced SO page says to use `dontKillMe` you only have `dontKill` here. That still fine?

sosiouxme: as long as it's not a jenkins build id, it should work.

PR #1779, opened 2019-04-09
build/ocp4: improve display of build plan

sosiouxme:
For force builds, include in the build description which RPMs and images will be built, like when it is incremental.
Update displayName to indicate how many things are (not) being built.
commonlib.shell(): return a bit more error output


sosiouxme:
@adammhaile @tbielawa PTAL

tbielawa:
/lgtm
seems legit to me

PR #1784, opened 2019-04-10
build/ocp: fix python-style comments in groovy

sosiouxme:
April Fools! I broke 3.x builds for a week!


sosiouxme:
@tbielawa if you would do the honors...

PR #1798, opened 2019-04-17
scheduled-jobs/build/ose-master-full: force build for reals

sosiouxme:
Actually pass true in the FORCE_BUILD param.
Also for both... add a space between the build number and title.


sosiouxme:
@adammhaile @tbielawa 
fix another :man_facepalming:

PR #1799, opened 2019-04-18
build/custom: doozer expects version to start with v

sosiouxme:


sosiouxme:
merging this quick fix to job that's currently broken
/cc @shiywang

PR #1804, opened 2019-04-18
DRY, elliottest=>elliott

sosiouxme:
DRY some repeat whitespace handling.
Looks like the path-less and user-less elliott works fine, let's make
that the norm.


tbielawa:
I see this here, given the problems we're having with refresh images right now I am hesitant to make changes there. Or would this fix some of that? I thought I saw you comment that the buildlib.elliott command was busted there too?

sosiouxme:
refresh-images invocation of elliott is already busted. this is to fix it.

sosiouxme:
@tbielawa I did a bug sweep with it and it seems to work. Merge?

sosiouxme:
The next step, of course, is to use `elliott find-builds --use-default-advisory rpm` in `refresh-errata` and do away with specifying the advisory at all.

PR #1813, opened 2019-05-01
Adding release job

adammhaile:
@tbielawa @sosiouxme PTAL
Obviously still missing some steps, but I've added placeholders and this should cover what's at least possible now.

### DIFF
--- a/jobs/build/release/release.groovy
+++ b/jobs/build/release/release.groovy
@@ -1,0 +1,130 @@
+buildlib = load("pipeline-scripts/buildlib.groovy")
+commonlib = buildlib.commonlib
+
+
+// dump important tool versions to console
+def stageVersions() {
+    sh "oc version"
+    sh "doozer --version"
+    sh "elliott --version"
+}
+
+def stageValidation() {
+    echo "Verifying payload does not already exist"
+    res = commonlib.shell(
+        returnAll: true,
+        script: "oc adm release info ${params.NAME}"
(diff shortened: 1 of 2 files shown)


tbielawa:
net-new so I'm going to hulk smash green button

jobs/build/release/release.groovy
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+def stageGenPayload() {
+    // build metadata blob
+    def metadata = "{\"description:\" \"${params.DESCRIPTION}\""
+    if (params.ADVISORY != "") {
+        metadata += ", \"advisory_url\": \"https://errata.devel.redhat.com/advisory/${params.ADVISORY}\""
+    }
+    metadata += "}"
```
sosiouxme: blurgh. I hate that Jenkins doesn't allow sane use of json. Humans shouldn't have to read/write json.

I'd rather not see all the escaping though. Could be avoided like so:
```
    def advisory = params.ADVISORY ? """ "advisory_url": "https://errata.devel.redhat.com/advisory/${params.ADVISORY}", """ : ''
    def metadata = """{${advisory} description:" "${params.DESCRIPTION}"} """
```
That's _assuming_ we can rely on the description not to have any double quotes... If that's even vaguely possible we should probably `writeJson` and `readJson` instead. I'm tempted to write `commonlib.jsonString` that does that.

PR #1817, opened 2019-05-02
fix for calling oc sync job

adammhaile:


jobs/build/release/release.groovy
```diff
@@ -1,6 +1,9 @@
 buildlib = load("pipeline-scripts/buildlib.groovy")
 commonlib = buildlib.commonlib
 
+def param(type, name, value) {
+    return [$class: type + 'ParameterValue', name: name, value: value]
```
tbielawa: This is defined in buildlib already. Why so wet?

https://github.com/openshift/aos-cd-jobs/blob/master/pipeline-scripts/buildlib.groovy#L688-L690

adammhaile: oh duh... forgot I had that pulled in
one moment

PR #1819, opened 2019-05-02
update oc command

adammhaile:
@tbielawa

### DIFF
--- a/jobs/build/release/release.groovy
+++ b/jobs/build/release/release.groovy
@@ -1,6 +1,8 @@
 buildlib = load("pipeline-scripts/buildlib.groovy")
 commonlib = buildlib.commonlib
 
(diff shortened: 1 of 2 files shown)


jobs/build/release/release.groovy
```diff
@@ -12,7 +14,7 @@ def stageValidation() {
     echo "Verifying payload does not already exist"
     res = commonlib.shell(
         returnAll: true,
-        script: "oc adm release info ${params.NAME}"
+        script: "${oc_cmd} adm release info ${params.NAME} ${kubeconfig}"
```
tbielawa: You doubled down on the kubeconfig I think. Looks like you include it on line 4 and then appended it to line 17 here. I think you only need it once (unless I'm not understanding the whole context of this)

PR #1822, opened 2019-05-06
[WIP] Add bug flag to find-bugs call

smunilla:


vfreex:
Closing due to the problem is not present.

PR #1831, opened 2019-05-10
Add GOTRACEBACK env var for full trace

tbielawa:
We had an `oc` error running the release job today but we didn't get the full trace. This forces full tracebacks if errors happen

https://buildvm.openshift.eng.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Foc_sync/14/console


sosiouxme:
bluh... why doesn't everything just use `buildlib.oc()`?

PR #1837, opened 2019-05-17
scheduled-jobs/build/rhcos-puddle: export REQUESTS_CA_BUNDLE

sosiouxme:
otherwise the job is broken and rhcos gets old version of openshift.


sosiouxme:
/cc @tbielawa @adammhaile 
self-merging since it's already busted and I can hardly make it worse.

PR #1843, opened 2019-05-24
Adding errata URL option

adammhaile:
@vikaslaad @smunilla @tbielawa @sosiouxme

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -43,6 +43,12 @@ node {
                         defaultValue: ""
                     ],
                     [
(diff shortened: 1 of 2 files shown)


jobs/build/release/Jenkinsfile
```diff
@@ -42,6 +42,12 @@ node {
                         $class: 'hudson.model.StringParameterDefinition',
```
tbielawa: Just gonna leave the old `ADVISORY` param in there for now?

smunilla: I'd say yes. We can get this URL from the ET API eventually.

PR #1847, opened 2019-05-28
Update find-bugs call for elliott upgrade

smunilla:


sosiouxme:
LGTM. And since we're not running sweeps ATM it can't really hurt...

PR #1849, opened 2019-05-28
rhcos-puddle: add puddle for 4.2 builds

yuqi-zhang:
Modify puddle job to also output 4.2-el8 builds as a separate stage.

Note: I based this off the conf here: https://github.com/openshift/aos-cd-jobs/blob/master/build-scripts/puddle-conf/atomic_openshift-4.2.el8.conf. This also does not affect the el7 based RPMS, and only intendeds to create a 4.2 puddle for RHCOS specific el8 packages.

### DIFF
--- a/scheduled-jobs/build/rhcos-puddle/Jenkinsfile
+++ b/scheduled-jobs/build/rhcos-puddle/Jenkinsfile
@@ -19,57 +19,60 @@ node {
     )
 
     def buildException = null
-    try {
-        buildlib.kinit()
-        stage("openshift rpm") {
-            buildlib.commonlib.shell(
-                '''
-                set -exuo pipefail
-                tmpdir=$(mktemp -d XXXXXXXXXX.distgit)
-                pushd $tmpdir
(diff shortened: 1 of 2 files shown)


yuqi-zhang:
cc @sosiouxme

sosiouxme:
LGTM, will try out a test run and merge.

sosiouxme:
Oh; it's not going to be very useful if it doesn't also rebuild the 4.2 openshift RPM.

yuqi-zhang:
Sorry, I thought the RPM copy was for something else. I've updated the loop to include both jobs (although I'm not sure if the syntaxing is correct as I'm not sure what the bash script is doing). PTAL

sosiouxme:
Looks good now. I'll merge it once openshift RPM will actually build (don't want to see constant notices that it's failing).

Future direction of this would be that this job is parameterized per release, joins the main jobs, and is kicked off as part of the ocp4 job. And eventually, that it includes signing the RPMs and creating a signed compose.

PR #1854, opened 2019-05-30
sign-artifacts: use workDir and push pub

sosiouxme:
Signatures are saved locally in the workDir.
Pushing pub not enterprise; it's actually a different script on use-mirror.
enterprise is for yum reposync. pub is for content.


sosiouxme:
best to ignore whitespace

PR #1856, opened 2019-05-31
sign-artifacts tweaks

tbielawa:
* Add SIGNATURE_NAME parameter for when we sign with two keys
* Copy signature-1 files to google storage
* Archive gs:// copy log file as well

### DIFF
--- a/jobs/signing/sign-artifacts/Jenkinsfile
+++ b/jobs/signing/sign-artifacts/Jenkinsfile
@@ -178,100 +178,112 @@ node {
             echo "Mirroring artifacts to mirror.openshift.com/pub/"
 
             dir(workDir) {
-                // ######################################################################
+                try {
+                    // ######################################################################
 
-                // the umb producer script should have generated two
-                // files. One is a 'sha26=.......' and one is
-                // 'sha256sum.txt.sig'
+                    // the umb producer script should have generated two
+                    // files. One is a 'sha26=.......' and one is
+                    // 'sha256sum.txt.sig'
 
-                // sha256=...........
-                //
-                // 1) JSON digest claims. They get mirrored to
-                // mirror.openshift.com/pub/openshift-v4/ using this directory
-                // structure:
-                //
-                // signatures/openshift/release/
-                //   -> sha256=<IMAGE-DIGEST>/signature-1
-                //
-                // where <IMAGE-DIGEST> is the digest of the payload image and
-                // the signed artifact is 'signature-1'
-                //
-                // 2) A message digest (sha256sum.txt) is mirrored to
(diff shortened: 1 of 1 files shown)


jobs/signing/sign-artifacts/Jenkinsfile
```diff
@@ -262,6 +265,13 @@ node {
                     }
                 }
             }
+
+            echo "Archiving artifacts in jenkins:"
```
sosiouxme: we would likely want these in a `finally {}` block in case there's partial success (one sig pushes, another doesn't).

jobs/signing/sign-artifacts/Jenkinsfile
```diff
@@ -206,12 +205,16 @@ node {
                 //
                 // where <RELEASE-NAME> is something like '4.1.0-rc.5'
 
-                // transform the file into a directory containing the file
+                // transform the file into a directory containing the
+                // file, mirror to google
                 commonlib.shell('''
                     for file in sha256=*; do
-                        mv $file signature-1
+                        mv $file ${params.SIGNATURE_NAME}
```
sosiouxme: won't work in a block with `'''`
needs to be `"""` and then backslash-escape `$file` everywhere

jobs/signing/sign-artifacts/Jenkinsfile
```diff
@@ -206,12 +205,16 @@ node {
...
                         mkdir $file
-                        mv signature-1 $file
+                        mv ${params.SIGNATURE_NAME} $file
+                        until gsutil cp -n -v -L cp.log -r $file gs://openshift-release/official/signatures/openshift/release; do
```
sosiouxme: should probably terminate at some point if it's not working?

jobs/signing/sign-artifacts/Jenkinsfile
sosiouxme: actually i suppose it won't care about `$file` as long as it's not in braces

jobs/signing/sign-artifacts/Jenkinsfile
sosiouxme: where da creds at?

tbielawa: re: creds: It's one of those `gsutil login` type things.

re: terminate: yeah.. I suppose so? I could add an increment and then `break` after it exceeds like.... 10 tries or something.

jobs/signing/sign-artifacts/Jenkinsfile
tbielawa: ack

PR #1858, opened 2019-06-06
build/ocp3/Jenkinsfile: fix git checkouts that can fail

sosiouxme:
Solves this problem where origin/HEAD is pointing to the branch it's trying to checkout:
```
+ git checkout -b enterprise-3.11 origin/enterprise-3.11
fatal: A branch named 'enterprise-3.11' already exists.
```


sosiouxme:
Going to self-merge this trivial change as 3.11 builds are dead without it and this isn't running regularly.

PR #1882, opened 2019-07-05
appregistry: ignore 4.2 failures until we get it sorted

sosiouxme:
For now, 4.2 errors are just noise. Turn them off and plan to reinstate once we have 4.2 app registry pushes working.


sosiouxme:
Tested in dev area and seems to work.

PR #1883, opened 2019-07-08
Start using the new aos-art-automation address

tbielawa:
In general this change makes it such that automated emails don't go
directly to the team email address. This hits most of the big
spammers. Changes follow this format.

* 'to' field is aos-art-automation+SOME_TOPIC@, OTHER_PEOPLE@
* 'replyTo' field is always the team email address
* 'from' will include aos-art-automation@


sosiouxme:
LGTM! Great idea to enable better filters!

PR #1889, opened 2019-07-16
Update to use release:gen-payload

adammhaile:


tbielawa:
this is lgtm. will merge after I review that MR

PR #1890, opened 2019-07-17
slove jenkins job partial failed no display problem

shiywang:
~I am still testing my result will paste log soon~

https://localhost:8888/job/hack/job/shiywang-aos-cd-jobs/job/build%252Fupdate-base-images/3/console

### DIFF
--- a/jobs/build/update-base-images/build.sh
+++ b/jobs/build/update-base-images/build.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -e
+set -o pipefail
+
 export REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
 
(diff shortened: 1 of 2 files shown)


jobs/build/update-base-images/build.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -e
```
shiywang: I belive comman.shell() already set this up, but from the jenkins log I didnt see any setup... so just add it here to double-check

shiywang:
@sosiouxme  PTAL

jobs/build/update-base-images/build.sh
```diff
@@ -1,5 +1,8 @@
...
+set -o pipefail
```
tbielawa: We should standardize on

`set -euxo pipefail`

https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

tbielawa:
lgtm after that shell script gets updated

shiywang:
Thanks for the comments, I am going to merge this and try it out

PR #1892, opened 2019-07-17
update Jenkinsfile to enable specifying DOOZER_DATA_PATH

shiywang:
…DATA_PATH

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -53,16 +53,10 @@ node {
                         defaultValue: 'images:rebase',
                     ],
                     [
-                        name: 'DOOZER_WORKING_DIR',
-                        description: '(Optional) Unless you want to override it, default is ${WORKSPACE}/doozer_working',
-                        $class: 'hudson.model.StringParameterDefinition',
-                        defaultValue: ""
-                    ],
-                    [
                         name: 'DOOZER_DATA_PATH',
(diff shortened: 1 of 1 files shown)


tbielawa:
I am very curious as to why you would want this.

shiywang:
@tbielawa  the use case is when I testing RHEL8 images in the future, we haven't had openshift-4.2-rhel8 branch something like this, so we might need to use a personal repo's branch.

the parameter already picked up in here https://localhost:8888/job/hack/job/shiywang-aos-cd-jobs/job/build%252Fcustom/10/console 

@tbielawa @sosiouxme PTAL

jobs/build/custom/Jenkinsfile
```diff
@@ -113,7 +129,7 @@ node {
                 if (rpms.toUpperCase() != "NONE") {
                     currentBuild.displayName += rpms.contains(",") ? " [RPMs]" : " [${rpms} RPM]"
                     currentBuild.description = "building RPM(s): ${rpms}\n"
-                    command = "--working-dir ${doozer_working} --group 'openshift-${params.BUILD_VERSION}' "
+                    command = "--working-dir ${doozer_working} --data-path ${doozer_data_path} --group 'openshift-${params.BUILD_VERSION}' "
```
sosiouxme: hmm, I realize it was this way when you found it, but the longer it gets, the more obvious... why isn't this just:
```suggestion
                    command = doozerOpts
```

and the same for the next couple changes

shiywang:
@sosiouxme updated, PTAL

PR #1899, opened 2019-07-25
make release.groovy can be reused by other jobs

shiywang:
@sosiouxme  this is the prerequisite of https://github.com/openshift/aos-cd-jobs/pull/1900


sosiouxme:
LGTM

sosiouxme:
I probably merged this too fast. I don't really love the idea of one job using another's code. Actually it will be super confusing to reference with the way things get moved around on the branches. And just conceptually I would rather release.groovy got pulled out into pipeline-scripts like buildlib.groovy and used from both jobs.

shiywang:
@sosiouxme no problems I will add it into my follow-up PR

PR #1900, opened 2019-07-25
add new pre-release job

shiywang:
@sosiouxme

### DIFF
--- a/jobs/build/pre-release/Jenkinsfile
+++ b/jobs/build/pre-release/Jenkinsfile
@@ -32,18 +32,6 @@ node {
                         defaultValue: ""
                     ],
                     [
-                        name: 'DESCRIPTION',
-                        description: 'Release description for metadata',
-                        $class: 'hudson.model.StringParameterDefinition',
-                        defaultValue: ""
-                    ],
-                    [
-                        name: 'PREVIOUS',
-                        description: 'Optional: Tag(s) (comma separated) of release this can upgrade FROM',
-                        $class: 'hudson.model.StringParameterDefinition',
-                        defaultValue: ""
-                    ],
-                    [
                         name: 'DRY_RUN',
                         description: 'Only do dry run test and exit.',
                         $class: 'BooleanParameterDefinition',
@@ -78,7 +66,7 @@ node {
 
             stage("validation") { prerelease.stageValidation(quay_url) }
 
-            stage("payload") { prerelease.stageGenPayload(quay_url) }
+            stage("payload") { prerelease.stageGenPayload(quay_url, "", "", "") }
 
             stage("tag prerelease") { prerelease.stageTagRelease(quay_url) }
 
--- a/jobs/build/pre-release/Jenkinsfile
+++ b/jobs/build/pre-release/Jenkinsfile
@@ -21,13 +21,7 @@ node {
                 parameterDefinitions: [
                     [
                         name: 'FROM_RELEASE_TAG',
-                        description: 'Build tag to pull from (i.e. 4.1.0-0.nightly-2019-04-22-005054)',
-                        $class: 'hudson.model.StringParameterDefinition',
-                        defaultValue: ""
-                    ],
-                    [
-                        name: 'NAME',
-                        description: 'Release name (i.e. 4.1.0-rc0)',
+                        description: 'Build tag to pull from (i.e. 4.1.0-0.nightly-2019-04-22-005054), pre-release job the release name is same as FROM_RELEASE_TAG',
                         $class: 'hudson.model.StringParameterDefinition',
                         defaultValue: ""
                     ],
@@ -64,16 +58,16 @@ node {
 
             stage("versions") { prerelease.stageVersions() }
 
-            stage("validation") { prerelease.stageValidation(quay_url, "") }
+            stage("validation") { prerelease.stageValidation(quay_url, ${params.FROM_RELEASE_TAG}, "") }
 
-            stage("payload") { prerelease.stageGenPayload(quay_url, "", "", "") }
+            stage("payload") { prerelease.stageGenPayload(quay_url, ${params.FROM_RELEASE_TAG}, ${params.FROM_RELEASE_TAG}, "", "", "") }
 
-            stage("tag prerelease") { prerelease.stageTagRelease(quay_url) }
+            stage("tag prerelease") { prerelease.stageTagRelease(quay_url, ${params.FROM_RELEASE_TAG}) }
 
             stage("client sync") { prerelease.stageClientSync("${params.FROM_RELEASE_TAG}",
             '/srv/pub/openshift-v4/clients/ocp-dev-preview/') }
 
-            stage("set client latest") { prerelease.stageSetClientLatest("${params.NAME}")}
+            stage("set client latest") { prerelease.stageSetClientLatest("${params.FROM_RELEASE_TAG}")}
         }
 
         dry_subject = ""
--- a/jobs/build/pre-release/Jenkinsfile
+++ b/jobs/build/pre-release/Jenkinsfile
@@ -2,7 +2,7 @@
 
 node {
     checkout scm
-    def prerelease = load("../release/release.groovy")
+    def prerelease = load("pipeline-scripts/release.groovy")
     def buildlib = prerelease.buildlib
     def commonlib = prerelease.commonlib
     def quay_url = "quay.io/openshift-release-dev/ocp-release-nightly"
@@ -98,9 +98,9 @@ node {
             to: "${params.MAIL_LIST_FAILURE}",
             replyTo: "aos-team-art@redhat.com",
             from: "aos-art-automation@redhat.com",
-            subject: "Error running OCP Release",
-            body: "Encountered an error while running OCP release: ${err}");
-        currentBuild.description = "Error while running OCP release:\n${err}"
+            subject: "Error running OCP Pre-Release",
+            body: "Encountered an error while running OCP pre release: ${err}");
+        currentBuild.description = "Error while running OCP pre release:\n${err}"
         currentBuild.result = "FAILURE"
         throw err
     }
--- a/jobs/build/pre-release/Jenkinsfile
+++ b/jobs/build/pre-release/Jenkinsfile
@@ -81,7 +81,7 @@ node {
 
             stage("payload") { prerelease.stageGenPayload(quay_url) }
 
-            stage("tag stable") { prerelease.stageTagStable(quay_url) }
+            stage("tag stable") { prerelease.stageTagRelease(quay_url) }
 
             // stage("wait for stable") { prerelease.stageWaitForStable() }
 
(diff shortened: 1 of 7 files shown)


jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+            // must be able to access remote registry for verification
+            buildlib.registry_quay_dev_login()
+
+            stage("versions") { prerelease.stageVersions() }
+
+            stage("validation") { prerelease.stageValidation(quay_url) }
+
+            stage("payload") { prerelease.stageGenPayload(quay_url) }
+
+            stage("tag stable") { prerelease.stageTagStable(quay_url) }
```
sosiouxme: should probably rename this since it's no longer destined for the stable stream... stageTagRelease ?

jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env groovy
+
+node {
+    checkout scm
+    def prerelease = load("../release/release.groovy")
```
sosiouxme: This would not actually work, it would need to be
```suggestion
    def prerelease = load("jobs/build/release/release.groovy")
```
... but let's not do that either, that's too coupled. Can you pull release.groovy into pipeline-scripts instead?

jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+                    artifactDaysToKeepStr: '',
+                    artifactNumToKeepStr: '',
+                    daysToKeepStr: '',
+                    numToKeepStr: '')),
+            [
+                $class: 'ParametersDefinitionProperty',
+                parameterDefinitions: [
+                    [
+                        name: 'FROM_RELEASE_TAG',
+                        description: 'Build tag to pull from (i.e. 4.1.0-0.nightly-2019-04-22-005054)',
```
sosiouxme: I would like this job to use the latest accepted when parameter is not specified (which would be most of the time). That way we can just run it on a schedule, and eventually, trigger on UMB -- no human providing parameters.

If that is hard in the short term, a human can fill the parameter for a while just to have something that works at all.

jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+                        $class: 'hudson.model.StringParameterDefinition',
+                        defaultValue: ""
+                    ],
+                    [
+                        name: 'NAME',
+                        description: 'Release name (i.e. 4.1.0-rc0)',
```
sosiouxme: Job should re-use the release tag unless this is specified. Should not need to specify.

jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+                        $class: 'hudson.model.StringParameterDefinition',
+                        defaultValue: ""
+                    ],
+                    [
+                        name: 'DESCRIPTION',
```
sosiouxme: unlikely we'll have any use for description or previous in this job, but if it makes code reuse easier, that's OK.

jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+
+            stage("validation") { prerelease.stageValidation(quay_url) }
+
+            stage("payload") { prerelease.stageGenPayload(quay_url) }
+
+            stage("tag stable") { prerelease.stageTagStable(quay_url) }
+
+            // stage("wait for stable") { prerelease.stageWaitForStable() }
+
+            stage("client sync") { prerelease.stageClientSync() }
```
sosiouxme: is this going to know to push to a different location?

Also, needs to call the job that updates the `latest` symlink.

jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+
+        }
+
+        dry_subject = ""
+        if (params.DRY_RUN) { dry_subject = "[DRY RUN] "}
+
+
+    } catch (err) {
+        commonlib.email(
```
sosiouxme: update text below to indicate pre-release

jobs/build/pre-release/Jenkinsfile
shiywang: done

jobs/build/pre-release/Jenkinsfile
shiywang: I checked on https://api.ci.openshift.org/console/project/ocp/browse/images/release, they have a tag called `latest` but not sure if it's 4.1 or 4.2, I think it should be worth asking them to provide a tag for us.

jobs/build/pre-release/Jenkinsfile
shiywang: I am not sure what do you mean, release job also need to manually specified, so how do we reuse?

jobs/build/pre-release/Jenkinsfile
shiywang: yeah, so the I use params.NAME pass into the latest symlink job, is that correct?

jobs/build/pre-release/Jenkinsfile
sosiouxme: say we are pre-releasing `registry.svc.ci.openshift.org/ocp/release:4.1.0-0.nightly-2019-07-24-213555`
then we are just creating `quay.io/openshift-release-dev/ocp-release-nightly:4.1.0-0.nightly-2019-07-24-213555`
there is no need to give it a new name, unless we want to for some reason

jobs/build/pre-release/Jenkinsfile
sosiouxme: yes, although as above that NAME will usually not be specified

jobs/build/pre-release/Jenkinsfile
shiywang: done

jobs/build/pre-release/Jenkinsfile
shiywang: done

shiywang:
@sosiouxme PTAL

pipeline-scripts/release.groovy
```diff
@@ -10,48 +10,50 @@ def stageVersions() {
...
-        cmd += "--previous \"${params.PREVIOUS}\" "
+    cmd += "--from-release=registry.svc.ci.openshift.org/ocp/release:${from_release_tag} "
+    if (previous != "") {
+        cmd += "--previous \"${previous}\" "
     }
-    cmd += "--name ${params.NAME} "
-    cmd += "--metadata '${metadata}' "
-    cmd += "--to-image=${quay_url}:${params.NAME} "
+    cmd += "--name ${name} "
+    if (description != "") {
```
sosiouxme: description has only ever been empty, but we do want the errata url. actually, i don't see why this is conditional at all; it wasn't before.
```suggestion
```

shiywang:
Hi @sosiouxme, I change pre-release job to use 4.2.0-0.nightly for client symlink job, but I got these errors:

``` 
GOTRACEBACK=all oc --config=/home/jenkins/kubeconfigs/art-publish.kubeconfig adm release new --from-release=registry.svc.ci.openshift.org/ocp/release:4.2.0-0.nightly-2019-07-25-114728 --name 4.2.0-0.nightly --metadata '{"description": ""}' --to-image=quay.io/openshift-release-dev/ocp-release-nightly:4.2.0-0.nightly --dry-run=true 
[Pipeline] sh
+ set +x
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x4f9f70]

goroutine 79 [running]:
bufio.(*Reader).fill(0xc0019621e0)
	/opt/rh/go-toolset-1.11/root/usr/lib/go-toolset-1.11-golang/src/bufio/bufio.go:100 +0xe0
bufio.(*Reader).WriteTo(0xc0019621e0, 0x2fcf3e0, 0xc00033cff0, 0x7fb6877dd270, 0xc0019621e0, 0x802f01)
	/opt/rh/go-toolset-1.11/root/usr/lib/go-toolset-1.11-golang/src/bufio/bufio.go:496 +0x10c
io.copyBuffer(0x2fcf3e0, 0xc00033cff0, 0x2fca0a0, 0xc0019621e0, 0x0, 0x0, 0x0, 0x2fc4b10, 0x2, 0x1)
```

it used to work like this:

```
GOTRACEBACK=all oc --config=/home/jenkins/kubeconfigs/art-publish.kubeconfig adm release new --from-release=registry.svc.ci.openshift.org/ocp/release:4.2.0-0.nightly-2019-07-25-114728 --name 4.2.0-0.nightly-2019-07-25-114728 --to-image=quay.io/openshift-release-dev/ocp-release-nightly:4.2.0-0.nightly-2019-07-25-114728 --dry-run=true 
[Pipeline] sh
+ set +x
info: Found 99 images in release
info: Loading sha256:2864cbf6924dff88489e3f88b02251c195c23272b51e7c9ef1648800ffc60a33 cluster-authentication-operator
```

so basically `oc release new` has three parameters: `--from-release`, `--name` and `--to-image`
shoudn't `name` and `to-image` be same ?

shiywang:
@sosiouxme PTAL https://localhost:8888/job/hack/job/shiywang-aos-cd-jobs/job/build%252Fpre-release/22/console

sosiouxme:
> Hi @sosiouxme, I change pre-release job to use 4.2.0-0.nightly for client symlink job, but I got these errors:
> [...]
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x4f9f70]

When you see this, assume it is a flake and run it again. We don't know what exactly causes it yet. But if anything golang panics, that's a problem in the tool (oc in this case), not in how we're invoking it.

> so basically `oc release new` has three parameters: `--from-release`, `--name` and `--to-image`
> shoudn't `name` and `to-image` be same ?

For this job, `name` should be `from-release` without the registry, and `to-image` should be the quay registry plus `name`. We'll use the same name all along. I think this is what we discussed but just wanted to put it in writing here.

jobs/build/pre-release/Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+            // must be able to access remote registry for verification
+            buildlib.registry_quay_dev_login()
+
+            stage("versions") { prerelease.stageVersions() }
+
+            stage("validation") { prerelease.stageValidation(quay_url, from_release_tag, "") }
+
+            stage("payload") { prerelease.stageGenPayload(quay_url, from_release_tag, from_release_tag, "", "", "") }
+
+            stage("tag prerelease") { prerelease.stageTagRelease(quay_url, from_release_tag) }
```
sosiouxme: I think this should be skipped in pre-release. AIUI (@tbielawa help me out here) this was used in `release` job because we actually gave the release in quay a different name, and needed to tag that back to the `release` repo in CI. But here, we're just reusing the name, not creating anything new. Doing this tag would actually overwrite the original release image with its copy.

tbielawa: :+1:

PR #1904, opened 2019-07-29
Remove specific emails from snapshot email

adammhaile:
The snapshot email was only ever sending to me, @smunilla, and @jupierce - none of us are involved anymore.


tbielawa:
Close. Should be more like the other emails we changed lately though.

https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/release/Jenkinsfile#L130-L131

adammhaile:
@tbielawa <tbielawa@redhat.com> That's fine with me as long as it gets
updated. I mostly wanted to bring attention to it.

Issue #1936, opened 2019-08-29
refresh-images job will be failed if building AMI is checked.


shiywang:
https://github.com/openshift/aos-cd-jobs/blob/cf5a56746a67ec306533a67f2c578deaa1e6adfe/jobs/build/refresh-images/Jenkinsfile#L278

I don't think that image still exists, if we want to get version, maybe we can use `openshift-enterprise-cli` as an alternative ?

base on this job https://localhost:8888/job/aos-cd-builds/job/build%252Frefresh-images/540/console
although check on buildAMI was not my intention.

shiywang:
@sosiouxme

sosiouxme:
it doesn't exist, and i have no idea what is invoking this. someone should definitely investigate. however when i saw this last, the real build worked, and this was just a failure at the end... right?

shiywang:
@sosiouxme  so the failure is due to I checked `build_ami (golden images)`. and I check it failed at that stage the first job, second job I didnt check that option. 

so if build_ami really doesn't use anymore or not that frequent we move it to lower prioprity.

PR #1937, opened 2019-08-30
change image to openshift-enterprise-cli

shiywang:
Fixes https://github.com/openshift/aos-cd-jobs/issues/1936


sosiouxme:
`openshift-enterprise-cli` didn't exist until 3.10
It turns out there was an `openshift-enterprise-docker` image until the mass docker renaming back in April 2018. It was renamed to just `openshift-enterprise`. So we should use that. Although since we haven't noticed before, I'd say we don't actually need the `BUILD_AMI` option any more!

PR #1938, opened 2019-08-30
Use commonlib.buildURL

joepvd:
And correct a widespread typo


tbielawa:
@joepvd Somebody else maintains the `alderan` files. We shouldn't be changing those, even if they are just tyop fixes. Your other fixes though, like under `jobs` and `pipeline-scripts` and `scheduled-jobs` are good!

We need to fix this up in the future by having our own repo instead of sharing one.

sosiouxme:
LGTM now please merge monday when we're around to notice failures (not that any are expected).

PR #1943, opened 2019-09-03
buildvm_backup job

jupierce:


jobs/maintenance/buildvm_backup/build.groovy
```diff
@@ -0,0 +1,102 @@
...
+    if (tarRes.returnStatus == 0) {
+        error("Error creating local tar")
+    }
+
+    scpRes = commonlib.shell(
+            returnAll: true,
+            script: "scp ${tarballPath} ${backupPlan.destHost}:${tarballPath}"
+    )
+
+    if (scpRes.returnStatus == 0) {
```
tbielawa: `!=` ?

jobs/maintenance/buildvm_backup/build.groovy
```diff
@@ -0,0 +1,102 @@
...
+        echo "Would have run:\n$ {cmds.join('\n$ ')}"
+        return
+    }
+
+    tarRes = commonlib.shell(
+            returnAll: true,
+            script: cmds.join('\n')
+    )
+
+    if (tarRes.returnStatus == 0) {
```
tbielawa: `!=` ?

jobs/maintenance/buildvm_backup/build.groovy
```diff
@@ -0,0 +1,102 @@
...
+}
+
+def stageRunBackup() {
+
+    res = commonlib.shell(
+            returnAll: true,
+            script: "hostname"
+    )
+
+    if (res.returnStatus == 0) {
```
tbielawa: `!=` ?

jobs/maintenance/buildvm_backup/build.groovy
```diff
@@ -0,0 +1,104 @@
...
+        '/home/jenkins/.kube',
+        '/home/jenkins/credentials',
+        '/home/jenkins/kubeconfigs',
+        '/home/jenkins/go',
+
+        // Jenkins war
+        '/usr/lib/jenkins',
+
+        // Jenkins configuration, plugins, etc
+        '/mnt/nfs/jenkins_home',
```
tbielawa: Not so certain about this one. I'm running a `du` on that right now and it has been running for about a minute. We might need to clean up that location.

tbielawa: Yeah it's still running.

tbielawa: Hm. I know what we need here. To run the actual jenkins backup script. It picks out the useful stuff. `/root/jenkins-backup.sh` has it.

> usage: jenkins-backup.sh /path/to/jenkins_home archive.tar.gz

I invoke it like this

> ./jenkins-backup.sh /mnt/nfs/jenkins_home `pwd`/2019-09-03-1.tar.gz

That omits log files. Which in theory won't be a huge issue because we're going to be jamming those into splunk soon anyway.

jupierce:
@tbielawa updated to only pull in the configs. That should be sufficient to restore the job definitions if everything was lost.

PR #1945, opened 2019-09-05
ART-947: fix release job ADVISORY parameter and remove ERRATA_URL par…

vfreex:
…ameter

1. If `ADVISORY` is not given, the value will be retrieved from `ocp_build_data` with `elliott`.
2. Verify the advisory exists and is in `QE` state and a live ID is associated.
3. `ERRATA_URL` parameter is removed. The live URL is constructed from the advisory information.
4. Print console logs with fancy emojis.

### DIFF
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -10,7 +10,9 @@ def stageVersions() {
     sh "elliott --version"
 }
 
-def stageValidation(quay_url, name, advisory) {
+Map stageValidation(String quay_url, String name, int advisory = 0) {
+    def retval = [:]
+    def version = commonlib.extractMajorMinorVersion(name)
     echo "Verifying payload does not already exist"
     res = commonlib.shell(
             returnAll: true,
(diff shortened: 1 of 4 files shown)


pipeline-scripts/release.groovy
```diff
@@ -21,24 +22,61 @@ def stageValidation(quay_url, name, advisory) {
...
+    echo "Verifying advisory ${advisoryInfo.id} (https://errata.engineering.redhat.com/advisory/${advisoryInfo.id}) status"
+    if (advisoryInfo.status != 'QE') {
+        error("🚫 Advisory ${advisoryInfo.id} is not in QE state.")
+    }
+    echo "✅ Advisory ${advisoryInfo.id} is in QE state."
+
+    // Extract liveID (if exists) from advisory info
+    // FIXME: Not sure if this is a stable way.
+    // If the advisory has a live ID assigned, let's say RHSA-2019:2594-03,
+    // advisoryInfo.errata_id should be 2547 and advisoryInfo.fulladvisory be RHSA-2019:2594-03
```
shiywang: do you mean `2594` I don't understand where this 2547 from?

vfreex: This is a mistake. I will fix it.

vfreex: Done

vfreex:
I just realized `pre-release` job also uses the `stageValidation()` function defined in `release.groovy`. Need to fix that either.

vfreex:
Changed the signature of `stageValidation` function back so that it won't break `pre-release` job.
Removed `VERSION` parameter, which can be extracted from the `NAME` parameter.

Please take another look. Thanks

pipeline-scripts/release.groovy
```diff
@@ -21,24 +23,61 @@ def stageValidation(quay_url, name, advisory) {
...
+    retval.advisoryInfo = advisoryInfo
+
+    echo "Verifying advisory ${advisoryInfo.id} (https://errata.engineering.redhat.com/advisory/${advisoryInfo.id}) status"
+    if (advisoryInfo.status != 'QE') {
+        error("🚫 Advisory ${advisoryInfo.id} is not in QE state.")
+    }
+    echo "✅ Advisory ${advisoryInfo.id} is in QE state."
+
+    // Extract live ID (if exists) from advisory info
+    // FIXME: Not sure if this is a stable way.
```
sosiouxme: That seems fine to me. We don't really have a clearer signal from the API and it's unlikely to change.

advisory with live ID:
```
    "errata_id": 2681, 
    "fulladvisory": "RHBA-2019:2681-02", 
    "id": 46049, 
    "old_advisory": "RHBA-2019:46049-02", 
```

advisory without:
```
    "errata_id": 46143, 
    "fulladvisory": "RHBA-2019:46143-01", 
    "id": 46143, 
    "old_advisory": null, 
```

pipeline-scripts/release.groovy
```diff
@@ -21,24 +23,61 @@ def stageValidation(quay_url, name, advisory) {
...
+    // If the advisory has a live ID assigned, let's say RHSA-2019:2594-03,
+    // advisoryInfo.errata_id should be 2594 and advisoryInfo.fulladvisory be RHSA-2019:2594-03
+    // while advisoryInfo.id remains the old internal advisory ID.
+    if (advisoryInfo.errata_id != advisoryInfo.id && advisoryInfo.fulladvisory && advisoryInfo.old_advisory) {
+        retval.liveID = (advisoryInfo.fulladvisory =~ /^(RH(?:B|S)A-\d+:\d+)-\d+$/)[0][1] // remove "-XX" suffix
```
sosiouxme: slightly simpler...
```suggestion
        retval.liveID = (advisoryInfo.fulladvisory =~ /^(RH[EBS]A-\d+:\d+)-\d+$/)[0][1] // remove "-XX" suffix
```
(We haven't used an RHEA in the past for $reasons but I hate to restrict things arbitrarily...)

pipeline-scripts/release.groovy
```diff
@@ -21,24 +23,61 @@ def stageValidation(quay_url, name, advisory) {
         error("Payload ${name} already exists! Cannot continue.")
     }
 
-    // AMH - This may be optional?
-    if (advisory != "") {
-        echo "Verifying advisory exists"
-        res = commonlib.shell(
+    if (!advisory) {
+        echo "Getting current advisory for OCP $version from build data..."
+        res = withEnv(["_VERSION=$version"]) {
```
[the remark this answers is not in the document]
vfreex: To prevent shell command injection though build parameters, although it is running internally by trusted people.

pipeline-scripts/release.groovy
vfreex: I plan to talk with Errata developers next week and see if they have better ideas.

pipeline-scripts/release.groovy
vfreex: Good suggestion! 👍

vfreex:
@sosiouxme Thanks for the review. I'll rebase it on Monday.

pipeline-scripts/release.groovy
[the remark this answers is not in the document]
sosiouxme: yeah, if someone untrusted can run jobs or trick us into doing it with XSS or something... then we're kinda screwed regardless of parameter safety. They could replay a job with `sh("rm -rf $HOME")` or whatever they like.

vfreex:
@sosiouxme @shiywang Thanks for the review!

PR #1948, opened 2019-09-06
Fix path inside RPM & allow expansion to occur before other steps

jupierce:
- Quoting in the bash prevented the shell from globbing. Now resolving the actual RPM name early.
- The location x86 binaries changed from 'openshift' to 'openshift-clients'


sosiouxme:
@jupierce  the package name in 4.2 is `openshift-clients`, the package in 4.1 is still `openshift` (RPM name is the same in both cases) -- both going to still work?

jupierce:
Nice catch! Would not have worked on 4.1. Updated to support different paths in the RPM.

sosiouxme:
seems legit

PR #1950, opened 2019-09-06
ART-777: Send UMB message when release job completes

vfreex:
The message includes the release image JSON got from `4-stable` stream and a advisory number associated with the release image. It is of a format that is similar to [product-build.build.complete](https://pagure.io/fedora-ci/messages/blob/master/f/schemas/product-build.build.complete.yaml) but haven't been standardized yet.

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -94,6 +94,7 @@ node {
             advisory = params.ADVISORY? Integer.parseInt(params.ADVISORY.toString()) : 0
             previous = "${params.PREVIOUS}"
             String errata_url
+            Map release_obj
 
             // must be able to access remote registry for verification
             buildlib.registry_quay_dev_login()
@@ -105,13 +106,14 @@ node {
             }
             stage("payload") { release.stageGenPayload(quay_url, name, from_release_tag, description, previous, errata_url) }
             stage("tag stable") { release.stageTagRelease(quay_url, name) }
-            stage("wait for stable") { release.stageWaitForStable() }
+            stage("wait for stable") { release_obj = release.stageWaitForStable() }
             stage("get release info") {
                 release_info = release.stageGetReleaseInfo(quay_url, name)
             }
             stage("client sync") { release.stageClientSync('4-stable', 'ocp') }
             stage("advisory update") { release.stageAdvisoryUpdate() }
             stage("cross ref check") { release.stageCrossRef() }
+            stage("send release message") { release.sendReleaseCompleteMessage(release_obj, advisory, errata_url) }
         }
 
         dry_subject = ""
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -1,3 +1,5 @@
+import groovy.json.*
+
 buildlib = load("pipeline-scripts/buildlib.groovy")
 commonlib = buildlib.commonlib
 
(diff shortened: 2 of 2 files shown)


vfreex:
Depends on #1945

jobs/build/release/Jenkinsfile
```diff
@@ -97,14 +91,18 @@ node {
             name = "${params.NAME}"
             from_release_tag = "${params.FROM_RELEASE_TAG}"
             description = "${params.DESCRIPTION}"
-            advisory = "${params.ADVISORY}"
+            advisory = params.ADVISORY? Integer.parseInt(params.ADVISORY.toString()) : 0
```
thiagoalessio: Hi @vfreex just curious, why did you turn the advisory into an integer?

pipeline-scripts/release.groovy
```diff
@@ -21,24 +23,64 @@ def stageValidation(quay_url, name, advisory) {
...
+                script: "elliott --group=openshift-${version} get --json - --use-default-advisory image",
+            )
+        if(res.returnStatus != 0) {
+            error("🚫 Advisory number for OCP $version couldn't be found from ocp_build_data.")
+        }
+    } else {
+        echo "Verifying advisory ${advisory} exists"
+        res = commonlib.shell(
+                returnAll: true,
+                script: "elliott --group=openshift-${version} get --json - -- ${advisory}",
```
thiagoalessio: same here

pipeline-scripts/release.groovy
```diff
@@ -21,24 +23,64 @@ def stageValidation(quay_url, name, advisory) {
...
-    // AMH - This may be optional?
-    if (advisory != "") {
-        echo "Verifying advisory exists"
+    if (!advisory) {
+        echo "Getting current advisory for OCP $version from build data..."
         res = commonlib.shell(
                 returnAll: true,
-                script: "elliott get ${advisory}"
-        )
+                script: "elliott --group=openshift-${version} get --json - --use-default-advisory image",
```
thiagoalessio: even though `commonlib.shell` is used under the hood, i've noticed that in other jobs `buildlib.elliott` is used whenever we want to perform `elliott` commands.
https://github.com/openshift/aos-cd-jobs/blob/master/pipeline-scripts/buildlib.groovy#L107-L111
usage example: https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/sweep/Jenkinsfile#L45

 don't know the reasons for that, but would recommend that you make use of it as well, for consistency.

pipeline-scripts/release.groovy
```diff
@@ -21,24 +23,64 @@ def stageValidation(quay_url, name, advisory) {
...
+    //     "errata_id": 2681,
+    //     "fulladvisory": "RHBA-2019:2681-02",
+    //     "id": 46049,
+    //     "old_advisory": "RHBA-2019:46049-02",
+    // - advisory without:
+    //     "errata_id": 46143,
+    //     "fulladvisory": "RHBA-2019:46143-01",
+    //     "id": 46143,
+    //     "old_advisory": null,
+    if (advisoryInfo.errata_id != advisoryInfo.id && advisoryInfo.fulladvisory && advisoryInfo.old_advisory) {
```
thiagoalessio: optional nitpick:
to extract this complex conditional into a function with a meaningful name, something like:
```groovy
def hasLiveId(advisoryInfo) {
    return advisoryInfo.errata_id != advisoryInfo.id && advisoryInfo.fulladvisory && advisoryInfo.old_advisory
}

// ...

if (hasLiveID(advisoryInfo)) {
    // ....
}
```

jobs/build/release/Jenkinsfile
sosiouxme: that's part of the first commit, which is actually from a different PR that I just merged (however I think this may need a rebase). really only need to review the second commit :)

vfreex: This is only my personal preference. The advisory ID returned by `elliott get --json` is `int` and I just personally prefer static typing than dynamic typing.

pipeline-scripts/release.groovy
vfreex: I haven't realized that method. `commonlib.shell()` is frequently used in this file so  I just copied it. I think we should make a better use of `buildlib.elliott`.

pipeline-scripts/release.groovy
vfreex: Maybe I can move it to commonlib?

vfreex:
Thanks for the review. I will rebase and update my code soon. (got crazy as a release manager recently)

vfreex:
@tbielawa Could you also take a look?

tbielawa:
This looks good to me!
/lgtm

PR #1953, opened 2019-09-06
crio: remove oct provision calls

haircommander:
origin-ci-tool is effectively deprecated, and crio is trying to switch to rhel8. because the package names are different, we are currently failing test_ami* calls. Fix this by not installing ansible on ami test (i.e. not calling the deprecated oct tool at this stage), but instead expect the ami creator to have installed ansible already.

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/sjb/config/test_cases/test_pull_request_crio_ami_rhel.yml
+++ b/sjb/config/test_cases/test_pull_request_crio_ami_rhel.yml
@@ -12,11 +12,6 @@ extensions:
       script: |-
         oct prepare user
         sed -i 's/User ec2-user/User origin/g' ./.config/origin-ci-tool/inventory/.ssh_config
-    - type: "host_script"
-      title: "bootstrap remote host"
-      script: |-
-        oct bootstrap host
-        oct bootstrap node
     - type: "script"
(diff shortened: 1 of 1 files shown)


haircommander:
/assign @stevekuznetsov 
PTAL

mrunalp:
:+1:

stevekuznetsov:
While it's deprecated unless you are moving off of it I would suggest just changing the code -- [here](https://github.com/openshift/origin-ci-tool/blob/2b40f3e11aadb569dc9c0c9fb90e7273658ce6ed/oct/ansible/oct/roles/dependencies/tasks/install_dependencies.yml#L80-L90) is an example of distribution-dependent installation and [here](https://github.com/openshift/origin-ci-tool/blob/2b40f3e11aadb569dc9c0c9fb90e7273658ce6ed/oct/ansible/oct/roles/dependencies/tasks/install_dependencies.yml#L10) is the `ansible` installl.

haircommander:
Ok, I'll probably close this in favor of some version of https://github.com/openshift/origin-ci-tool/pull/180

PR #1956, opened 2019-09-11
Initial iptables output lockdown helper

jupierce:


hacks/iptables/buildvm-scripts/canttouchthat.py
```diff
@@ -0,0 +1,112 @@
...
+def make_output_args(cidr):
+    return ['OUTPUT',
+            '-d', cidr,
+            '-j', 'ACCEPT',
+            '-m', 'comment',
+            '--comment', RULE_COMMENT
+            ]
+
+@click.command()
+@click.option('-n', '--output-networks', metavar='FILE',
```
tbielawa: Consider the following

https://click.palletsprojects.com/en/7.x/arguments/#file-arguments

vfreex:
Is it possible to manage whitelisted CIDRs using `ipset` rather than ordinary rules?

tbielawa:
> Is it possible to manage whitelisted CIDRs using `ipset` rather than ordinary rules?

From our conversation on slack

* `ipset` looks really cool - https://wiki.archlinux.org/index.php/Ipset
* For right now we don't need to look at that
* In the future, by using it, we could get some real business value in the way it lets you create named groups of rules and such
* I've only scratched the peak of the docs for it, but I am actually excited about it

PR #1963, opened 2019-09-17
odo_sync: Add token to GitHub API request

joepvd:


tbielawa:
squash this down and then we're `lgtm`

PR #1967, opened 2019-09-19
Reuse ose repo to speed up merge / lock github activity

jupierce:


jobs/build/merge_ocp/Jenkinsfile
```diff
@@ -21,12 +21,16 @@ node {
...
+                    booleanParam(
+                        name: 'CLEAN_CLONE',
+                        defaultValue: false,
+                        description: 'Force all git repos to be re-pulled'
+                    ),
+                    booleanParam(
+                        name: 'SCHEDULE_INCREMENTAL',
+                        defaultValue: true,
+                        description: 'If changes are detected, schedule an incremental build (4.x only)'
+                    ),
```
tbielawa: Something I'd add here would be like we have in a few other jobs
https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/ocp4/Jenkinsfile#L76
`commonlib.suppressEmailParam(),`
Won't send emails if it's not in the main `aos-cd-jobs` place. Not required, just a suggestion.

tbielawa:
/lgtm

my comment is only a suggestion.

PR #1969, opened 2019-09-20
refresh-images: fix partial rebuilds

sosiouxme:
When refreshing only specific images, --ignore-missing-base only works
if the parent image was successfully built at the time the dockerfile
was previously updated. The failure in the case where that's not true is
really confusing. --latest-parent-version will point at the most recent
successful build of the parent which is usually what we want.

### DIFF
--- a/jobs/build/refresh-images/Jenkinsfile
+++ b/jobs/build/refresh-images/Jenkinsfile
@@ -167,18 +167,14 @@ Proceed?
 
                 /**
                  * Determine if something other than the whole list should be built.
-                 * Note that --ignore-missing-base causes the rebase not to update a member image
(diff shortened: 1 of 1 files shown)


sosiouxme:
@thiagoalessio you saw this before, now i ran into it too.
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Frefresh-images/555/console

sosiouxme:
testing with https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Frefresh-images/2/console

PR #1972, opened 2019-09-23
[ART-903] Stop master => master merge from origin => ose

thiagoalessio:

PR #1973, opened 2019-09-23
Enabling 4.3 merges

jupierce:


thiagoalessio:
does that obsoletes https://github.com/openshift/aos-cd-jobs/pull/1972 ?

PR #1975, opened 2019-09-25
Remove Gabe from devex jenkins email lists

tbielawa:
@gabemontero @akram 

This is a follow up to the MR against ocp-build-data which removed Gabe and Adam from the maintainers lists. Adam wasn't present, so nothing to do for removing him here.


tbielawa:
updated email again to the team list!

PR #1983, opened 2019-10-02
Improve owner feedback if there is a merge error into distgit.

jupierce:
In the past, distgit commits were virtually guranteed since the upstream
content was definitive. Now, OLM manifest manipulation can cause
errors if the upstream data is properly populated. As such, we have seen
a rash of emails going out to owners claiming reconciliation problems
when, in fact, they are issues with OLM metadata upstream.

For example, in record.log:
```
distgit_commit_failure|image=openshift/ose-local-storage-operator|message=/mnt/workspace/jenkins/working/aos-cd-builds/build%2Focp4/doozer_working/distgits/containers/local-storage-operator/manifests/4.3/local-storage-operator.v4.3.0.clusterserviceversion.yaml does not exist as defined in art.yaml|owners=aos-storage-devel@redhat.com|distgit=containers/local-storage-operator|
```

A new email will now be sent if the job detects these records.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -845,7 +845,8 @@ def get_failed_builds(Map record_log, Boolean fullRecord=false) {
 //   dockerfile: /tmp/doozer-uEeF2_.tmp/distgits/jenkins-slave-maven-rhel7-docker/Dockerfile
 //   owners: bparees@redhat.com
 //   distgit: rpms/jenkins-slave-maven-rhel7-docker
-//   sha: 1b8903ef72878cd895b3f94bee1c6f5d60ce95c3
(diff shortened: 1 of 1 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -945,7 +966,16 @@ The reconciled (downstream OCP) Dockerfile can be viewed here:
  - https://pkgs.devel.redhat.com/cgit/${distgit}/tree/Dockerfile?id=${val.sha}
 
 Please direct any questions to the Automated Release Tooling team (#aos-art on slack).
-        """);
+        """
+        } else {
+            error("Unable to determine notification reason; something is broken"
```
sosiouxme: Needs a closing paren
Also, probably ought to tell them what to do about it... which I think is something along the lines of "hang tight and see if ART fixes it, but complain if it continues"

jupierce: Nice catch!
I've also updated the message.

PR #1988, opened 2019-10-08
OpenShift-Ansible 3.9 Install pip Directly

patrickdillon:
This is the 3.9 version of this PR openshift/release#4589

Centos pip version is out of date when installing from package manager. This pulls pip directly.

### DIFF
--- a/sjb/generated/test_branch_origin_extended_conformance_install_39.xml
+++ b/sjb/generated/test_branch_origin_extended_conformance_install_39.xml
@@ -165,7 +165,7 @@ for image in &#39;registry.svc.ci.openshift.org/ci/clonerefs:latest&#39; &#39;re
         fi
     done
 done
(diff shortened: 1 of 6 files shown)


sjb/generated/test_branch_origin_extended_conformance_install_39.xml
```diff
@@ -165,7 +165,7 @@ for image in &#39;registry.svc.ci.openshift.org/ci/clonerefs:latest&#39; &#39;re
         fi
     done
 done
-clonerefs_args=\${CLONEREFS_ARGS:---repo=openshift,image-registry=release-3.9 --repo=openshift,openshift-ansible=release-3.9 --repo=openshift,kubernetes-metrics-server=release-3.9 --repo=openshift,origin-web-console-server=release-3.9 --repo=openshift,aos-cd-jobs=master }
+clonerefs_args=\${CLONEREFS_ARGS:---repo=openshift,openshift-ansible=master --repo=openshift,aos-cd-jobs=master }
```
patrickdillon: @jcpowermac pointed out that this has a reference to master. I think this is being updated because of the change to test_pull_request_openshift_ansible_tox.yml which looks to me like an unused job. It seems to me we should either update this as so, or completely remove test_pull_request_openshift_ansible_tox.yml if it is not being used.

PR #1989, opened 2019-10-09
Differentiate email subjects for error / success reconciliations

jupierce:


pipeline-scripts/buildlib.groovy
```diff
@@ -952,6 +953,7 @@ ${val.failure}
 
         """
         } else if ( val.sha ) {
+            email_subject = "Changed Dockerfile reconciled for ${val.image} in OCP v${buildVersion}"
```
sosiouxme: Great, but since people are expecting notifications to have nothing but bad news could we break that expectation with:

`SUCCESS: Changed Dockerfile reconciled for...`

jupierce: Done. Added corresponding FAILURE for symmetry.

sosiouxme:
:shipit:

PR #1993, opened 2019-10-14
build/oc_sync: Accept version next to stream

joepvd:
Currently, when running build/oc_sync, one can specify the stream (e.g. stable-4). The job figures out automatically what the latest build is doing.

With multiply Y-streams in support, this mechanism does not support setting e.g. a 4.1.Z ocp release out when a newer 4.2.Z exists. This has been a stop-gap solution to fix an immediate problem. With ART-1117, a better design is sought and implemented.

### DIFF
--- a/jobs/build/oc_sync/publish-clients-from-payload.sh
+++ b/jobs/build/oc_sync/publish-clients-from-payload.sh
@@ -1,25 +1,43 @@
 #!/bin/bash
 set -eux
+
 export MOBY_DISABLE_PIGZ=true
+
 WORKSPACE=$1
 STREAM=$2
 MIRROR=$3
+VERSION="$4"
 OC_MIRROR_DIR="/srv/pub/openshift-v4/clients/$MIRROR/"
 
 SSH_OPTS="-l jenkins_aos_cd_bot -o StrictHostKeychecking=no use-mirror-upload.ops.rhcloud.com"
 
-# get latest release from release-controller API
-wget https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/${STREAM}/latest -O latest
-
-#extract pull_spec
-PULL_SPEC=`jq -r '.pullSpec' latest`
-if [ "$MIRROR" = "ocp-dev-preview" ]; then
-  # point at the published pre-release that will stay around -- registry.svc.ci gets GCed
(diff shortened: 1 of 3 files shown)


jobs/build/oc_sync/publish-clients-from-payload.sh
```diff
@@ -1,55 +1,114 @@
...
-PULL_SPEC=`jq -r '.pullSpec' latest`
-if [ "$MIRROR" = "ocp-dev-preview" ]; then
-  # point at the published pre-release that will stay around -- registry.svc.ci gets GCed
-  PULL_SPEC="${PULL_SPEC/registry.svc.ci.openshift.org\/ocp\/release/quay.io/openshift-release-dev/ocp-release-nightly}"
-fi
 
-#extract name
-VERSION=`jq -r '.name' latest`
+main() {
+  set -euxo pipefail
```
sosiouxme: This seems like it ought to stay global. It has more or less the same effect being at the beginning but... it just looks wrong and make me think.

jobs/build/oc_sync/publish-clients-from-payload.sh
```diff
@@ -1,55 +1,114 @@
...
+  # Command line arguments
+  WORKSPACE="$1"
+  STREAM="$2"
+  MIRROR="$3"
+
+  # Global variables
+  OC_MIRROR_DIR="/srv/pub/openshift-v4/clients/$MIRROR/"
+  export MOBY_DISABLE_PIGZ=true
```
sosiouxme: I know why this is here, but I think a brief explanation or link is in order.

jobs/build/oc_sync/publish-clients-from-payload.sh
```diff
@@ -37,48 +56,60 @@ get_url() {
...
-cd ${TMPDIR}
+  local pullspec outdir
+  pullspec="$1"
+  outdir="$2"
 
-OUTDIR=${TMPDIR}/${VERSION}
-mkdir -p ${OUTDIR}
-pushd ${OUTDIR}
+  mkdir -p "$outdir"
+  pushd "$pullspec"
```
sosiouxme: errr... what? ITYM
```suggestion
  pushd "$outdir"
```

sosiouxme:
@joepvd ^^ also I think this needs a new title...

sosiouxme:
pretty sure this is obsolete with how client syncs are done now :)

PR #1996, opened 2019-10-17
Honor freeze_automation in group.yml for ocp4

jupierce:
https://github.com/openshift/doozer/pull/121 must merge first.

This allows you to set builds.permitted in group.yml to either 'none', 'any', or 'manual'. If `none`, the pipeline will now exit with UNSTABLE and not make any attempt to modify distgit  or build. If `manual`, the pipeline will verify that a human executed the job (directly or indirectly via propagation).

### DIFF
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -103,12 +103,13 @@ node {
 
     currentBuild.description = ""
     try {
-        stage("initialize") { build.initialize() }
 
         sshagent(["openshift-bot"]) {
             // To work on private repos, buildlib operations must run
             // with the permissions of openshift-bot
             lock("github-activity-lock-${params.BUILD_VERSION}") {
+                stage("initialize") { build.initialize() }
+                buildlib.assertBuildPermitted(doozerOpts)
                 stage("build RPMs") { build.stageBuildRpms() }
                 stage("build compose") { build.stageBuildCompose() }
                 stage("update dist-git") { build.stageUpdateDistgit() }
@@ -119,6 +120,12 @@ node {
         }
         stage("report success") { build.stageReportSuccess() }
     } catch (err) {
+
+        if (!buildlib.isBuildPermitted(doozerOpts)) {
+            echo 'Exiting because this build is not permitted: ${err}'
+            return
+        }
+
         currentBuild.description += "\n-----------------\n\n${err}"
         currentBuild.result = "FAILURE"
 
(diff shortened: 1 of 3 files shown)


tbielawa:
This seems legit to me
/lgtm

jobs/build/ocp4/Jenkinsfile
```diff
@@ -119,6 +119,13 @@ node {
         }
         stage("report success") { build.stageReportSuccess() }
     } catch (err) {
+
+        if ( buildPlan.is_permitted == false ) {
+            echo 'Exiting because this build is not permitted: ${err}'
```
sosiouxme: Could it go in the `currentBuild.description` too?

jobs/build/ocp4/build.groovy
```diff
@@ -240,6 +243,33 @@ def planBuilds() {
         currentBuild.description += "error during plan builds step:\n${err.getMessage()}\n"
         throw err
     }
+
+    // check whether the group should be built right now
```
sosiouxme: This is OK as far as it goes, but can we extract this to commonlib and use it for the `sweep` job as well? That's another thing we don't want running during release work.

jobs/build/ocp4/build.groovy
```diff
@@ -240,6 +243,33 @@ def planBuilds() {
...
+
+    echo "Group's freeze_automation flag: ${buildPlan.freeze_automation}"
+    echo "Builder email: ${buildPlan.builderEmail}"
+
+    if (buildPlan.freeze_automation == "yes") {
+        echo "All automation are currently disabled in group.yml."
+        buildPlan.is_permitted = false
+    }
+
+    if (buildPlan.freeze_automation == "manual" && buildPlan.builderEmail == null) {
```
vfreex: The `builderEmail` part is like a hack. It doesn't donate a `-1` but would be better to handle this situation better.

jupierce: Unfortunately, this is the most elegant way I know of given the current plugins. Happy to be proven wrong, though.

jobs/build/ocp4/Jenkinsfile
jupierce: This is now done in `assertBuildPermitted`

jobs/build/ocp4/build.groovy
jupierce: I put it in buildlib with the same end goal - since it was doozer / build based.

sosiouxme:
thanks!

PR #1998, opened 2019-10-18
Disable pigz during client extraction

tbielawa:
Re: https://bugzilla.redhat.com/show_bug.cgi?id=1716550
Re: https://jira.coreos.com/browse/ART-1133

Should fix race condition we experience during releases


thiagoalessio:
I had to search what was that weird name:
> pigz, which stands for parallel implementation of gzip,
is a fully functional replacement for gzip that exploits
multiple processors and multiple cores to the hilt when compressing data.

PR #2001, opened 2019-10-21
ART-1006: Fetch OLM operators NVRs from advisories

thiagoalessio:
Related ticket: https://jira.coreos.com/browse/ART-1006

Currently our `build/appregistry` job only fetches the latest metadata build. With all those concurrent releases going on, we need to have the ability to pick precise NVRs to build metadata from.

I split this PR in 2 sections, "refactoring" and "new feature", and created explicit merges of each of those sections to make it easier for reviewers:
```
*   803fbfd9d Merge branch 'New-ART-1006-new-feature' into New-ART-1006
|\
| * 62aacf1a2 (new-feature) Fetch NVRs from advisories depending on STREAM
| * 14c546e27 Collect {component} name of images
| * c7040bbd8 Add "validate params" stage
| * ac385f1f0 Add new SOURCE_ADVISORIES parameters
|/
*   98e9d7b32 Merge branch 'New-ART-1006-refactoring' into New-ART-1006
|\
| * a53f3d7f5 (refactoring) Rename ADVISORY parameter to TARGET_ADVISORY
| * a6e7db932 Extract `attachToAdvisory` logic from stage
| * 03a9663a6 Extract `getMetadataNVRs` logic from stage
| * b72b04aa0 Move metadata advisory attachment to own stage
| * 93efa1bac Extract `pushToOMPS` logic from stage
| * 00f8190ff Extract `getImagesData` logic from stage
| * 48ab55859 Remove duplication of common elliott arguments
| * 1fef9ac99 Remove duplication of common doozer arguments
| * d24e8bd2e Access params.BUILD_VERSION through buildVersion
|/  
*
```

#### Refactoring

I needed to move some code around in order to make the addition of the new feature more intuitive. There is much more that could be refactored, but I needed to stop due to time constraints.

But the **TL;DR** is extracting some low-level details of certain operations out of the stage definition, to allow us to see "the big picture" when looking at stages.

#### New feature

Introduces a new build parameter: `SOURCE_ADVISORIES` that should be populated with advisories that contain OLM operators attached.

We read the builds attached to `SOURCE_ADVISORIES`, and filter out everything that is not an OLM operator. That way we have the exact NVRs of attached operators.

And those are the NVRs we pass to `doozer` to build metadata containers.

`SOURCE_ADVISORIES` and `TARGET_ADVISORY` only make sense if `STREAM` is either `"stage"` or `"prod"`, so there is a small validation stage right at the beginning.

#### ~~Still missing~~

~~Any `doozer` API changes introduced by https://jira.coreos.com/browse/ART-1005 will need to be added here.~~
Using changes introduced in doozer's ART-1005: https://github.com/openshift/aos-cd-jobs/pull/2001/commits/c3279967a29091b2e91e510a7b7cd1ede5e08e74

### DIFF
--- a/jobs/build/appregistry/Jenkinsfile
+++ b/jobs/build/appregistry/Jenkinsfile
@@ -203,7 +203,7 @@ node {
         sshagent(["openshift-bot"]) {
             stage("validate params") {
                 if (!validate(params)) {
-                    sh "exit 1"
+                    error "Parameter validation failed"
                 }
             }
             stage("fetch appregistry images") {
--- a/jobs/build/appregistry/Jenkinsfile
+++ b/jobs/build/appregistry/Jenkinsfile
@@ -294,28 +294,30 @@ node {
                 }
             }
             stage("attach metadata to advisory") {
-                    if (params.METADATA_ADVISORY) {
-
-                        def metadata_nvrs = getMetadataNVRs(operatorData.collect { it.name })
-
-                        elliott """
-                            change-state -s NEW_FILES
-                            -a ${params.METADATA_ADVISORY}
-                            ${params.DRY_RUN ? "--noop" : ""}
-                        """
-
-                        attachToAdvisory(params.METADATA_ADVISORY, metadata_nvrs)
-
-                        /*
-                        // this would be convenient, except that we don't have a way
-                        // to set the CDN repos first, and can't move to QE without that.
-                        elliott """
-                            change-state -s QE
-                            -a ${params.METADATA_ADVISORY}
-                            ${params.DRY_RUN ? "--noop" : ""}
-                        """
-                        */
+                    if (!params.METADATA_ADVISORY) {
+                        currentBuild.description += "\nskipping attach to advisory."
+                        return
                     }
+
+                    def metadata_nvrs = getMetadataNVRs(operatorData.collect { it.name })
+
+                    elliott """
+                        change-state -s NEW_FILES
+                        -a ${params.METADATA_ADVISORY}
+                        ${params.DRY_RUN ? "--noop" : ""}
+                    """
+
+                    attachToAdvisory(params.METADATA_ADVISORY, metadata_nvrs)
+
+                    /*
+                    // this would be convenient, except that we don't have a way
+                    // to set the CDN repos first, and can't move to QE without that.
+                    elliott """
+                        change-state -s QE
+                        -a ${params.METADATA_ADVISORY}
+                        ${params.DRY_RUN ? "--noop" : ""}
+                    """
+                    */
             }
         }
     } catch (err) {
--- a/jobs/build/appregistry/Jenkinsfile
+++ b/jobs/build/appregistry/Jenkinsfile
@@ -1,5 +1,5 @@
 @NonCPS
-def processImages(lines) {
+def parseAndFilterOperators(lines) {
     def data = []
     lines.split().each { line ->
         // label, name, nvr, version, component
@@ -164,7 +164,7 @@ node {
         }.flatten()
     }
 
-    def replaceNVRsByFoundInAdvisories = { images, advisoriesNVRs ->
+    def findImageNVRsInAdvisories = { images, advisoriesNVRs ->
         images.collect {
             image -> [
                 name: image.name,
@@ -209,11 +209,11 @@ node {
             }
             stage("fetch appregistry images") {
                 def lines = getImagesData(params.IMAGES.trim())
-                operatorData = processImages(lines)
+                operatorData = parseAndFilterOperators(lines)
 
                 if (params.STREAM in ["stage", "prod"]) {
                     def advisoriesNVRs = fetchNVRsFromAdvisories(params.OLM_OPERATOR_ADVISORIES)
-                    operatorData = replaceNVRsByFoundInAdvisories(operatorData, advisoriesNVRs)
+                    operatorData = findImageNVRsInAdvisories(operatorData, advisoriesNVRs)
 
                     if (operatorData.any { it.nvr == null }) {
                         currentBuild.description += """\n
@@ -228,7 +228,7 @@ node {
                         2. Attach missing operators to at least one of the provided advisories: ${params.OLM_OPERATOR_ADVISORIES}
                         3. Limit the expected operators in IMAGES parameter: ${operatorData.findAll { it.nvr }.collect { it.name }.join(",")}
                         """
-                        sh "exit 1"
+                        error "operators not found"
                     }
                 }
 
--- a/jobs/build/appregistry/Jenkinsfile
+++ b/jobs/build/appregistry/Jenkinsfile
@@ -260,7 +260,7 @@ node {
                         def token = retrieveBotToken()
                         for (def i = 0; i < operatorData.size(); i++) {
                             def build = operatorData[i]
-                            def metadata_nvr = doozer("operator-metadata:latest-build ${build.name}")
+                            def metadata_nvr = getMetadataNVRs([build.name])
 
                             def response = [:]
                             try {
--- a/jobs/build/appregistry/Jenkinsfile
+++ b/jobs/build/appregistry/Jenkinsfile
@@ -129,7 +129,9 @@ node {
     def validate = { params ->
         if (params.STREAM in ["stage", "prod"]) {
             if (!params.SOURCE_ADVISORIES || !params.TARGET_ADVISORY) {
-                echo "ERROR: SOURCE_ADVISORIES and TARGET_ADVISORY parameters are required for selected STREAM."
+                currentBuild.description += """\n
+                ERROR: SOURCE_ADVISORIES and TARGET_ADVISORY parameters are required for selected STREAM.
+                """
                 return false
             }
         }
--- a/jobs/build/appregistry/Jenkinsfile
+++ b/jobs/build/appregistry/Jenkinsfile
@@ -80,12 +80,12 @@ node {
                         defaultValue: 'dev',
                     ],
                     string(
-                        name: 'SOURCE_ADVISORIES',
+                        name: 'OLM_OPERATOR_ADVISORIES',
                         description: 'One or more advisories where OLM operators are attached\n* Required for "stage" and "prod" STREAMs',
                         defaultValue: '',
                     ),
                     string(
-                        name: 'TARGET_ADVISORY',
+                        name: 'METADATA_ADVISORY',
                         description: 'Advisory to attach corresponding metadata builds\n* Required for "stage" and "prod" STREAMs',
                         defaultValue: '',
                     ),
@@ -128,9 +128,9 @@ node {
 
     def validate = { params ->
         if (params.STREAM in ["stage", "prod"]) {
-            if (!params.SOURCE_ADVISORIES || !params.TARGET_ADVISORY) {
+            if (!params.OLM_OPERATOR_ADVISORIES || !params.METADATA_ADVISORY) {
                 currentBuild.description += """\n
-                ERROR: SOURCE_ADVISORIES and TARGET_ADVISORY parameters are required for selected STREAM.
+                ERROR: OLM_OPERATOR_ADVISORIES and METADATA_ADVISORY parameters are required for selected STREAM.
                 """
                 return false
             }
@@ -211,7 +211,7 @@ node {
                 operatorData = processImages(lines)
 
                 if (params.STREAM in ["stage", "prod"]) {
-                    def advisoriesNVRs = fetchNVRsFromAdvisories(params.SOURCE_ADVISORIES)
+                    def advisoriesNVRs = fetchNVRsFromAdvisories(params.OLM_OPERATOR_ADVISORIES)
                     operatorData = replaceNVRsByFoundInAdvisories(operatorData, advisoriesNVRs)
 
                     if (operatorData.any { it.nvr == null }) {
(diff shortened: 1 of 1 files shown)


jobs/build/appregistry/Jenkinsfile
```diff
@@ -120,6 +126,16 @@ node {
 
     def skipPush = params.SKIP_PUSH
 
+    def validate = { params ->
+        if (params.STREAM in ["stage", "prod"]) {
+            if (!params.SOURCE_ADVISORIES || !params.TARGET_ADVISORY) {
+                echo "ERROR: SOURCE_ADVISORIES and TARGET_ADVISORY parameters are required for selected STREAM."
```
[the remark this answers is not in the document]
thiagoalessio: yes! bd46858da

jobs/build/appregistry/Jenkinsfile
```diff
@@ -168,9 +199,33 @@ node {
     try {
         def operatorData = []
         sshagent(["openshift-bot"]) {
+            stage("validate params") {
+                if (!validate(params)) {
+                    sh "exit 1"
```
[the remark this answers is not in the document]
thiagoalessio: yes! 73de7c210

jobs/build/appregistry/Jenkinsfile
```diff
@@ -168,9 +199,33 @@ node {
...
+                    if (operatorData.any { it.nvr == null }) {
+                        echo """
+                        ERROR: The following operators were not found in provided advisories.
+                        ${operatorData.findAll { !it.nvr }.collect { it.name }.join(",")}
+
+                        Possible solutions:
+                        1. Add more advisories in SOURCE_ADVISORIES parameter, that have the missing operators attached
+                        2. Attach missing operators to at least one of the provided advisories: ${params.SOURCE_ADVISORIES}
+                        3. Limit the expected operators in IMAGES parameter: ${operatorData.findAll { it.nvr }.collect { it.name }.join(",")}
+                        """
```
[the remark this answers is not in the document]
thiagoalessio: yes! c5a5a2842

jobs/build/appregistry/Jenkinsfile
```diff
@@ -136,10 +152,25 @@ node {
...
+        commonlib.cleanCommaList(advisories).split(",").collect { advisory ->
+            readJSON(text: elliott("get --json - ${advisory}")).errata_builds.values().flatten()
+        }.flatten()
+    }
+
+    def replaceNVRsByFoundInAdvisories = { images, advisoriesNVRs ->
+        images.collect {
+            image -> [
+                name: image.name,
+                nvr: advisoriesNVRs.find { it.startsWith(image.component) }
```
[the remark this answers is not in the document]
thiagoalessio: not sure if I understood the case you are pointing out @sosiouxme ...
`advisoriesNVRs` should contain component names + nvr, right? something like:
```
cluster-logging-operator-container-v4.2.1-201910171619
```
And `image.component` would be for example `cluster-logging-operator-container`.
groovy-console

So I can't see when the case you described could happen ... even if people name their images with `something-container`, the corresponding component (attached to the advisory) would have a weird name, `something-container-container-v1.2.3`

jobs/build/appregistry/Jenkinsfile
```diff
@@ -113,37 +119,125 @@ node {
...
+        """
+    }
+
+    def fetchNVRsFromAdvisories = { advisories ->
+        commonlib.cleanCommaList(advisories).split(",").collect { advisory ->
+            readJSON(text: elliott("get --json - ${advisory}")).errata_builds.values().flatten()
+        }.flatten()
+    }
+
+    def replaceNVRsByFoundInAdvisories = { images, advisoriesNVRs ->
```
vfreex: This part is like a hack and I think it is unnecessary and error-prone. I am thinking why you need this function for replacing rather than directly use the return value `advisoriesNVRs` of `fetchNVRsFromAdvisories()`. Here is my thought:

IIUC, it is used to
1. filter out operators that are not in the list of `IMAGES` from `advisoriesNVRs`, where `IMAGES` are dist-git names (rather than brew component name).
2. supply arguments for `elliott operator-metadata:latest-build NAME`, where `NAME` is dist-git name.

So the root cause is you want to use this function to get dist-git name mapping for operators on advisory, right?

For item 1, considering why this `IMAGES` parameter exists: To build metadata for a subset of operators in case of previous failed builds. In my opinion, it is better to change the parameter to something like `OPERATOR_NVRS` to let user explicitly specify operator NVRs rather than dist-git names. Building with dist-git names sounds like the behavior is up to the time you run the job.

For item 2, I think `doozer operator-metadata:build` should be updated to give you a NVR to tracker later, so that you can directly use the returned NVR rather than fetch with `doozer operator-metadata:latest-build` afterwards because the latter also depends on timing and is possible to attach the wrong build.

Hence, I'd like to propose a change both to the PR and the doozer command for a more robust enhancement.
 
@sosiouxme Would also hear the thoughts from you.

jobs/build/appregistry/Jenkinsfile
```diff
@@ -203,41 +291,31 @@ node {
                             error "${errors}"
                         }
                     }
-                } else {
-                    if (params.ADVISORY) {
+                }
```
vfreex: Would like to short circuit if `STREAM` is not `dev`.

jobs/build/appregistry/Jenkinsfile
```diff
@@ -166,19 +260,13 @@ node {
                         def token = retrieveBotToken()
                         for (def i = 0; i < operatorData.size(); i++) {
                             def build = operatorData[i]
-                            def metadata_nvr = buildlib.doozer("--group openshift-${params.BUILD_VERSION} operator-metadata:latest-build ${build.name}", [capture: true])
+                            def metadata_nvr = doozer("operator-metadata:latest-build ${build.name}")
```
vfreex: You might want to reuse `getMetadataNVRs` here.

jobs/build/appregistry/Jenkinsfile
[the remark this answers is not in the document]
vfreex: A more robust way is to query the dist-git name from Brew. But I think it is better to get rid of the use of distgit repo name in this job. See my comments.

jobs/build/appregistry/Jenkinsfile
```diff
@@ -79,8 +80,13 @@ node {
...
-                        description: 'Should not be filled if STREAM is "dev"',
+                        name: 'SOURCE_ADVISORIES',
+                        description: 'One or more advisories where OLM operators are attached\n* Required for "stage" and "prod" STREAMs',
+                        defaultValue: '',
+                    ),
+                    string(
+                        name: 'TARGET_ADVISORY',
+                        description: 'Advisory to attach corresponding metadata builds\n* Required for "stage" and "prod" STREAMs',
                         defaultValue: '',
                     ),
```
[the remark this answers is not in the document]
thiagoalessio: agree, too vague. fixed: https://github.com/openshift/aos-cd-jobs/pull/2001/commits/299378a66aec4d2fb7e05edeb0a714347e740c83

jobs/build/appregistry/Jenkinsfile
thiagoalessio: oh yes, forgot that one ... fixed: https://github.com/openshift/aos-cd-jobs/pull/2001/commits/b5a85fcb8c62267f3ded0e28491512e3716f0ec6

jobs/build/appregistry/Jenkinsfile
```diff
@@ -113,37 +119,125 @@ node {
     def workDir = "${env.WORKSPACE}/workDir"
     buildlib.cleanWorkdir(workDir)
 
-    currentBuild.description = "Collecting appregistry images for ${params.BUILD_VERSION}"
-    currentBuild.displayName += " - ${params.BUILD_VERSION}"
+    def buildVersion = params.BUILD_VERSION
+
+    currentBuild.description = "Collecting appregistry images for ${buildVersion}"
+    currentBuild.displayName += " - ${buildVersion}"
```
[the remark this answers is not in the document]
thiagoalessio: totally! will make our lives easier. fixed: https://github.com/openshift/aos-cd-jobs/pull/2001/commits/b6550c1d9ad25ecd0f420942a24d75bc0b7dca14

jobs/build/appregistry/Jenkinsfile
```diff
@@ -113,37 +119,125 @@ node {
...
 
     def skipPush = params.SKIP_PUSH
 
+    def validate = { params ->
```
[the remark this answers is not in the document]
thiagoalessio: I used lambdas because most functions needed to access `buildlib` or `commonlib`, and those are not available outside `node { }`. Also because with lambdas I no longer needed to whitelist stuff on the groovy sandbox, nor use the `@NonCPS` decorator.

but I agree that 3 of them don't need anything I mentioned above, were just kept as lambdas for consistency with the rest, and could just be converted to regular functions:
* `validate`
* `replaceNVRsByFoundInAdvisories`
* `pushToOMPS`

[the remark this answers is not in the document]
vfreex: Makes sense.

jobs/build/appregistry/Jenkinsfile
sosiouxme: Let me take this piece by piece as it's definitely complicated... and complicated logic isn't a good sign, but the domain itself really is a bit convoluted. Probably the dual branches of logic for dev vs stage/prod are the most confusing, and this may be a landmine for anyone that tries to edit this...

> I am thinking why you need this function for replacing rather than directly use the return value advisoriesNVRs of fetchNVRsFromAdvisories(). 

The first problem to solve is determining which images are even OLM operators (only a few extras are). The doozer command does that for all or a subset of configured images (identified by distgit which is our primary key). It looks only at the config and Dockerfile, not what has actually been built.

That's a bit dev-focused, but I don't want to completely invert the logic for non-dev and work from NVR back to distgit, as a separate branch of logic. I think actually the only thing I'd change here (aside from commenting these "methods") is the name. It's really just a lookup using the components doozer identified as operators. How about:
```suggestion
    def findImageNVRsInAdvisories = { images, advisoriesNVRs ->
```
Probably best to also rename `processImages` to something more specific like `parseAndFilterOperators` as it's not just processing the output but also filtering it to just operators.

> So the root cause is you want to use this function to get dist-git name mapping for operators on advisory, right?

Well, and to determine which things are operators, whether we have advisories to draw from or not.

> For item 1, considering why this IMAGES parameter exists: To build metadata for a subset of operators in case of previous failed builds. 

I'd say rather that it exists to limit the scope of the job to those images, whether it's looking them up, rebuilding metadata, or pushing/adding it.

> In my opinion, it is better to change the parameter to something like OPERATOR_NVRS to let user explicitly specify operator NVRs rather than dist-git names. Building with dist-git names sounds like the behavior is up to the time you run the job.

That would force the user to look up NVRs for stuff. Humans are bad at that, tools should do that. Humans are thinking in terms of abstract images, not specific builds. The distgit is our primary key.

> For item 2, I think doozer operator-metadata:build should be updated to give you a NVR back for tracking later, so that you can directly use the returned NVR rather than fetch with doozer operator-metadata:latest-build afterwards because the latter also depends on timing and is possible to attach the wrong build.

I agree I'd like to see that enhancement, although doozer doesn't know beforehand the NVR it is building (the release is not specified), and determining it from the successful task is awkward (maybe you can parse `koji_builds = ...` out of the result, `brew taskinfo` gives you nothing).

But I think that's a nice to have, not a need, because we're going to be looking up the metadata container that matches a particular operator NVR and stream, and we should never have two of those going at the same time, so there shouldn't be a timing issue.

jobs/build/appregistry/Jenkinsfile
```diff
@@ -113,37 +119,126 @@ node {
...
+                        echo """
+                        ERROR: The following operators were not found in provided advisories.
+                        ${operatorData.findAll { !it.nvr }.collect { it.name }.join(",")}
+
+                        Possible solutions:
+                        1. Add more advisories in OLM_OPERATOR_ADVISORIES parameter, that have the missing operators attached
+                        2. Attach missing operators to at least one of the provided advisories: ${params.OLM_OPERATOR_ADVISORIES}
+                        3. Limit the expected operators in IMAGES parameter: ${operatorData.findAll { it.nvr }.collect { it.name }.join(",")}
+                        """
+                        sh "exit 1"
```
sosiouxme: It works, but it feels like an abuse of `sh`. How about
```suggestion
                        error "operators not found"
```

jobs/build/appregistry/Jenkinsfile
```diff
@@ -113,37 +119,126 @@ node {
...
-                    --group 'openshift-${params.BUILD_VERSION}'
-                    ${include}
-                    images:print
-                    --label 'com.redhat.delivery.appregistry'
-                    --short '{label},{name},{build},{version}'
-                """, [capture: true]
+            }
+            stage("fetch appregistry images") {
+                def lines = getImagesData(params.IMAGES.trim())
                 operatorData = processImages(lines)
```
sosiouxme: maybe...
```suggestion
                operatorData = parseAndFilterOperators(lines)
```

jobs/build/appregistry/Jenkinsfile
vfreex: @sosiouxme Thanks for the detailed explanation. Those are only some thoughts when I review the PR, but don't donate a `-1`. Some improvements can definitely be done piece by piece.

jobs/build/appregistry/Jenkinsfile
thiagoalessio: https://github.com/openshift/aos-cd-jobs/pull/2001/commits/fd4a8d7a47f4ab8f48002c716abd322010da6de3

jobs/build/appregistry/Jenkinsfile
thiagoalessio: https://github.com/openshift/aos-cd-jobs/pull/2001/commits/fd4a8d7a47f4ab8f48002c716abd322010da6de3

PR #2002, opened 2019-10-22
Use new arches code from doozer config:read-group

tbielawa:
Instead of curling github every time we run the build-sync job, we should get the arches from doozer instead.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -820,14 +820,10 @@ def parse_record_log( working_dir ) {
 // @return arches <List>: A list of the arches built for this branch
 def branch_arches(String branch) {
     echo("Fetching group config for '${branch}'")
-    def groupConfigSource = "https://raw.githubusercontent.com/openshift/ocp-build-data/${branch}/group.yml"
-    def groupConfig = sh(
-        returnStdout: true,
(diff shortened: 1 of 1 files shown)


tbielawa:
> Might save a tick if you pass in `doozerOpts` to reuse workingDir.

I suspect that's true for certain values of `tick`. In the bigger picture it'll be a negligible improvement.

jupierce:
pre-req is pending https://github.com/openshift/doozer/pull/121 is still pending.

jupierce:
whelp.. the old code will probably work fine for this use case :)

tbielawa:
> pre-req is pending [openshift/doozer#121](https://github.com/openshift/doozer/pull/121) is still pending.

My test on buildvm showed otherwise?

```
[jenkins@buildvm tmp.2sFGky5FWu]$ doozer --group=openshift-4.3 config:read-group arches --yaml
2019-10-22 14:18:30,382 INFO Cloning config data from https://github.com/openshift/ocp-build-data.git
2019-10-22 14:18:31,197 INFO Using branch from group.yml: rhaos-4.3-rhel-7
- x86_64
- ppc64le
- s390x
```

https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/tbielawa-aos-cd-jobs/job/build%252Fbuild-sync/59/console old code working

PR #2004, opened 2019-10-24
make signedComposeStateQE retry 3 times

shiywang:
Here's the interesting thing I found today when the singed-compose first attempt to move QE, elliott throw errors I guess from errata:

```
Running shell script via commonlib.shell:
elliott --group=openshift-3.10 change-state --state QE --use-default-advisory rpm
[Pipeline] sh
+ set +x
2019-10-24 02:34:00,782 INFO Cloning config data from https://github.com/openshift/ocp-build-data.git
2019-10-24 02:34:01,604 INFO Using branch from group.yml: rhaos-3.10-rhel-7
Default advisory detected: 47576
Erratum 47576: base: Validation failed: Errata Must complete RPMDiff
```

But the previous step RPM diffs resolved (which means rpmdiff.py pass). I don't have enough env to reproduce this problem... so adding a retry logic here hopefully could migrate the problems a little bit.
Screen Shot 2019-10-24 at 15 02 26

https://localhost:8888/job/aos-cd-builds/job/build%252Fsigned-compose/83/console

### DIFF
--- a/jobs/build/signed-compose/build.groovy
+++ b/jobs/build/signed-compose/build.groovy
@@ -103,7 +103,12 @@ def signedComposeRpmdiffsResolved(advisory) {
 // you're having trouble getting a build to get signed in a reasonable
 // amount of time (10 minutes).
 def signedComposeStateQE() {
-    buildlib.elliott("${elliottOpts} change-state --state QE ${advisoryOpt}")
+    def seconds = 0
+    retry(3) {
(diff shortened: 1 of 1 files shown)


vfreex:
BTW for the failure, I think the `rpmdiff resolved` step might return earlier than it should due to missing some checks. This is only an assumption because I didn't read that part of code.

PR #2006, opened 2019-10-24
[ART-1146] Enable 4.4 merges

thiagoalessio:
Not sure if this is all that is needed, took inspiration from @jupierce 's 4.3 work: https://github.com/openshift/aos-cd-jobs/pull/1973

### DIFF
--- a/pipeline-scripts/commonlib.groovy
+++ b/pipeline-scripts/commonlib.groovy
@@ -13,17 +13,19 @@ ocp3Versions = [
     "3.1",
 ]
 
-ocp4DefaultVersion = "4.2"
+ocp4DefaultVersion = "4.3"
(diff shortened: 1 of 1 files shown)


jupierce:
lgtm, but merge is unfortunately timing specific to when product engineering officially makes master track 4.4.

jupierce:
actually, I think https://github.com/openshift/aos-cd-jobs/pull/1973 removed the last vestige of `ocp4MasterVersion` . If you want to remove that value altogether, I think we can merge.

thiagoalessio:
@jupierce yeah, grepping for `ocp4MasterVersion` on the repo showed only that occurrence, nothing else is using it. now is gone ;D

PR #2017, opened 2019-10-29
Require action on failed RPM Diffs

tbielawa:
Fixes: https://jira.coreos.com/browse/ART-1200
"signed-compose incorrectly considering 'failed' statuses"

Failed rpm diffs now cause script to exit non-0. This is picked up by
the Jenkins job and handled the same way as incomplete rpmdiffs. Users
are emailed and information is included indicating there is a failure.

### DIFF
--- a/jobs/build/signed-compose/rpmdiff.py
+++ b/jobs/build/signed-compose/rpmdiff.py
@@ -39,9 +39,9 @@ def rpmdiffs_resolved(advisory):
     pp(rpmdiffs)
 
     for diff in rpmdiffs:
-        if diff['attributes']['status'] in ['INFO', 'WAIVED', 'NEEDS_INSPECTION']:
+        if diff['attributes']['status'] in ['INFO', 'WAIVED', 'PASSED']:
             completed_diffs.append(diff)
-        if diff['attributes']['status'] in ['FAILED']:
+        if diff['attributes']['status'] in ['NEEDS_INSPECTION', 'FAILED']:
             failed_diffs.append(diff)
         else:
             incomplete_diffs.append(diff)
@@ -50,13 +50,13 @@ def rpmdiffs_resolved(advisory):
         pass
     elif failed_diffs:
         print("One or more RPM Diffs FAILED and require inspection")
-            for diff in failed_diffs:
-                url = "https://rpmdiff.engineering.redhat.com/run/{}/".format(
-                    diff['attributes']['external_id'])
-                print("{status} - {nvr} - {url}\n".format(
-                    status=diff['attributes']['status'],
-                    nvr=diff['relationships']['brew_build']['nvr'],
-                    url=url))
+        for diff in failed_diffs:
+            url = "https://rpmdiff.engineering.redhat.com/run/{}/".format(
+                diff['attributes']['external_id'])
+            print("{status} - {nvr} - {url}\n".format(
+                status=diff['attributes']['status'],
+                nvr=diff['relationships']['brew_build']['nvr'],
+                url=url))
         # This will exit non-0 on its own after other checks
     else:
         print("All RPM diffs have been resolved")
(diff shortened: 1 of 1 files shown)


jobs/build/signed-compose/rpmdiff.py
```diff
@@ -33,18 +33,31 @@ def rpmdiffs_resolved(advisory):
     rpmdiffs = advisory.externalTests(test_type='rpmdiff')
     completed_diffs = []
     incomplete_diffs = []
+    failed_diffs = []
 
     print("Current RPM Diff status data:")
     pp(rpmdiffs)
 
     for diff in rpmdiffs:
-        if diff['attributes']['status'] in ['INFO', 'WAIVED', 'FAILED', 'NEEDS_INSPECTION']:
```
sosiouxme: You'll notice I'm the one that added `'FAILED', 'NEEDS_INSPECTION'`. What was happening _before_ was that if something came up with those statuses, it was considered incomplete, and so the job sat there and waited for them to change state until it timed out. These are complete, they're just failed.

jobs/build/signed-compose/rpmdiff.py
```diff
@@ -33,18 +33,31 @@ def rpmdiffs_resolved(advisory):
...
+        if diff['attributes']['status'] in ['INFO', 'WAIVED', 'NEEDS_INSPECTION']:
```
sosiouxme: `NEEDS_INSPECTION` is effectively `FAILED`. I don't know why there's a distinction.

jobs/build/signed-compose/rpmdiff.py
```diff
@@ -33,18 +33,31 @@ def rpmdiffs_resolved(advisory):
...
+        if diff['attributes']['status'] in ['FAILED']:
+            failed_diffs.append(diff)
         else:
             incomplete_diffs.append(diff)
 
     if incomplete_diffs:
         pass
+    elif failed_diffs:
+        print("One or more RPM Diffs FAILED and require inspection")
+            for diff in failed_diffs:
```
sosiouxme: why is this indented?

tbielawa: Haste

jobs/build/signed-compose/rpmdiff.py
vfreex: IIRC, the rpmdiff results from Errata Tool has less number of statuses than what you see on rpmdiff.
For example, `FAILED` in Errata Tool may means `FAILED`, `NEEDS_INSPECTION`, or the rpmdiff test is not finished. For raw rpmdiff status, you need to check `diff['attributes']['external_status']` instead.
I don't have the full list of valid statuses. Be carefully if some statuses are missing.

vfreex: Also, you might want to use `set` rather than `list` in the `in` clause:
``` python
if diff['attributes']['status'] in { 'state1', 'state2' }:
```

vfreex: Well, I got all valid RPMDiff statuses from Errata's source code. I made lists of them for Elliott in `elliott rpmdiff show` PR: https://github.com/openshift/elliott/pull/86/files#diff-fb7e98bd08dd54fd472c838e06d6224cR98-R103

IMO you need add the `PASSED` status here (`INELIGIBLE` is not used in RPMDiff actually).

vfreex: Not really. I played around the Errata API when implementing `elliott rpmdiff show`, finding out that  Errata does use more concrete statuses in its API (unlike what you see on the Web UI):
https://github.com/openshift/elliott/pull/86/files#diff-fb7e98bd08dd54fd472c838e06d6224cR75-R85

jobs/build/signed-compose/rpmdiff.py
tbielawa: Moved those to the `failed diffs` section.

tbielawa:
I think I fixed up what you two pointed out @sosiouxme @vfreex

vfreex:
/lgtm
I'd like to merge this and test it with the next OCP release.

PR #2029, opened 2019-11-07
[ART-1122] make set_client_latest multi-version aware

shiywang:
https://jira.coreos.com/browse/ART-1122

I have a few assumptions here: (base on its only 1 point card, otherwise, it will be much more complicated than this)

1. set_client_latest job was invoked automatically by release/pre-release job so there won't be any version degraded, each RELEASE version should be bigger than the previous one, eg:  `4.1.14 -> 4.1.15 -> 4.1.16`

2. this is only for ocp folder `OC_MIRROR_DIR` not for other client like odo, crc... and not for ocp-dev-preview, cuz that one involves sth string like `4.3.0-0.nightly-2019-11-07-010532`

### DIFF
--- a/jobs/build/set_client_latest/set-latest.sh
+++ b/jobs/build/set_client_latest/set-latest.sh
@@ -18,14 +18,30 @@ TMPDIR=${WORKSPACE}/tools
 mkdir -p "${TMPDIR}"
 cd ${TMPDIR}
 
-# create latest symlink
-ln -svf ${RELEASE} latest
+LESS=$(cat ${RELEASE} |awk -F '.' '{print $1 "." $2 "."  $3 "-0"}')
+ABOVE=$(cat ${RELEASE} |awk -F '.' '{print $1 "." $2+1 "."  $3 "-0"}')
+RELEASE_DIR=$(cat ${RELEASE} |awk -F '.' '{print $1 "." $2}')
+Z_LATEST=$(curl -X GET --fail -G https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest --data-urlencode "in=>${LESS} <${ABOVE}" |jq  '.name')
+Y_LATEST=$(curl -X GET --fail -G https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest |jq  '.name')
+
+if ["${RELEASE}" != "${Z_LATEST}"]; then
+    echo "Release ${RELEASE} is not the latest z-stream compare to ${Z_LATEST} on payload 4-stable/latest"
+    exit 1
+fi
+
+# create latest-{4.y} symlink
+ln -svf ${RELEASE} latest-${RELEASE_DIR}
+
+# if ${RELEASE} match the latest y stream, then update latest symlink
+if ["${RELEASE}" == "${Y_LATEST}"]; then
+    ln -svf latest-${RELEASE_DIR} latest
+fi
 
 #sync to use-mirror-upload
 rsync \
     -av --delete-after --progress --no-g --omit-dir-times --chmod=Dug=rwX \
     -e "ssh -l jenkins_aos_cd_bot -o StrictHostKeyChecking=no" \
-    latest \
+    latest-${RELEASE_DIR} \
     use-mirror-upload.ops.rhcloud.com:${OC_MIRROR_DIR}
 
 # kick off full mirror push
--- a/jobs/build/set_client_latest/set-latest.sh
+++ b/jobs/build/set_client_latest/set-latest.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -ux
+set -euxo pipefail
 
 SSH_OPTS="-l jenkins_aos_cd_bot -o StrictHostKeychecking=no use-mirror-upload.ops.rhcloud.com"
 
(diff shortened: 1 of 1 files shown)


jobs/build/set_client_latest/set-latest.sh
```diff
@@ -18,14 +18,30 @@ TMPDIR=${WORKSPACE}/tools
 mkdir -p "${TMPDIR}"
 cd ${TMPDIR}
 
-# create latest symlink
-ln -svf ${RELEASE} latest
+LESS=$(cat ${RELEASE} |awk -F '.' '{print $1 "." $2 "."  $3 "-0"}')
```
jupierce: Output after `echo` fix:
```
++ echo 4.3.0-0.nightly-2019-11-01-215341
++ awk -F . '{print $1 "." $2 "."  $3 "-0"}'
+ LESS=4.3.0-0-0
```
Suggesting to split on both `.` and `-`
```suggestion
LESS=$(cat ${RELEASE} |awk -F '[.-]' '{print $1 "." $2 "."  $3 "-0"}')
```

jobs/build/set_client_latest/set-latest.sh
```diff
@@ -18,14 +18,30 @@ TMPDIR=${WORKSPACE}/tools
...
+ABOVE=$(cat ${RELEASE} |awk -F '.' '{print $1 "." $2+1 "."  $3 "-0"}')
+RELEASE_DIR=$(cat ${RELEASE} |awk -F '.' '{print $1 "." $2}')
+Z_LATEST=$(curl -X GET --fail -G https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest --data-urlencode "in=>${LESS} <${ABOVE}" |jq  '.name')
```
jupierce: 404 can be returned by this query. With `set -e`, this would cause your script to exit.
Using `||` logic will ensure that the error is swallowed and the echo takes precedence to feed jq. You also need `-r` for jq, or it will return something like `"4.2.4"` instead of `4.2.4`. 
```suggestion
$((curl -X GET --fail -G https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest --data-urlencode "in=>4.3.0-0 <4.4.0-0" || echo '{ "name": "none" }') | jq -r '.name')
```

jobs/build/set_client_latest/set-latest.sh
```diff
@@ -18,14 +18,30 @@ TMPDIR=${WORKSPACE}/tools
...
+Y_LATEST=$(curl -X GET --fail -G https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest |jq  '.name')
```
jupierce:
```suggestion
Y_LATEST=$(curl -X GET --fail -G https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4-stable/latest | jq -r '.name')
```

jobs/build/set_client_latest/set-latest.sh
```diff
@@ -18,14 +18,30 @@ TMPDIR=${WORKSPACE}/tools
...
+if ["${RELEASE}" != "${Z_LATEST}"]; then
+    echo "Release ${RELEASE} is not the latest z-stream compare to ${Z_LATEST} on payload 4-stable/latest"
+    exit 1
+fi
+
+# create latest-{4.y} symlink
+ln -svf ${RELEASE} latest-${RELEASE_DIR}
+
+# if ${RELEASE} match the latest y stream, then update latest symlink
+if ["${RELEASE}" == "${Y_LATEST}"]; then
```
jupierce:
```suggestion
if [[ "${RELEASE}" == "${Y_LATEST}" ]]; then
```

jobs/build/set_client_latest/set-latest.sh
```diff
@@ -18,14 +18,30 @@ TMPDIR=${WORKSPACE}/tools
...
+    ln -svf latest-${RELEASE_DIR} latest
+fi
 
 #sync to use-mirror-upload
 rsync \
     -av --delete-after --progress --no-g --omit-dir-times --chmod=Dug=rwX \
     -e "ssh -l jenkins_aos_cd_bot -o StrictHostKeyChecking=no" \
-    latest \
+    latest-${RELEASE_DIR} \
```
jupierce: What is `latest` is also updated?

shiywang: So IIUC, `latest` is considered as a "secondary symlink"
We have **(before)**
``` 
latest-4.2 -> 4.2.4/
latest-4.1  -> 4.1.23/
latest -> latest-4.2
```
When a new version `4.3.0/` being created, *since* `if [[ "${RELEASE}" == "${Y_LATEST}" ]]`
we will do 
1. create `latest-4.3` symlink.
2. update "secondary symlink" point to `latest-4.3`

**(after)**
```
latest-4.2 -> 4.2.4/
latest-4.1  -> 4.1.23/
latest-4.3  -> 4.3.0/
latest -> latest-4.3
```

shiywang: yeah, `latest` also need to be rsynced

shiywang:
@jupierce updated, PTAL

jobs/build/set_client_latest/set-latest.sh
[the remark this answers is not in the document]
jupierce: @shiywang - Thanks for adding the -eo pipefail, but you will also need to change the curl invocation as described above.

jupierce:
@shiywang I tried to make this multi-arch aware too in https://github.com/openshift/aos-cd-jobs/pull/2031 . It includes your commits. 
The logic was complex enough I thought it needed to run on the upload-mirror system itself.

jobs/build/set_client_latest/set-latest.sh
```diff
@@ -18,13 +18,30 @@ TMPDIR=${WORKSPACE}/tools
 mkdir -p "${TMPDIR}"
 cd ${TMPDIR}
```
vfreex: This message is for line 10 (just adding here because I can't comment to line 10 where you don't make a change):
Could you also change the `curl` command line to include the `--fail` option so that it will fail the script if the response status code is not 2xx?
After adding that, the `| grep -q "404 Not Found"` and  line 12-15 can also be removed.

jobs/build/set_client_latest/set-latest.sh
```diff
@@ -18,13 +18,30 @@ TMPDIR=${WORKSPACE}/tools
...
```
vfreex: Better to quote `${TMPDIR}`.

vfreex:
I think this PR is effectively rewritten by https://github.com/openshift/aos-cd-jobs/pull/2031

PR #2031, opened 2019-11-08
Add latest-major.minor links and multi-arch support

jupierce:
Running in single arch mode: 
```
16:38 $ mkdir -p /srv/pub/openshift-v4/clients/ocp/4.2.4

16:38 $ ./set-v4-client-latest.sh 4.2.4 ocp all
Can't find x86_64 arch; assuming single arch mode is still in place
'latest-4.2' -> '4.2.4'
latest-4.2  now points to 4.2.4 in /srv/pub/openshift-v4/clients/ocp
'latest' -> '4.2.4'
Overall latest now points to 4.2.4 in /srv/pub/openshift-v4/clients/ocp

16:38 $ ls -l /srv/pub/openshift-v4/clients/ocp
total 4
drwxrwxr-x. 2 jupierce jupierce 4096 Nov  8 16:38 4.2.4
lrwxrwxrwx. 1 jupierce jupierce    5 Nov  8 16:38 latest -> 4.2.4
lrwxrwxrwx. 1 jupierce jupierce    5 Nov  8 16:38 latest-4.2 -> 4.2.4

16:38 $ mkdir -p /srv/pub/openshift-v4/clients/ocp/4.2.5

16:39 $ ./set-v4-client-latest.sh 4.2.5 ocp all
Can't find x86_64 arch; assuming single arch mode is still in place
'latest-4.2' -> '4.2.5'
latest-4.2  now points to 4.2.5 in /srv/pub/openshift-v4/clients/ocp
'latest' -> '4.2.5'
Overall latest now points to 4.2.5 in /srv/pub/openshift-v4/clients/ocp

16:39 $ ls -l /srv/pub/openshift-v4/clients/ocp
total 8
drwxrwxr-x. 2 jupierce jupierce 4096 Nov  8 16:38 4.2.4
drwxrwxr-x. 2 jupierce jupierce 4096 Nov  8 16:39 4.2.5
lrwxrwxrwx. 1 jupierce jupierce    5 Nov  8 16:39 latest -> 4.2.5
lrwxrwxrwx. 1 jupierce jupierce    5 Nov  8 16:39 latest-4.2 -> 4.2.5

16:39 $ ./set-v4-client-latest.sh 4.2.4 ocp all
Can't find x86_64 arch; assuming single arch mode is still in place
'latest-4.2' -> '4.2.4'
latest-4.2  now points to 4.2.4 in /srv/pub/openshift-v4/clients/ocp
'latest' -> '4.2.4'
[...]

PR #2035, opened 2019-11-12
Make release multi-arch aware

jupierce:
Includes https://github.com/openshift/aos-cd-jobs/pull/2031
Goes hand in hand with https://github.com/openshift/doozer/pull/143


tbielawa:
:+1:  to the `build-sync` changes

PR #2036, opened 2019-11-12
improve signed-compose job

sosiouxme:
* [ART-1127] give it an option to not add new builds.

### DIFF
--- a/jobs/build/signed-compose/build.groovy
+++ b/jobs/build/signed-compose/build.groovy
@@ -22,6 +22,11 @@ def initialize(advisory) {
 
 // Set the advisory to the NEW_FILES state (allows builds to be added)
 def signedComposeStateNewFiles() {
+    if (params.DRY_RUN) {
+        echo("Skipping advisory state change for dry run.")
+        echo("would have run: ${elliottOpts} change-state --state NEW_FILES ${advisoryOpt}")
+        return
+    }
     buildlib.elliott("${elliottOpts} change-state --state NEW_FILES ${advisoryOpt}")
 }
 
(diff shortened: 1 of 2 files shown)


jobs/build/signed-compose/build.groovy
```diff
@@ -22,12 +22,21 @@ def initialize(advisory) {
...
+        return
+    }
     buildlib.elliott("${elliottOpts} change-state --state NEW_FILES ${advisoryOpt}")
 }
 
 // Search brew for, and then attach, any viable builds. Do this for
 // EL7 and EL8.
 def signedComposeAttachBuilds() {
+    if (!params.ATTACH_BUILDS) {
+        echo "Job configured not to attach builds; continuing using builds already attached"
```
tbielawa: tiny thing, but (almost) all the other `echo`s use `echo(msg)` (looks like I missed `Action may be required`).

jobs/build/signed-compose/build.groovy
```diff
@@ -200,7 +220,7 @@ def signedComposeNewComposeEl8() {
     if ( params.DRY_RUN ) {
         currentBuild.description += "\nDry-run: EL8 Compose not actually built"
         echo("Packages which would have been added to rhaos-${params.BUILD_VERSION}-rhel-8-image-build tag:")
-        def testTagCmd = cmd + " --test"
+        def testTagCmd = tagCmd + " --test"
```
tbielawa: Was this busted before?

tbielawa:
left comments, but overall
/lgtm

jobs/build/signed-compose/build.groovy
sosiouxme: yes, but only if you did a dry run.

jobs/build/signed-compose/build.groovy
sosiouxme: i don't mind changing it, but maybe we should establish a jenkins pipeline style guide :-P

PR #2040, opened 2019-11-14
hardcode logging proxy image

richm:
This seems to fix logging CI


richm:
@mtnbikenc @jcantrill 

It turns out that `l_os_non_standard_reg_url` is the culprit - for now, just hardcode the proxy image - add it to the pile of tech debt to figure out a more elegant solution

richm:
well, I was partially right - removing `l_os_non_standard_reg_url` makes origin install work - but I had the proxy image wrong

richm:
@jupierce I think we're working on the same thing . . .

richm:
> @jupierce I think we're working on the same thing . . .

no, nevermind

richm:
I manually applied the patch to https://github.com/openshift/openshift-ansible/pull/11918 and it passed - please merge ASAP

PR #2042, opened 2019-11-15
Temp fix to push client to new multi-arch location

jupierce:


jobs/build/oc_sync/publish-clients-from-payload.sh
```diff
@@ -0,0 +1,70 @@
...
+    if [[ $count -lt 4 ]]; then
+      sleep 5
+    else
+      return "$exit_code"
+    fi
+  done
+}
+
+# kick off full mirror push
+retry ssh ${SSH_OPTS} timeout 15m /usr/local/bin/push.pub.sh openshift-v4 -v
```
sosiouxme: could this please be limited to
```suggestion
retry ssh ${SSH_OPTS} timeout 15m /usr/local/bin/push.pub.sh "openshift-v4/$ARCH/clients/$CLIENT_TYPE/$VERSION" -v
```
so that we only attempt to push what we just rsynced?

(I have a PR pending to do this across the board, just don't want to propagate this)

PR #2046, opened 2019-11-15
[ART-1258] scope mirror pushes to client dir

sosiouxme:
Pushing all of openshift-v4 on use-mirrors risks pushing unintended content. It's also needlessly checking the entire directory structure for updates.

This change scopes pushes as narrowly as possible to what needs to be updated. It also corrects some bad patterns that got repeated.


tbielawa:
/lgtm
None of the `rsync`'s changed, which would have made me much more nervous. Nice little `retry` there, too.

PR #2049, opened 2019-11-17
[ART-1252] remove refresh-image jobs and move attach advisory to custom job

shiywang:


jobs/build/custom/Jenkinsfile
```diff
@@ -209,6 +215,30 @@ node {
                 }
             }
 
+            stage ('Attach Images') {
+                if (rpms.toUpperCase() == "NONE" && 
+                    params.IMAGES == "" &&
+                    params.IMAGE_MODE == "update-dockerfile") {
```
vfreex: I don't think it is reasonable to to restrict the condition for attaching builds.
Suppose we need to rebuild an image with newer RPMs/sources to catch up a CVE fix, we need to rerun the `build/custom` job with `SIGNED` option checked. It would be much convenient if this job can attach the build after job completes.

jobs/build/custom/Jenkinsfile
```diff
@@ -71,6 +71,12 @@ node {
                         $class: 'hudson.model.StringParameterDefinition',
                         defaultValue: "",
                     ],
+                    [
+                        name: 'ADVISORY_ID',
```
vfreex: If this is only for image advisory ID, then it should be named `IMAGE_ADVISORY` or something.

jobs/build/custom/Jenkinsfile
```diff
@@ -209,6 +215,30 @@ node {
...
+                        def attach = ADVISORY_ID == "default" ? "--use-default-advisory image" : "--attach ${ADVISORY_ID}"
+                        try {
+                            buildlib.elliott """
+                            --group 'openshift-${OSE_MAJOR}.${OSE_MINOR}'
+                            find-builds
+                            --kind image
+                            ${attach}
+                            """
+                        } catch ( attach_err ) {
+                            throw attach_err
```
vfreex: Don't need to catch the exception here.

jobs/build/custom/Jenkinsfile
```diff
@@ -209,6 +215,30 @@ node {
...
+                        } 
+                    } else {
+                        echo 'Skipping stage...since no adviosry...'
+                    }
+                } else {
+                    echo 'Skipping stage...since not a full image build...'
```
vfreex: I don't think attaching image builds should be limited to full product build.

sosiouxme:
LGTM

sosiouxme:
Actually if we did this with 4.y it would attach all the builds, rhel7 and rhel8. That's not really what we want, but this was more intended for 3.y. Once we have https://jira.coreos.com/browse/ART-676 I'd like to enhance this with an `EXTRAS_ADVISORY_ID` to sweep the image builds into the two different advisories. Not going to worry about it for now, this is really just decommissioning refresh-images.

PR #2053, opened 2019-11-19
ART-1210: Auto run multiarch pre-release job for 4.3 and 4.4

vfreex:
Will run pre-prelease job in parallel for multiple versions and arches.

Test run with multiarch support:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/poll-payload/52/console

### DIFF
--- a/scheduled-jobs/build/poll-payload/Jenkinsfile
+++ b/scheduled-jobs/build/poll-payload/Jenkinsfile
@@ -4,63 +4,71 @@
 import groovy.transform.Field
 import java.io.FileNotFoundException
 import java.net.URLEncoder
+import groovy.json.JsonOutput
 
-// The list of release stream names to be polled
-@Field final RELEASE_STREAMS = ["4.3.0-0.nightly"]
 
-// A map of release stream names and actions
+@Field commonlib
+
+// A map of release stream names to be polled and and actions to be taken
 @Field final ACTIONS = [
-    "4.3.0-0.nightly": { releaseStream, changed, latestRelease, previousRelease ->
-        if (!changed)
-            return
-        build(
-            job: '/aos-cd-builds/build%2Fpre-release',
-            parameters: [
-                string(name: 'FROM_RELEASE_TAG', value: latestRelease.name),
-                // booleanParam(name: 'DRY_RUN', value: true),
-                booleanParam(name: 'MIRROR', value: true),
-                booleanParam(name: 'SUPPRESS_EMAIL', value: true),
-            ],
-        )
-    },
+    "4.3.0-0.nightly": this.&startPreReleaseJob,
+    "4.3.0-0.nightly-s390x": this.&startPreReleaseJob,
+    "4.3.0-0.nightly-ppc64le": this.&startPreReleaseJob,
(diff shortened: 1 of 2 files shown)


vfreex:
Updated for multiarch.

scheduled-jobs/build/poll-payload/Jenkinsfile
```diff
@@ -4,27 +4,37 @@
 import groovy.transform.Field
 import java.io.FileNotFoundException
 import java.net.URLEncoder
+import groovy.json.*
```
sosiouxme: bit broad, is there a need for anything beyond `JsonOutput`?

thiagoalessio: +1 let's import only what we need ;D

sosiouxme:
aside from the conflict, this will need some rework for the separate arch-specific release-controllers.

vfreex:
OK.

vfreex:
Finished. Please do another review. Thanks.
@thiagoalessio @sosiouxme

vfreex:
> @vfreex LGTM except I'm mildly worried about these complaints: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/poll-payload/57/console
> 
> ```
> Invoking action for 4.3.0-0.nightly
> expected to call org.kohsuke.groovy.sandbox.impl.SandboxedMethodClosure.call but wound up catching WorkflowScript.startPreReleaseJob; see: https://jenkins.io/redirect/pipeline-cps-method-mismatches/
> [Pipeline] echo (hide)
> Invoking action for 4.4.0-0.nightly
> expected to call org.kohsuke.groovy.sandbox.impl.SandboxedMethodClosure.call but wound up catching WorkflowScript.startPreReleaseJob; see: https://jenkins.io/redirect/pipeline-cps-method-mismatches/
> ```
> 
> Not digging too deep but I think this is what happens when you use iterator methods; I'm not sure exactly what effect it has, other than apparently raising an error that makes the build come out UNSTABLE.

I noticed this warning during development. I believe it is a false positive warning when calling a method via a pointer, similar to https://issues.jenkins-ci.org/browse/JENKINS-59492. Those functions are called correctly without any issues.

The `UNSTABLE` result is not set by the warning. https://github.com/openshift/aos-cd-jobs/pull/2053/files#diff-b89f72b7074288d1245c7ece87537d33R99 and https://github.com/openshift/aos-cd-jobs/pull/2053/files#diff-b89f72b7074288d1245c7ece87537d33R114 set the build result to `UNSTABLE` when it fails to call the release controller API or downstream `pre-release` job fails. Currently many non-x86_64 API endpoints return 404, such as https://openshift-release-ppc64le.svc.ci.openshift.org/api/v1/releasestream/4.3.0-0.nightly-ppc64le/latest because no accepted payload exists.

sosiouxme:
Thanks for the explanation. Hm, I'd like it to be OK to have missing endpoints (just ignore them) and report success but it doesn't seem like a reason to hold this up.

PR #2055, opened 2019-11-20
[ART-1231] adding sign-artifacts to support multiarch

shiywang:
I have a question for how to test this PR which I left as a comment in https://jira.coreos.com/browse/ART-1231

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -127,6 +127,7 @@ node {
                     dry_run: params.DRY_RUN,
                     env: "prod",
                     key_name: "redhatrelease2",
+                    arch: arch,
                 )
             }
         }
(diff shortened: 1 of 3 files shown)


tbielawa:
I think it would be pertinent to include the architecture in the message we send over the bus to the robosig. This PR gets of the essential stuff set up correctly, I can include the change I just mentioned in my upcoming signing jobs PR.

tbielawa:
> I think it would be pertinent to include the architecture in the message we send over the bus to the robosig. This PR gets of the essential stuff set up correctly, I can include the change I just mentioned in my upcoming signing jobs PR.

Unless our `release name` fields will explicitly include things like the arch? `4.2.0-s390x` ?

sosiouxme:
Our release names should include the arch, yeah. I'm a little concerned someone will forget. In any case, they'll need to be distinguished somehow in Cincinnati, and that's how we planned to do it.

PR #2079, opened 2019-12-06
Fix up RHCOS sync handling of arches

sosiouxme:
Number of issues.

1. release browser dir structure changed in 4.3+
2. destination on mirror needs to be arch-specific
3. files need to be renamed to include name and arch instead of rhcos build id

This does that.
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Frhcos_sync/9/

### DIFF
--- a/build-scripts/use-mirror/rhcossync.sh
+++ b/build-scripts/use-mirror/rhcossync.sh
@@ -1,10 +1,13 @@
 #!/bin/bash
 set -exo pipefail
 
-# RHCOS version, like 42.80.20190828.2
-VERSION=
-# Where to put this on the mirror, such as'4.2' or 'pre-release'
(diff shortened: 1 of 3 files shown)


build-scripts/use-mirror/rhcossync.sh
```diff
@@ -64,6 +67,10 @@ function downloadImages() {
     for img in $(<${SYNCLIST}); do
 	curl -L --retry 5 -O $img
```
jupierce: You might want to add a `--fail` or it will eventually pull a Justin.

sosiouxme: done, thanks!

PR #2085, opened 2019-12-13
Use UTC timestamp as OCP 4 release number

vfreex:
Currently the OCP 4 release number (e.g. `201912130945`) is based on a local timezone. Considering both Red Hat Bugzilla and Errata Tool are using UTC, our life would be easier if the release number uses UTC as well. With UTC timestamp, we can quickly determine if a bugfix is included in a build at first glance.


sosiouxme:
Having build timestamp mismatch job timestamp is still confusing; I'd rather just have buildvm on UTC.

joepvd:
BuildVM is using UTC now. Closing.

PR #2086, opened 2019-12-13
[WIP] [ART-1351] Use ${VERSION}-${ARCH} tags for Quay labels

sosiouxme:
Remove assumption that release-name == quay-tag on release job - ensure
quay tag is always arch-specific.


sosiouxme:
WIP was superseded by more successful impl

PR #2094, opened 2020-01-07
Preserve shas for nightlies on pre-release

jupierce:
context: https://coreos.slack.com/archives/CJARLA942/p1578428831151000?thread_ts=1578423101.125900&cid=CJARLA942

### DIFF
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -117,11 +117,15 @@ def getArchSuffix(arch) {
 
 def stageGenPayload(dest_repo, release_name, dest_release_tag, from_release_tag, description, previous, errata_url) {
     // build metadata blob
-    def metadata = "{\"description\": \"${description}\""
-    if (errata_url) {
-        metadata += ", \"url\": \"${errata_url}\""
+
+    def metadata = ''
+    if (description || errata_url) {
+        metadata = "{\"description\": \"${description}\""
+        if (errata_url) {
+            metadata += ", \"url\": \"${errata_url}\""
+        }
+        metadata += "}"
     }
-    metadata += "}"
 
     def arch = getReleaseTagArch(from_release_tag)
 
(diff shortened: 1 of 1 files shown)


pipeline-scripts/release.groovy
```diff
@@ -117,11 +117,15 @@ def getArchSuffix(arch) {
 
 def stageGenPayload(dest_repo, release_name, dest_release_tag, from_release_tag, description, previous, errata_url) {
     // build metadata blob
-    def metadata = "{\"description\": \"${description}\""
-    if (errata_url) {
-        metadata += ", \"url\": \"${errata_url}\""
+
+    def metadata = ''
+    if (description || errata_url) {
+        metadata = "{\"description\": \"${description}\""
```
wking: Might be easier with [`JsonOutput`][1]?  I'm trying to work up code for that approach...

[1]: https://groovy-lang.org/json.html#_jsonoutput

wking: Something like:

```groovy
def metadata = {}
if (description) {
  metadata.description = description
}
if (errata_url) {
  metadata.url = errata_url
}
```

and then later:

```groovy
if (metadata) {
  cmd += "--metadata '" + JsonOutput.toJson(metadata) + "' "
}
```

or some such?

sosiouxme: just because it's in groovy doesn't mean using it in a pipeline is easy. jenkins sandbox makes it really hard to do anything sane with json.

pipeline-scripts/release.groovy
```diff
@@ -138,7 +142,12 @@ def stageGenPayload(dest_repo, release_name, dest_release_tag, from_release_tag,
         cmd += "--previous \"${previous}\" "
     }
     cmd += "--name ${release_name} "
-    cmd += "--metadata '${metadata}' "
+
+    // Adding metadata will change the payload sha. Ideally, nightlies and
+    // pre-release shas match, so don't add it unnecessarily.
```
sosiouxme: This seems fine so I'm gonna merge it, but does it accomplish the goal? Won't `oc adm release new` create a new sha regardless? Well, maybe that was in the tl;dr. Let's see.

jupierce: Clayton believed it should be smart enough to preserve the sha if nothing is really changing.

PR #2095, opened 2020-01-08
Adding jobs for 4.5

jupierce:


scheduled-jobs/build/bug-sweeps/4.5/Jenkinsfile
```diff
@@ -0,0 +1,28 @@
+properties(
```
thiagoalessio: IIRC, @sosiouxme wants to get rid of scheduled sweeps, in favor of sweeping bugs after building something, so bugs and builds walk together ...

jupierce:
@thiagoalessio 
- nice catch on reposync. I've updated the job so it will not need independent attention in the future. 
- the puddle confs are inputs to the puddle building process (urls built here: https://github.com/openshift/aos-cd-jobs/blob/bccb39672cba0ccbb33efcdf9baf113dca875342/jobs/build/ocp4/build.groovy#L81-L83) and are necessary (until we refactor) for each release.
- I didn't want to buck the trend on bug sweeps until they were completely deleted :)

PR #2099, opened 2020-01-11
Scan all sources for changes to schedule incrementals

jupierce:


jobs/build/merge_ocp/Jenkinsfile
```diff
@@ -152,29 +171,30 @@ node {
...
+                                        echo "${downstream} openshift/ose was already up to date. Nothing to merge."
                                     }
                                 }
 
                                 successful.add(version)
                                 echo "Success running ${version} merge"
+
+                                if (params.SCHEDULE_INCREMENTAL && scheduleBuild) {
+                                    build(
+                                        job: 'build%2Focp4',
```
sosiouxme: Actually... I don't like this. Say a change comes in. A build gets kicked off. Until that build at least gets done updating distgit, this job is going to keep kicking off more builds (which will immediately block, and presumably finish quickly doing nothing once unblocked). That just seems really untidy. There will be a lot of fluff job runs. Is there any clever way to prevent builds being scheduled when already running?

sosiouxme: Nevermind this, lock check above should prevent spamming ocp4.

PR #2103, opened 2020-01-15
Adjust rhcos sync for multiarch builds

jupierce:


miabbott:
Looks reasonable to me :+1:

PR #2106, opened 2020-01-21
Reflect build/ocp4 result in scheduled builds

thiagoalessio:
As a triage manager, it is confusing to have failed builds when the subsequent jobs exited with "unstable". It takes some clicks to find out if it is a real failure or something we can safely ignore (i.e. `freeze_automation`)

I see that `b.result` is being appended to currentBuild's description, but maybe we could have a better use for that space, such as the reason for failure/unstable.

This PR is more of a conversation starter ... I'd be happy even if we don't merge it, but end up with a good explanation for why we opted for doing this:
```
currentBuild.result = (b.result == "SUCCESS") ? "SUCCESS" : "FAILURE"
```

cc @joepvd

### DIFF
--- a/scheduled-jobs/build/ocp-4.1-full/Jenkinsfile
+++ b/scheduled-jobs/build/ocp-4.1-full/Jenkinsfile
@@ -14,6 +14,6 @@ def b = build(
     ],
 )
 
-currentBuild.result = (b.result == "SUCCESS") ? "SUCCESS" : "FAILURE"
+currentBuild.result = b.result
 currentBuild.displayName = "[${b.result}] ${b.displayName}"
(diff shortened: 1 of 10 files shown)


joepvd:
Thanks for opening this discussion, @thiagoalessio :+1: 

Zooming in on the use case I ran into as triage manager:

If the `freeze_automation` flag is set, the job will discover that there is nothing to be done. Doing nothing finishes successfully. I'd find it appropriate to have the build status set to success in that case.

From `man true`: `do nothing, successfully`.

thiagoalessio:
didn't get much traction :( ... closing it for now, as my fork got outdated.

PR #2111, opened 2020-02-06
update-base-images: use most recent builds of ansible-runner and nodejs-6

yselkowitz:


sosiouxme:
dislike, but only because there ought to be a registry tag to point at to get the latest freshmaker build. but alright, let's do this.

yselkowitz:
freshmaker deliberately does NOT create additional tags for its builds, and as these are no longer publicly maintained, there really isn't anything simpler to which to point.

PR #2115, opened 2020-02-11
ART-1475: allow ignoring lock in custom builds

sosiouxme:
at the same time, narrow the scope of "github activity" lock so it does
not include extraneous activity, and make sure parallel puddle/composes
still do not conflict.

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -70,6 +70,12 @@ node {
                         $class: 'hudson.model.BooleanParameterDefinition',
                         defaultValue: false
                     ],
+                    [
+                        name: 'IGNORE_LOCKS',
+                        description: 'Do not wait for other builds in this version to complete (use only if you know they will not conflict)',
+                        $class: 'hudson.model.BooleanParameterDefinition',
+                        defaultValue: false
+                    ],
                     commonlib.suppressEmailParam(),
                     [
                         name: 'MAIL_LIST_SUCCESS',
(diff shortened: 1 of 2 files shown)


jupierce:
Would sweep running simultaneously be problematic? That was my only concern.

vfreex:
LGTM as a workaround.

The github activity lock is too coarse-grained. at lease in build/ocp4 job it locks the whole job run. I think we should split it into smaller locks and avoid unnecessary locks in the future. For example, we can split it to source rebase lock, image mirroring lock, etc. We can have even more fine-grained per-image lock in the long term.

sosiouxme:
I think running "sweep" concurrently is ok. Ought to be idempotent, and seems low risk if I'm wrong.
Concurrency is hard :) I would prioritize simplicity over optimization. Although the initial impetus for a lock here is that a rebase fails when a competing rebase updates distgit first, there really are a lot of other dependencies that could conceivably go awry given concurrent updates. The safest route is definitely to lock the whole ocp4 job. I thought a while before scoping it down at all; but the things that happen after the build seem safe enough to remove (and I'm tired of e.g. waiting for the CVP to run so appregistry can finish so the next container build can begin). Let's see how that goes.

vfreex:
Yes, Running "sweep" concurrently is definitely ok. 

I had very bad experience with the ocp4 job, which locked the custom build/appregistry job for so long (a few hours) and I was hard to debug image build failures.

I would consider locking rebase is absolutely needed, otherwise we may get a git push failure. Image build doesn't seem to be locked, since Brew/OSBS are tracking the source using commit hash, while it may cause the `latest` tag be updated by another job but do we really need to care about the latest tag? Also I am little worry about the image sync because I not sure if it tracks the built images using NVR or just `latest-build` (I didn't read that part of code carefully).

sosiouxme:
image sync is effectively `latest-build` (may actually be looking at tags in brew registry, but should normally be the same). it doesn't look at what was in the build, it just syncs everything latest. in that sense it's idempotent too -- only rolls forward.

i think this will address your problems so going to merge it for now... can refine later if it seems warranted.

PR #2121, opened 2020-02-14
crio: use podman-docker on Fedora

giuseppe:
Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/sjb/config/common/test_cases/crio.yml
+++ b/sjb/config/common/test_cases/crio.yml
@@ -108,10 +108,10 @@ post_actions:
       then
           sudo yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
           sudo yum install -y jq docker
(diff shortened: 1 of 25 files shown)


rhatdan:
NICE
LGTM
Now if we can get rid of all users of Docker.

sjb/config/common/test_cases/crio.yml
```diff
@@ -108,10 +108,10 @@ post_actions:
       then
           sudo yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
           sudo yum install -y jq docker
+          sudo systemctl start docker
```
dobbymoodge: Would `systemctl enable --now docker` be better here?

giuseppe: yes thanks, that is better, I'll add a patch for it

PR #2138, opened 2020-02-28
Demo: Automatic publish to PyPI

vfreex:
Automatically publish elliott and doozer to PyPI when a new release tag is created.

See [openshift-pipelines/README.md](openshift-pipelines/README.md) for instructions.


sosiouxme:
tl;dr
but it seems unlikely to cause problems and i trust you know what you're doing :)

PR #2139, opened 2020-03-02
Allow links to be managed by Cincinnati data

jupierce:
In addition to legacy function of setting `latest` and `latest-4.x` link to a specified release, this implementation allows the periodic management of links on the mirror to point links to releases in the Cincinnati graph.
Invocations which specify a channel name instead of a specific release can now be invoked.
`./set-v4-client-latest.sh fast-4.2 ocp fast any`  - means that the fast-4.2 channel should be queried for all architectures. For each arch, the latest release in Cincinnati should be linked as `fast-4.2` . If 4.2 happened to be latest MINOR version released, `fast` would also be linked to point to the release.

### DIFF
--- a/jobs/build/set_client_latest/Jenkinsfile
+++ b/jobs/build/set_client_latest/Jenkinsfile
@@ -18,8 +18,8 @@ node {
                 $class: 'ParametersDefinitionProperty',
                 parameterDefinitions: [
                     [
-                        name: 'RELEASE',
-                        description: 'Release Version on mirror.openshift.com to set to latest (e.g. 4.1.0)',
+                        name: 'CHANNEL_OR_RELEASE',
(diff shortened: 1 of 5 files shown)


jobs/build/set_client_latest/Jenkinsfile
```diff
@@ -29,6 +29,12 @@ node {
                         $class: 'hudson.model.StringParameterDefinition',
                         defaultValue: "ocp"
                     ],
+                    [
+                        name: 'LINK_NAME',
+                        description: 'Link name to set',
```
tbielawa: Can you include context in the description on what this affects? Is this closely tied to the `CHANEL_OR_RELEASE` parameter value? Any possibility of invalid input going in here causing heck on the mirror if it isn't checked for sanity first? Or, do the updates to the script account for bogus inputs fairly well?

tbielawa: Nice update on the description there

PR #2142, opened 2020-03-03
Fix declarative pipeline for helm_sync and odo_sync

thiagoalessio:
Since we rarely run those jobs, we didn't notice the mistake ... but today i tried to use helm_sync and it exploded, of course.


sosiouxme:
noticed this today on odo_sync :)

PR #2150, opened 2020-03-11
Switch signing-job log archiver to use rclone

tbielawa:
Previously we used `rsync`. Lately we've seen the time require for the
signing job to increase to several hours. Virtually all of that time
is during the log syncing part of the job.

This switches the log archiving to use the `rclone` tool. This should
speed things up for several reasons:

* No more `s3fusefs` abstraction layer implementing unnecessary system
  calls against the 'mounted' s3 bucket
* Only considering files created in the last week to copy
* `rclone` seems very fast in general

The required `rclone` configuration has been stored in `bitwarden`

----

I chose `1 week` for the syncing period consideration because I am fairly certain we try to sign **something** at least once a week. I considered `72h` originally but thought this would be safer. Thoughts?

### DIFF
--- a/jobs/signing/sign-artifacts/Jenkinsfile
+++ b/jobs/signing/sign-artifacts/Jenkinsfile
@@ -325,12 +325,24 @@ node {
     }
 
     stage('log sync'){
+        buildArtifactPath = env.WORKSPACE.replaceFirst('/working/', '/builds/')
+        echo "Artifact path (source to sync): ${buildArtifactPath}"
         if ( !params.DRY_RUN ) {
-            buildArtifactPath = env.WORKSPACE.replaceFirst('/working/', '/builds/')
-            dirName = buildArtifactPath.split('/')[-1]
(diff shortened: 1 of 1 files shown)


sosiouxme:
`1 week` is fine. we sign pre-releases several times a day so we could tighten it a bit if it matters but i bet it won't.

tbielawa:
> LGTM but would like to see a test run before slamming it in there... i worked hard to make this job fully testable :)

Caaaaaaaaaaan do. The dry-run part should accomplish that. I'll get back with a log real quick.

tbielawa:
> > LGTM but would like to see a test run before slamming it in there... i worked hard to make this job fully testable :)
> 
> Caaaaaaaaaaan do. The dry-run part should accomplish that. I'll get back with a log real quick.

As discussed on slack, I ran into issues with this due to jenkins automatic path shortening so I couldn't fully test it out. I'm merging this. We can revert if it actually does create issues.

PR #2153, opened 2020-03-13
Open cincinnati PRs on release

jupierce:
Example hack run of the method: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jupierce_prs/46/console

Example PRs:
https://github.com/jupierce/cincinnati-graph-data/pull/37
https://github.com/jupierce/cincinnati-graph-data/pull/38
https://github.com/jupierce/cincinnati-graph-data/pull/39

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -193,6 +193,10 @@ node {
                 }
             }
 
+            stage("channel prs") {
+                release.openCincinnatiPRs(NAME, errata_url)
+            }
+
         }
 
         dry_subject = ""
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -4,6 +4,9 @@ commonlib = load("pipeline-scripts/commonlib.groovy")
 commonlib.initialize()
 slacklib = commonlib.slacklib
 
+// Kubeconfig allowing ART to interact with api.ci.openshift.org
+ciKubeconfig = "/home/jenkins/kubeconfigs/art-publish.kubeconfig"
+
 GITHUB_URLS = [:]
 GITHUB_BASE_PATHS = [:]
 
@@ -844,14 +847,25 @@ def parse_record_log( working_dir ) {
 //
 // @param branch <String>: The name of the branch to get configs
 //   for. For example: 'openshift-4.3'
+// @param gaOnly <boolean>: If you only want group arches and do not care about arches_override.
 //
 // @return arches <List>: A list of the arches built for this branch
-def branch_arches(String branch) {
+def branch_arches(String branch, boolean gaOnly=false) {
     echo("Fetching group config for '${branch}'")
-    def d = doozer("--group=${branch} config:read-group --yaml arches",
+
+    // Check if arches_override has been specified. This is used in group.yaml
+    // when we temporarily want to build for CPU architectures that are not yet GA.
+    def arches_override = doozer("--group=${branch} config:read-group --yaml arches_override --default '[]'",
+            [capture: true]).trim()
+    def arches_override_list = readYaml(text: arches_override)
+    if ( !gaOnly && arches_override_list ) {
+        return arches_override_list
+    }
+
+    def arches = doozer("--group=${branch} config:read-group --yaml arches",
 		   [capture: true]).trim()
-    def datas = readYaml(text: d)
-    return datas
+    def arches_list = readYaml(text: arches)
+    return arches_list
 }
 
 // Search the build log for failed builds
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -460,4 +460,229 @@ def signArtifacts(Map signingParams) {
     )
 }
 
+/**
+ * Opens a series of PRs against the cincinnati-graph-data GitHub repository.
+ *    Specifically, a PR for each channel prefix (e.g. candidate, fast, stable) associated with the specified release
+ *    and the next minor channels (major.minor+1) IFF those channels currently exist.
+ * @param releaseName The name of the release (e.g. "4.3.6")
+ * @param errata_url The errata associated with the release.
+ * @param ghorg For testing purposes, you can call this method specifying a personal github org/account. The
+ *        openshift-bot must be a contributor in your fork of cincinnati-graph-data.
+ */
+def openCincinnatiPRs(releaseName, errata_url, ghorg = 'openshift') {
+    def (major, minor) = commonlib.extractMajorMinorVersionNumbers(releaseName)
+    if ( major != '4' ) {
+        error("Unable to open PRs for unknown major minor: ${major}.${minor}")
+    }
+    def minorNext = "${minor.toInteger() + 1}"
+    dir(env.WORKSPACE) {
+        sshagent(["openshift-bot"]) {
+
+            // PRs that we open will be tracked in this file.
+            prs_file = "${env.WORKSPACE}/prs.txt"
+            sh "rm -f ${prs_file} && touch ${prs_file}"  // make sure we start with a clean slate
+
+            sh "git clone git@github.com:${ghorg}/cincinnati-graph-data.git"
+            dir('cincinnati-graph-data/channels') {
+                def prefixes = [ "candidate", "fast", "stable"]
+                if ( major == '4' && minor == '1' ) {
+                    prefixes = [ "prerelease", "stable"]
+                }
+
+                prURLs = [:]  // Will map channel to opened PR
+                for ( String prefix : prefixes ) {
+                    channel = "${prefix}-${major}.${minor}"
+                    channelFile = "${channel}.yaml"
+                    upgradeChannel = "${prefix}-${major}.${minorNext}"
+                    upgradeChannelFile = "${upgradeChannel}.yaml"
+
+                    channelYaml = [ name: channel, versions: [] ]
+                    if (fileExists(channelFile)) {
+                        channelYaml = readYaml(file: channelFile)
+                    } else {
+                        // create the channel if it does not already exist
+                        writeFile(file: channelFile, text: "name: ${channel}\nversions:\n" )
+                    }
+
+                    /**
+                     * @param name - The release name
+                     * @param version - a list of versions in a channel
+                     * @return Returns true if name or name+<arch> exists in list
+                     */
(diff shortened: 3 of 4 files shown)


jupierce:
Needs agreement from OTA. sdodson to arrange meeting.

jupierce:
Updated to reflect: https://docs.google.com/document/d/1B2HxFP_2yhAbAXxuhkaWkqEh3I0ScZqe-N4syoDAvt8/edit?ts=5e78fcf7#

pipeline-scripts/buildlib.groovy
```diff
@@ -1337,4 +1337,138 @@ def getChanges(yamlData) {
...
+                            if [[ "${addToChannel}" == "true" ]]; then
+                                echo >> ${channelFile}    # add newline
+                                echo '# Errata: ${errata_url}' >> ${channelFile}    # add link to errata for reference
+                                echo '- ${releaseName}' >> ${channelFile}   # add the entry
+                                git add ${channelFile}
+                            fi
+                            if [[ "${addToUpgradeChannel}" == "true" ]]; then
+                                echo >> ${upgradeChannelFile}    # add newline
+                                echo '# Errata: ${errata_url}' >> ${upgradeChannelFile}    # add link to errata for reference
+                                echo '- ${releaseName}' >> ${upgradeChannelFile}   # add the entry
```
sosiouxme: This is going to add it at the end of the file. It's perhaps a minor thing but if you look at e.g. https://github.com/openshift/cincinnati-graph-data/blob/master/channels/candidate-4.3.yaml the 4.2 "upgrade from" stream is separate from the 4.3 stream. OTA team likes this to be human-readable. Perhaps we could `sed` the line in before the first {major}.{minor}.0 line?

sosiouxme: The other problem that's evident in that file is that upgrades from 4.2 to 4.3 require the `+amd64` prefix for as long as 4.3 isn't shipping Z. No idea how to handle that.

pipeline-scripts/buildlib.groovy
```diff
@@ -1337,4 +1337,138 @@ def getChanges(yamlData) {
     return changed
 }
 
+/**
+ * Opens a series of PRs against the cincinnati-graph-data GitHub repository.
+ *    SetA: A PR for each channel (e.g. candidate, fast, stable) associated with the major.minor of the release.
+ *    SetB: A PR for each channel (e.g. candidate, fast, stable) associated with the major.(minor+1) of the release.
+            SetB is only created if the major.(minor+1) channels already exist.
```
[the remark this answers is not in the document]
jupierce: Who keeps comments current?! :) 
Will update in next push.

jobs/build/release/Jenkinsfile
```diff
@@ -193,6 +193,10 @@ node {
                 }
             }
 
+            stage("open prs") {
```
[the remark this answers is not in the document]
jupierce: Next push.

pipeline-scripts/buildlib.groovy
```diff
@@ -1337,4 +1337,138 @@ def getChanges(yamlData) {
...
+            prs_file = "${env.WORKSPACE}/prs.txt"
+            sh "rm -f ${prs_file} && touch ${prs_file}"  // make sure we start with a clean slate
+
+            sh "git clone git@github.com:${ghorg}/cincinnati-graph-data.git"
+            dir('cincinnati-graph-data/channels') {
+                def prefixes = [ "candidate", "fast", "stable"]
+                if ( major == '4' && minor == '1' ) {
+                    prefixes = [ "prerelease", "stable"]
+                }
+                for ( String prefix : prefixes ) {
```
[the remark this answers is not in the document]
jupierce: I've had luck with this on the current version of Jenkins.

pipeline-scripts/buildlib.groovy
jupierce: @sosiouxme https://github.com/openshift/doozer/pull/197 for consideration on how to accomplish this.

jupierce: The new push does not use https://github.com/openshift/doozer/pull/197 . Instead, it checks for the existence of the next minor's latest release in the various api.ci namespaces to determine if that release was built / released for each architecture.

jupierce: Current push contains this latter strategy.

pipeline-scripts/release.groovy
```diff
@@ -460,4 +460,229 @@ def signArtifacts(Map signingParams) {
...
+                            set -o xtrace
+                            if git checkout origin/${branchName} ; then
+                                echo "The branch ${branchName} already exists in cincinnati-graph-data. No additional PRs will be created."
+                                exit 0
+                            fi
+                            git branch -f ${branchName} origin/master
+                            git checkout ${branchName}
+                            if [[ "${addToChannel}" == "true" ]]; then
+                                echo >> ${channelFile}    # add newline
+                                echo '# Errata: ${errata_url}' >> ${channelFile}    # add link to errata for reference
```
sosiouxme, line 649: to be crystal clear which release the comment refers to, how about:
```suggestion
                                echo '# ${releaseName} Errata: ${errata_url}' >> ${channelFile}    # add link to errata for reference
```

pipeline-scripts/release.groovy
```diff
@@ -460,4 +460,229 @@ def signArtifacts(Map signingParams) {
...
+                                echo '- ${releaseName}' >> ${channelFile}   # add the entry
+                                git add ${channelFile}
+                            fi
+                            if [[ "${addToUpgradeChannel}" == "true" ]]; then
+                                echo >> ${upgradeChannelFile}    # add newline
+                                echo '# Errata: ${errata_url}' >> ${upgradeChannelFile}    # add link to errata for reference
```
sosiouxme, line 655: same here. although i think this whole comment might be superfluous for an "upgrade from"?

also, this is still going at the end of the file... is OTA team going to be OK with upgrades being intermingled with channel contents?

pipeline-scripts/release.groovy
jupierce: Next push

pipeline-scripts/release.groovy
jupierce: sed'ing it in would be tricky (but I'll admit to not having sed-fu). Given the following file:
```
....
versions:
- 4.2.0
- 4.2.1

# Something really important about 4.3.0
- 4.3.0
```

If I'm inserting 4.2.99 above the first 4.3 instance, my simple approach would yield:
```
versions:
- 4.2.0
- 4.2.1

# Something really important about 4.3.0
- 4.2.99
- 4.3.0
```

Whitespace and comments will wreak havoc unless I make it fairly complex. 

I've added a new commit which tries to keep things separate, but also simple, by adding 4.x-1 versions directly under `versions:` Example PR: https://github.com/jupierce/cincinnati-graph-data/pull/42

pipeline-scripts/release.groovy
```diff
@@ -460,4 +460,235 @@ def signArtifacts(Map signingParams) {
...
+
+                                /**
+                                 * @param name - The name of the release to look for
+                                 * @param arch - The architecture you want to know if the release is defined for
+                                 * @return Returns true if the release specified has support for the specified architecture
+                                 */
+                                def existsForArch = { name, arch ->
+                                    def suffix = getArchSuffix(arch)
+                                    // If there is an imagestream in the arch's api.ci namespace, it was built for that arch by ART.
+                                    // Check for both is/release and is/release-<ach> (the latter being an older style of release naming)
```
sosiouxme, line 550:
```suggestion
                                    // Check for both is/release and is/release-<arch> (the latter being an older style of release naming)
```
... which we will never use again, so I'm wondering if it really should be looking for both.

pipeline-scripts/release.groovy
```diff
@@ -460,4 +460,235 @@ def signArtifacts(Map signingParams) {
...
+
+                                /**
+                                 * For each of the arches in the current release, see if a counterpart exists for the
+                                 * latest 4.y+1 release. If it is found, record that we are going to add an arch specific
+                                 * entry for the various channels (e.g. ["4.y.z+amd64", "4.y.z+s390x"] if both builds have
+                                 * x86_64 & s390x, but only 4.y.z has ppc64le)
+                                 */
+                                currentReleaseArches = buildlib.branch_arches('openshift-${major}.${minor}', true)
+                                for ( String arch : currentReleaseArches ) {
+                                    if ( existsForArch(lastNextVer, arch) ) {
```
sosiouxme, line 566: i'm wondering if the existence of the next release's imagestream is a sufficient condition to begin adding upgrade-ables in its channel. if i'm understanding correctly, a 4.4 RC release right now would generate a PR adding it into the 4.5 channels as an upgrade-able; the 4.5 channels would have nothing but 4.4 releases for months until we have a 4.5 RC, which would surely look confusing. i would think we don't want to add to upgrade channels until there's something there to upgrade *to*. we could either check the release-controller `4-stable` stream for the presence of something or the contents of the channel to see if there is anything for 4.5 before we start adding 4.4 entries. WDYT?

pipeline-scripts/release.groovy
jupierce: Just my personal paranoia shining through.. that we will switch back for some reason. Or run one accidentally and just run with it because we are out of time.

pipeline-scripts/release.groovy
jupierce: There are two branches which try to prevent early creation / addition of 4.y to the 4.y+1 stream in that scenario. 
1. The 4.y+1 channel file must already exist: https://github.com/openshift/aos-cd-jobs/blob/a1ddb31b999258389035c5b7744fb33375df3302/pipeline-scripts/release.groovy#L526 . It will not be created by a 4.y release. 
2. If the file does exist, it must contain a 4.y+1 entry before the code will consider adding a 4.y entry: https://github.com/openshift/aos-cd-jobs/blob/a1ddb31b999258389035c5b7744fb33375df3302/pipeline-scripts/release.groovy#L540 . In your scenario, even if the 4.5 files exist, if they don't have a 4.5 entry, then `lastNextVer` in the code should be an empty string and skip setting `releasesForUpgradeChannel = true`.

sosiouxme: Thanks, I did not see the second part. Seems like you have it covered.

PR #2164, opened 2020-03-19
Retry and pause when repairing/sweeping bugs

tbielawa:
Re: https://issues.redhat.com/browse/ART-1677

Sometimes the bugzilla API flakes out and gives us a network proxy
error. So, let's just retry the call a few times, give it a minute
between attempts just to be safe.

### DIFF
--- a/jobs/build/sweep/Jenkinsfile
+++ b/jobs/build/sweep/Jenkinsfile
@@ -57,24 +57,31 @@ node {
     def kind = (major == '4') ? 'image' : 'rpm'
 
     stage("Repair bug state") {
-        try {
-            currentBuild.description += "* Moving attached bugs in MODIFIED state to ON_QA...\n"
-            buildlib.elliott "--group=openshift-${version} repair-bugs --use-default-advisory ${kind} --auto" 
-        } catch (elliottErr) {
-            currentBuild.description = "Error repairing:\n${elliottErr}"
-            throw elliottErr
+        currentBuild.description += "* Moving attached bugs in MODIFIED state to ON_QA...\n"
+        retry (3) {
+            try {
+                buildlib.elliott "--group=openshift-${version} repair-bugs --use-default-advisory ${kind} --auto"
+            } catch (elliottErr) {
+                echo("Error repairing (will retry a few times):\n${elliottErr}")
+                sleep(time: 1, unit: 'MINUTES')
+                throw elliottErr
+            }
         }
     }
 
     stage("Sweep bugs") {
-        try {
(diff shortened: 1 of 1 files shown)


jobs/build/sweep/Jenkinsfile
```diff
@@ -59,7 +59,10 @@ node {
     stage("Repair bug state") {
         try {
             currentBuild.description += "* Moving attached bugs in MODIFIED state to ON_QA...\n"
-            buildlib.elliott "--group=openshift-${version} repair-bugs --use-default-advisory ${kind} --auto" 
+            retry (3) {
+                buildlib.elliott "--group=openshift-${version} repair-bugs --use-default-advisory ${kind} --auto"
+                sleep(time: 1, unit: 'MINUTES')
```
joepvd: Would be nice not to wait when it succeeds.

tbielawa: Whoops! That's a good point.

tbielawa:
@joepvd OK, I think I got it like you were saying to try out. What do you think now?

jobs/build/sweep/Jenkinsfile
```diff
@@ -57,24 +57,31 @@ node {
...
     stage("Repair bug state") {
-        try {
-            currentBuild.description += "* Moving attached bugs in MODIFIED state to ON_QA...\n"
-            buildlib.elliott "--group=openshift-${version} repair-bugs --use-default-advisory ${kind} --auto" 
-        } catch (elliottErr) {
-            currentBuild.description = "Error repairing:\n${elliottErr}"
-            throw elliottErr
+        retry (3) {
+            try {
+                currentBuild.description += "* Moving attached bugs in MODIFIED state to ON_QA...\n"
```
joepvd: This can:
- end up in build log
- before the `retry` block, so it only shows once

jobs/build/sweep/Jenkinsfile
```diff
@@ -57,24 +57,31 @@ node {
...
+                buildlib.elliott "--group=openshift-${version} repair-bugs --use-default-advisory ${kind} --auto"
+            } catch (elliottErr) {
+                sleep(time: 1, unit: 'MINUTES')
+                currentBuild.description += "Error repairing (will retry a few times):\n${elliottErr}"
```
joepvd: Just log the retry. No need for it to end up in the description.

sosiouxme:
Second joep's advice to not have retries add to build description.

Otherwise, LGTM

tbielawa:
Moved the `retries` and `description` around a bit. Should be good to go now!

PR #2167, opened 2020-03-20
Hack'n'Hustle: Auto update pipeline jobs and rebuild images on merge

vfreex:
This adds 2 OpenShift Pipeline jobs `aos-cd-jobs-poll-master` and `update-jobs`.

`aos-cd-jobs-poll-master` is a scheduled job which polls the `master` branch of `aos-cd-jobs` repo
and triggers `update-jobs` job on merge.

`update-jobs` checks changed files from changelog then reconfigures jobs and other API objects defined in `openshift-pipelines/{jobs,infra}` when there is a change.
It also rebuilds the images if Dockerfiles in `openshift-pipelines/images` are changed.

Actual job code is defined in `/vars` with Jenkins Shared Library. The directory is moved to https://github.com/openshift/art-ci-toolkit.git with PR https://github.com/openshift/art-ci-toolkit/pull/5/.


sosiouxme:
@vfreex it's taken an embarrassingly long time for me to start looking at this. And I have to admit I don't remember what you said when you mentioned you had worked on this.

Before I look deeper, I'd appreciate a summary of the strategy here. It sounds like you want to set up the infrastructure so that we can move everything to using shared libraries, instead of copying them out to all the job branches like we do now. Is that right? That seems like a great goal. With this in place, do we move everything out of `pipeline-scripts` at once, or does each job need to be individually updated, or... how does that work? Or am I completely missing something...

vfreex:
@sosiouxme Sorry for keeping your question open for so long. I didn't pay much attention to GitHub PRs when my laptop was broken. Will explain to you later.

vfreex:
We already have 4 jobs running on our PSI Jenkins (created on more previous Hack'n'Hustle days): https://github.com/openshift/aos-cd-jobs/tree/master/openshift-pipelines/jobs. They are used to automatically release Elliott and Doozer to PyPI on tag creation. I think most of you guys have noticed that. Each .yaml file represents one Jenkin job, where we can embed a Jenkinsfile or reference an external Jenkinsfile there. Those jobs run workloads in dynamic OpenShift pods on PSI, so there is a Dockerfile (https://github.com/openshift/aos-cd-jobs/tree/master/openshift-pipelines/images) for the jenkins-agent with Elliott, Doozer, ocp-build-data-validator installed.
The 2 PRs are about to automate more:  Whenever the Dockerfile is changed, it will automatically trigger the image rebuild. Whenever the job .yaml file is changed, auto-update the job configuration on our PSI Jenkins.
My intention is people who want to create a new job on our PSI Jenkins, he just needs to drop a new .yaml in that directory, modify the Dockerfile in anyway he wants, the job will be automatically created and container images are rebuilt as well. That would make the life easier.
About the Jenkins shared library https://github.com/openshift/art-ci-toolkit.git:  @thiagoalessio already uses that for Elliott/Doozer/ocp-build-data-validator CI. I just keep that practice and found it very useful for sharing common code among jobs. For example, the logic of releasing Elliott and Doozer is basically the same. So I keep the logic in the shared library but call the library functions in each job's Jenkinsfile with different parameters.

PR #2181, opened 2020-04-02
Request upgrade tests with slack

jupierce:
Right now it seems that cluster-bot will ignore requests directly from art-bot. So notifying art-team as an intermediary until I can figure that out.

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -158,9 +158,13 @@ node {
                     testLines << "test upgrade ${from_release} ${NAME} ${mode}"
                     testIndex++
                 }
-                def slackChannel = slacklib.to('#team-art')
-                slackChannel.say("Hi @release-artists . A new release is ready and needs some upgrade tests to be triggered. "
+                try {  // don't let a slack outage break the job
+                    def slackChannel = slacklib.to('#team-art')
+                    slackChannel.say("Hi @release-artists . A new release is ready and needs some upgrade tests to be triggered. "
                         + "Please open a chat with @cluster-bot and issue each of these lines individually:\n${testLines.join('\n')}")
+                } catch(ex) {
+                    echo "slack notification failed: ${ex}"
+                }
                 currentBuild.description += "\n@cluster-bot requests:\n${testLines.join('\n')}\n"
             }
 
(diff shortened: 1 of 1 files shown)


jobs/build/release/Jenkinsfile
```diff
@@ -207,6 +207,24 @@ node {
                 }
             }
 
+            stage("request upgrade tests") {
```
sosiouxme, line 210: Why wait until everything else is done? They should start upgrade tests as soon as the release exists, i.e. right before `stage("wait for stable")`

jobs/build/release/Jenkinsfile
```diff
@@ -207,6 +207,24 @@ node {
...
+                }
+                def testIndex = 0
+                def testLines = []
+                for ( String from_release : previousList) {
+                    mode = modeOptions[testIndex % modeOptions.size()]
+                    testLines << "test upgrade ${from_release} ${NAME} ${mode}"
+                    testIndex++
+                }
+                def slackChannel = slacklib.to('#team-art')
+                slackChannel.say("Hi @art-team . A new release is ready and needs some upgrade tests to be triggered. Please open a chat with @cluster-bot and issue each of these lines individually:\n${testLines.join('\n')}")
```
sosiouxme, line 225: If this is done immediately, it would make sense to scope it to the `@release-artists`.
```suggestion
                slackChannel.say("Hi @release-artists . A new release is ready and needs some upgrade tests to be triggered. Please open a chat with @cluster-bot and issue each of these lines individually:\n${testLines.join('\n')}")
```
And also, add it to the `currentBuild.description` (why not?).

sosiouxme: Also, break up this line, it makes the review impossibly wide :stuck_out_tongue_winking_eye:

jobs/build/release/Jenkinsfile
jupierce: Great insight. Next push.

jobs/build/release/Jenkinsfile
jupierce: I originally had release-artist, but I also thought they might struggle with the limit of ten jobs at a time - making it require a bit of team work. But I think you are right, we can start with artist and adapt if it is a problem.  Next push.

PR #2185, opened 2020-04-09
tarball-sources: change project from RCM to EXD SP Cloud Distribution

shiywang:


shiywang:
@vfreex unfortunately, this tcl script doesn't work very well, I am not sure why it that, I spend the whole day trying to figure out but still haven't, so I plan to use curl instead of this jira tool and I met some other problems, like escape path inside the json etc... still wip
/hold

shiywang:
I wrote a cronjob to temporarily resolve this problem, will add more code once I finished this week's release.

PR #1974, opened 2019-09-23
Add 4.3 puddle configurations

jupierce:

PR #2193, opened 2020-04-17
Add 4.6 puddle configs

tbielawa:
https://issues.redhat.com/browse/ART-1632

This was based off the example in the template card on JIRA which references https://github.com/openshift/aos-cd-jobs/pull/1974/files

Note that one of the files in there doesn't exist anymore, `build-scripts/puddle-conf/errata-puddle-4.3-signed.conf`. In 8389eaaf6436149288f427c19b00d0cc528f10af @sosiouxme removed many of the signed files because:

> Now that we no longer build composes directly from errata, but instead from a tag, these are obsolete.


tbielawa:
updated

PR #2194, opened 2020-04-17
Add ose4.6 jobs for branch cut

tbielawa:
https://issues.redhat.com/browse/ART-1636

### DIFF
--- a/pipeline-scripts/commonlib.groovy
+++ b/pipeline-scripts/commonlib.groovy
@@ -12,6 +12,7 @@ ocp4DefaultVersion = "4.5"
 
 // All buildable versions of ocp4
 ocp4Versions = [
+    "4.6",
     "4.5",
     "4.4",
     "4.3",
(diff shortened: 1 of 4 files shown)


pipeline-scripts/commonlib.groovy
```diff
@@ -8,10 +8,11 @@ ocp3Versions = [
 ]
 
 // Which version of ocp4 build parameters should show by default
-ocp4DefaultVersion = "4.5"
+ocp4DefaultVersion = "4.6"
```
sosiouxme, line 11: this is vestigial... see https://github.com/openshift/aos-cd-jobs/pull/2198

pipeline-scripts/commonlib.groovy
```diff
@@ -21,6 +22,7 @@ ocp4Versions = [
 
 // Which versions should undergo merges from origin->ose
 ocpMergeVersions = [
+    "4.6",
```
sosiouxme, line 25: i would wait to do this until we have the builds running fairly smoothly or we'll likely be running failures every 15 minutes.
maybe this should be broken out into a separate step in the GA playbook...

tbielawa:
Updated

PR #2202, opened 2020-04-22
Cincinnati PR opening changes

jupierce:
- OTA has requested that the link provided be the internal errata tool URL (since the live-id doesn't show anything until the advisory ships) and that we do not put the advisory information in the YAML.
- OTA also asked that we only create the PRs after x86_64 release is accepted by release controller. 
- Extract PR opening into separately runnable job


vfreex:
I'll merge it and try it with next 4.4 RC or GA promotion.
/lgtm

PR #2204, opened 2020-04-24
Programmatic previous

jupierce:
Depends on https://github.com/openshift/doozer/pull/208

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -48,9 +48,9 @@ node {
                     ],
                     [
                         name: 'PREVIOUS',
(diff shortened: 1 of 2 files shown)


thiagoalessio:
```
[jenkins@buildvm ~]$ which doozer
~/.local/bin/doozer
[jenkins@buildvm ~]$ doozer --version
Doozer v1.0.19
Python v3.6.8 (default, Jun 11 2019, 15:15:01)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]

The Doozers don't mind their buildings being eaten;
if the Fraggles didn't eat the constructions,
the Doozers would run out of building space,
and if they ran out of building space,
they would have to move away from Fraggle Rock
or else they would die.
```

PR #2207, opened 2020-04-27
Generate 'releases' for send-umb-message from commonlib.ocp4Versions

tbielawa:
Discovered during 4.6 branch cut when we accidentally missed adding
4.6 to the send-umb-message job

### DIFF
--- a/pipeline-scripts/commonlib.groovy
+++ b/pipeline-scripts/commonlib.groovy
@@ -16,6 +16,12 @@ ocp4Versions = [
     "4.1",
 ]
 
+// Send UMB messages for these new nightlies, yields a list like:
+//
+//     [4-stable, 4.1.0-0.nightly, 4.2.0-0.nightly, 4.3.0-0.nightly, ...]
+ocp4SendUMBVersions = ocp4Versions.collect([]) { it + ".0-0.nightly" }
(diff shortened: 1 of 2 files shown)


pipeline-scripts/commonlib.groovy
```diff
@@ -16,6 +16,12 @@ ocp4Versions = [
     "4.1",
 ]
 
+// Send UMB messages for these new nightlies, yields a list like:
+//
+//     [4-stable, 4.1.0-0.nightly, 4.2.0-0.nightly, 4.3.0-0.nightly, ...]
+ocp4SendUMBVersions = ocp4Versions.collect([]) { it + ".0-0.nightly" }
+ocp4SendUMBVersions.add(0, "4-stable")
```
sosiouxme, line 23: I think this could be 
```suggestion
ocp4SendUMBVersions = ocp4Versions.collect(["4-stable"]) { it + ".0-0.nightly" }
```
... but it's not important :)

4-stable will likely end up needing special treatment anyway.

tbielawa: THAT'S how that method works. Nice. The docs weren't very helpful with that part. I'm going to make that change, push it, and then merge this.

PR #2212, opened 2020-05-01
[WIP] signed-compose fixups

sosiouxme:


jobs/build/signed-compose/rpmdiff.py
```diff
@@ -14,13 +14,19 @@ def rpmdiffs_ran(advisory):
...
+    nvrs_with_diffs = set()
     for diff in rpmdiffs:
         print_diff(diff)
+        nvrs_with_diffs.add(diff['relationships']['brew_build']['nvr'])
         if diff['attributes']['status'] in ['QUEUED_FOR_TEST', 'RUNNING', 'PENDING']:
             not_finished_diffs.append(diff)
 
-    if not_finished_diffs:
-        print("There are {} rpmdiffs which have not ran or completed yet".format(
+    if any(build not in nvrs_with_diffs for builds in advisory.errata_builds.values() for build in builds):
```
joepvd, line 24: The logic per line ration is rather high here :) Would a refactor into a multiline function with a descriptive name be an option?

sosiouxme: but it's so obvious!!

sosiouxme:
will not spend more effort on this moribund job.

PR #2213, opened 2020-05-02
commonlib.shell: safely run in parallel

sosiouxme:
There are a number of places I'd like to parallelize but couldn't because commonlib.shell files would be scrambled by race conditions. Now it can be done.

~~Note this does add some noise in the console about locking, and will add tiny delays obtaining and releasing the global lock (hopefully not significant compared to the execution time of the shell scripts).~~

I followed Justin's suggestion of using `AtomicInteger` and symlinking cleverness to avoid the need for any lock.

There are a few other minor tweaks I added to clarify comments and handle edge cases better.

### DIFF
--- a/pipeline-scripts/commonlib.groovy
+++ b/pipeline-scripts/commonlib.groovy
@@ -211,7 +211,15 @@ def safeArchiveArtifacts(List patterns) {
     }
 }
 
-shellIndex = 0
+
+import java.util.concurrent.atomic.AtomicInteger
+shellCounter = new AtomicInteger()
+
+@NonCPS
+def getNextShellCounter() {
+    return shellCounter.incrementAndGet()
+}
+
 /**
  * Wraps the Jenkins Pipeline sh step in order to get actually useful exceptions.
  * Because https://issues.jenkins-ci.org/browse/JENKINS-44930 is ridiculous.
@@ -222,8 +230,9 @@ shellIndex = 0
  * (0 for success, 1 for failure).
  * If your code really cares about the rc being right then do not use.
  *
+ * @param alwaysArchive true will always archive shell artifacts; default is to do so only on failure
  * @param returnAll true causes to return map with stderr, stdout, combined, and returnStatus
- * @param otherwise same as sh() command
+ * otherwise params are same as sh() step
  * @return same as sh() command, unless returnAll set
  * @throws error (unless returnAll or returnStatus set) with useful message (and archives)
  */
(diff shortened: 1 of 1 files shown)


jupierce:
For https://stackoverflow.com/q/61553123 , I suspect the issue is the odd way in which Jenkins groovy code executes parallel with custom user space green threads: 
From https://github.com/jenkinsci/workflow-cps-plugin:
```
All program logic is run inside a “CPS VM thread”, which is just a Java thread pool that can run binary methods and figure out which continuation to do next. The parallel step uses “green threads” (also known as coöperative multitasking): it records logical thread (~ branch) names for various actions, but does not literally run them simultaneously. The program may seem to perform tasks concurrently, but only because most steps run asynchronously, while the VM thread is idle, and they may overlap in time. No Java thread is consumed except during the typically brief intervals when Groovy code is actually being run on the VM thread. The executor widget only displays an entry for the “flyweight” executor on the master node when the VM thread is busy; normally it is hidden.
```

So I think the single thread actually doing work can re-enter the re-entrant lock whenever it wants. 

If you did want to avoid the global `lock()` messages, I think using [AtomicInteger](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html#incrementAndGet--) would work in this situation.

sosiouxme:
AtomicInteger would work great if it weren't for the race in removing the directory vs putting stuff in it.

If the issue is the same thread being used... I wonder if a non-reentrant lock would work? Or if that would just deadlock the thread :)

jupierce:
Do we not clean workspaces? If not, maybe something like:
```
AtomicInteger shellCounter = new AtomicInteger()
@NonCPS
def getNextShellCounter() {
    return shellCounter.incrementAndGet()
}

def shell(...) {
    dir (env.WORKSPACE) {
        shellDir = "shell_${currentBuild.buildNumber}"
        sh "mkdir -p ${shellDir}"
        sh "ln -sfn ${shellDir} shell"
        sh "rm -rf `ls --ignore shell --ignore ${shellDir}`"  // remove shell info from other jobs.
        ...
        threadShellIndex = getNextShellCounter()
        def filename = "${shellDir}/sh.${threadShellIndex}."
    }
}
```

sosiouxme:
That would nuke the checked out code/scripts, but yes, I think some cleverness along those lines could work.

sosiouxme:
Updated with new non-locking strategy!

pipeline-scripts/commonlib.groovy
```diff
@@ -236,17 +244,23 @@ def shell(arg) {
...
 
-    // ensure a clean space to save output files
-    if (!shellIndex++) { sh("rm -rf shell") }
-
-    def filename = "shell/sh.${shellIndex}." + truncScript.replaceAll( /[^\w-.]+/ , "_").take(80)
-    sh("mkdir -p ${filename}")
-    filename += "/sh.${shellIndex}"
+    // put shell result files in a subdir specific to this shell call
+    def threadShellIndex = getNextShellCounter()
+    def filename = "shell/sh.${threadShellIndex}." + truncScript.replaceAll( /[^\w-.]+/ , "_").take(80)
```
jupierce, line 249: If `shell(...)` is used inside of a `dir(){}` block, I think it will tend to drop shellDir's around the filesystem. Making these absolute paths with `env.WORKSPACE` might be worthwhile.

pipeline-scripts/commonlib.groovy
```diff
@@ -236,17 +244,23 @@ def shell(arg) {
...
+    filename += "/sh.${threadShellIndex}"
+
+    // concurrency-safe creation of output dir and removal of previous
+    def shellDir = "shell.${currentBuild.number}"
+    sh """#!/bin/bash +x
+        mkdir -p ${shellDir}
+        ln -sfn ${shellDir} shell
+        mkdir -p ${filename}
+        rm -rf shell.foo `ls -d shell.* | sed -e '/^${shellDir}\$/d' `
```
jupierce, line 258: `rm -rf` in my experience is a safe rc=0 operation when no arguments are supplied.

sosiouxme: hahah, i can't believe i didn't try it to see!

pipeline-scripts/commonlib.groovy
sosiouxme: Seemed worthwhile to me. Turned into a bigger project than expected but hopefully the outcome is actually clearer.

PR #2223, opened 2020-05-13
Prevent duplicate upgrade edges

jupierce:
This cause an brief outage in Cincinnati where are duplicate entry
in PREVIOUS caused the graph builder to fail. There is a fix
underway in Cincy, but we should avoid it here for clarity as
well.
Artists should not need to be concerned with whether the release
in flight is already in Cincinnati or not.


jupierce:
@joepvd added logging statement: https://github.com/openshift/aos-cd-jobs/pull/2223/files#diff-cbfda85ed8407cced5ac8810a72001f0R149

PR #2225, opened 2020-05-15
Advisories improvements

joepvd:
Various small improvements, see individual commits for details

**uncomment arch_override**
When prepping a new release, generally the extra arches should get enabled again. This is automation of a manual step. I believe our tooling should get better in understanding when to build extra arches and when not, yet until that time...

**optionally trigger full build**
When experimental arches are enabled again, image builds are expected when there is a parent-child relation between images, and a child image gets built because of automation. This ensures a full image build is started, is not waited on, to help with the process. (Not tested, I expect to need to make changes here.)

**parse out jira text**
I spend some time updating the release jira texts with links to advisories. This blurps a pastable text into the build log.

### DIFF
--- a/jobs/build/advisories/Jenkinsfile
+++ b/jobs/build/advisories/Jenkinsfile
@@ -215,7 +215,7 @@ node {
                         cmd += "sed -e 's/freeze_automation:.*/freeze_automation: no/' -i group.yml ; "
                     }
                     if (params.ENABLE_ARCH_OVERRIDE) {
-                        cmd += "gawk --include inplace '/^#? *arches_override:$/{flag=1;print \$NF;next}; NF<=1{flag=0}; flag {print $(NF-1), \$NF;next}; 1'"
+                        cmd += '''gawk --include inplace '/^#? *arches_override:$/{flag=1;print $NF;next}; NF<=1{flag=0}; flag {print $(NF-1), $NF;next}; 1' '''
                     }
                     cmd += """
                         git diff ;
--- a/jobs/build/advisories/Jenkinsfile
+++ b/jobs/build/advisories/Jenkinsfile
@@ -234,6 +234,27 @@ node {
                         )
                     }
                     echo "out: ${out}"
+
+                    jira_cmd = """
+                    if command -v yq; then
+                      rm -rf ocp-build-data;
+                      git clone --single-branch --branch openshift-${params.VERSION} git@github.com:openshift/ocp-build-data.git;
+                      cd ocp-build-data;
+                      yq -r '
+                        .advisories
+                        | keys_unsorted[] as \$k
+                        | "* *\\(\$k)*: https://errata.devel.redhat.com/advisory/\\(.[\$k])"
+                      ' group.yml;
+                    fi || true
+                    """
+                    try {
+                        commonlib.shell(
+                            returnStdout: true,
+                            script: jira_cmd
+                        )
+                    } catch (err) {
+                        echo "Ignoring error ${err}"
+                    }
                 }
             }
             stage("trigger full build") {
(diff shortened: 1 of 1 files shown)


jobs/build/advisories/Jenkinsfile
```diff
@@ -202,6 +214,9 @@ node {
                     if (params.ENABLE_AUTOMATION) {
                         cmd += "sed -e 's/freeze_automation:.*/freeze_automation: no/' -i group.yml ; "
                     }
+                    if (params.ENABLE_ARCH_OVERRIDE) {
+                        cmd += "gawk --include inplace '/^#? *arches_override:$/{flag=1;print \$NF;next}; NF<=1{flag=0}; flag {print $(NF-1), \$NF;next}; 1'"
```
sosiouxme, line 218: Make sure to test this :)
Embedding code with `$` in a GString has always seemed problematic. Since nothing here is using interpolation, why not use a `'''` string?
```suggestion
                        cmd += '''gawk --include inplace '/^#? *arches_override:$/{flag=1;print $NF;next}; NF<=1{flag=0}; flag {print $(NF-1), $NF;next}; 1' '''
```

jobs/build/advisories/Jenkinsfile
```diff
@@ -219,6 +234,44 @@ node {
                         )
                     }
                     echo "out: ${out}"
+
+                    jira_cmd = """
+                    if command -v yq; then
+                      rm -rf ocp-build-data;
+                      git clone --single-branch --branch openshift-${params.VERSION} git@github.com:openshift/ocp-build-data.git;
+                      cd ocp-build-data;
+                      yq -r '
```
sosiouxme, line 243: I could really use a comment here. What is `yq` and what is it doing here?

jobs/build/advisories/Jenkinsfile
```diff
@@ -219,6 +234,44 @@ node {
                         )
                     }
                     echo "out: ${out}"
+
+                    jira_cmd = """
+                    if command -v yq; then
+                      rm -rf ocp-build-data;
+                      git clone --single-branch --branch openshift-${params.VERSION} git@github.com:openshift/ocp-build-data.git;
```
sosiouxme, line 241: .... why re-clone this?

jobs/build/advisories/Jenkinsfile
```diff
@@ -219,6 +234,44 @@ node {
...
+                    } catch (err) {
+                        echo "Ignoring error ${err}"
+                    }
+                }
+            }
+            stage("trigger full build") {
+                try {
+                    if (params.TRIGGER_FULL_BUILD) {
+                        build(
+                            job: "../aos-cd-builds/build%2F${params.VERSION}",
```
sosiouxme, line 264: This rebuilds RPMs and really ends up doing more than we need for this use case. Plus I'm still leaning toward getting rid of the full build jobs.
Could this just trigger the custom build with images only? And only if there is an `arches_override` section...

vfreex:
I don't think editing arch_overrides in this job is reasonable. The job, literally, is for creating advisories. We create advisories and record them to ocp-build-data, it's fairly reasonable. But doing more stuffs doesn't make much sense to me.

sosiouxme:
We can rename it "begin-release" if that would make more sense as we add more capabilities. The sequence for creating advisories is likely to change as we now can automate everything non-exceptional. I think in the near future we will no longer mark the end of a previous release with new advisories; we might only create advisories at the point where we hand off a release to QE. We could trigger this on a timer to create a release and populate it with artifacts from the most recent successful+accepted build.

That said, I don't know exactly where the re-enable-arch-overrides and full rebuild logic would fit into that. For now, while we're still on the existing cadence, here could make sense.

joepvd:
Most has been included in other ways. Closing stale PR.

PR #2226, opened 2020-05-15
fix flaky jira tool to swtich use self-implement one

shiywang:
I fixed some of the jira-go flaky issue by deprecating old tools and implemnt a new simple one at [here.](https://github.com/openshift/art-tools/pull/1)

the new tool currently manually stored at `~/.local/bin/jirago` on buildvm

Here's the pass log:
https://localhost:8888/job/hack/job/shiywang-aos-cd-jobs/job/build%252Ftarball-sources/127/console


shiywang:
@sosiouxme @joepvd I guess it's a trivial one and it doesn't break another job, I will merge it then to see what how it goes

PR #2227, opened 2020-05-16
[ART-1799] Fix signature directory structure

thiagoalessio:


jobs/signing/sign-artifacts/Jenkinsfile
```diff
@@ -249,18 +249,21 @@ node {
                             mv \$file ${params.SIGNATURE_NAME}
                             mkdir \$file
                             mv ${params.SIGNATURE_NAME} \$file
+                            cp -r \$file release@\$file
+                            cp -r \$file ocp-release@\$file
+                            cp -r \$file ocp-release-nightly@\$file
                             i=1
-                            until ${gsutil} cp -n -v -L cp.log -r \$file gs://openshift-release/${googleStoragePath}/signatures/openshift/release; do
+                            until ${gsutil} cp -n -v -L cp.log -r release@\$file gs://openshift-release/${googleStoragePath}/signatures/openshift; do
```
wking, line 256: Per [the spec][1], isn't the goal to match the pullspec's namespace and repository?  Taking [4.4.4][2] as an example, the signature URI should end with `.../openshift-release-dev/ocp-release@sha256=baa687f29b0ac155d8f4c6914056d36d68f343feb9c1e82b46eef95819d00be5/signature-1` (and similarly for `signature-2`, etc., if we have multiple signatures at some point.  But that would be `ocp-release@sha256=...`, not `release@sha256=...`. 

[1]: https://github.com/containers/image/blob/175bf8b8f9ad897bdc10761e11b466d00f516a63/docs/signature-protocols.md#path-structure
[2]: https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.4.4/release.txt

wking: Ah, I see you have some `ocp-release@...` stuff down below.  What do we push out under `release@`?

thiagoalessio:
@tbielawa should I close this PR, or merge it?

thiagoalessio:
alright, so closing for now, since we've been signing stuff every week and i hadn't heard any complains so far ;D

PR #2239, opened 2020-05-27
[ART-1850] Disable RHCOS 4.1

thiagoalessio:
Related ticket: [ART-1850](https://issues.redhat.com/browse/ART-1850)


sosiouxme:
I'd prefer to rip out 4.1 in one commit https://github.com/openshift/aos-cd-jobs/pull/2243
So that we don't gradually remove bits related to it and have to remember them all later if we need 4.1 back.

thiagoalessio:
Agree! Closed in favor of https://github.com/openshift/aos-cd-jobs/pull/2243

PR #2240, opened 2020-05-27
[ART-1858] verify payload as promotion preflight check

joepvd:
Before promoting a nightly, it needs to be checked that the images in the advisory match the ones in the payload. This uses `elliott verify-payload` as an input validator.

There might be false positives when payload images are in RHSAs. For this reason, the operator is asked for input.

### DIFF
--- a/jobs/build/pre-release/Jenkinsfile
+++ b/jobs/build/pre-release/Jenkinsfile
@@ -132,7 +132,7 @@ node {
             def CLIENT_TYPE = "ocp-dev-preview"
 
             stage("validation") {
-                release.stageValidation(quay_url, dest_release_tag, -1, params.PERMIT_PAYLOAD_OVERWRITE)
+                release.stageValidation(quay_url, dest_release_tag, -1, params.PERMIT_PAYLOAD_OVERWRITE, params.PERMIT_ALL_ADVISORY_STATES, params.FROM_RELEASE_TAG, arch)
             }
 
             stage("build payload") {
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -41,7 +41,7 @@ def destReleaseTag(String releaseName, String arch) {
  *      Valid advisories must be in QE state and have a live ID so we can
  *      include in release metadata the URL where it will be published.
  */
-Map stageValidation(String quay_url, String dest_release_tag, int advisory = 0, boolean permitPayloadOverwrite = false, boolean permitAnyAdvisoryState = false) {
+Map stageValidation(String quay_url, String dest_release_tag, int advisory = 0, boolean permitPayloadOverwrite = false, boolean permitAnyAdvisoryState = false, String nightly, String arch) {
     def retval = [:]
     def version = commonlib.extractMajorMinorVersion(dest_release_tag)
     echo "Verifying payload does not already exist"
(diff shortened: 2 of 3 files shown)


jobs/build/pre-release/Jenkinsfile
```diff
@@ -132,7 +132,7 @@ node {
             def CLIENT_TYPE = "ocp-dev-preview"
 
             stage("validation") {
-                release.stageValidation(quay_url, dest_release_tag, -1, params.PERMIT_PAYLOAD_OVERWRITE)
+                release.stageValidation(quay_url, dest_release_tag, -1, params.PERMIT_PAYLOAD_OVERWRITE, params.PERMIT_ALL_ADVISORY_STATES, params.FROM_RELEASE_TAG, arch)
```
vfreex, line 135: I don't think we actually attach images to advisories for dev previews. @sosiouxme Can you confirm this?

pipeline-scripts/release.groovy
```diff
@@ -116,6 +116,22 @@ Map stageValidation(String quay_url, String dest_release_tag, int advisory = 0,
         error("🚫 Advisory ${advisoryInfo.id} doesn't seem to be associated with a live ID.")
     }
 
+    if (arch == 'amd64' || arch == 'x86_64') {
```
vfreex, line 119: All arches should be validated.

pipeline-scripts/release.groovy
```diff
@@ -116,6 +116,22 @@ Map stageValidation(String quay_url, String dest_release_tag, int advisory = 0,
...
+        echo "Verifying payload"
+        res = commonlib.shell(
+            returnAll: true,
+            script: "elliott --group=openshift-${version} verify-payload registry.svc.ci.openshift.org/ocp/release:${nightly} ${advisory}"
```
Ximinhan, line 123: now each arch has it's own directory path like
registry.svc.ci.openshift.org/ocp-s390x/release-s390x:4.5.0-0.nightly-s390x-2020-09-14-030917
registry.svc.ci.openshift.org/ocp-ppc64le/release-ppc64le:4.5.0-0.nightly-ppc64le-2020-09-14-124036
registry.svc.ci.openshift.org/ocp/release:4.5.0-0.nightly-2020-09-14-124053

jobs/build/pre-release/Jenkinsfile
sosiouxme: You're right, we do not, so you can't validate that part, or most of the rest of the things we might validate during a release... you do have to match the method signature though, I think that's all this does.

pipeline-scripts/release.groovy
sosiouxme: well; yes. note, though, that some images don't build in all arches, and missing entries are replaced in the release image by dummy images (from the `cli` image) so non-x86 are likely to be "missing" some of the contents that are attached to the advisory. `elliott verify-payload` needs to get a little smarter to handle those.

pipeline-scripts/release.groovy
sosiouxme: true but as noted above this only currently works correctly with x86.

pipeline-scripts/release.groovy
```diff
@@ -116,6 +116,22 @@ Map stageValidation(String quay_url, String dest_release_tag, int advisory = 0,
...
+        )
+        if (res.returnStatus != 0) {
+            def cd = currentBuild.description
+            currentBuild.description = "${currentBuild.description} - INPUT REQUIRED"
+            slackChannel = slacklib.to(version)
+            slackChannel.failure("Verify-payload failed. User input required to proceed")
+            input 'Advisory contents does not match payload. Proceed anyway?'
```
sosiouxme, line 130: I'm not sure there's a place for the user to check what didn't match - does that show up in the console logs? Anyway it would help to include the output from `verify-payload` (it creates a json file) in this user prompt.

sosiouxme (commented):
Really the only reason this had to wait was that `verify-payload` didn't work properly with partial payloads at the time. This looks pretty good to me, just one refinement suggested.

sosiouxme (changes requested):
Please add to the `promote` job too

pipeline-scripts/release.groovy
sosiouxme: although i tried it out last time i was promoting stuff and it did not flag anything about the non-x86 arches. maybe you already made it smart enough?

Ximinhan:
also update promote job https://github.com/openshift/aos-cd-jobs/pull/2377

PR #2244, opened 2020-05-28
Add refine-candidate/pkgs-in-rhel

joepvd:
```
pkgs-in-rhel: Script to find packages that are also released in rhel
Usage:
  pkgs-in-rhel <ocp-version> <rhel-version>
Example:
  pkgs-in-rhel 4.5 8
This will:
- Query the latest builds from brew tag rhaos-4.5-rhel-8-candidate
- Look up the package names of these builds
- Find the latest build in rhel-8-build-base
- Output a CSV with the information if the package can be found in rhel
The resulting list can form a basis to look for build pruning from the rhaos tag
```

### DIFF
--- a/hacks/refine-candidate/pkgs-in-rhel
+++ b/hacks/refine-candidate/pkgs-in-rhel
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+# List packages from rhaos that are also in rhel
+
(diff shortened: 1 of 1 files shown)


vfreex:
Considering it is a hack script, I'd like to merge it first.
/lgtm

PR #2249, opened 2020-06-03
ART-1791: signed-compose: Only look at specified RPM advisory

vfreex:
This PR might not provide a full fix of
https://issues.redhat.com/browse/ART-1791, but can reduce
the possibility of tagging 2 builds of the same component
simultaneously.

Instead of all builds on open (not-shipped-yet) advisories,
only look at the RPM advisory recorded in ocp-build-data, or override it
in the build parameter.

Test job run:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu-aos-cd-jobs/job/build%252Fsigned-compose/15/

### DIFF
--- a/jobs/build/signed-compose/Jenkinsfile
+++ b/jobs/build/signed-compose/Jenkinsfile
@@ -22,6 +22,12 @@ node {
                 parameterDefinitions: [
                     commonlib.ocpVersionParam('BUILD_VERSION'),
                     [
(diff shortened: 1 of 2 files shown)


vfreex:
After plashet is landed, we may not need signed compose job anymore.

sosiouxme:
given https://issues.redhat.com/browse/ART-1946 there is no need for signed-compose.

PR #2258, opened 2020-06-15
Delete doozerWorking dir after job finishes

joepvd:


sosiouxme:
Seems fine; might save some space in hack jobs. I think these jobs all clear the workspace anyway when they start?

joepvd:
Yep, they get cleaned away when a job starts. But when multiple, parallel jobs happen to get another executor, the artifacts linger around longer than they should:

du

PR #2265, opened 2020-06-20
rhcos puddle job -> el8 rebuilds job

sosiouxme:
Whenever openshift and openshift-clients packages are built for el7, this job
takes the same content and rebuilds it under an el8 buildroot. RHCOS is the
only consumer of these builds.
    
NOTE: There are no concurrency controls because it is expected to be called                    
from within a job that already blocks concurrency per version.


joepvd:
Have never been aware that this roundabout is necessary and in use :)

I was wondering if there is anything against [this approach](https://github.com/openshift/ocp-build-data/compare/openshift-4.6...joepvd:rhel-8-rpms?expand=1). I am [about to find out](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fcustom/1400/console) :crossed_fingers:

PR #2269, opened 2020-07-02
[ART-1973] Build OLM Bundles for all OCP 4 versions

thiagoalessio:


pipeline-scripts/buildlib.groovy
```diff
@@ -784,16 +784,31 @@ def sync_images(major, minor, mail_list, build_number) {
...
+            results.add build(job: 'build%2Folm_bundle', propagate: false, parameters:
+                [ param('String', 'BUILD_VERSION', fullVersion) ]  // https://stackoverflow.com/a/53735041
+            )
+        }
+    } else {
+        parallel "build-sync": {
+            results.add build(job: 'build%2Fbuild-sync', propagate: false, parameters:
+                [ param('String', 'BUILD_VERSION', fullVersion) ]  // https://stackoverflow.com/a/53735041
+            )
+        }, appregistry: {
```
shiywang, line 803: `appregistry: {`  this by not having quotes intentionally?

pipeline-scripts/buildlib.groovy
```diff
@@ -784,16 +784,31 @@ def sync_images(major, minor, mail_list, build_number) {
...
 
-    parallel "build-sync": {
-        results.add build(job: 'build%2Fbuild-sync', propagate: false, parameters:
-            [ param('String', 'BUILD_VERSION', fullVersion) ]  // https://stackoverflow.com/a/53735041
-        )
-    }, metadata: {
-        results.add build(job: metadataJob, propagate: false, parameters:
-            [ param('String', 'BUILD_VERSION', fullVersion) ]  // https://stackoverflow.com/a/53735041
-        )
+    if (fullVersion == '4.6') {
```
shiywang, line 788: How about 4.6 beyond? like 4.7 4.8..., if it's the same as 4.6 maybe we can let `if` condition to be fullVersion less than 4.6 and `else` to be the above equal 4.6?

shiywang:
others LGTM

pipeline-scripts/buildlib.groovy
thiagoalessio: Hi @shiywang we still don't know what should happen after 4.6, maybe they won't even be built by ART ... that's why I left anything different from 4.6 to just fall in the "else" block. but we have no idea how it will look like.

shiywang: okay, thanks for explaining to me, I guess we can always correct the course if something changed.

PR #2273, opened 2020-07-06
Ensure brewroot contains all rpms before assuming signed

jupierce:
Previously, plashet would just check to see if a directory existed for signed RPMs and assume that the directory structure was fully populated. Now, it will poll to make sure the signed RPMs match the number of unsigned RPMs before returning that the nvr is signed.

### DIFF
--- a/build-scripts/plashet.py
+++ b/build-scripts/plashet.py
@@ -226,7 +226,7 @@ def is_signed(config, nvr):
         all_rpm_count = len(glob.glob(f'{unsigned_base}/**/*.rpm', True))
 
         if all_rpm_count != (signed_rpm_count * 2):
-            logger.info('fFound incomplete signed rpm directory for {nvr}; brewroot may still be being built.')
+            logger.info(f'Found incomplete signed rpm directory for {nvr}; brewroot may still be being built.')
             return False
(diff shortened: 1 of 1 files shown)


build-scripts/plashet.py
```diff
@@ -211,7 +212,25 @@ def is_signed(config, nvr):
...
+        # the unsigned directories. This eliminates a potential race condition between a nvr
+        # being signed and the time it takes to populate the brewroot directories.
+
+        signed_rpm_count = len(glob.glob(f'{signed_base}/**/*.rpm', True))
+        # Note the structure brewroot has signed under the unsigned directory, so the next
+        # glob will find both signed and unsigned.
+        all_rpm_count = len(glob.glob(f'{unsigned_base}/**/*.rpm', True))
+
+        if all_rpm_count != (signed_rpm_count * 2):
+            logger.info('fFound incomplete signed rpm directory for {nvr}; brewroot may still be being built.')
```
sosiouxme, line 229: ITYM
```suggestion
            logger.info(f'Found incomplete signed rpm directory for {nvr}; brewroot may still be being built.')
```

otherwise looks good... nice solution.

PR #2274, opened 2020-07-06
custom / ocp4 - always build a plashet

sosiouxme:
plashets are cheap and quick, i think we're ready to just build them unconditionally.


sosiouxme:
didn't mean to put the branch on origin. moved to https://github.com/openshift/aos-cd-jobs/pull/2275

PR #2275, opened 2020-07-06
custom / ocp4 - always build a plashet

sosiouxme:
plashets are cheap and quick, i think we're ready to just build them unconditionally.

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -31,8 +31,14 @@ node {
                         defaultValue: ""
                     ],
                     [
+                        name: 'COMPOSE',
+                        description: 'Build plashets/compose (always true if building RPMs)',
+                        $class: 'hudson.model.BooleanParameterDefinition',
+                        defaultValue: false
+                    ],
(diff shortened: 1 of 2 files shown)


sosiouxme:
Good point, I'm not sure anybody really knew about `REPO` anyway (I myself had forgotten, though that's a somewhat better approach than ocp4 FORCE if you need a new plashet).

If something new gets tagged in, our scan-sources will try to rebuild whatever uses it... so for normal incrementals, we should always want a new plashet. However there have certainly been times when we were finalizing a release and didn't want to pull in something new... so maybe this should be another flag in the custom job to override if you don't in fact want a new plashet.

sosiouxme:
I added the COMPOSE option to custom, but made it default to false unless you're building RPMs. Plashets take more time now that there are four, and most of the time when testing stuff out we don't need one and don't want to wait. So "RPMs: REPO" => COMPOSE basically.

But it makes sense to me for incremental builds to always create a plashet and pull in whatever has been tagged. Use the custom job if you want finer-grained control.

PR #2277, opened 2020-07-09
Use brew tags where possible

joepvd:
With brew tags, it is better controllable which images are for public consumption, rather than a floating docker tag. This PR:
- seperates data from script
- allows for pullspec to be queried based on a package name and brew tag
- allows for the above logic to be overridden by a docker pullspec


sosiouxme:
I would have pointed at https://issues.redhat.com/browse/ART-2182  as the way forward, at least for `-released` tags.

PR #2279, opened 2020-07-09
update-base-images: enable ubi8 images on aarch64

yselkowitz:


yselkowitz:
@sosiouxme looks like images successfully built.  Ready to merge?

PR #2280, opened 2020-07-10
ART-2004: mirror images targeting ocp priv namespaces on api.ci

vfreex:
build_sync job currently only assumes an arch suffix. It should take all files like src_dest.* / image_stream.*.yaml into consideration.

Also adds `BREW_EVENT_ID` parameter to sync last images of the given moment (requires https://github.com/openshift/doozer/pull/236).

Test run: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Fbuild-sync/8/console

### DIFF
--- a/jobs/build/build-sync/Jenkinsfile
+++ b/jobs/build/build-sync/Jenkinsfile
@@ -42,6 +42,12 @@ node {
                                             defaultValue: "",
                                     ],
                                     [
+                                            name        : 'BREW_EVENT_ID',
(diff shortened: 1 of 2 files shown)


sosiouxme:
Just noticed the part about "(requires openshift/doozer#236)." Oops. But - that will only be a problem if we specify the optional param, so I'll leave it that way until doozer can catch up.

PR #2281, opened 2020-07-10
[ART-2007] Set SCHEDULE_INCREMENTAL default value to false

thiagoalessio:
In order to prevent unintented build triggers.


thiagoalessio:
:+1:

PR #2283, opened 2020-07-14
Refactor advisories

joepvd:
See here for an example run: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Fadvisories/26/console

### DIFF
--- a/jobs/build/advisories/Jenkinsfile
+++ b/jobs/build/advisories/Jenkinsfile
@@ -5,8 +5,9 @@ import java.text.SimpleDateFormat
 node {
     wrap([$class: "BuildUser"]) {
         checkout scm
-        def buildlib = load("pipeline-scripts/buildlib.groovy")
-        def commonlib = buildlib.commonlib
+        def lib = load("advisories.groovy")
+        def buildlib = lib.buildlib
+        def commonlib = lib.commonlib
 
         def dateFormat = new SimpleDateFormat("yyyy-MMM-dd")
         def date = new Date()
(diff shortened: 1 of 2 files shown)


jobs/build/advisories/Jenkinsfile
```diff
@@ -250,56 +177,12 @@ node {
...
-                        out = "DRY RUN mode, command did not run"
-                    } else {
-                        out = buildlib.elliott(cmd, [capture: true])
-                    }
-                    echo "out: ${out}"
+            stage("add placeholder bugs to advisories") {
+                lib.ADVISORIES.each {
+                    if (it.key == "rpm" && params.VERSION.startsWith("3.")) { return }
+                    if (it.key == "image" && params.VERSION.startsWith("4.")) { return }
+                    lib.create_placeholder(it.key)
```
sosiouxme, line 185: actually RHSAs will never need placeholders - by nature they must have bugs. 
we could probably invert this logic a bit:
```suggestion
                    if (it.key == "rpm" && params.VERSION.startsWith("4.")) { lib.create_placeholder(it.key) }
                    if (it.key == "image" && params.VERSION.startsWith("3.")) { lib.create_placeholder(it.key) }
                    if (it.key == "extras") { lib.create_placeholder(it.key) }
```

jobs/build/advisories/advisories.groovy
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    kind = "image"
+    if (name ==~ /rpm/) {
+        kind = "rpm"
+    }
+
+    type = "RHBA"
+    if (name ==~ /rhsa/) {
+        type = "RHSA"
```
sosiouxme, line 23: creating an RHSA currently requires specifying for what, and adds the flaw bug in addition to the tracker (which doesn't happen with an ordinary add), so there's some complexity around it.

i actually think... we are going to end up turning our regular advisories into RHSAs as needed rather than separating out the CVEs. so this expansion may not be necessary.

joepvd:
Thanks! I updated the pull request to only the refactor, without the behavior change. That can be added later more easily when it is clear how are processes will look like.

jobs/build/advisories/advisories.groovy
joepvd: Is the requirement of specifying what for by elliott, or by Errata Tool?

shiywang: current this will not work because specifying bug is needed, @vfreex do you think we need to add an option like `--mode cve` to `find-bugs` then remove this limitation? 
```
raise ElliottFatalError(
"When creating an RHSA, you must provide a list of bug id(s) using one or more `--bug` options.")
```

vfreex: @shiywang I am ok with removing the limitation, but since @sosiouxme proposed we can attach all bugs (no matter it is CVE or not) to an RHBA then convert it to RHSA later, I would like to hold off this change until we decide to do that.

sosiouxme: right, it's an elliott requirement not an ET requirement. but in any case i think we will be re-working elliott a bit regarding CSVs and not need the ready-made RHSAs.

jobs/build/advisories/Jenkinsfile
```diff
@@ -250,56 +175,13 @@ node {
...
-                    } else {
-                        out = buildlib.elliott(cmd, [capture: true])
-                    }
-                    echo "out: ${out}"
+            stage("add placeholder bugs to advisories") {
+                lib.ADVISORIES.each {
+                    if (it.key == "rpm" && params.VERSION.startsWith("3.")) { return }
+                    if (it.key == "image" && params.VERSION.startsWith("4.")) { return }
+                    if (it.key.contains('rhsa') { return }
+                    lib.create_placeholder(it.key)
```
thiagoalessio, line 184:
```
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
WorkflowScript: 184: expecting ')', found 'lib' @ line 184, column 21.
                       lib.create_placeholder(it.key)
```

PR #2284, opened 2020-07-15
ART-2000: fix the custom job to also build an el8 plashet

vfreex:
When the ocp4 job builds composes, it builds plashets for el7 and el8.
The custom job uses a different section of code in buildlib and only builds el7.

Consolidated the code into buildlib and fixed the custom job to also build el8.

Test job runs:
- https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Fcustom/4/console
- https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Focp4/3/

### DIFF
--- a/jobs/build/custom/Jenkinsfile
+++ b/jobs/build/custom/Jenkinsfile
@@ -192,7 +192,8 @@ node {
                         } else {
                             echo 'Building 4.x plashet'
                             // For 4.x, use plashets
-                            buildlib.buildBuildingPlashet(version, release)
+                            buildlib.buildBuildingPlashet(version, release, 7)  // build el7 plashet
+                            buildlib.buildBuildingPlashet(version, release, 8)  // build el8 plashet
                         }
                     }
                 }
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -119,7 +119,9 @@ node {
                 stage("update dist-git") { build.stageUpdateDistgit() }
                 stage("build images") { build.stageBuildImages() }
             }
-            stage("mirror RPMs") { build.stageMirrorRpms() }
+            lock("mirroring-rpms-lock-${params.BUILD_VERSION}") {
+                stage("mirror RPMs") { build.stageMirrorRpms() }
+            }
             stage("sync images") { build.stageSyncImages() }
             stage("sweep") {
                 if (!params.DRY_RUN) {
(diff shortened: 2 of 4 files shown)


jupierce:
@vfreex looks like a great start. Though it doesn't look like the ocp4 test run succeeded. I'm also concerned that the current implementation would break ocp4's build.stageMirrorRpms .

sosiouxme:
ocp4 failed because it calls out to a job that's missing, before it gets to the plashet. so, inconclusive.

vfreex:
@jupierce nice catch! `stageMirrorRpms` uses `rpmMirror.localPlashetPath` and `rpmMirror.plashetDirName` so I should keep them.

I also found the the following issues with stageMirrorRpms.  Should we fix them (in a separate effort)?
- ocp4 job only mirrors el7 plashet. el8 plashet is never mirrored.
- custom job doesn't mirror anything.
- stageMirrorRpms firstly rsyncs the plashet to a shared staging dir before copies to the final destination, but seems it is not protected by a lock.

vfreex:
Running ocp4 hack build https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Focp4/4/console

vfreex:
Seems the hack job passed the plashet creation and rpm mirroring stages. @jupierce @sosiouxme PTAL.

jupierce:
@vfreex - we can open a separate card / PR if desired, but I'd just as soon fix the issues you mention with this work.
- We have not historically mirrored the .el8 plashet, but we should start doing so with 4.6 transitioning to RHEL8 base for everything.
- I'm hoping the custom job will go away soon for 4.x. If no one has complained yet, I'd say we don't worry about mirroring. 
- `firstly rsyncs the plashet to a shared staging dir before copies to the final destination, but seems it is not protected by a lock` . That is a subtle point -- and I hate relying on subtlety :) . It would cost us nothing to put a release specific lock around this so that it can't bite us in the future as we refine jobs.

vfreex:
@jupierce Yeah, custom job doesn't seem required after we always build signed. But let's keep it until we officially deprecate it.

I added a new commit to this PR to wrap the mirroring RPMs stage with a lock.

sosiouxme:
I think the custom job will always have a role, but greatly diminished (not a part of standard release work).

PR #2290, opened 2020-07-21
Record plashet state information in plashet directory

jupierce:
Write operations logs and plashet.yml (metadata about the operation) to the plashet directory. This will allow us to review what happened during assembly and also provide information to the consumers of the plashet.

### DIFF
--- a/build-scripts/plashet.py
+++ b/build-scripts/plashet.py
@@ -12,15 +12,22 @@ import requests
 import ssl
 import time
 import wrapt
+import sys
+import yaml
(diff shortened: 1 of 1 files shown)


vfreex:
Seems a more robust way (to avoid race conditions) is call `koji_api.getLastEvent` first then use that event ID for every `getLatestBuilds` call in the same command invocation.

jupierce:
I like it!

jupierce:
@vfreex updated with the suggestion.

PR #2298, opened 2020-07-31
ART-997: Sweep (or don't sweep) bugs into correct advisory

vfreex:
Requires https://github.com/openshift/elliott/pull/142.

@sosiouxme We are still creating placeholder bugs for extras advisories. Do we need to change the advisoires job or it is ok to keep the placeholder bug?


sosiouxme:
elliott 1.0.8 is deployed.

@vfreex it does not _hurt_ to have the placeholder bug on extras. All the same, it would be best if the advisories job didn't create one, since we'll pretty much always have bugs for changes.

PR #2300, opened 2020-07-31
Allow direction promotion of nightlies

jupierce:
If ART is asked to release a nightly for a specific customer, DIRECT_RELEASE_NIGHTLY should
- rebuild (with potential PREVIOUS) and mirror the nightly
- sign the nightly
- open Cincinnati PRs for the nightly

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -30,7 +30,7 @@ node {
                     ],
                     [
                         name: 'NAME',
-                        description: 'Release name (e.g. 4.1.0)',
+                        description: 'Release name (e.g. 4.1.0); Leave blank for direct nightly.',
                         $class: 'hudson.model.StringParameterDefinition',
                         defaultValue: ""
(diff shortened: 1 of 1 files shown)


jobs/build/release/Jenkinsfile
```diff
@@ -249,14 +304,37 @@ node {
             }
 
             buildlib.registry_quay_dev_login()  // chances are, earlier auth has expired
+
             stage("mirror tools") {
-              retry(3) {
-                release.stagePublishClient(quay_url, dest_release_tag, release_name, arch, CLIENT_TYPE)
-              }
+                if (params.DIRECT_RELEASE_NIGHTLY) {
+                    return
```
sosiouxme, line 310: i think the signing will fail if we don't publish the client.

jupierce: Didn't look like much of a downside to allow the clients to be published for this. Removed short-circuit for the stage.

PR #2311, opened 2020-08-05
ART-1844 - Fix missing fields in json claim signing request

tbielawa:
Adds the specified ".optional.creator" field per:

https://github.com/containers/image/blob/master/docs/containers-signature.5.md#optionalcreator

https://issues.redhat.com/browse/ART-1844


tbielawa:
Merging this. @thiagoalessio be aware when you run the promotion jobs. This should work, but, worst case scenario, you only have to revert it!

PR #2312, opened 2020-08-05
Changing feature candidate to be named

jupierce:


jobs/build/release/Jenkinsfile
```diff
@@ -138,16 +138,22 @@ node {
     direct_release_nightly = false
     detect_previous = true
     candidate_pr_only = false
+    release_offset = params.RELEASE_OFFSET.toInteger()
```
sosiouxme, line 141: `toInteger()` would throw exception when it's empty for hotfix.

jupierce: Nice catch! Updated.

PR #2324, opened 2020-08-13
Check mirroring results on push

jupierce:
Observed during release job and apparently failure to push to subordinate mirrors. It did not raise an error and caused the signing job to fail.

### DIFF
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -361,7 +361,13 @@ done
 
     if ( ! params.DRY_RUN ) {
         commonlib.shell(script: rsync_cmd)
-        commonlib.shell(script: "ssh -o StrictHostKeyChecking='no' ${MIRROR_HOST} timeout 15m /usr/local/bin/push.pub.sh openshift-v4")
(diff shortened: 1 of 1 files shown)


pipeline-scripts/release.groovy
```diff
@@ -361,7 +361,13 @@ done
 
     if ( ! params.DRY_RUN ) {
         commonlib.shell(script: rsync_cmd)
-        commonlib.shell(script: "ssh -o StrictHostKeyChecking='no' ${MIRROR_HOST} timeout 15m /usr/local/bin/push.pub.sh openshift-v4")
+        timeout(time: 60, unit: 'MINUTES') {
```
joepvd, line 364: :thinking: Wouldn't four times retrying with the timeout kept to 15 mins be statistically more likely to yield a positive result?

joepvd: Not sure, but I think there might be a reason the timeout was defined in the process that runs on `use-mirror`. `push.pub.sh` might get reparented to PID1 when jenkins kills the ssh process, and continue to run.

thiagoalessio (approved):
/lgtm

PR #2329, opened 2020-08-19
release: be tolerant of padding in string params

sosiouxme:


joepvd (commented):
Get the impression that `trim: true` can be added to a string parameter definition, which might be a good thing to do for all string parameters in all jobs. See e.g. https://issues.jenkins-ci.org/browse/JENKINS-47115

PR #2335, opened 2020-08-24
Add tekton job `link-pull-request-to-jira`

vfreex:
We can use this job as a start point for exploring Tekton.

This pipeline job is from the demo in the ART knowledge share on July 9, 2020.
Original repo https://github.com/vfreex/example-tekton-pipelines. See https://drive.google.com/file/d/1tfil9bQ10n9a15YzlRbqeEE0pnxeao48/view for the recording.

### DIFF
--- a/tekton-pipelines/README.md
+++ b/tekton-pipelines/README.md
@@ -1,0 +1,216 @@
+# Tekton Demo: Put the link to GitHub Pull Request to JIRA
+
+## Prerequisites
+- A Kubernetes cluster version 1.15+ or OpenShift cluster 3.11+.
+  Use [Minikube][] to runs a single-node Kubernetes cluster in a virtual machine on your personal computer.
+
+- Tekton Pipelines
+    ``` sh
+    # Kubernetes
+    kubectl apply -f https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml
+    # OpenShift
+    oc apply -f https://storage.googleapis.com/tekton-releases/pipeline/latest/release.notags.yaml
+    ```
+    ``` sh
+    # check for installation
+    kubectl get pods --namespace tekton-pipelines
+    kubectl get svc -n tekton-pipelines tekton-dashboard
+    ```
+- Tekton Triggers
+    ``` sh
+    kubectl apply -f https://storage.googleapis.com/tekton-releases/triggers/latest/release.yaml
+    ```
+    ``` sh
+    # check for installation
+    kubectl get pods --namespace tekton-pipelines
+    ```
+- (Optional) Tekton Dashboard
+    ``` sh
+    # Kubernetes
+    kubectl apply -f https://github.com/tektoncd/dashboard/releases/latest/download/tekton-dashboard-release.yaml
+    # OpenShift
+    oc apply -f https://github.com/tektoncd/dashboard/releases/latest/download/openshift-tekton-dashboard-release.yaml
+    ```
+    ``` sh
+    # check for installation
+    kubectl get pods --namespace tekton-pipelines
+    kubectl get svc -n tekton-pipelines tekton-dashboard
+    ```
+- (Optional) Tekton CLI
+    ```sh
+    # Linux
+    curl -LO https://github.com/tektoncd/cli/releases/download/v0.10.0/tkn_0.10.0_Linux_x86_64.tar.gz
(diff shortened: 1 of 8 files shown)


tekton-pipelines/README.md
```diff
@@ -0,0 +1,216 @@
...
+    -H 'X-GitHub-Event: pull_request' \
+    -H 'X-Hub-Signature: sha1=13eaa0168f8d8efcdf5189ea75b782cf89809de6' \
+    -H 'Content-Type: application/json' \
+    -d '{"action": "opened", "head_commit":{"id":"master"},"repository":{"url": "https://github.com/tektoncd/triggers"}}' \
+    http://localhost:8080
+  ```
+  See [Webhook event payloads][] for the format.
+## (Optional) Proxy GitHub webhook behind a secure firewall with smee.io
+- Create a new channel at https://smee.io/.
+- Run `smee` CLI behind your firewall.
```
sosiouxme, line 197: i think we just need the smee client running in a container and handling all our external events generically. i already wanted this for art-bot. listening for github webhooks is an obvious use case. i think we can just provide a configmap for mapping known endpoints to internal targets. if this gets serious enough we could look into [hosting our own smee.io](https://github.com/probot/smee.io#deploying-your-own-smeeio) on say api.ci... if DPTP hasn't already set up a similar relay. smee.io is public and so is this repo, so we'd need to be careful about our traffic since basically anyone could figure out how to watch it and spoof it. we'll need to think that through at some point.

sosiouxme (approved):
This is great to start with; I think once more of us start getting involved we'll want to think through the organization of it a bit more. I didn't initially like dividing everything up by function, would lean toward keeping everything related to a pipeline in one place. On the other hand dividing up by function might contribute to reusability. It's probably too early to think much about it (also maybe there are tools and patterns from others who've already thought it through...)

if any of these comments inspire you to make changes now, feel free, otherwise let's just merge this and iterate.

tekton-pipelines/tasks/jira-set-link-to-pr.yaml
```diff
@@ -0,0 +1,89 @@
...
+    #   value: http://192.168.64.1:11087
+    script: |
+      #!/usr/bin/env python3
+      import os
+      import re
+      import requests
+      from requests.auth import HTTPBasicAuth
+      from urllib.parse import urlparse, quote
+
+      r = requests.get("$(params.GITHUB_API_URL)/repos/$(params.REPO_FULL_NAME)/pulls/$(params.PR_NUMBER)")
```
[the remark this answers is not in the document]
vfreex: Right, the script is put inline for demonstration purpose. I agree we'd better to have a dedicated place for the pipeline scripts, such as having a standalone python project,  and bake them into the image.

Are you suggesting creating that python project from this PR, or in a separate PR? Also I am wondering if https://github.com/openshift/art-ci-toolkit is good place for those scripts or keep using aos-cd-jobs?

tekton-pipelines/README.md
vfreex: Right, smee.io can be watched and spoofed and it doesn't provide any obvious security measures by itself. 

In this demo we are relying on the GitHub interceptor https://github.com/openshift/aos-cd-jobs/pull/2335/files#diff-7e0d17996c28fa4a4296278fb9565a8aR10-R16 to verify if the webhook request is actually sent by GitHub (by checking the signature generated from the secret defined at https://github.com/openshift/aos-cd-jobs/pull/2335/files#diff-b8176e4e814062ec7824d9ab9b698c9aR7). The content is still visible by anyone viewing smee.io but I am not sure if that part should also be protected...

tekton-pipelines/tasks/jira-set-link-to-pr.yaml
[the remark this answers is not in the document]
sosiouxme: actually i think art-ci-toolkit would be better. leave aos-cd-jobs to be mostly jenkins-focused.

tekton-pipelines/README.md
sosiouxme: i don't think we care about the webhook being observable as long as it can't be spoofed. so that seems fine; i wasn't aware of that detail, thanks.

vfreex:
merging this first before making any further improvements :)

PR #2343, opened 2020-09-04
Wrap up automatic firewall enforcement

tbielawa:
New build job 'enforce-firewall':

    Checks if the firewall rules are enforcing. If they are not
    then they will be reapplied. If the rules are reapplied by
    this job then a notification will be sent over slack to the
    #team-art channel.

Simply makes a quick connection to yahoo.com to see if the network is open. Times out after 15 seconds.


thiagoalessio (approved):
/lgtm

PR #2346, opened 2020-09-09
[ART-2059] new sweep flow

joepvd:
The sweep job will grow a new boolean option ATTACH_BUGS which is false by default.

When false, the sweep job will use elliott find-bugs --qe to transition MODIFIED bug state without changing advisory attachment.
When true, the sweep job will use elliott find-bugs to attach only bugs in ON_QA or VERIFIED states to the default advisories.
The current invocation of repair-bugs can be removed (will be handled by the --qe sweep).
When the ocp4 and custom jobs invoke the "sweep" job they will invoke it with ATTACH_BUGS false (so, no change needed).
The sweep job with ATTACH_BUGS true can be invoked manually or (eventually) by the advisories job to fill advisories for a release.

Also, extract the common `DRY_RUN` parameter into a method in `commonlib`. An optional argument with the description is accepted (yet a generic one is used when called without argument).

### DIFF
--- a/jobs/build/sweep/Jenkinsfile
+++ b/jobs/build/sweep/Jenkinsfile
@@ -63,9 +63,10 @@ node {
 
     commonlib.checkMock()
 
-    def version = params.BUILD_VERSION
-    doozerOpts = "--group openshift-${version}"
     stage("Init") {
+        version = params.BUILD_VERSION
+        doozerOpts = "--group openshift-${version}"
+
         echo "Initializing bug sweep for ${version}. Sync: #${currentBuild.number}"
         currentBuild.displayName = "${version} bug sweep"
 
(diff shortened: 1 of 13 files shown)


jobs/build/sweep/Jenkinsfile
```diff
@@ -75,34 +83,53 @@ node {
...
     }
 
-    currentBuild.description = "Repairing state and sweeping new bugs.\n"
     def (major, minor) = commonlib.extractMajorMinorVersionNumbers(version)
-    def kinds = major >= 4 ? ["image", "extras"] : ["rpm"]
 
-    stage("Repair bug state") {
-        currentBuild.description += "* Moving attached bugs in MODIFIED state to ON_QA...\n"
-        for (kind in kinds) {
+    currentBuild.description = "Repairing state and sweeping new bugs.\n"
```
sosiouxme, line 88: There is no separate "repairing state" now so I'd just say:
```suggestion
    currentBuild.description = "Sweeping bugs.<br/>"
```
(Output should be HTML now)

jobs/build/sweep/Jenkinsfile
```diff
@@ -75,34 +83,53 @@ node {
...
+
+    if (params.ATTACH_BUGS) {
+        stage("Attach bugs to advisories") {
```
sosiouxme, line 91: I don't like when stages vary between runs - it makes the stage view drop previous runs with mismatched stages. Could this whole stage be `Sweep bugs` and include both branches, and just update the description according to which branch is taken?

sosiouxme (changes requested):
minor nits, otherwise looks good!

sosiouxme (approved):

PR #2348, opened 2020-09-11
Rename release job to promote job

joepvd:
The word **release** is a tad overloaded. To release a payload can mean
multiple things, some of which this job is concerned with. This is a
pull request to figure out a better name.

What the job does, is:
- to give a payload a name
- to protect the payload from garbage collection
- to extract other artifacts and persist them
- to sign all artifacts
- to kick off processes that publish the artifacts to customers

I think _to promote_ is a good word to summarize these actions.

### DIFF
--- a/jobs/build/promote/Jenkinsfile
+++ b/jobs/build/promote/Jenkinsfile
@@ -8,10 +8,7 @@ node {
     def commonlib = release.commonlib
     def slacklib = commonlib.slacklib
     def quay_url = "quay.io/openshift-release-dev/ocp-release"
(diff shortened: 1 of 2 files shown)


sosiouxme (approved):
Cool :)
We'll keep "release" around until we don't think we're likely to look at it any more, then can just remove.

sosiouxme:
You'll want to point docs at the new job...

sosiouxme:
aw man, i am not too bright. this doesn't incorporate the release changes from the last couple days.

PR #2355, opened 2020-09-15
Look for latest 4.6 builds in rhel-8 tag instead of rhel-7

thiagoalessio:
Since 4.6 started being built on rhel-8, `bundles` stopped finding the real latest-builds, and it is trying to fetch whatever is latest in the `rhel-7` tag.
Not an ideal solution, but we only run this job for 4.6 at the moment; Hacky patch to get bundles working again and gain us some time to think about a proper way to infer the correct brew tag.


sosiouxme:
short term fix. would prefer looking it up from the `group.yml` (which will suit us until we have operators on multiple OSes). but this will do for now.

PR #2359, opened 2020-09-18
[ART-2182] update-base-images: Use actually released images

joepvd:
The `-released` brew tags are deprecated and will soon be removed.

This PR ensures the base images are based off of the formally released images, by using registry.redhat.io as the registry. 

The code change has been run here:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Fupdate-base-images/11/console

### DIFF
--- a/jobs/build/update-base-images/pullspecs.yaml
+++ b/jobs/build/update-base-images/pullspecs.yaml
@@ -16,13 +16,13 @@ jboss.openjdk18.rhel7:
   pullspec: jboss/openjdk18-rhel7:latest
   user: "185"
 rhel7:
-  pullspec: rhel7:7-released
+  pullspec: registry.redhat.io/rhel7:7.8
   user: "0"
 ubi7:
-  pullspec: ubi7:7-released
+  pullspec: registry.redhat.io/ubi7:7.8
   user: "0"
 ubi8:
-  pullspec: ubi8:8-released
+  pullspec: registry.redhat.io/ubi8:8.2
   user: "0"
 rhscl.nodejs.6.rhel7:
   user: "1001"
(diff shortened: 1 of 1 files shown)


jobs/build/update-base-images/pullspecs.yaml
```diff
@@ -16,13 +16,13 @@ jboss.openjdk18.rhel7:
   pullspec: jboss/openjdk18-rhel7:latest
   user: "185"
 rhel7:
-  pullspec: rhel7:7-released
+  pullspec: registry.redhat.io/rhel7:7.8
   user: "0"
 ubi7:
-  pullspec: ubi7:7-released
+  pullspec: registry.redhat.io/ubi7:7.8
```
sosiouxme, line 22: fine for now, but any particular reason you didn't go for `:latest`?
normally that's bad practice but for these base images, the builds seem to increase in version monotonically, i think it would be ok.
https://catalog.redhat.com/software/containers/ubi7/5c6cbf69dd19c77a158f2893?tag=all

sosiouxme (approved):

jobs/build/update-base-images/pullspecs.yaml
joepvd: This was the canonical example from the documentation.

I'd argue that us being intentional about an upgrade is better than us being surprised by a downgrade.

I do not know the support policies around `ubi7` releases, and if it is conceivable if release `minor - 1` could see an emergency fix, which would result in `:latest` pointing at old stuff. But if the chances of the latter are deemed negligable, I am convinced :+1:

sosiouxme: To some degree it doesn't matter - we yum update everything on top of it anyway. Just given that, I think it would be misleading to put a specific version when we'll always update RPMs to latest.

UBI is not RHEL... I don't think there is any interest in maintaining separate streams for minor versions or EUS releases (I see there are freshmaker builds but I'm pretty sure those only bump what's already latest). The older versions are there if someone wants to pin them and move intentionally but I'm pretty sure latest really is always latest.

So I think I do want ubi[78]:latest

PR #2361, opened 2020-09-21
el8-rebuild: Rebuild when default branch is rhel7

joepvd:
Since changing the base to rhel-8 with ocp4.6, rpms do not need to be rebuilt anymore for 4.6 and newer.

This PR checks for the value of `branch` in `group.yml`, and builds it when the default branch ends with `rhel-7` (and not when it ends with `rhel-8`)

4.6: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Fel8-rebuilds/6/console
4.2: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Fel8-rebuilds/5/console

### DIFF
--- a/jobs/build/el8-rebuilds/Jenkinsfile
+++ b/jobs/build/el8-rebuilds/Jenkinsfile
@@ -51,32 +51,35 @@ node {
 
     def version = params.BUILD_VERSION
     currentBuild.displayName += " - ${version}"
-    try {
-        buildlib.kinit()
-        def builds = [
-            "openshift": {
-                stage("openshift RPM") {
-                    commonlib.shell("./rebuild_rpm.sh openshift ${version}")
+    def branch = buildlib.doozer("--group openshift-${version} config:read-group branch", [capture: true]).trim()
+    if (branch.endsWith('-rhel-7')) {
+        try {
+            buildlib.kinit()
+            def builds = [
+                "openshift": {
+                    stage("openshift RPM") {
+                        commonlib.shell("./rebuild_rpm.sh openshift ${version}")
+                    }
+                },
+                "clients": {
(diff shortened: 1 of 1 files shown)


jobs/build/el8-rebuilds/Jenkinsfile
```diff
@@ -51,32 +51,35 @@ node {
...
     def version = params.BUILD_VERSION
     currentBuild.displayName += " - ${version}"
-    try {
-        buildlib.kinit()
-        def builds = [
-            "openshift": {
-                stage("openshift RPM") {
-                    commonlib.shell("./rebuild_rpm.sh openshift ${version}")
+    def branch = buildlib.doozer("--group openshift-${version} config:read-group branch", [capture: true]).trim()
+    if (branch.endsWith('-rhel-7')) {
```
sosiouxme, line 55: what isn't obvious here is that we are in fact still building the two RPMs for rhel7 first and then rebuilding for rhel8. See https://github.com/openshift/ocp-build-data/blob/128e7f81fe8b37af5c0a0b5816597471b9a29adf/rpms/openshift.yml#L18-L20 and https://github.com/openshift/ocp-build-data/blob/128e7f81fe8b37af5c0a0b5816597471b9a29adf/rpms/openshift-clients.yml#L12-L14

We need the RPMs for both since we support RHEL 7 workers. It would seem a little less crazy to build on rhel8 first for 4.6+, or to try to build both at the same time, but the simplest _logic_... is to keep doing what we've been doing.

sosiouxme (changes requested):
Do not want!

PR #2362, opened 2020-09-22
Skip advisory creation during promotion

joepvd:
With the new flow, the creation of advisories is postponed to the day the advisories are prepared for consideration of QE. This PR removes the advisory creation job as part of the promotion.


thiagoalessio (approved):
/lgtm

PR #2367, opened 2020-09-24
Automation move

joepvd:
At `release` time, we do not create advisories anymore. But we do want to enable automation under some circumstances.

This PR removes the automation enablement from the `advisories` job, and moves it to the `release` job.

In the `release` job, the default behavior is to enable automation if the following is all true:
- the architecture of the nightly is x86_64
- it is a Standard or RC release

The default behavior can be overridden by an explicit Yes or No.

This PR also improves the job description somewhat, and allows for the special value of `-1` for `ADVISORIES` to demand less of the outside world for testing purposes (the job won't fail on advisories being in the wrong state).

A test run can be found here: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Frelease/16/console

### DIFF
--- a/jobs/build/release/Jenkinsfile
+++ b/jobs/build/release/Jenkinsfile
@@ -96,13 +96,12 @@ node {
                         description: [
                                 'Unfreeze automation to enable building and sweeping into the new advisories',
                                 '<b>Default</b>: Enable for Standard and RC releases, if arch is x86_64',
-                            ].join('\n'),
+                            ].join('<br/>'),
                         choices: [
                             'Default',
                             'Yes',
                             'No',
                         ].join('\n'),
-                        defaultValue: true
                     ),
                     booleanParam(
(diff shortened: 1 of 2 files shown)


jobs/build/release/Jenkinsfile
```diff
@@ -82,6 +91,19 @@ node {
...
+                        description: [
+                                'Unfreeze automation to enable building and sweeping into the new advisories',
+                                '<b>Default</b>: Enable for Standard and RC releases, if arch is x86_64',
+                            ].join('\n'),
+                        choices: [
+                            'Default',
+                            'Yes',
+                            'No',
+                        ].join('\n'),
+                        defaultValue: true
```
sosiouxme, line 105: `choice` doesn't really have a `defaultValue` does it? anyway it wouldn't be boolean here

sosiouxme (approved):
LGTM

PR #2372, opened 2020-09-30
appregistry: Add advisories to description

joepvd:
Currently, it takes time to find the right jobs from the jenkins search
interface when one wants to look at what happened in the past. This
change adds the advisory IDs in the job's description, so that can be
used to query.


thiagoalessio (approved):
/lgtm

very useful :tada:

PR #2375, opened 2020-10-09
[ART-2032] Use doozer & elliott as submodules

thiagoalessio:
In addition to this PR, the following change had to be made in the [build multi-stage branch configuration](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/configure) in order to automatically fetch submodules when cloning.

build-multistage-config

*NOTE:* In order to play around with it (or even in the future, if this PR gets merged) y'all will need to perform the same change in your respective [hack](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/) space.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -115,7 +115,7 @@ def cleanWhitespace(cmd) {
 // The following commands will run automatically every time one of our jobs
 // loads buildlib (ideally, once per pipeline)
 VENV = "${env.WORKSPACE}/art-venv"
-commonlib.shell(script: "python3 -m venv --symlinks ${VENV}")
+commonlib.shell(script: "python3 -m venv --system-site-packages --symlinks ${VENV}")
 commonlib.shell(script: "${VENV}/bin/pip3 install --upgrade pip")
 commonlib.shell(script: "${VENV}/bin/pip3 install -r art-tools/doozer/requirements.txt")
 commonlib.shell(script: "${VENV}/bin/pip3 install -r art-tools/elliott/requirements.txt")
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -126,7 +126,7 @@ def doozer(cmd, opts=[:]){
             return commonlib.shell(
                     returnStdout: opts.capture ?: false,
                     alwaysArchive: opts.capture ?: false,
-                    script: "doozer --datastore prod --cache-dir /mnt/workspace/jenkins/doozer_cache ${cleanWhitespace(cmd)}")
+                    script: "${VENV}/bin/python3 art-tools/doozer/doozer --datastore prod --cache-dir /mnt/workspace/jenkins/doozer_cache ${cleanWhitespace(cmd)}")
         }
     }
 }
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -119,9 +119,13 @@ DOOZER_BIN = "${VENV}/bin/python3 art-tools/doozer/doozer"
 ELLIOTT_BIN = "${VENV}/bin/python3 art-tools/elliott/elliott"
 
 commonlib.shell(script: "python3 -m venv --system-site-packages --symlinks ${VENV}")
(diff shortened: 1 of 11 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -111,13 +111,22 @@ def cleanWhitespace(cmd) {
     )
 }
 
+// Preparing venv for ART tools (doozer and elliott)
+// The following commands will run automatically every time one of our jobs
+// loads buildlib (ideally, once per pipeline)
+VENV = "${env.WORKSPACE}/art-venv"
+commonlib.shell(script: "python3 -m venv --symlinks ${VENV}")
```
sosiouxme, line 118: I suspect we would want:
```suggestion
commonlib.shell(script: "python3 -m venv --system-site-packages --symlinks ${VENV}")
```
... in order to get the system `certifi`; also that would allow us to install fewer things in the venv.

The downside is that we wouldn't get a newer version of deps unless we specified newer versions in requirements.txt, so we might not test locally with what we get on buildvm. Maybe that just means we need to put explicit versions in requirements? Really we'd have the same problem with venvs anyway, they won't get updates unless we're explicit...

sosiouxme (changes requested):
One concern, the approach LGTM... that was surprisingly easy!

Definitely want to make sure everyone knows this is coming, probably wait for a thursday to do it?

Also I found some references that don't use buildlib.doozer:
```
jobs/build/buildvm-maint/Jenkinsfile:                        sh "doozer --working-dir ${doozer_working} --group sync-3.9 images:push --to-defaults"
jobs/build/buildvm-maint/Jenkinsfile:                        sh "doozer --working-dir ${doozer_working} --group sync-misc images:push --to-defaults"
jobs/build/buildvm-maint/Jenkinsfile:                        sh "doozer --working-dir ${doozer_working} --group sync-3.7 images:push --to-defaults"
jobs/build/build-sync/Jenkinsfile:        sh "which doozer"
pipeline-scripts/release.groovy:    sh "doozer --version"
pipeline-scripts/buildlib.groovy:                    script: "doozer --datastore prod --cache-dir /mnt/workspace/jenkins/doozer_cache ${cleanWhitespace(cmd)}")
```

elliott too:
```
jobs/build/sweep/Jenkinsfile:        sh "which elliott"
jobs/build/promote/Jenkinsfile:                        commonlib.shell(script: "elliott advisory-images -a ${advisory} > ${filename}")
pipeline-scripts/release.groovy:    sh "elliott --version"
pipeline-scripts/release.groovy:                script: "elliott --group=openshift-${version} get --json - -- ${advisory}",
pipeline-scripts/release.groovy:                script: "elliott --group=openshift-${version} get --json - --use-default-advisory image",
```

Guess we could find any remaining landmines faster by just uninstalling doozer and elliott from jenkins user for a while. They're awfully useful to have there though.

pipeline-scripts/buildlib.groovy
sosiouxme: Also... if PyPI is down/blocked... pip3 would fail trying to reach it right? Even if there's nothing to update, is there a way to try to figure that out that doesn't fail? It would be nice to ignore PyPI outages.

Maybe it's as simple as ignoring whether these succeed or not (presumably doozer/elliott will immediately choke if missing deps).

thiagoalessio:
Great, thanks a lot for the review! I guess now all usages of doozer and elliott are covered: https://github.com/openshift/aos-cd-jobs/pull/2375/commits/09d4dab92c5e45a031e1305cfb4a894bfa8c8aec

jupierce:
lgtm - I'd recommend we add a `bump-doozer.sh` and `bump-elliott.sh` as part of this which updates the submodules to align with the remote master. Each is simple enough to do, but a pain to remember. Something like 
`git submodule update --remote --merge <path to submodule>`

sosiouxme (approved):

thiagoalessio:
@jupierce I didn't like the `submodule update ... --merge` option, at least locally, it messed up the submodule tree and it left the submodules in a deatached HEAD state, so I created this script instead: https://github.com/openshift/aos-cd-jobs/pull/2375/commits/f682436098e8468b79ba3a59a6d9f19214b66155

PR #2378, opened 2020-10-15
ART-1906: Use separate signing advisory for each release

vfreex:
Signing advisories will be read from group.yml instead of hardcoded 54765.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -1477,13 +1477,16 @@ def getChanges(yamlData) {
  * @param release  The 4.x release field (usually a timestamp)
  * @param el_major The RHEL major version (7 or 8)
  * @param include_embargoed If true, the plashet will include the very latest rpms (embargoed & unembargoed). Otherwise Plashet will only include unembargoed historical builds of rpms
- * @param auto_signing_advisory The signing advisory to use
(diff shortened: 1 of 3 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -1483,7 +1483,7 @@ def getChanges(yamlData) {
  *
  * }
  */
-def buildBuildingPlashet(version, release, el_major, include_embargoed, auto_signing_advisory=54765) {
+def buildBuildingPlashet(version, release, el_major, include_embargoed, auto_signing_advisory) {
```
sosiouxme, line 1486: if there is no signing advisory in the group.yml `auto_signing_advisory` will be empty and signing won't occur. i'm thinking of 4.2 which still builds... we can either set it to `54765` if not given, or we can throw an error at this point (and update 4.2 `group.yml` to use `54765`). I think I'd prefer continuing to default if not specified.

sosiouxme (changes requested):
minor issue, otherwise LGTM

pipeline-scripts/buildlib.groovy
vfreex: Right, let's continue to use the default signing advisory if not specified.

sosiouxme (approved):
Thanks!

PR #2385, opened 2020-10-26
Prevent creation of OLM advisory for OCP 4.6

thiagoalessio:


jobs/build/advisories/Jenkinsfile
```diff
@@ -116,7 +116,9 @@ node {
                 lib.create_advisory("rpm")
                 if (params.VERSION.startsWith("4")) {
                     lib.create_advisory("extras")
-                    lib.create_advisory("metadata")
+                    if (params.VERSION != "4.6") {
```
joepvd, line 119: We'll run into this with 4.7 again.

joepvd (changes requested):

jobs/build/advisories/Jenkinsfile
vfreex: Right, better to use a condition like `>= 4.6`. See `commonlib.extractMajorMinorVersionNumbers` to convert the version string to numbers.

thiagoalessio:
obsolete by https://github.com/openshift/aos-cd-jobs/commit/67b302888b8a09db5ef373558b79f0a1037ee221

PR #2386, opened 2020-10-26
[ART-2336] Add `FORCE_BUILD` option for OLM bundles

thiagoalessio:
Only makes sense with https://github.com/openshift/doozer/pull/314
(if doozer PR gets merged, need to update submodule before merging this PR)


sosiouxme (approved):
LGTM

PR #2389, opened 2020-10-27
images-health: Tell when broken

joepvd:
We rely on the `images-health` job to know if we are good to ship. This PR announces in slack if the job is broken.


sosiouxme (approved):
Seems good, although it wouldn't have helped today since the problem was it couldn't speak on the channel :)

I added a branching Jira step to add the channel and add art-bot to it.

PR #2391, opened 2020-10-27
Don't need tbielawa in this file now

tbielawa:


thiagoalessio (approved):
/lgtm

thiagoalessio:
i'd rather keep you receiving emails from aos-cd-jobs, so you don't forget your humble origins :rofl:

PR #2403, opened 2020-11-02
images-health: Say if good

joepvd:
Notification spam, or information to celebrate?

I think I'd prefer to see if images health is good, over a silence.


thiagoalessio (approved):
/lgtm

PR #2405, opened 2020-11-03
[ART-1858] verify payload as promotion preflight check

Ximinhan:
Before promoting a nightly, it needs to be checked that the images in the advisory match the ones in the payload. This uses elliott verify-payload as an input validator.

There might be false positives when payload images are in RHSAs. For this reason, the operator is asked for input.


sosiouxme (approved):
There, I fixed it
(I hope)

PR #2414, opened 2020-11-12
sjb/config/test_cases: Bypass AMI validation

mtnbikenc:


mtnbikenc:
https://issues.redhat.com/browse/CORS-1593

PR #2416, opened 2020-11-12
Disable resume on jobs

joepvd:
After a jenkins incident earlier this week, a job attempted to resume,
and did not get further. This blocked scheduled runs of that job. This
change will prevent that behavior, and ensure that jobs don't silently
stop to run.


sosiouxme (approved):
Seems legit. Thanks for fixing some funky formatting too.

PR #2417, opened 2020-11-13
sjb/config/common: Fix built openshift-ansible install

mtnbikenc:
With the updated base image, tito is now generating builds named with '.el7_9'.

Observed [here](https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_extended_builds_311/205/console).

PR #2418, opened 2020-11-13
test_cases: Fix additional openshift-ansible installs

mtnbikenc:
Follow up to #2417


mtnbikenc:
https://issues.redhat.com/browse/CORS-1593
/cc @sdodson

PR #2419, opened 2020-11-17
custom: still push v3 images to reg-aws

sosiouxme:
recent change made the image push part happen only for v4. we want it back for v3 too.


shiywang:
@sosiouxme QE provided me a new token which I will try it out today then I think we can merge this once the new token works.

shiywang:
ref: https://github.com/openshift/aos-cd-jobs/pull/2421

PR #2421, opened 2020-11-17
update username&password on jenkins with a more persistent cred, 

shiywang:
update username&password on jenkins with a more persistent cred, chage registry to registry.stage.redhat.io


sosiouxme:
@shiywang i don't understand why we would login to quay via registry.stage.redhat.io - do you have a conversation you can link to for where that came from? it's the first i'm hearing it.

shiywang:
@anpingli  Could you please explain that?

anpingli:
@shiywang  I don't think we need to the registry.stage.redhat.io when mirror image to quay.io.  
@sosiouxme BTW,  QE can use  brew.registry.redhat.io image directly.  We can stop pushing images to quay.io/openshift-qe-optional-operator now

sosiouxme:
we don't need this then. thanks.

PR #2424, opened 2020-11-18
add option to skip payload creation in promote job

vfreex:


joepvd:
Let's run this branch in a hack space :+1:

PR #2426, opened 2020-11-18
[ART-2353] Clean metadata advisory before attaching builds

thiagoalessio:
To avoid a mix of `.stage` and `.prod` builds in the same advisory.

### DIFF
--- a/jobs/build/appregistry/appregistry.groovy
+++ b/jobs/build/appregistry/appregistry.groovy
@@ -152,6 +152,23 @@ def getMetadataNVRs(operatorNVRs, stream) {
     return doozer("operator-metadata:latest-build --stream ${stream} ${nvrFlags}").split()
 }
 
+def removeDifferentStreamBuilds(advisory) {
+    def attachedBuilds = fetchNVRsFromAdvisories(advisory)
+    def differentStreamBuilds = attachedBuilds.findAll {
+        it -> it.indexOf(".${params.STREAM}") == -1
+    }
+    def elliottCmdBuildFlags = []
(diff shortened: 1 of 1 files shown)


joepvd (changes requested):
If only a single operator needs to get rebuilt and attached, now all others will get removed. Not sure that's the behavior I'd expect.

jobs/build/appregistry/appregistry.groovy
```diff
@@ -348,6 +352,7 @@ def stageAttachMetadata(operatorBuilds) {
         ${params.DRY_RUN ? "--noop" : ""}
     """
 
+    cleanAdvisoryBuilds(params.METADATA_ADVISORY)
```
vfreex, line 355: As @joepvd mentioned, if we only rebuild specific metadatas, this is going to remove all metadata builds. Would like to have a smarter way (e.g. filter by environment name then remove them with elliott)

vfreex (changes requested):

jobs/build/appregistry/appregistry.groovy
vfreex: Or a simple approach: add an option to opt out this behavior if checked.

thiagoalessio:
Thanks @vfreex and @joepvd for the review! I changed the strategy: I list all currently attached builds, and remove _only_ the ones that don't have the current `STREAM` on its NVR.
So, if we are running it with `STREAM=prod`, it would remove all builds that have no `".prod"` on the NVR.

sosiouxme (approved):
nice

PR #2434, opened 2020-11-24
Random refactors

joepvd:


pipeline-scripts/buildlib.groovy
```diff
@@ -126,6 +126,11 @@ def setup_venv() {
 
 }
 
+setup_doozer() {
+    doozerWorking = "${env.WORKSPACE}/doozer_working"
+    env.DOOZER_WORKING_DIR = doozerWorking
```
sosiouxme, line 128: hopefully you tested this... does this env var actually get passed in to shell scripts? I thought you needed `withEnv` (https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#setting-environment-variables).

If the working dir isn't set, doozer will still function, at least until it fills up `/tmp` which won't take long... please make sure a working dir is set :)

jobs/build/build-sync/Jenkinsfile
```diff
@@ -78,7 +78,7 @@ node {
 
     stage("Version dumps") {
         buildlib.doozer "--version"
-        echo "${buildlib.DOOZER_BIN}"
+        sh "which doozer"
```
sosiouxme, line 81: prefer `command -v` see https://stackoverflow.com/a/677212
```suggestion
        sh "command -v doozer"
```

sosiouxme (changes requested):
@joepvd sorry for the delay, these still seem like useful updates, care to rebase? Likely also a good idea to coordinate with Sid as https://github.com/openshift/aos-cd-jobs/pull/2475 seems in the same area.

PR #2437, opened 2020-11-25
Bump doozer

joepvd:


sosiouxme:
Pretty sure this will be covered with https://github.com/openshift/aos-cd-jobs/pull/2444

PR #2439, opened 2020-11-27
load commonlib from buildlib

Ximinhan:
The pull-payload job failed due to can not load commonlib correctly, I thought it caused by missing init step, so use the same method in other script to import.


Ximinhan:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/ximhan-commonlib/5/console

PR #2440, opened 2020-11-27
load commonlib from function

Ximinhan:
The pull-payload job failed due to can not load commonlib correctly, move these part into a function.


Ximinhan:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/ximhan-commonlib/5/console

PR #2445, opened 2020-11-30
olm_bundle: make bundle_nvrs global

sosiouxme:
[When in use today](https://coreos.slack.com/archives/GV3UMNACV/p1606721562360500?thread_ts=1606719582.358700&cid=GV3UMNACV) it appears that defining this var in one stage and using it in the next doesn't work. Hopefully this makes it global? I haven't had a chance to test yet.


thiagoalessio (approved):
/lgtm

thiagoalessio:
I thought that by simply omitting `def`, `bundle_nvrs` would be considered global ... apparently not :disappointed:

PR #2446, opened 2020-12-01
promote: only allow skipping individual validations

sosiouxme:
We do need the result of the validation (really, the advisory info) even if one or more validation must be skipped. Instead of allowing skipping the whole stage, prompt individually and allow retrying but not skipping the whole thing.

I was trying for a minimal change since this is so hard to test.

### DIFF
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -126,19 +126,18 @@ Map stageValidation(String quay_url, String dest_release_tag, int advisory = 0, 
         error("🚫 Advisory ${advisoryInfo.id} doesn't seem to be associated with a live ID.")
     }
 
+    slackChannel = slacklib.to(version)
     if (arch == 'amd64' || arch == 'x86_64') {
         echo "Verifying payload"
         res = commonlib.shell(
                 returnAll: true,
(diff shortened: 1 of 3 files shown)


pipeline-scripts/release.groovy
```diff
@@ -152,8 +152,12 @@ Map stageValidation(String quay_url, String dest_release_tag, int advisory = 0,
             script: "${buildlib.ELLIOTT_BIN} --group=openshift-${version} verify-attached-bugs ${advisoryInfo.id}",
         )
         if(res.returnStatus != 0) {
+            def cd = currentBuild.description
+            currentBuild.description = "${currentBuild.description} - INPUT REQUIRED"
+            slacklib.to(version).failure("elliott verify-attached-bugs failed. User input required to proceed")
             def htmlLines = res.stdout.replaceAll('\n', '<br/>\n')
-            error("🚫 Bug verification failed with the following output:<br>\n${htmlLines}")
+            input "Bug verification failed with the following output; proceed anyway?<br>\n${htmlLines}"
```
jupierce, line 159: Consider using commonlib.inputRequired?

sosiouxme: good suggestion, thanks, did that.

PR #2448, opened 2020-12-03
Fix plashet 3.11 creation

joepvd:
Fixes:
```
hudson.remoting.ProxyException: groovy.lang.MissingPropertyException: No such property: version for class: WorkflowScript
```


thiagoalessio (approved):
/lgtm

it won't fix plashets fully, but these changes will get rid of the current error.

joepvd:
As the plashet PR has been reverted because of upstream changes needed as well, I'll close this one.

PR #2450, opened 2020-12-03
WIP [ART-2048] Workspace cleanup before and after build

thegreyd:
### Summary
This change is to add pre/post build workspace cleanup steps, using workspace cleanup plugin. It was suggested in the story to use it at the top, but after reading, the step only takes the build state in account when it's run after the build. There's a question for preBuildCleanup which I'm not sure we can use in our scenario.

### Links
- [Jira link](https://issues.redhat.com/browse/ART-2048)
- [Workspace cleanup step](https://plugins.jenkins.io/ws-cleanup/) and [params](https://www.jenkins.io/doc/pipeline/steps/ws-cleanup/) - documentation is pretty sparse 

### Todos
- [ ] [preBuildCleanup](https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.wrapper.WrapperContext.preBuildCleanup) - this step looks like it's part of Job DSL plugin, I'm not sure if we do have that or can use the step outside of it.
- [ ] There are several places where we use `buildlib.cleanWorkdir(workDir)` usually when initializing, cleaning up doozer, or at the end of the job. Usage varies across jobs. Would have to see if they can be replaced with this
- [ ] Run and test the job
- [ ] Get a list of (active?) jobs we would like to see updated with these steps


sosiouxme:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Focp4/28/console is a test run... `No such DSL method 'preBuildCleanup' found among steps` - is it only valid for declarative pipeline? Would we need an added plugin?

PR #2457, opened 2020-12-09
promote: skipVerifyBugs if next is not release

joepvd:
Currently, bugs are verified at Promote time if they have been VERIFIED
in the next minor stream. The reason for this is to ensure customers do
not regress when upgrading their clusters. For the newest minor release,
this leads to unnecessary prompts during the promote job.

This proposal uses the ocp4ReleaseState from commonlib to determine if
any architecture in the next minor version has been marked as "release".
This is done at the moment where we are close enough to shipping to
build and use signed rpms.

The conjecture of this patch is that that moment resembles when we ought
to pay attention to what has been verified in the next release to ensure
customers do not regress for a bug when upgrading their clusters.

### DIFF
--- a/jobs/build/promote/Jenkinsfile
+++ b/jobs/build/promote/Jenkinsfile
@@ -170,6 +170,7 @@ node {
     detect_previous = true
     candidate_pr_only = false
     is_4stable_release = true
+    next_is_prerelease = false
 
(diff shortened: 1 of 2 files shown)


sosiouxme:
i think that is a good way to distinguish released vs non-released, yes.

sosiouxme (approved):

PR #2461, opened 2020-12-11
ocp3: remove a lot of cruft around build mode

sosiouxme:
when OCP 3 was the latest development, we had a bunch of logic around when to merge from master versus a release branch, on top of managing an update stream for OpenShift Online. that's all irrelevant now. master is OCP 4, fixes need to be in release branches.

this request removes a lot of that irrelevant logic, hopefully making it easier to figure out what is going on.


joepvd (commented):
-470 lines ✨ 
Looks good, think https://github.com/openshift/aos-cd-jobs/blob/0df44035d371b0d3dc5a8a65d8b4aff60427b328/scheduled-jobs/build/ocp-3.11/Jenkinsfile#L16 can now also disappear.

sosiouxme:
good point @joepvd thanks! merging now.

PR #2463, opened 2020-12-14
[ART-2422][ART-2530] Notify 'stage' and 'prod' appregistry runs

thiagoalessio:


jobs/build/appregistry/Jenkinsfile
```diff
@@ -145,10 +146,36 @@ node {
                     appregistry.stageAttachMetadata(operatorBuilds)
                 }
             }
+            stage("slack notification to release channel") {
+                if (params.STREAM != 'stage' || params.STREAM != 'prod') {
```
sosiouxme, line 150: this will always skip the notif. how about:
```suggestion
                if (params.STREAM == 'dev') {
```

PR #2465, opened 2020-12-14
[ART-2422][ART-2533] Send Slack notification when tarball-sources runs

thiagoalessio:


jobs/build/tarball-sources/Jenkinsfile
```diff
@@ -142,9 +143,24 @@ node {
...
+                *:heavy_check_mark: tarball-sources sent to CLOUDDST*
+                CLOUDDST JIRA: ${jiraCardURL}
+                buildvm job: ${commonlib.buildURL('console')}
+                """)
+            }
         }
 
     } catch (err) {
+        slacklib.to(version).say("""
+        *:heavy_exclamation_mark: tarball-sources failed*
```
sosiouxme, line 160: seems likely to get lost in the noise if we don't call out... we do want some action to be taken here
```suggestion
        *:heavy_exclamation_mark: @release-artists tarball-sources failed*
```

sosiouxme (approved):

PR #2467, opened 2020-12-15
[ART-2460] update eus-4.6 channel in cincinnati PR

Ximinhan:
For 4.6 releases we need to update both stable-4.6 and eus-4.6 when create cincinnati PRs, and I think when comes to 4.7 will still have stable-4.7 but eus-4.x will hold until eus-4.10 created.

### DIFF
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -579,6 +579,10 @@ def signArtifacts(Map signingParams) {
     )
 }
 
+def isSupportEUS(ocpVersion) {
+  return ocpVersion in commonlib.eusVersions
+}
+
 /**
  * Opens a series of PRs against the cincinnati-graph-data GitHub repository.
  *    Specifically, a PR for each channel prefix (e.g. candidate, fast, stable) associated with the specified release
@@ -616,7 +620,7 @@ def openCincinnatiPRs(releaseName, advisory, candidate_only = false,ghorg = 'ope
             if ( major == 4 && minor == 1 ) {
                 prefixes = [ "prerelease", "stable"]
             }
-            if ( major == 4 && minor >= 6 ) {
+            if ( isSupportEUS(commonlib.extractMajorMinorVersion) ) {
                 prefixes = [ "candidate", "fast", "stable", "eus"]
             }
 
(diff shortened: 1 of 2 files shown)


pipeline-scripts/release.groovy
```diff
@@ -616,6 +616,9 @@ def openCincinnatiPRs(releaseName, advisory, candidate_only = false,ghorg = 'ope
             if ( major == 4 && minor == 1 ) {
                 prefixes = [ "prerelease", "stable"]
             }
+            if ( major == 4 && minor >= 6 ) {
```
joepvd, line 619: From what I've been able to gather, we need to have EUS channels for 4.6 (and later, 4.10), but not for 4.7, 4.8, ....

thiagoalessio: may I suggest that we create something like `eusVersions = [...]` or a function `def isEUS(...)` in buildlib or commonlib, and use that to make comparisons instead?

Ximinhan: What I understand is that currently only for 4.6 update eus channel with the same content we updated in stable channel.

pipeline-scripts/release.groovy
```diff
@@ -706,6 +711,19 @@ def openCincinnatiPRs(releaseName, advisory, candidate_only = false,ghorg = 'ope
...
 
+                            break
+                        case 'eus':
+                            // currently put eus change in a seperate PR as same as stable channel
+                            labelArgs = "-l 'do-not-merge/hold'"
+                            pr_messages << "Please merge within 48 hours of ${internal_errata_url} shipping live OR a Cincinnati-first release."
+
+                            if (prURLs.containsKey('prerelease')) {
+                                pr_messages << "This should provide adequate soak time for prerelease channel PR ${prURLs.prerelease}"
+                            }
```
sosiouxme, line 722: prerelease channel was only for 4.1 ... this can be omitted as it will never be relevant.

pipeline-scripts/release.groovy
sosiouxme: Yes. I think a `commonlib.eusVersions` next to `commonlib.ocp4Versions` would be perfect.

sosiouxme (changes requested):

sosiouxme (approved):
thanks LGTM

PR #2472, opened 2020-12-17
[ART-1188] enable publish rpms job

Ximinhan:


sosiouxme (approved):
as you can see, I added a commit with what I consider improvements.

* no more use of `sudo` (root should not be involved)
* using `rsync` instead of `rm` and `scp`
* some minor cleanup

PR #2475, opened 2020-12-21
ART-2048 Workspace cleanup

thegreyd:
### Summary
Change to add pre/post build workspace cleanup steps, using workspace cleanup plugin. It was suggested in the story to use it at the top, but after reading, the step only takes the build state in account when it's run after the build. Also delete tmp workspace folders.

### Links
- [Jira link](https://issues.redhat.com/browse/ART-2048)
- [Workspace cleanup step](https://plugins.jenkins.io/ws-cleanup/) and [params](https://www.jenkins.io/doc/pipeline/steps/ws-cleanup/)
- [Sample run of ocp4 job with the commonlib function](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Focp4/11/console)

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -1174,6 +1174,19 @@ def getGroupBranch(doozerOpts) {
     return branch
 }
 
+def cleanWorkspace() {
+    cleanWs(cleanWhenFailure: false, notFailBuild: true)
+    dir("${workspace}@tmp") {
+        deleteDir()
+    }
+    dir("${workspace}@script") {
+        deleteDir()
+    }
+    dir("${workspace}@libs") {
+        deleteDir()
+    }
+}
+
 WORKDIR_COUNTER=0 // ensure workdir cleanup can be invoked multiple times per job
 def cleanWorkdir(workdir, synchronous=false) {
     // get a fresh workdir; removing the old one can be synchronous or background.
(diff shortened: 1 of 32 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -1174,6 +1174,19 @@ def getGroupBranch(doozerOpts) {
     return branch
 }
 
+def cleanWorkspace() {
+    cleanWs(cleanWhenFailure: false, notFailBuild: true)
+    dir("${workspace}@tmp") {
```
vfreex, line 1179: Sometimes the dir is named as `@<number>` or `@<number>@tmp`, would like to use a wildcard to delete them as well.

vfreex (changes requested):
Thanks. I have a request to delete `@<number>` and `@<number>@tmp` directories as well.

pipeline-scripts/buildlib.groovy
thegreyd: Yeah, so as I understand after reading [here](https://issues.jenkins.io/browse/JENKINS-30231), these are additional workspaces created when concurrent builds are running on the same node. In that case I would assume the workspace dirs (like job@2) get deleted with cleanWs statement, and other dirs like job@2@tmp get deleted with subsequent statements. But that is what I'm assuming, it would be nice if I can actually test that..

thegreyd: hey @vfreex do you think we can merge this? And then we can see if the issue persist?

vfreex (approved):
Thank you. Let's try it out!

PR #2476, opened 2020-12-23
plashet from-tags: Add option to inherit

joepvd:
For 3.11, we depend on brew inheritance to build the OSE repositories.
The plashet implementation does not descend into the inheritance tree.

This commit adds the boolean option `--inherit` to `plashet from-tags`
to look into children tags as well. Defaults to False.

Test runs:
- [custom 3.11](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Fcustom/2/console): includes `--inherit`, resulting [plashet](http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/plashets/3.11/3.11.348-1/x86_64/os/Packages/) includes the needed packages
- [custom 4.2](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Fcustom/3/console) does not incl;ude `--inherit` in the plashet invocation.

### DIFF
--- a/build-scripts/plashet.py
+++ b/build-scripts/plashet.py
@@ -459,7 +459,9 @@ class KojiWrapper(wrapt.ObjectProxy):
 @click.option('--event', required=False, default=None, help='The brew event for the desired tag states')
 @click.option('--include-embargoed', default=False, is_flag=True,
               help='If specified, embargoed/unshipped RPMs will be included in the plashet')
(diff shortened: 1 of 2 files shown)


thiagoalessio (approved):
/lgtm

sosiouxme:
yeah, makes good sense. we stopped using inheritance in 4.x so didn't think about it.

i don't particularly want our automation picking up new inherited things and signing them (in fact i'm not sure ET would allow that) but i don't think any inherited tags are getting new content so hopefully that won't be an issue.

PR #2484, opened 2021-01-06
appregistry/ocp4_scan/olm_bundle: move load to buildvm2

sosiouxme:


sosiouxme:
updated to kinit before running

thiagoalessio (approved):
/lgtm

PR #2490, opened 2021-01-12
Don't untag atomic-openshift when ocp3 fails

vfreex:
This behavior is inconsistent with ocp4 build, and causes mixed rpm versions in plashet.


joepvd:
+1 for removing this logic, it has caught me by surprise as well.

PR #2503, opened 2021-01-14
pipeline-scripts: enable ARM and Z in 3.11 plashets

yselkowitz:
These are used to maintain internal resources.


thiagoalessio (approved):
/lgtm

didn't know we built 3.11 for these arches

yselkowitz:
ART does not but we rely upon to support internal resources.

PR #2504, opened 2021-01-14
use updated doozer with mysql backend

sosiouxme:
incorporate https://github.com/openshift/doozer/pull/247 - write build results to mariadb instead of simpledb.


sosiouxme:
BTW @adarshtri requested that we push results to the dev DB until he can do some art-dash updates to use the prod DB properly. Don't expect art-dash contents to see updates right away. (I hope I understood that correctly.)

PR #2508, opened 2021-01-20
parallel the download process

Ximinhan:
an other way is to use brew watch-logs to grep installing log instead of download all the log files.


thiagoalessio (approved):
/lgtm

PR #2512, opened 2021-01-21
ART-2389: sync opm to mirror

vfreex:
Extract and mirror opm binaries to mirror, store their sha256 digests to sha256sum.txt as well to get signed.

Tested with following hack jobs:

promote:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Fpromote/7/

oc_sync:
- 4.7.0-fc.3 x86_64: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Foc_sync/10/
- 4.6.12 s390x: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Foc_sync/9/
- 4.6.12 x86_64: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Foc_sync/8/

### DIFF
--- a/jobs/build/oc_sync/publish-clients-from-payload.sh
+++ b/jobs/build/oc_sync/publish-clients-from-payload.sh
@@ -1,12 +1,22 @@
 #!/bin/bash
-set -eux
+set -euo pipefail
 export MOBY_DISABLE_PIGZ=true
+
+if [ $# -lt 4 ]; then
+    echo "Usage: $0 WORKSPACE VERSION CLIENT_TYPE PULL_SPEC" >&2
+    exit 1
+fi
+
 WORKSPACE=$1
 VERSION=$2
 CLIENT_TYPE=$3
 PULL_SPEC=$4
 
-ARCH=$(skopeo inspect docker://${PULL_SPEC} -config | jq .architecture -r)
+
+GOTRACEBACK=all oc version --client
+
+ARCH=$(skopeo inspect docker://${PULL_SPEC} --config | jq .architecture -r)
+
 if [[ "${ARCH}" == "amd64" ]]; then
     ARCH="x86_64"
 fi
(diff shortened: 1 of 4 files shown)


joepvd (approved):
/lgtm

jobs/build/oc_sync/publish-clients-from-payload.sh
```diff
@@ -1,12 +1,22 @@
 #!/bin/bash
```
joepvd, line 1: As the `promote` job copies quite some of the logic in this file, would it make sense to move it to `pipeline-scripts`, and have the `promote` job invoke that as well?

sosiouxme: i seem to recall the contexts were _just different enough_ to make it not worthwhile (we barely use `oc_sync` at all)

jobs/build/oc_sync/publish-clients-from-payload.sh
```diff
@@ -24,44 +34,99 @@ else
...
-    #  - openshift-install-mac-4.3.0-0.nightly-s390x-2020-01-06-081137.tar.gz
-    #  ...
-    # So, match, and store in a group, any non-digit up to the point we find -DIGIT. Ignore everything else
-    # until we match (and store in a group) one of the valid file extensions.
-    if [[ "$f" =~ ^([^0-9]+)-[0-9].*(tar.gz|tgz|bz|zip)$ ]]; then
-        # Create a symlink like openshift-client-linux.tgz => openshift-client-linux-4.3.0-0.nightly-2019-12-06-161135.tar.gz
-        ln -sfn "$f" "${BASH_REMATCH[1]}.${BASH_REMATCH[2]}"
+    MAJOR=$(echo "$VERSION" | cut -d . -f 1)
+    MINOR=$(echo "$VERSION" | cut -d . -f 2)
+    if [ "$MAJOR" -eq 4 ] && [ "$MINOR" -le 6 ]; then
```
sosiouxme, line 85: I do not believe we want to publish this for releases before 4.6. can you please exclude them?

pipeline-scripts/release.groovy
```diff
@@ -397,11 +396,56 @@ for f in *.tar.gz *.bz *.zip *.tgz ; do
...
+    BINARIES=(opm)
+    PLATFORMS=(linux)
+    if [ "$ARCH" == "x86_64" ]; then  # For x86_64, we have binaries for macOS and Windows
+        BINARIES+=(darwin-amd64-opm windows-amd64-opm)
+        PLATFORMS+=(mac windows)
+    fi
+
+    MAJOR=$(echo "$VERSION" | cut -d . -f 1)
+    MINOR=$(echo "$VERSION" | cut -d . -f 2)
+    if [ "$MAJOR" -eq 4 ] && [ "$MINOR" -le 6 ]; then
```
sosiouxme, line 416: same here, obviously

sosiouxme (changes requested):
@vfreex other than this, seems ready to go... could merge perhaps before the next 4.7 fc?

jobs/build/oc_sync/publish-clients-from-payload.sh
vfreex: I hoped we can use a consolidated logic to do this, but the promote job has its own oc_sync implementation rather than calling the oc_sync job. Don't really know why we have different implementation so I won't think it is a good idea to refactor in this PR.

jobs/build/oc_sync/publish-clients-from-payload.sh
vfreex: ack. will update it in the next push.

joepvd (approved):
/lgtm

sosiouxme (approved):

PR #2519, opened 2021-01-26
ART-2445 Automate checks for Blocker Bugs

thegreyd:
- https://issues.redhat.com/browse/ART-2445

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -152,7 +152,10 @@ class PrepareReleasePipeline:
             f"--target_release {target_release_str}",
         ]
         _LOGGER.debug("Running command: %s", cmd)
-        subprocess.run(cmd, check=True, universal_newlines=True)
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -38,8 +38,11 @@ def __init__(
         self.working_dir = Path(working_dir).absolute()
         self.dry_run = dry_run
         self.release_version = tuple(map(int, self.release_name.split(".", 2)))
+        self.target_release = (
```
vfreex, line 41: `target_release` is a concept only valid to bugzilla. would like to keep this a local variable rather than a class field since the name is confusing.

vfreex (commented):
This looks good-ish, although I would prefer to implement this using the API rather than the command line invocation. If there's any authentication issue, in my memory it just prints nothing rather print out an error.

thegreyd:
Moving this to elliott! New PR https://github.com/openshift/elliott/pull/177

PR #2522, opened 2021-02-03
Fix a jira clone failure if description is null

vfreex:
An error occurs when cloning 3.11 jira subtasks:

```
{'description': 'Operation value must be a string'}
```
This is because some template issues have empty descriptions, which become `null`
when retrieving from the API. But the create issue API doesn't allow null for the description.
Let's just convert it to empty string.


vfreex:
Since this PR is quick and obvious, I am merging it. Let me know if you have any concerns.

PR #2520, opened 2021-01-28
[ART-2459] Add operator-sdk_sync job

thiagoalessio:
New sync job for `operator-sdk` client.

Screenshot from 2021-01-28 13-45-26

```
use-mirror-upload:/srv/pub/openshift-v4
$ ls -lsh x86_64/clients/operator-sdk/
total 0
0 drwxr-xr-x. 2 jenkins_aos_cd_bot jenkins_aos_cd_bot 25 Jan 28 07:48 4.7.0
0 lrwxrwxrwx. 1 jenkins_aos_cd_bot jenkins_aos_cd_bot  5 Jan 28 07:49 latest -> 4.7.0

use-mirror-upload:/srv/pub/openshift-v4
$ ls -lsh x86_64/clients/operator-sdk/4.7.0
total 65M
65M -rwxr-x--x. 1 jenkins_aos_cd_bot jenkins_aos_cd_bot 65M Jan 28 07:49 operator-sdk
```
```
use-mirror-upload:/srv/pub/openshift-v4
$ ls -lsh s390x/clients/operator-sdk/
total 0
0 drwxr-xr-x. 2 jenkins_aos_cd_bot jenkins_aos_cd_bot 25 Jan 28 07:49 4.7.0
0 lrwxrwxrwx. 1 jenkins_aos_cd_bot jenkins_aos_cd_bot  5 Jan 28 07:49 latest -> 4.7.0

use-mirror-upload:/srv/pub/openshift-v4
$ ls -lsh s390x/clients/operator-sdk/4.7.0
total 64M
64M -rwxr-x--x. 1 jenkins_aos_cd_bot jenkins_aos_cd_bot 64M Jan 28 07:49 operator-sdk
```
```
use-mirror-upload:/srv/pub/openshift-v4
$ ls -lsh ppc64le/clients/operator-sdk/
total 0
0 drwxr-xr-x. 2 jenkins_aos_cd_bot jenkins_aos_cd_bot 25 Jan 28 07:49 4.7.0
0 lrwxrwxrwx. 1 jenkins_aos_cd_bot jenkins_aos_cd_bot  5 Jan 28 07:49 latest -> 4.7.0

use-mirror-upload:/srv/pub/openshift-v4
$ ls -lsh ppc64le/clients/operator-sdk/4.7.0
total 62M
62M -rwxr-x--x. 1 jenkins_aos_cd_bot jenkins_aos_cd_bot 62M Jan 28 07:49 operator-sdk
```

PR #2524, opened 2021-02-09
[ART-2459] Update operator-sdk_sync job

thiagoalessio:
Follow up of https://github.com/openshift/aos-cd-jobs/pull/2520

Extracts `operator-sdk` version to name the tarballs, and creates the following structure:

```
$ for arch in x86_64 s390x ppc64le; do ls /srv/pub/openshift-v4/${arch}/clients/operator-sdk/4.7.0-rc.0 ; done
operator-sdk-v1.3.0-ocp-linux-x86_64.tar.gz
operator-sdk-v1.3.0-ocp-linux-s390x.tar.gz
operator-sdk-v1.3.0-ocp-linux-ppc64le.tar.gz
```

Open question to @jmrodri: It is still unclear to me from where we should extract the `darwin` binaries.

### DIFF
--- a/jobs/build/operator-sdk_sync/Jenkinsfile
+++ b/jobs/build/operator-sdk_sync/Jenkinsfile
@@ -19,20 +19,25 @@ pipeline {
 
     parameters {
         string(
-            name: 'BUILD_TAG',
-            description: "Build of ose-operator-sdk from which the contents should be extracted.<br/>" +
-                         "Examples:<br/>" +
-                         "<ul>" +
-                         "<li>v4.7.0-202101261648.p0</li>" +
-                         "<li>v4.7.0</li>" +
-                         "<li>v4.7</li>" +
-                         "</ul>",
+            name: 'OCP_VERSION',
+            description: 'Under which directory to place the binaries.<br/>' +
+                         'Examples:<br/>' +
+                         '<ul>' +
+                         '<li>4.7.0</li>' +
+                         '<li>4.7.0-rc.0</li>' +
+                         '</ul>',
             defaultValue: '',
             trim: true,
         )
         string(
-            name: 'VERSION',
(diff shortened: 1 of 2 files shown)


jmrodri:
@thiagoalessio maybe I missed a step in setting things up for darwin. Beause we're not doing anything specific downstream to build darwin, I think I assumed that the same process that builds images for the other arches would also build for darwin. For now I say we skip darwin and fix that up later.

jmrodri (approved):
/lgtm

jobs/build/operator-sdk_sync/Jenkinsfile
```diff
@@ -42,39 +47,62 @@ pipeline {
...
                 script {
-                    def manifestList = readJSON(
-                        text: sh(
-                            script: "skopeo inspect --raw docker://${imagePath}:${params.BUILD_TAG}",
-                            returnStdout: true,
-                        )
-                    ).manifests
                     manifestList.each {
-                        sh "rm -rf ./${params.VERSION} && mkdir ./${params.VERSION}"
+                        def sdkArch = it.platform.architecture == 'amd64' ? 'x86_64' : it.platform.architecture
```
yselkowitz, line 71: This won't work when aarch64 (arm64 in go-speak) is enabled.

sosiouxme: Good to know, we have that translation in multiple places :-/ We'll probably want a centralized go-2-RH translation method.

sosiouxme (approved):

jobs/build/operator-sdk_sync/Jenkinsfile
rcbaus: I am looking at this @sosiouxme where should this go-2-RH method live? thinking groovy or python? Looking to help out here.

PR #2526, opened 2021-02-10
Promote job - Upgrade test message fixes

thegreyd:
## Summary
- Do not print "upgrade test" slack message for non-x64 arches
- Print sorted list of upgrade versions from most recent ones
- Add comment to not use "mirror" variant with upgrade tests since it's broken possibly - https://coreos.slack.com/archives/CBN38N3MW/p1612989185114600?thread_ts=1612969025.078300&cid=CBN38N3MW

## Test run
[hack job url]


thiagoalessio (approved):
/lgtm

PR #2529, opened 2021-02-16
[ART-2683] Add job 'build/kam_sync'

thiagoalessio:
Sync binaries to <https://mirror.openshift.com/pub/openshift-v4/clients/kam>

### DIFF
--- a/jobs/build/kam_sync/Jenkinsfile
+++ b/jobs/build/kam_sync/Jenkinsfile
@@ -1,0 +1,70 @@
+#!/usr/bin/env groovy
+node {
+    checkout scm
+    load("pipeline-scripts/commonlib.groovy").describeJob("kam_sync", """
+        -----------------------------------
+        Sync OpenShift kam client to mirror
+        -----------------------------------
+        http://mirror.openshift.com/pub/openshift-v4/clients/kam/
+
+        Timing: This is only ever run by humans, upon request.
+    """)
+}
+
+pipeline {
+    agent any
+    options { disableResume() }
+
+    parameters {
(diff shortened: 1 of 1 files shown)


jobs/build/kam_sync/Jenkinsfile
```diff
@@ -0,0 +1,70 @@
...
+    parameters {
+        string(
+            name: "VERSION",
+            description: "Desired version name. Example: v1.0.0",
+            defaultValue: "",
+            trim: true,
+        )
+        string(
+            name: "SOURCES_LOCATION",
+            description: "Example: http://download.eng.bos.redhat.com/staging-cds/developer/openshift-gitops-kam/1.0.0-78/signed/all/",
```
joepvd, line 27: If the artifacts are on the url, they are also on nfs: `tree /mnt/redhat/staging-cds/developer/openshift-gitops-kam/1.0.0-78/signed/all/`. This is made use of in https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/helm_sync/Jenkinsfile.

joepvd (approved):
/lgtm.
Left comment for potential change, but I'd be good with this implementation as well.

jobs/build/kam_sync/Jenkinsfile
thiagoalessio: oh, good to know! i guess i'll just keep the URL for now and observe how the next requests will look like (if they do like `odo`, every request requires changes to the job, but if `kam` people place consistent requests i'll surely prefer to use nfs)

PR #2540, opened 2021-03-04
ART-2731 tkn_sync Create signed binaries.

adarshtri:


sosiouxme:
We don't need `BREW_BUILD_ID` any longer, and the example for the version ought to include the `-release` as well

sosiouxme:
LGTM
trial: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Ftkn_sync/6/

PR #2542, opened 2021-03-09
Bug 1930378: operator-sdk_sync: Extract darwin artifact

joepvd:
Darwin binaries are in the x86_64 image, extract those. And do some rearranging to make that part easy.

Test run from before enabling the syncing:
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Foperator-sdk_sync/15/console


thiagoalessio (approved):
/lgtm

PR #2548, opened 2021-03-17
Remove Release pipeline

joepvd:
The Release job has been superseded by the Promote job. With this
commit, the Release job gets removed from the pipeline.


thiagoalessio (approved):
/lgtm

vfreex:
This LGTM. @sosiouxme Do you think it is ok to remove this job?

joepvd:
The proper timing is when 4.5 is EOL. Closing.

sosiouxme:
I would rather not remove it at this time. The logs/artifacts for this job in Jenkins are a record of how things got into customer hands (ridiculously old though they may be now). It looks like the last release we used it on was 4.5; I think once we release 4.8 and 4.5 is EOL, that is the right time to get rid of this.
/hold

PR #2549, opened 2021-03-18
rhcos_sync: Add installer symlinks

joepvd:
Some customer portals continue to link to `rhcos-installer`, see e.g. https://cloud.redhat.com/openshift/install/power/user-provisioned.

This PR ensures that all symlinks that start with `rhcos-live-*` have a counterpart `rhcos-installer-*` that points to the same file.

Also, change tabs to spaces to standardize.


sosiouxme (approved):
/LGTM

PR #2556, opened 2021-03-25
ART-2596: Add fastly ip ranges to firewall

vfreex:
Fetches Fastly IP ranges from the API and adds it to the firewall.

When this PR is merged, https://github.com/openshift/aos-cd-jobs/pull/2556 can be merged to revert https://github.com/openshift/aos-cd-jobs/commit/2431925d9f2bcf851516ad9d80c7009e0c3b7647.

### DIFF
--- a/hacks/iptables/buildvm-scripts/canttouchthat.py
+++ b/hacks/iptables/buildvm-scripts/canttouchthat.py
@@ -1,7 +1,4 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
-
-from __future__ import print_function
+#!/usr/bin/env python3
 
 import click
 import subprocess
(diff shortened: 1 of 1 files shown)


hacks/iptables/buildvm-scripts/canttouchthat.py
```diff
@@ -216,6 +223,11 @@ def main(output_networks, enforce, dry_run, clean):
         print('Would have written the following direct rules to {}'.format(DIRECT_XML_PATH))
         print_direct_rules(direct)
 
+def get_fastly_ip_ranges():
+    response = requests.get("https://api.fastly.com/public-ip-list")
```
sosiouxme, line 227: are we guaranteed that the firewall allows `api.fastly.com`?

Also, probably wise to allow some retries.

But we can watch and see if any refinement is needed.

PR #2558, opened 2021-03-25
[ART-2747] add release mark for security

Ximinhan:
add release mark for security


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -350,6 +350,10 @@ def create_release_jira(self, advisories: Dict[str, int]):
 
         def fields_transform(fields):
             labels = set(fields.get("labels", []))
+            # change summary title for security
+            if fields.get("summary"):
+                if "product security" in fields["summary"]:
+                    fields["summary"] = f"{self.release_name} [{self.release_date}][Wed] Notify product security to review"
```
vfreex, line 356: I don't think overriding the summary is a good idea. Better to just prepend the release name and date.

vfreex (changes requested):

PR #2559, opened 2021-03-25
Blocker bug check retry 

thegreyd:
This enhances the blocker bug check in the promote job. Currently there is no option to retry check, once blocker bugs are found, which results in aborting the release or proceeding anyway. This uses the retryAbort method which we use for example, in checking for stable release in RC, to provide similar functionality.

todo: untested code. Do a test run.

### DIFF
--- a/jobs/build/promote/Jenkinsfile
+++ b/jobs/build/promote/Jenkinsfile
@@ -212,7 +212,7 @@ node {
             group = "openshift-${major}.${minor}"
             
             commonlib.retryAbort("Waiting for Blocker Bugs to be resolved", taskThread, 
-                                    "Blocker Bugs found for release. Do not proceed until confirmation from ARTists/bug owners that it is okay to proceed. If the bugs are not blockers, coordinate to remove the blocker+ flag on the bug(s). RETRY when blocker(s) is finally resolved.") {
+                                    "Blocker Bugs found for release; do not proceed without resolving. See https://github.com/openshift/art-docs/blob/master/4.y.z-stream.md#handling-blocker-bugs") {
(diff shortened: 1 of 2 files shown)


jobs/build/promote/Jenkinsfile
```diff
@@ -210,45 +210,10 @@ node {
...
-                    )
-                    if (!resp) {
-                        currentBuild.result = 'ABORTED'
-                        error('Aborting because release blocker bugs found.')
-                    }
-                }
-            } else {
-                echo "No blocker bugs found. Proceeding"
+            commonlib.retryAbort("Waiting for Blocker Bugs to be resolved", taskThread, 
+                                    "Blocker Bugs found for release. Do not proceed until confirmation from ARTists/bug owners that it is okay to proceed. If the bugs are not blockers, coordinate to remove the blocker+ flag on the bug(s). RETRY when blocker(s) is finally resolved.") {
```
sosiouxme, line 215: Rather than putting a lot of detail here I'd rather just refer to the doc.
```suggestion
                                    "Blocker Bugs found for release; do not proceed without resolving. See https://github.com/openshift/art-docs/blob/master/4.y.z-stream.md#handling-blocker-bugs") {
```

sosiouxme:
just saw "todo: untested code. Do a test run." sorry. i tested that it compiles. harder to test properly :) but if it fails next time, at least it fails early.

sosiouxme:
although my [boring test run](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Fpromote/11/console) looks ok in this area.

PR #2561, opened 2021-03-30
olm_bundles - Validate advisories and group

thegreyd:
Idiot proof olm_bundles job so an ARTist knows when he messed up! for [example](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Folm_bundle/4159/parameters/)

Make sure group(4.7) is substring of advisory title version(4.7.4), and both advisory version are exact matches.

Test runs:
- fail case 1 group not found in advisory: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Folm_bundle/6/console
- fail case 2 advisory minor version differ: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Folm_bundle/7/console
- success: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Folm_bundle/9/console

### DIFF
--- a/jobs/build/olm_bundle/Jenkinsfile
+++ b/jobs/build/olm_bundle/Jenkinsfile
@@ -91,6 +91,14 @@ pipeline {
                 }
             }
         }
+        stage('Make sure advisories belong to a single group') {
+            when {
+                expression { ! params.EXTRAS_ADVISORY.isEmpty() }
+            }
+            steps {
+                olm_bundles.validate_advisories(params.EXTRAS_ADVISORY, params.METADATA_ADVISORY, params.BUILD_VERSION)
+            }
+        }
         stage('Get advisory builds') {
             when {
                 expression { ! params.EXTRAS_ADVISORY.isEmpty() }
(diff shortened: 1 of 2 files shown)


joepvd:
/lgtm

Thanks for building in this safeguard!

Spent a bit of time to look if there is a way of determining the version of a given advisory, as that seems potentially brittle. I learnt that an advisory does not seem to have a concept of target release.

One scenario in which this could go wrong, is when advisories are recycled for `4.y.(n+1)`, and not all advisory titles have been updated yet. 

One way of approaching this more structurally, would be to look at the attached placeholder bug and compare their `target_version`. To get the placeholder bug, some string comparisons need to happen anyway. This would complicate things, and I am not sure if it is worth the effort. Flagging when not all advisories have the same version could be called a feature :)

I am good with shipping this nominal improvement.

PR #2573, opened 2021-04-08
Promote: Run `attach-cve-flaws`

joepvd:
Currently, before promotion, the release manager needs to ensure that
flaw bugs are attached to advisories, and advisories are changed to
RHSAs if needed. If the operator omits this step, the released artifacts
will contain broken links.

This change ensures that, at the time of promotion, the advisories are
changed in kind, foregoing potential operator omissions.

With this change, a leak is introduced. Where *Promote* is only about
the payload, it now operates also on the other advisories. My impression
is that that is acceptable.

### DIFF
--- a/jobs/build/promote/Jenkinsfile
+++ b/jobs/build/promote/Jenkinsfile
@@ -175,6 +175,8 @@ node {
     release_offset = params.RELEASE_OFFSET?params.RELEASE_OFFSET.toInteger():0
     def (major, minor) = commonlib.extractMajorMinorVersionNumbers(params.FROM_RELEASE_TAG)
 
+    group = "openshift-${major}.${minor}"
+
     if (params.RELEASE_TYPE.startsWith('1.')) { // Standard X.Y.Z release
         release_name = "${major}.${minor}.${release_offset}"
         ga_release = true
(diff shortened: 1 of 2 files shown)


jobs/build/promote/Jenkinsfile
```diff
@@ -333,6 +333,20 @@ node {
                     errata_url = retval.errataUrl
                 }
             }
+            stage("add cve flaw bugs") {
+                if (params.DRY_RUN || advisory == -1) {
+                    return
+                }
+                if (arch != 'x86_64') {
+                    return
+                }
```
vfreex, line 342: This will require us to promote x86_64 first then other arches, right?
I guess we can remove this check since it will be no-ops if flaw bugs are already attached.

jobs/build/promote/Jenkinsfile
```diff
@@ -333,6 +333,20 @@ node {
...
+                if (major == 4 && !is_4stable_release) {
+                    return
+                }
+                ["rpm", "image", "extras"].each {
+                    script: "${buildlib.ELLIOTT_BIN} -g ${group} attach-cve-flaws --use-default-advisory ${it}"
+                }
```
vfreex, line 348: Sometimes we don't have rpm or extras advisories for a release. I'll prefer to only attach flaw bugs to the main image advisory, or ignore the errors if those advisories don't exist or got closed.

vfreex (changes requested):
I love this idea. Just some minor issues.

jobs/build/promote/Jenkinsfile
sosiouxme: this is a good point because actually the only problem here is with the payload advisory; its URL gets written into the payload image, so we need it to be correct at that time. the other ones can change without any problem.

i think the goal is to automate all this, so i like running it on all (BTW we could see bugs attached to the metadata advisory at some point also). but to yuxiang's point some advisories might not be shipping, so i'd probably wrap each in a retry/skip to query the human if it doesn't work.

sosiouxme: ... which is almost what you updated it with, great :) i think i would do the `retrySkipAbort` on each individual advisory though (or perhaps run in parallel and wrap that?)

jobs/build/promote/Jenkinsfile
```diff
@@ -333,6 +333,19 @@ node {
...
                 }
             }
+            stage("add cve flaw bugs") {
+                if (advisory == -1) {
+                    return
+                }
+                if (major == 4 && !is_4stable_release) {
+                    return
+                }
+                release.getAdvisoryIds().each {
```
thegreyd, line 343: Trying to understand this step, does getAdvisoriesIds() return ["image","rpm","extras"]? Kind of a nitpick but if we're reading group.yml in a previous step, might as well pass in advisories explicitly rather than ask elliott to read again?

joepvd: It gets the advisory IDs. See https://github.com/openshift/aos-cd-jobs/pull/2573/files#diff-d2bd219940ec77d5275acd297f229e62357751e91311290812987988a687c9d2R237.

joepvd:
Tested it out, and now it does what says on the tin: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/jdelft-aos-cd-jobs/job/build%252Fpromote/7/console

PR #2574, opened 2021-04-08
[ART-2816] Use `doozer beta:rpms:build` to build rpms

vfreex:
https://github.com/openshift/doozer/pull/382 provides a new implementaion of rpm build without using tito.

We'd like to test it with a full product release.

Merge this when it is ready to test.


sosiouxme (approved):
doozer PR has merged. obviously this PR needs to be paired with a submodule bump. this seems fairly low-risk and could really go anytime, but...

can we test this usage out with a few forced builds in 4.8 and ask the requesters to make sure this fits their need and doesn't break existing assumptions, before we roll it out in general?

PR #2578, opened 2021-04-12
Temporarily decommission buildvm2

joepvd:


thiagoalessio (approved):
/lgtm

at least to prepare bundles for this week, we can revert this later, once buildvm2 is back.

PR #2581, opened 2021-04-14
Resume from mirror artifacts

thegreyd:
This was hacky way to accomplish what I wanted to - which was running remaining steps for 3 failed promote jobs, which failed at the mirror binaries stage.

But the general idea is to have all the different stages in a dropdown, to resume from. Implementation wise I'm thinking of storing the stages in an ordered list/map of some kind and then setting previous stage skip variables in an automated fashion. Ideas/feedback welcome

### DIFF
--- a/jobs/build/promote/Jenkinsfile
+++ b/jobs/build/promote/Jenkinsfile
@@ -116,7 +116,7 @@ node {
                         description: 'Select stage to resume from. Useful to execute remaining steps in the case of a failed promote job.',
                         choices: [
                                 'Not Applicable',
-                                'Mirror binaries',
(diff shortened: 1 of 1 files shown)


thegreyd:
@sosiouxme I think I got a verbal lgtm from you, can you put it here :)

sosiouxme (commented):
otherwise LGTM

joepvd:
Bike shedding alert:
I think I will not be the only one who will get confused by these very similarly named variables. Could you find them a prefix, so that also in speaking it is clear what somebody means, and make them lowercase or CamelCase, to make them visually different?

sosiouxme (approved):

PR #2591, opened 2021-04-23
[ART-2807] Ask build type when ocp4 is manually triggered

thiagoalessio:


jobs/build/ocp4/build.groovy
```diff
@@ -87,6 +88,37 @@ def initialize() {
     return planBuilds()
 }
 
+def setBuildType() {
+    if (!isTriggeredByOrganicLifeForms()) {
```
sosiouxme, line 92: :rofl:

jobs/build/ocp4/build.groovy
```diff
@@ -87,6 +88,37 @@ def initialize() {
...
+        echo "Assuming *PROD* build, triggered by automation"
+        scratch = false
+        return
+    }
+    askBuildType()
```
sosiouxme, line 97: should not do this if hooman checked the `SCRATCH` option; only "are you sure" on prod builds.

jobs/build/ocp4/build.groovy
```diff
@@ -87,6 +88,37 @@ def initialize() {
...
+        )
+        scratch = res == 'TEST'
+    }
+}
+
+@NonCPS
+def isTriggeredByOrganicLifeForms() {
+    currentBuild.rawBuild.getCauses().collect {
+        it.getClass().getCanonicalName().tokenize('.').last()
+    }.contains('UserIdCause')
```
sosiouxme, line 119: buildlib.isBuildPermitted just checks that there's a real user that kicked off the request. i don't know which is the "best" way but it seems like they ought to work the same...

sosiouxme (changes requested):
what about the `custom` job? maybe logic can be extracted into buildlib.

ugh, `ocp4` still has no `README`. `custom` does though.

thiagoalessio:
> what about the `custom` job? maybe logic can be extracted into buildlib.
done:
- extracted logic to buildlib: https://github.com/openshift/aos-cd-jobs/pull/2591/commits/30f95056f97b8a702ff9289d3329cc27c31863d0
- use it on custom job too: https://github.com/openshift/aos-cd-jobs/pull/2591/commits/478bc81c0b928e9370a2796f2f24a2e9c8ed7adc

thiagoalessio:
> ugh, ocp4 still has no README. custom does though.

no problem, writing one now ...

pipeline-scripts/buildlib.groovy
```diff
@@ -1564,6 +1564,39 @@ def get_owners(doozerOpts, images, rpms=[]) {
     return yamlData
 }
 
+def setBuildType() {
+    if (!isTriggeredByOrganicLifeForms()) {
+        echo "Assuming *PROD* build, triggered by automation"
+        scratch = false
```
vfreex, line 1570: I am concerned about the use of global variable `scratch` here. People have to remember to define this global variable when calling from a Jenkinsfile. I would suggest returning this value instead.

Also the function name `setBuildType` sounds confusing. Would prefer a name more specific like `getBuildPurpose` `shouldSkipTag` or something.

pipeline-scripts/buildlib.groovy
```diff
@@ -1564,6 +1564,39 @@ def get_owners(doozerOpts, images, rpms=[]) {
...
+        return
+    }
+    if (!scratch) {
+        askBuildType()
+    }
+}
+
+def askBuildType() {
```
vfreex, line 1578: Same here, would prefer a name more specific like `askBuildPurpose` `askSkipTag` or something.

vfreex (changes requested):
Some minor issues otherwise LGTM.

joepvd:
I think this approach has become obsolete with the concept of assemblies: https://issues.redhat.com/projects/ART/issues/ART-2924. Can we close this, @thiagoalessio?

thiagoalessio:
yeah, obsolete @joepvd ! closing it now

PR #2599, opened 2021-04-26
[ART-2752] add alert when all images build failed in ocp4

Ximinhan:
if all images build failed then send an alert to sclack and skip the following steps
such error messages like:
```
:warning: All of 4 image builds failed in ocp4 job
openshift-enterprise-pod:https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=36484036
ironic:https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=36484064
ose-installer:https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=36484070
ose-installer-artifacts:n/a
```

### DIFF
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -148,15 +148,38 @@ node {
                 }
 
                 stage("update dist-git") { build.stageUpdateDistgit() }
-                stage("build images") { build.stageBuildImages() }
+                stage("build images") {
+                    build.stageBuildImages()
+                    branch=build.version.branch
+                    imagelist=build.buildPlan.imagesIncluded.split(',')
+                    for (i = 0; i < imagelist.size(); i++) {
+                        out = commonlib.shell(
+                            returnStdout: true,
+                            script: "brew list-tagged --latest-n=2 --quiet ${branch} | grep ${imagelist[i]} | grep .p0 | wc -l"
+                        )
+                        builds = commonlib.shell(
+                            returnStdout: true,
+                            script: "brew list-tagged --latest-n=2 --quiet ${branch}| grep ${imagelist[i]} | cut -d ' ' -f 1"
+                        )
+                        if (out.contains("1")){
+                            slacklib.to(commonlib.extractMajorMinorVersion(params.BUILD_VERSION)).say("""
+                                *:alert:(test message) we may catch a newly-embargoed build ${imagelist[i]} in ocp4 build*
+the two recents builds are:
+${builds}
+                            """)
+                        }
+                    }
+                }
             }
             lock("mirroring-rpms-lock-${params.BUILD_VERSION}") {
                 stage("mirror RPMs") { build.stageMirrorRpms() }
             }
-            stage("sync images") { build.stageSyncImages() }
-            stage("push qe quay images") { build.stagePushQEImages() }
-            stage("sweep") {
-                buildlib.sweep(params.BUILD_VERSION)
+            if (!buildlib.allImagebuildfailed){
+                stage("sync images") { build.stageSyncImages() }
+                stage("push qe quay images") { build.stagePushQEImages() }
+                stage("sweep") {
+                    buildlib.sweep(params.BUILD_VERSION)
+                }
             }
         }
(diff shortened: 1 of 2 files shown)


jobs/build/ocp4/Jenkinsfile
```diff
@@ -148,15 +148,38 @@ node {
                 }
 
                 stage("update dist-git") { build.stageUpdateDistgit() }
-                stage("build images") { build.stageBuildImages() }
+                stage("build images") {
+                    build.stageBuildImages()
```
sosiouxme, line 152: The "alert on embargoes" part here should really be a separate commit (maybe a separate PR) from the "deal with image failures" sections later.

Also could the new logic go inside `stageBuildImages`? I prefer to keep this Jenkinsfile high-level.

jobs/build/ocp4/Jenkinsfile
```diff
@@ -148,15 +148,38 @@ node {
...
+                    branch=build.version.branch
+                    imagelist=build.buildPlan.imagesIncluded.split(',')
+                    for (i = 0; i < imagelist.size(); i++) {
+                        out = commonlib.shell(
+                            returnStdout: true,
+                            script: "brew list-tagged --latest-n=2 --quiet ${branch} | grep ${imagelist[i]} | grep .p0 | wc -l"
```
sosiouxme, line 158: a lot of problems in this one line.

in general, i do not like to see bash doing this much work unless it is in a bash script. Here you have groovy to do the work for you, and it has much more precise tools for processing output. so does python if you prefer.

note that this invokes brew to list _all_ components in the tag, once for every image (and then again below). this is not a quick command. it should be called _once_, and processed into a lookup hash, then the loop can iterate against that hash. the hash might look like:

```
  [
    "foo-container": ["v4.8.0-202104270102.p0"],  // there might only be one ever built
    "bar-container": ["v4.8.0-202104270102.p0", "v4.8.0-202104270304.p1"], // here's one that just went private
    "baz-container": ["v4.8.0-202104270102.p1", "v4.8.0-202104270304.p0"], // here's one that just went public
  ...]
```
(note all three of these would have `wc -l` of 1, but only one is what we're looking for.)
showing groovy notation of course - but invoking a small python script could be good too (then you could also choose to make koji api calls rather than invoke brew).

finally, `grep ${imagelist[i]}` could be a very inaccurate match for two reasons. first, the distgit name does not necessarily match the brew component at all. second, it could match more than one (for example `foo` and `foo-operator`).

instead, invoke doozer to get a list of the components you are looking for:
`doozer -i ${imagesIncluded} images:print --short '{component}'`
then you can lookup each in the hash built above.

jobs/build/ocp4/build.groovy
```diff
@@ -347,6 +348,12 @@ def stageBuildImages() {
         def r = buildlib.determine_build_failure_ratio(recordLog)
         if (r.total > 10 && r.ratio > 0.25 || r.total > 1 && r.failed == r.total) {
             echo "${r.failed} of ${r.total} image builds failed; probably not the owners' fault, will not spam"
+            if (r.failed == r.total) {
```
sosiouxme, line 351: there's one case when this won't catch a total failure; when there's only one image building.

so i'd move this outside the `if` above to catch that case too.

jobs/build/ocp4/build.groovy
```diff
@@ -347,6 +348,12 @@ def stageBuildImages() {
...
+                allImagebuildfailed = true
+                commonlib.slacklib.to(params.BUILD_VERSION).say("""
+                    *:warning:(test-run) all of ${r.total} image builds failed in ocp4 job*
```
sosiouxme, line 354: obviously, remove `test-run` before merging :)
also could this link to the job?

sosiouxme (changes requested):

jobs/build/ocp4/Jenkinsfile
Ximinhan: I got it, I will split this part in another PR and think of a better way to collect data.

vfreex (approved):
/lgtm
but i didn't test this.

thiagoalessio (approved):
/lgtm

PR #2610, opened 2021-05-06
WIP: Enable aarch64

yselkowitz:
Until now, x86_64/amd64 was the only enabled architecture with differing 
nomenclatures.  However, aarch64/arm64 has the same issue, which 
accounts for most of the complications in adding it.

Please check this carefully for mistakes and missing pieces.  There are 
some things that probably need to be in place first before this can be 
merged.

/cc @sosiouxme


sosiouxme:
Thanks, I had not realized how many places we had this sort of thing.
I think we can centralize this a little bit instead of repeating the same idea everywhere :)

yselkowitz:
Please also double-check based on what lands in https://github.com/openshift/release/pull/18595

sosiouxme:
following up in https://github.com/openshift/aos-cd-jobs/pull/2620

PR #2615, opened 2021-05-10
[ART-1517] Ensure filename length does not exceed max

thiagoalessio:


pipeline-scripts/commonlib.groovy
```diff
@@ -310,7 +310,12 @@ Console Log: ${buildURL('console')}
         to = args.to ?: args.cc ?: args.bcc ?: "NOBODY"
         to = to.replaceAll(/[^@.\w]+/, "_")
         subject = args.get("subject", "NO SUBJECT").replaceAll(/\W+/, "_")
-        filename = String.format("email/email%03d-%s-%s.txt", ++emailIndex, to, subject)
+
+        def filename_ext = ".txt"
+        def max_filename_len = 255 - filename_ext.size()
```
sosiouxme, line 315: wouldn't this still be too long when you add the extension back on? perhaps:
```suggestion
        def max_filename_len = 251 - filename_ext.size()
```

sosiouxme (changes requested):

PR #2618, opened 2021-05-13
[ART-2626] add slack alert for embargoed build

Ximinhan:
use doozer detect-embargo command to check whether latest build become embargoed


jobs/build/ocp4/build.groovy
```diff
@@ -366,6 +366,19 @@ def stageBuildImages() {
 
         buildlib.doozer "${doozerOpts} images:streams mirror"
     }
+    // check new embargoed build
+    out = commonlib.shell(
+        returnAll: true,
+        script: "${env.WORKSPACE}/build-scripts/querybuilds.py -v ${version.stream} -t ${version.branch} -i ${buildPlan.imagesIncluded}"
```
vfreex, line 372: In stead of inventing a new script, could you have a look at `doozer detect embargo` and see if it applicable?

vfreex (changes requested):

jobs/build/ocp4/build.groovy
Ximinhan: thanks for the info, looks like doozer detect-embargo don't support images list as args ? and for this task I think it just need a fast and light method to do a simple check in ocp4 job.

vfreex: The problem is that `.p1` in the release field only means "the source of the image is embargoed". An image should also be considered embargoed if it contains embargoed rpms (".p1" or tagged with `-embargoed`).
This is somewhat complicated. So please don't duplicate the logic here.

Run `doozer detect-embargo` for help. If the existing Doozer command is not powerful enough for this use case, you can extend it.

vfreex (changes requested):
See my comment inline.

jobs/build/ocp4/build.groovy
vfreex: I am not sure what you mean by `images list as args` here. doozer detect-embargo supports detecting by nvrs or release payloads.

jobs/build/ocp4/build.groovy
```diff
@@ -366,6 +366,17 @@ def stageBuildImages() {
 
         buildlib.doozer "${doozerOpts} images:streams mirror"
     }
+    // check new embargoed build
+    imagelist = buildPlan.imagesIncluded.replaceAll(',', ' ')
+    report = buildlib.doozer("${doozerOpts} -q detect-embargo images --version ${version.stream} ${imagelist}", [capture: true]).trim()
```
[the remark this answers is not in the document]
sosiouxme: `doozer detect-embargo images` doesn't exist yet, did you mean to link a PR for that? I think it's a good idea, but I think you'll want it to look up the last two builds and compare them.

jobs/build/ocp4/build.groovy
```diff
@@ -366,6 +366,17 @@ def stageBuildImages() {
...
+    if (report.returnStatus != 2) {
```
sosiouxme, line 372: don't trust the exact return status; it doesn't always seem to match what bash returns. really all you can safely check here is whether it's zero or not.

sosiouxme (changes requested):
Aside from the comments below, I see a few things about ART-2626 that aren't realized here.

Mainly, this lacks any awareness of whether an embargoed build is new. Just looking at the most recent build and seeing it embargoed doesn't tell you whether that's new. Just looking at the two most recent builds and seeing an embargo change doesn't tell you whether there was already an alert. I do not want more than one alert when a build turns from unembargoed to embargoed. So per the Jira, it first needs to check doozer's `state.yaml` for _successful_ builds (not just attempted builds -- they might fail and spam us repeatedly). Only for those should it be comparing the last two builds to see if status changed and sending an alert.

The other thing is that I'd like to know about embargo changes in RPMs as well as images. I think you'd need a `doozer detect-embargo rpms`? The two are really pretty independent, might be a separate PR.

jobs/build/ocp4/build.groovy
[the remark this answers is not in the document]
Ximinhan: yes, I have a PR for doozer to implement that

sosiouxme (changes requested):
`doozer detect-embargo` should detect whether any most recent build is embargoed.

That doesn't indicate whether this represents a change from the previous state. What we need to check here is whether a new build from this run of `ocp4` introduced an embargoed fix not present in the previous build. Otherwise, this would spam the channel every time there's a run of `ocp4`, until the embargo lifts.

PR #2620, opened 2021-05-14
aarch64 enablement

sosiouxme:
Adding on to [yaakov's PR here](https://github.com/openshift/aos-cd-jobs/pull/2610):

> Until now, x86_64/amd64 was the only enabled architecture with differing
> nomenclatures. However, aarch64/arm64 has the same issue, which
> accounts for most of the complications in adding it.

I'm still reviewing and looking carefully at what's affected here... I think there's a doozer piece too, but this is close.

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -57,8 +57,9 @@ class PrepareReleasePipeline:
             else:
                 arch = "x86_64"
             if ":" not in nightly:
-                # prepend pullsepc URL to nightly name
-                arch_suffix = "" if arch == "x86_64" else "-" + arch
+                # prepend pullspec URL to nightly name
+                # TODO: proper translation facility between brew and go arch nomenclature
+                arch_suffix = "" if arch == "x86_64" else "-amd64" if arch == "aarch64" else "-" + arch
                 nightly = f"registry.ci.openshift.org/ocp{arch_suffix}/release{arch_suffix}:{nightly}"
             self.candidate_nightlies[arch] = nightly
         self.elliott_working_dir = self.working_dir / "elliott-working"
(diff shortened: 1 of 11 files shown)


sosiouxme:
https://github.com/openshift/doozer/pull/406 supplies the related doozer changes around syncing images. this should include that.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -52,11 +52,14 @@ def __init__(
...
+            elif "arm64" in nightly:
+                arch = "aarch64"
             else:
                 arch = "x86_64"
             if ":" not in nightly:
-                # prepend pullsepc URL to nightly name
-                arch_suffix = "" if arch == "x86_64" else "-" + arch
+                # prepend pullspec URL to nightly name
+                # TODO: proper translation facility between brew and go arch nomenclature
+                arch_suffix = "" if arch == "x86_64" else "-amd64" if arch == "aarch64" else "-" + arch
```
yselkowitz, line 62: `arch_suffix = "" if arch == "x86_64" else "-arm64" if arch == "aarch64" else "-" + arch

iow `s/amd64/arm64/' in this line

sosiouxme: thanks, i must be blind

sosiouxme:
```suggestion
                arch_suffix = "" if arch == "x86_64" else "-arm64" if arch == "aarch64" else "-" + arch
```

sosiouxme:
still want to test out a thing or two but i think it's just about ready.

sosiouxme:
I'd like to get to the point where I can get the imagestream for the ARM nightly updated before merging this. Right now we're facing problems compiling some essential components as well as infra issues muddying the water so I can't test that properly. This is probably fine but I'm waiting to see.

PR #2623, opened 2021-05-17
Supply assembly argument to all doozer invocations

jupierce:
Requires https://github.com/openshift/doozer/pull/408 .
Casts a broad net and includes assembly in every doozer invocation. This will be refined later.


vfreex (approved):
LGTM. Requires https://github.com/openshift/doozer/pull/408 before merge.

/lgtm

vfreex:
Merging this (Luke has approved us taking some risk and merging)

PR #2646, opened 2021-06-15
ART-3059 Prepare release - Default advisory mode

thegreyd:
This change is so that we can use the prepare-release job when we're releasing RCs and preparing advisories upto GA. With every subsequent RC, we run prepare release in default advisory mode to sweep builds and bugs.

Test run: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Fprepare-release/9/

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -29,7 +29,8 @@ class PrepareReleasePipeline:
         working_dir: str,
         jira_username: str,
         jira_password: str,
-        dry_run: bool=False,
+        default_advisories: bool = False,
+        dry_run: bool = False,
     ) -> None:
         _LOGGER.info("Initializing and verifying parameters...")
         self.config = config
@@ -37,6 +38,7 @@ class PrepareReleasePipeline:
         self.release_date = date
         self.package_owner = package_owner or self.config["advisory"]["package_owner"]
         self.working_dir = Path(working_dir).absolute()
+        self.default_advisories = default_advisories
         self.dry_run = dry_run
         self.release_version = tuple(map(int, self.release_name.split(".", 2)))
         self.group_name = (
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -192,6 +198,38 @@ def create_advisory(self, type: str, kind: str, impetus: str) -> int:
...
+        subprocess.run(cmd, check=True, universal_newlines=True, cwd=self.working_dir)
+        # update advisory numbers
+        with open(repo / "group.yml", "r") as f:
+            group_config = f.read()
+        
+        advisories = {}
+        for kind in ["rpm", "image", "extras", "metadata"]:
+            match = re.search(
+                fr"^\s+{kind}:\s*([0-9]+)$", group_config, flags=re.MULTILINE
+            )
```
vfreex, line 226: I think it is better to use yaml module to parsee group.yml here so we don't need to hardcode the advisory kinds.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -192,6 +198,38 @@ def create_advisory(self, type: str, kind: str, impetus: str) -> int:
         advisory_num = int(match[1])
         return advisory_num
 
+    def get_advisories(self) -> Dict[str, int]:
+        repo = self.working_dir / "ocp-build-data-push"
+        shutil.rmtree(repo, ignore_errors=True)
```
vfreex, line 203: better to remove the repo in `save_advisories` function as well.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -71,32 +73,48 @@ def run(self):
...
+                # don't create placeholder bugs for OCP 4 image advisory and OCP 3 rpm advisory
+                if (
+                    not advisory
+                    or self.release_version[0] >= 4
+                    and kind == "image"
+                    or self.release_version[0] < 4
+                    and kind == "rpm"
+                ):
+                    continue
+                self.create_and_attach_placeholder_bug(kind, advisory)
```
vfreex, line 117: In the future we can improve this logic by only adding placeholder bugs if not bugs are swept into an advisory to ensure advisories are ready for QE. But for now it seems to ok to only create placeholder bugs for new advisories.

vfreex (approved):
Generally LGTM. I added a few suggestions.

/lgtm

pyartcd/pyartcd/pipelines/prepare_release.py
thegreyd: It was already doing it :)

pyartcd/pyartcd/pipelines/prepare_release.py
thegreyd: Done. save_advisories can also use that but going to defer that for the future.

thegreyd:
Looks good with new test run. Merging

PR #2654, opened 2021-06-17
[ART-3056] Don't fail if there are no bundles to build

thiagoalessio:
The pipeline already counts with stage conditions to prevent failure when there is nothing to be built, but they were beeing fooled by an odd behavior when splitting by `\n`:
```
groovy:000> doozer_output = "".split()
===> []
groovy:000> doozer_output.size()
===> 0
groovy:000> doozer_output = "".split("\n")
===> []
groovy:000> doozer_output.size()
===> 1
groovy:000>
```

So the list wasn't _technically_ empty, and consequentially the stage conditions weren't being met.


joepvd (approved):
/lgtm

thegreyd:
Noting that I reverted the change https://github.com/openshift/aos-cd-jobs/pull/2657

PR #2665, opened 2021-06-23
ART-3083 Prepare release - Perform sweep at the end

thegreyd:
Prepare release job sometimes fails due to errors in sweeping bugs/builds. These are considered risky steps. In case of job failure, next steps are performed manually by the artist and sometimes missed (https://coreos.slack.com/archives/CB95J6R4N/p1624464663042000). 

The subsequent steps of sending notification email to qe, and attaching placeholder bugs don't really rely on the success of the sweep. This change is to move sweep at the end, after those steps.


joepvd (approved):
/lgtm

PR #2673, opened 2021-06-25
WIP: Promote validations report and don't fail early

thegreyd:
There are several promote validations we run, some of them are independent, some of them are dependent. When a validation fails, it shouldn't block other validations from being run, so we get a report at the end. Then the ARTist can fix/retry/skip/abort as needed


thiagoalessio:
🤔 that might be true for GA, but I don't think we'll ever want to promote a z-stream if the payload advisory is not in `QE` state.

thegreyd:
Correct. Right now the later validations don't run if advisory is not in qe.. I still want to see the results of elliott verify payload, verify attached bugs etc before it errors out and gives me the option of skip/retry/abort

thegreyd:
Ignore the branch name :) I don't intend to skip this test, just to provide the result of all validations and then fail

PR #2681, opened 2021-07-01
create versionless symlinks for linux-amd64 binaries

jupierce:
Copy logic changes in https://github.com/openshift/aos-cd-jobs/pull/2678


yselkowitz:
/lgtm
sorry for missing this

PR #2685, opened 2021-07-05
Revert "keep check=False"

vfreex:
This reverts commit 36791e1298522fe502b9d6e42d3f182007c43ca0.

Errors shouldn't be ignored. We trust the status of prepare-release job so that if it finishes with success, we know everything is fine. I know if prepare-release fails it’s hard to run again, but that issue will gone when we finish the assemblies work.


thiagoalessio (approved):
/lgtm

to quote The Zen of Python:
> Explicit is better than implicit.

PR #2692, opened 2021-07-15
[ART-3136] embed validate rhsa

joepvd:
Make the new `elliott verify-rhsa` an informing stage at the end of the promote job. If there is anything to look at. let is knmow in the release specific channel.


vfreex (changes requested):
I remember @sfowl said this command is only for validating RHSAs for z-stream releases. In case of y-stream release, could you add an option to bypass the check?

thegreyd:
@vfreex it works for small advisories, so for GA it would work for rpm RHSA but not for image RHSA. I don't think getting an error message is an issue.. @joepvd maybe add a comment somewhere about this detail, so we remember it in the future?

sfowl:
Aside from the size issue, the Alerts for y-stream releases don't handle the first fix rule well, e.g.:

```
E: Tracker 1954940 is missing flaw(s) 1954736 on the erratum's bug list
  > Add the missing flaw to the erratum's bug list.

  > Note: This does not apply in cases when Y-stream erratum contains a Y-stream blocker
  > bug for a flaw that was previously fixed in Z-stream.
  > In such cases, the flaw must not appear in the erratum, only the Y-stream tracker.
  > Therefore, in case of the Y-stream errata, you need to manually verify all reported
  > issues to see if they were previously fixed in the product.
```
I.e. It still generates an alert and has a note for manual review. This might improve in SFM2 in the future.

Maybe it's worthwhile adding an option to bypass the check for y-stream releases until that time.

jobs/build/promote/Jenkinsfile
```diff
@@ -683,6 +683,27 @@ node {
...
+                }
+                if (advisory == -1) {
+                    return
+                }
+                if (major == 4 && !is_4stable_release) {
+                    return
+                }
+                release.getAdvisoryIds().each {
+                    res = commonlib.shell(
+                        script: "${buildlib.ELLIOTT_BIN} --group ${group} verify-rhsa ${it}",
```
sfowl, line 699: s/verify-rhsa/validate-rhsa/ ?

Also I don't think "--group ${group} " is strictly required.

joepvd:
@vfreex Could you review again?

PR #2707, opened 2021-07-29
art-2519: add check upstream step in image_health job

Ximinhan:
use doozer images sub-command to gather upstream pr report and only check on Monday's health check job to avoid spam message, The mapping of owner and redhat email address is still imperfect so by now not send email by bot.

### DIFF
--- a/jobs/scanning/images-health/Jenkinsfile
+++ b/jobs/scanning/images-health/Jenkinsfile
@@ -17,6 +17,17 @@ node() {
                         defaultValue: true,
                         description: "If false, output will only be sent to console"
                     ),
+                    choice(
+                        name: 'ALERT_DATE', 
+                        choices: [
+                            'Monday',
+                            'Tuesday',
+                            'Wednesday', 
+                            'Thursday',
+                            'Friday'
+                        ].join('\n'), 
+                        description: 'In which day check upstream prs'
+                    ),
                     commonlib.mockParam(),
                 ]
             ],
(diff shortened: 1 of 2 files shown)


jobs/scanning/images-health/Jenkinsfile
```diff
@@ -58,5 +56,32 @@ node() {
             currentBuild.result = "FAILURE"
             throw exception  // gets us a stack trace FWIW
         }
-    }
+
+        Date today = new Date();
+        Calendar c = Calendar.getInstance();
+        c.setTime(today);
+        if (c.get(Calendar.DAY_OF_WEEK) == 2 ) {  // to avoid spam message only check once on Mon
```
vfreex, line 63: Magic number `2` can be avoided by using constant like `Calendar.MONDAY`.

Also, how can I test the message if today is not Monday?

I would like to have a parameter to control whether to send the message, then move the weekday check to `scheduled_jobs/run_images-health` so we will have a central place to control how frequently this job should be triggered and how frequently we should send a message, without losing the flexibility to force send a message.

vfreex (changes requested):

jobs/scanning/images-health/Jenkinsfile
```diff
@@ -58,5 +67,33 @@ node() {
...
             throw exception  // gets us a stack trace FWIW
         }
+        
+        def week = [1:'Sunday', 2:'Monday', 3:'Tuesday', 4:'Wednesday', 5:'Thursday', 6:'Friday', 7:'Saturday']
+        Date today = new Date();
+        Calendar c = Calendar.getInstance();
+        c.setTime(today);
+        if ( week[c.get(Calendar.DAY_OF_WEEK)] == params.ALERT_DATE ) {  // to avoid spam message only check once on Mon
+            withCredentials([string(credentialsId: 'openshift-bot-token', variable: 'GITHUB_TOKEN')]) {
+                slackChannel = slacklib.to(BUILD_VERSION)
```
sosiouxme, line 77: we should have some kind of "summary" to #aos-art per the Jira, so it seems like these should be gathered for that purpose. however i'm really not sure what i'd want the summary to look like. a dump of hundreds of PRs is probably not it...

PR #2712, opened 2021-08-03
Optionally update "latest" symlink on operator-sdk sync

thiagoalessio:
we'll usually do it for 4.8 on Mondays, then 4.7 on Tuesdays, but we want to keep "latest" pointing to the latest 4.8, not the latest 4.7 we published later.

### DIFF
--- a/jobs/build/operator-sdk_sync/Jenkinsfile
+++ b/jobs/build/operator-sdk_sync/Jenkinsfile
@@ -41,6 +41,11 @@ pipeline {
             defaultValue: '',
             trim: true,
         )
+        booleanParam(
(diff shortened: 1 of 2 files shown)


vfreex:
I am ok with this PR but actually we can programmatically determine whether the current version is the latest or not: https://github.com/openshift/aos-cd-jobs/blob/1daf61bf8c8f3dae88fc8585aabc93fde257026d/pipeline-scripts/commonlib.groovy#L100

sosiouxme:
Looks fine to me, but needs rebase. Fix and merge?

thiagoalessio:
fixed & merged with human intervention for now, just because we need to use it nearly every week. but +1 for automating this.

PR #2721, opened 2021-08-13
Remove `StrictHostKeyChecking=no` from SSH invocations

vfreex:
It's not needed because our all CI environments has recorded the host key in known_hosts
and we don't really want to risk MIIT attacks.


joepvd:
🎉 
/lgtm

PR #2722, opened 2021-08-13
[ART-3069] olm_bundle job assembly awareness

vfreex:
1. Add `ASSEMBLY` and `OPERATOR_NVRS` parameters to `olm_bundle` job.
2. `ocp4`, `custom`, and `rebuild` jobs automatically call `olm_bundle` job to prebuild bundles for operator NVRs that they just built. We don't have to wait for release artists to pin the operator NVRs to the assembly.
3. Manually run `olm_bundle` job with `ASSEMBLY` will build bundles for operators that are currently selected by the assembly.
4. Improve troubleshooting capability by not capturing stdout/stderr for Doozer and Elliott.

Requires https://github.com/openshift/doozer/pull/498

### DIFF
--- a/jobs/build/ocp4/build.groovy
+++ b/jobs/build/ocp4/build.groovy
@@ -467,15 +467,22 @@ def stageSyncImages() {
         echo "No built images to sync."
         return
     }
-    if(buildPlan.dryRun) {
-        echo "Not syncing images in a dry run."
-        return
+
+    def record_log = buildlib.parse_record_log(doozerWorking)
+    def records = record_log.get('build', [])
+    def operator_nvrs = []
+    for (record in records) {
+        if (record["has_olm_bundle"] != '1' || record['status'] != '0' || !record["nvrs"]) {
+            continue
+        }
+        operator_nvrs << record["nvrs"].split(",")[0]
     }
     buildlib.sync_images(
         version.major,
         version.minor,
         "aos-team-art@redhat.com",
-        currentBuild.number
+        params.ASSEMBLY,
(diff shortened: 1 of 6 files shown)


jobs/build/ocp4/build.groovy
```diff
@@ -467,15 +467,22 @@ def stageSyncImages() {
         echo "No built images to sync."
         return
     }
-    if(buildPlan.dryRun) {
-        echo "Not syncing images in a dry run."
-        return
+
+    def record_log = buildlib.parse_record_log(doozerWorking)
```
jupierce, line 471: I wonder if this could be made more like scan-sources instead of relying on record logs. e.g. a verb smart enough that if any operator / operand image in the assembly is newer than the available bundle matching the NVR of the operator, it rebuilds the bundle. 
`olm-bundle:scan-and-rebuild`.

vfreex: It is possible. PR https://github.com/openshift/doozer/pull/498 made a change that if we run `doozer --group $group --assembly $name olm-bundle:rebase-and-build` without `--force` option or `NVRs`, it will check if the bundle for assembly-selected operator exists, and only rebase and build if that doesn't exist.

This part is to pre-build the bundles for operators that haven't been pinned to the assembly config. The built bundles will not be selected unless their operators are selected.

vfreex:
@jupierce Hey, could you look at this PR again? We want to integrate olm_bundle build with prepare-release job.

jupierce (approved):

PR #2724, opened 2021-08-18
Add explict name of the next z-stream for which the merge window is open

deads2k:
On my first patch manager rotation, I mis-interpretted the message about
what release was open for changes.  This explicitly names the version
which is open.

DANGER: I've never coded groovy and I don't know how to compile or test this.  It looked a like java, so I'm hoping the libraries match closely.


deads2k:
/hold

DANGER: I've never coded groovy and I don't know how to compile or test this. It looked a like java, so I'm hoping the libraries match closely.

I didn't see a "compile like this" in the repo readme, but if there is one, feel free to point me there.

PR #2725, opened 2021-08-19
ART-3377 Rhcos sync from release tag

thegreyd:
https://issues.redhat.com/browse/ART-3377
Test run: <>


vfreex (commented):
This job is called by `promote` job. needs to change promote job as well otherwise promote job will break.

PR #2731, opened 2021-08-23
WIP: Run olm_bundle as part of prepare-release

thegreyd:


thegreyd:
@vfreex What do you think would be the best way to get the advisory ids to trigger the job? Should prepare release spit out some output that we consume? or a separate elliott/doozer command?

thegreyd:
elliott seems to me the right place, so the change would be to pass in assembly to the job, which would pass it to elliott find-builds --attach --assembly {assembly_name}

vfreex:
@thegreyd With https://github.com/openshift/aos-cd-jobs/pull/2722, you can pass the assembly name via `ASSEMBLY` parameter. That PR hasn't been merged yet, let me try ask Justin to get another review.

vfreex:
Elliott find-builds doesn't find olm bundles at the moment, but olm-bundle job can attach the bundles by NVRs.

vfreex:
Also it would be nice if we run `elliott verify-attached-operators` after calling olm_bundle.

PR #2734, opened 2021-08-26
art-3078: Add create text only advisory

Ximinhan:
Extend advisory job to be able to used create text only advisory if needed, need to pass through all required parameters for advisory creation and related bug creation, because each use cases are different, don't have a template to use without modification but the default value have a basic instance of what needed, some minor version changes need human to check and confirm.

### DIFF
--- a/jobs/build/advisories/Jenkinsfile
+++ b/jobs/build/advisories/Jenkinsfile
@@ -97,6 +97,42 @@ node {
                             ].join(","),
                             trim: true
                         ),
+                        booleanParam(
+                            name: 'TEXT_ONLY',
+                            description: 'Create one text only advisory, the following params should modified manually if this is true',
+                            defaultValue: false,
+                        ),
+                        string(
+                            name: 'SYNOPIS'
+                            description: 'Synopsis value for text only advisory',
+                            defaultVale: "OpenShift Container Platform 4.7.z notification of delayed upgrade path to 4.8"
+                        ),
+                        string(
+                            name: 'TOPIC'
+                            description: 'Topic value for text only advisory',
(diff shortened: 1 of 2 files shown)


jobs/build/advisories/Jenkinsfile
```diff
@@ -135,7 +175,7 @@ node {
                         draft << "- ${it.key}: https://errata.devel.redhat.com/advisory/${it.value}"
                     }
                     body = draft.join('\n')
-                    if (params.DRY_RUN) {
+                    if (params.DRY_RUN || params.TEXT_ONLY) {
```
vfreex, line 178: I think we need live IDs for text-only advisories (until we get the permission to assign live IDs by ourselves).

jobs/build/advisories/advisories.groovy
```diff
@@ -67,4 +67,32 @@ def create_placeholder(kind) {
...
+      "--topic ${params.TOPIC}",
+      "--description ${params.DESCRIPTION}",
+      "--solution ${params.SOLUTION}",
+      "--bugtitle ${params.BUGTITLE}",
+      "--bugdescription ${params.BUGDESCRIPTION}",
+      "--assigned-to ${params.ASSIGNED_TO}",
+      "--manager ${params.MANAGER}",
+      "--package-owner ${params.PACKAGE_OWNER}"
+  ].join(' ')
+  if (params.DATE != null) {
```
vfreex, line 84: I think `params.DATE` will be empty string instead of `null` if the parameter is blank.
```suggestion
  if (params.DATE) {
```

vfreex (changes requested):

PR #2738, opened 2021-08-30
Use --reference-mode=source when creating payload

jupierce:
Without this parameter, the generated payload pullspecs will refer to the imagestream instead of the pullspecs in quay to which the imagestream refers.


sosiouxme (approved):
:shipit:

PR #2739, opened 2021-09-01
Fix olm_bundle job undefined operator_packages error

vfreex:
This allows us to continue to use this job with our "traditional"
approach.


joepvd (approved):
:)

/lgtm

PR #2740, opened 2021-09-01
ART-3306: prepare-release job enhancements

vfreex:
- Being able to rerun `prepare-release` job without side effects. It will not
  create new a JIRA ticket if there's already one defined in assembly
config but update it instead.
- Automatically building and attaching OLM bundles to the metadata advisory.
- Automatically validating OLM bundles.
- Supporting template JIRA subtasks.
- If there's an error in JIRA template, print a warning and keep the
  cloned JIRA as-is.
- Marking cloned JIRA tickets private if the template is private.
- New logic of creating placeholder bugs. It will only create and attach
  a placeholder bug if and only if an advisory doesn't have any bugs.
- Some Errata operations are unreliable. Now they will be automatically retried.
- Fix slack notification.

Note:
- Requires https://github.com/openshift/elliott/pull/226 and
  https://github.com/openshift/doozer/pull/505.
- If advisories are already defined in assembly config, this
  job will not create or update the existing advisories. You have to use
https://github.com/openshift/elliott/pull/225 to update the advisory
infos.

Test runs:
- 4.7.29: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Fprepare-release/50/console
- 4.6.44: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Fprepare-release/45/console

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -48,7 +48,7 @@ class PrepareReleasePipeline:
         self.release_name = None
         group_match = None
         if group:
-            group_match = re.fullmatch(r"openshift-(\d+).(\d)", group)
+            group_match = re.fullmatch(r"openshift-(\d+).(\d+)", group)
             if not group_match:
                 raise ValueError(f"Invalid group name: {group}")
         self.group_name = group
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -222,12 +222,13 @@ class PrepareReleasePipeline:
             for _, payload in self.candidate_nightlies.items():
                 self.verify_payload(payload, advisories["image"])
 
-        _LOGGER.info("Sending a notification to QE and multi-arch QE...")
-        if self.dry_run:
-            jira_issue_link = "https://jira.example.com/browse/FOO-1"
-        else:
-            jira_issue_link = jira_issue.permalink()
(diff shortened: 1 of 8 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -137,30 +152,37 @@ async def run(self):
...
+        _LOGGER.info("Ensuring JIRA ticket for release %s...", self.release_name)
+        jira_issue_key = group_config.get("release_jira")
+        jira_template_vars = {
+            "release_name": self.release_name,
+            "x": self.release_version[0],
+            "y": self.release_version[1],
+            "z": self.release_version[2],
+            "release_date": self.release_date,
+            "advisories": advisories,
+            "candiate_nightlies": self.candidate_nightlies,
```
joepvd, line 167: typo

joepvd (commented):
Looks good to me. Spotted one typo

pyartcd/pyartcd/pipelines/prepare_release.py
vfreex: Seems the typo is also in the JIRA template so we didn't notice that... I will hold off on the typo fix at this moment until we fix the releases for the next week so we can continue to use the job in my hackspace.

vfreex: Solved.

PR #2750, opened 2021-09-09
promote: Don't run verify-payload for null

vfreex:
When promoting an assembly and that assembly doesn't have reference nightlies, don't run `verify-payload`.


sosiouxme (approved):
thanks

PR #2754, opened 2021-09-13
Fix historical job view in ocp4

joepvd:
The stages of past jobs were not drawing because the "sweep" and "sync
images" stages are only created when needed. This change creates all
stages always. Also, make the logic of when a compose is created
somehwat more readable.

### DIFF
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -142,10 +142,10 @@ node {
                 // and announce on slack
 
                 stage("build compose") {
-                    if (buildlib.isBuildPermitted(doozerOpts) || joblib.buildPlan.buildRpms) {
+                    if (buildlib.getAutomationState(doozerOpts) in ["yes", "True"] || joblib.buildPlan.buildRpms) {
                         lock("compose-lock-${params.BUILD_VERSION}") {
                             joblib.stageBuildCompose()
-                            if(joblib.buildPlan.buildRpms) {
+                            if (joblib.buildPlan.buildRpms && buildlib.getAutomationState(doozerOpts) in ["scheduled", "False"]) {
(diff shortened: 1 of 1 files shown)


jobs/build/ocp4/Jenkinsfile
```diff
@@ -128,36 +128,31 @@ node {
...
 
-                if(buildlib.getAutomationState(doozerOpts) in ["no", "False"]){
-                    lock("compose-lock-${params.BUILD_VERSION}") {
-                        stage("build compose") { joblib.stageBuildCompose() }
-                    }
-                } else if(joblib.buildPlan.buildRpms){
-                    lock("compose-lock-${params.BUILD_VERSION}") {
-                        stage("build compose") {
+                stage("build compose") {
+                    if (buildlib.isBuildPermitted(doozerOpts) || joblib.buildPlan.buildRpms) {
```
sosiouxme, line 145: `getAutomationState` was used here due to the situation described in the comment; `isBuildPermitted` is more nuanced because it permits a human running the build to succeed during a freeze (but in that case we won't update the plashet unless we're building an RPM for it).

if you can think of a clearer way to explain this for future readers of the code, that'd be helpful. it's certainly a bit confusing.

sosiouxme (changes requested):
great, this annoys me all out of proportion to its importance. thank you.

one thing to fix up...

jobs/build/ocp4/Jenkinsfile
sosiouxme: possibly this exception just goes away at some point since we won't have merge windows and anyway would only be likely to run this job manually for an assembly

joepvd: Right. I believe this is fixed now.

thiagoalessio (approved):
/lgtm

PR #2755, opened 2021-09-13
Strip multiline messages to Slack

joepvd:
Example: https://coreos.slack.com/archives/GV64UQBPY/p1631527813006300


sosiouxme:
seems legit

PR #2764, opened 2021-09-17
Sync RHCOS deps to S3 mirror

jupierce:


jobs/build/rhcos_sync/build.groovy
```diff
@@ -101,7 +101,11 @@ def rhcosSyncMirrorArtifacts() {
             invokeOpts += " --nomirror"
     }
 
-    buildlib.invoke_on_use_mirror("rhcossync.sh", invokeOpts)
+    buildlib.invoke_on_use_mirror("rhcossync.sh", "-- ${invokeOpts} --basedir ${baseDir}")
+
+    withCredentials([aws(credentialsId: 's3-art-srv-enterprise', accessKeyVariable: 'AWS_ACCESS_KEY_ID', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY')]) {
+        commonlib.shell("${env.WORKSPACE}/jobs/build/rhcos_sync/S3-rhcossync.sh ${INVOKE_OPTS} --basedir ${s3MirrorBaseDir}")
```
sosiouxme, line 107: itym
```suggestion
        commonlib.shell("${env.WORKSPACE}/jobs/build/rhcos_sync/S3-rhcossync.sh ${invokeOpts} --basedir ${s3MirrorBaseDir}")
```

jobs/build/rhcos_sync/S3-rhcossync.sh
```diff
@@ -0,0 +1,191 @@
...
+
+  --nolatest       Do not update the 'latest' symlink after downloading
+  --test           Test inputs, but ensure nothing can ever go out to the mirrors
+  --nomirror       Do not run the push.pub script after downloading
+
+
+EOF
+}
+
+function downloadImages() {
```
sosiouxme, line 43: I see the necessity for downloading this again - other download is on use-mirror. but my experience with downloading this on buildvm previously was that it's unbearably slow. have you tried?

if there's a way to temp use s3 creds on use-mirror that _might_ be preferable but i hope i'm wrong because that doesn't seem like a good architecture. (and as far as i know, copying directly from one bucket to another would not improve it any)

sosiouxme (changes requested):
maybe try this out on actual buildvm with 4.9 rc2? (which doesn't look like it is going to survive)

jobs/build/rhcos_sync/S3-rhcossync.sh
jupierce: I tested copying from s3 to s3 and it was significantly faster from my laptop. So we have that going for us. 
I don't expect use-mirror to be around much longer, so we would need to provide our own staging location.

sosiouxme:
[did not work so well](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Frhcos_sync/21/console)
```
/mnt/workspace/jenkins/working/_lmeyer-dev_dev-build_rhcos_sync/jobs/build/rhcos_sync/S3-rhcossync.sh: line 43: /tmp/rhcos-synclist-21.txt: No such file or directory
```
after previous part of the job did
```
rm -f /tmp/rhcos-synclist-21.txt
```

... so sadly, didn't get to really check the download speed

jupierce:
_grumble_.. _grumble_.. encapsulation _grumble_.. _grumble_..

jupierce:
Updated to run the s3 mirroring first

jupierce:
@sosiouxme the script, in isolation, takes 30m on buildvm.

PR #2765, opened 2021-09-17
[ART-3331] Decouple mirror files from imagestream files in build-sync

jupierce:
The new assembly logic only generates a single mirroring list for each arch instead of one for public & private. The pipeline previously used mirroring files as a queue to determine which imagestream filenames to apply. This PR decouples the two and applies any imagestream file that `gen-payload` creates.


sosiouxme (approved):
needed a little fix :)

PR #2766, opened 2021-09-20
Fix and create triage scripts

joepvd:


hacks/triage/assembly-urls
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
+
+shopt -s extglob
+
+script="$(readlink $0)"
```
sosiouxme, line 5: so that it will work when not a symlink:
```suggestion
script="$(readlink $0)" || script="$0"
```

hacks/triage/nightlies
```diff
@@ -0,0 +1,88 @@
...
+for arch in "${arches[@]}"; do
+  if [[ -n "$version" && "$arch" == arm64 ]]; then
+    major="${version%.*}"
+    minor="${version#*.}"
+    if (( major == 4 && minor < 9 )); then
+      continue
+    fi
+  fi
+  build_link ${stream:-}
+done | qute-slurp
```
sosiouxme, line 88: guessing this only works with qutebrowser :) what about others?

sosiouxme (commented):
tried some things, overall seems pretty nice

hacks/triage/nightlies
joepvd: Oops! Should be fixed :)

hacks/triage/nightlies
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+halp() {
+  cat <<-EOHALP
+	nightlies: open web pages of release controller with a bunch of nightlies.
+	
+	nighlies -h|--help
```
sosiouxme, line 9:
```suggestion
	nightlies -h|--help
```

sosiouxme (approved):
rock on

PR #2767, opened 2021-09-20
Stop syncing stuff that does not move

joepvd:


sosiouxme (approved):
yeah... hard to imagine there's still a need for this.

PR #2769, opened 2021-09-22
[ART-3310] Fix FC promotion error

vfreex:
When promoting an FC using assembly, we may not have actual advisories created. `-1` is the placeholder advisory ID.

Existing code skips certain advisory related operations if `advisory` variable is `-1`. Assigning the image advisory in assembly config to global variable `advisory` should be able to fix this issue.

### DIFF
--- a/jobs/build/promote/Jenkinsfile
+++ b/jobs/build/promote/Jenkinsfile
@@ -196,6 +196,8 @@ node {
     next_is_prerelease = false
     def release_config = null
     def (major, minor) = [0, 0]
+    Map advisory_map = [:]
+    int advisory = 0
 
     // copy job params into normal vars, since params is an immutable map
     def skip = [
@@ -228,6 +230,9 @@ node {
         if (!params.VERSION) {
             error("You must explicitly specify a VERSION to promote a release for a non-stream assembly.")
         }
+        if (params.ADVISORY) {
+            error("Cannot override image advisory number for a non-stream assembly.")
+        }
         (major, minor) = commonlib.extractMajorMinorVersionNumbers(params.VERSION)
         group = "openshift-${major}.${minor}"
         // FIXME: Only support standard X.Y.Z and hotfix releases for now. RCs and FCs are not supported yet.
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

Do not particularly like the shorthands, as that is not groovy syntax I am using a lot, and have to stare and read docs to understand what is going on.

Surely this will become readable after a while :)

jobs/build/promote/Jenkinsfile
```diff
@@ -228,15 +230,19 @@ node {
         if (!params.VERSION) {
             error("You must explicitly specify a VERSION to promote a release for a non-stream assembly.")
         }
+        if (params.ADVISORY) {
+            error("Cannot override image advisory number for a non-stream assembly.")
+        }
         (major, minor) = commonlib.extractMajorMinorVersionNumbers(params.VERSION)
         group = "openshift-${major}.${minor}"
         // FIXME: Only support standard X.Y.Z and hotfix releases for now. RCs and FCs are not supported yet.
```
sosiouxme, line 238: aren't they?

jobs/build/promote/Jenkinsfile
```diff
@@ -228,15 +230,19 @@ node {
...
         release_config = buildlib.get_releases_config(group)?.releases?.get(params.ASSEMBLY)
         if (!release_config) {
             error("Assembly ${params.ASSEMBLY} is not defined in releases.yml.")
         }
-
-        direct_release_nightly = !release.getAdvisories(group)?.image  // Whether this assembly should go through Errata process is determined by group config advisories.
+        advisory_map = release.getAdvisories(group)?: [:]
+        advisory = advisory_map?.image?: -1
+        direct_release_nightly = advisory <= 0  // Whether this assembly should go through Errata process is determined by group config advisories.
```
sosiouxme, line 245: if we're talking about assemblies what do the group config advisories have to do with it?

sosiouxme (commented):
seems legit, just would like to make sure the comments are helpful

jobs/build/promote/Jenkinsfile
vfreex: I don't understand your meaning here.
This check makes sure advisory numbers are all defined in releases.yml. I don't think there's any use cases to specify image advisory using build parameter.

jobs/build/promote/Jenkinsfile
vfreex: Group config already has assembly group config merged on top of group.yml: https://github.com/openshift/doozer/blob/633736b5fd852bb4d6e82856f55291e03594789d/doozerlib/runtime.py#L304

PR #2771, opened 2021-09-22
prepare_release.py: create advisory live ids directly

sosiouxme:
we now have the ability to create our own live ids, so let's do that instead of asking docs.
UPDATE: now relies on [elliott creating live id](https://github.com/openshift/elliott/commit/9e4a6b3efb7eda5c9c4734d610d5644ca64cd298).

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -180,10 +180,6 @@ class PrepareReleasePipeline:
         _LOGGER.info("Updating ocp-build-data...")
         build_data_changed = await self.update_build_data(advisories, jira_issue_key)
 
-        if advisories_changed or build_data_changed:
-            _LOGGER.info("Sending an Errata live ID request email...")
-            self.send_live_id_request_mail(advisories)
-
         _LOGGER.info("Sweep builds into the the advisories...")
         for kind, advisory in advisories.items():
             if not advisory:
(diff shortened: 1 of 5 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -650,24 +650,13 @@ async def verify_attached_operators(self, *advisories: List[int]):
...
-{main_advisory}: https://errata.devel.redhat.com/advisory/{advisories[main_advisory]}
-
-This is the current set of advisories we intend to ship:
-"""
+    async def create_live_ids(self, advisories: Dict[str, int]):
         for kind, advisory in advisories.items():
-            content += (
-                f"- {kind}: https://errata.devel.redhat.com/advisory/{advisory}\n"
+            _LOGGER.info(f"Giving {kind} advisory {advisory} a Live ID.")
+            await exectools.cmd_assert_async(
```
joepvd, line 656: How does this return when a LiveID already has been assigned?

In order to have meaningful return codes, we might need to read if a LiveID has been assigned already, like this: https://github.com/openshift/aos-cd-jobs/blob/76d7ee16bb34dd81eb284ac5aaac53cd1ca9533b/pipeline-scripts/release.groovy#L120

joepvd (commented):
~This meeting could have been an email~
This email could have been an API call

joepvd:
Also, once an official API exists, we should use that: https://issues.redhat.com/browse/CLOUDWF-6578

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -650,24 +650,13 @@ async def verify_attached_operators(self, *advisories: List[int]):
...
+                f"curl -X POST --negotiate -u : https://errata.devel.redhat.com/errata/set_live_advisory_name/{advisory}\n"
```
vfreex, line 657: Would like to add `--fail` option to return non-zero code if response status is not 200-399.

pyartcd/pyartcd/pipelines/prepare_release.py
sosiouxme: it turns out to be idempotent AFAICS. checking the return code and maybe retrying would probably be a good idea though (and a slack msg if it fails - it's not critical). if it's any more complicated than this command (or even if that's all it is), might make a good elliott command

vfreex: I think we only need to request the live ID just after a new advisory is created. When updating an existing advisory, we don't need to request again.

vfreex: So this function can be chained after the create advisory logic.

sosiouxme: yeah this should just go in `elliott create`

pyartcd/pyartcd/pipelines/prepare_release.py
sosiouxme: did that in [elliott](https://github.com/openshift/elliott/commit/9e4a6b3efb7eda5c9c4734d610d5644ca64cd298)

sosiouxme:
I haven't run this but I think it's safe.

PR #2785, opened 2021-10-05
Upgrade test suggestion fix

thegreyd:
When in_flight_prev is an empty string, we get an extra line. Caught here https://coreos.slack.com/archives/C01UQM3PM3L/p1633427199193600


thiagoalessio (approved):
/lgtm

joepvd:
I believe this has been addressed recently in some other commits

PR #2789, opened 2021-10-13
[ART-3040] Publish changelog for pre-GA releases

jupierce:


pipeline-scripts/release.groovy
```diff
@@ -442,6 +442,37 @@ for f in *.tar.gz *.bz *.zip *.tgz ; do
...
+        rcArch = commonlib.getReleaseControllerArch(release_name)
+        stableStream = (rcArch=="amd64")?"4-stable":"4-stable-${rcArch}"
+        prevGA = "${major}.${prevMinor}.0"
+        outputDest = "${BASE_TO_MIRROR_DIR}/changelog.html"
+
+        // See if the previous minor has GA'd yet; e.g. https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.8.0
+        def check = httpRequest(
+            url: "${rcURL}/releasestream/${stableStream}/release/${prevGA}",
+            httpMode: 'GET',
+            validResponseCodes: '200,404',  // if we get 404, do not compute changelog yet; prev has not GA'd.
```
sosiouxme, line 459: won't this cause an exception for any other problems? could we try/catch/retry in case of network bobbles etc? in general it seems like it would be bad for computing the changelog to nuke the release...

jupierce: I did want it to get some attention if it started to fail, but nuking the release is too much. I've added a try/catch with a slack on failure.

PR #2796, opened 2021-10-21
[ART-2682] Verify flaw bugs are attached and associated with specific…

vfreex:
… builds

Requires https://github.com/openshift/elliott/pull/239


joepvd (changes requested):
This would need to be incorporated in `promote-assembly`...

vfreex:
@joepvd PR has been updated to be incorporated in promote-assembly.

joepvd (approved):

PR #2804, opened 2021-11-03
[ART-3253] justify overrides in promote job

vfreex:
When the "SKIP" option is offered, there should also be a text field for providing a justification, which is required if the user does indeed skip


joepvd (approved):
/lgtm

PR #2809, opened 2021-11-08
[ART-3383] Add check-bugs job + scheduled job

thiagoalessio:
Look for potential blocker bugs and regressions of a given OCP version and communicate findings to ARTists on Slack.

Depends on https://github.com/openshift/elliott/pull/244 for `check-regressions` stage.

_*NOTE:* requires real tests on buildvm. Since these jobs don't affect our release flow, I'd suggest we merge it rather sooner, and follow up on any issues that show up_

### DIFF
--- a/jobs/build/check-bugs/Jenkinsfile
+++ b/jobs/build/check-bugs/Jenkinsfile
@@ -49,6 +49,9 @@ pipeline {
         stage("check-regressions") {
             steps {
                 script {
+                    if (next_is_prerelease(params.BUILD_VERSION)) {
+                        return
+                    }
                     bugs = commonlib.shell(
                         returnStdout: true,
                         script: """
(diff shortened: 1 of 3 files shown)


jobs/build/check-bugs/Jenkinsfile
```diff
@@ -0,0 +1,108 @@
...
+                        script: """
+                            ${buildlib.ELLIOTT_BIN}
+                            --group openshift-${params.BUILD_VERSION}
+                            verify-bugs ${bugs}
+                        """.stripIndent().tr("\n", " ").trim()
+                    ).trim()
+                }
+            }
+        }
+        stage("slack-notification") {
```
joepvd, line 79: It'd be nice if the text would be a bit nicer:
- Special content when no issues were found (green check mark or so)
- only display `blocker_bugs` if that test has something to look at (no header for things we don't need to look at)
- Have clickable links on the bugs, not only text of their ID

joepvd (approved):
Of course I find ways to make the end result prettier. I'm good with this as-is, though, as it is a very nice addition!

thiagoalessio:
yeah, let's follow up in subsequent PRs. merging this one now.

PR #2814, opened 2021-11-10
Send images health Failures to aos-art

thegreyd:
This came out of a need to have @yselkowitz (and other devs) access to brew failed tasks so they can debug and fix them, especially on the occasion of golang bump where we would need to communicate brew builds

### DIFF
--- a/jobs/scanning/images-health/Jenkinsfile
+++ b/jobs/scanning/images-health/Jenkinsfile
@@ -47,14 +47,16 @@ node() {
 
     timestamps {
 
-        slackChannel = slacklib.to(BUILD_VERSION)
+        releaseChannel = slacklib.to(BUILD_VERSION)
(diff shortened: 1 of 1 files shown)


jobs/scanning/images-health/Jenkinsfile
```diff
@@ -47,23 +47,25 @@ node() {
 
     timestamps {
 
-        slackChannel = slacklib.to(BUILD_VERSION)
+        releaseChannel = slacklib.to(BUILD_VERSION)
+        aosArtChannel = slacklib.to("aos-art")
```
jupierce, line 51:
```suggestion
        aosArtChannel = slacklib.to("#aos-art")
```

jupierce (changes requested):

PR #2817, opened 2021-11-15
[ART-2722] create 4.x+1 dir to support CI when branch cut

Ximinhan:
during branch cut release version bumped but we haven't build image yet so keep create 4.x+1 dir with the same repos as 4.x latest.

### DIFF
--- a/scheduled-jobs/build/sync-for-ci/Jenkinsfile
+++ b/scheduled-jobs/build/sync-for-ci/Jenkinsfile
@@ -24,8 +24,13 @@ def runFor(group, dir, arch="x86_64") {
 
 @NonCPS
 def sortedVersions() {
+  // Return versions honoring semver, but exclude disabled
   def disabled = []
-  return commonlib.ocp4Versions.sort(false).findAll { !disabled.contains(it) }
+  def s = commonlib.ocp4Versions.sort(false) {
+    def (major, minor) = it.tokenize('.')
+    major*100000 + minor
+  }
+  return s.findAll { !disabled.contains(it) }
 }
 
 node() {


jobs/build/sync-for-ci/Jenkinsfile
```diff
@@ -127,15 +129,28 @@ node {
 
                 stage("push to s3") {
                     commonlib.syncRepoToS3Mirror("${LOCAL_SYNC_DIR}/", "${S3_SYNC_DIR}/")
+                    if ( commonlib.ocp4Versions.get(0) == ${REPOSYNC_DIR} ) {
```
sosiouxme, line 132: I don't think you can put a bare `${}` like that. Have you tried this in hack space?
probably ought to be:
```suggestion
                    if ( commonlib.ocp4Versions.get(0) == REPOSYNC_DIR ) {
```

jobs/build/sync-for-ci/Jenkinsfile
```diff
@@ -80,6 +80,8 @@ node {
     LOCAL_SYNC_DIR = "${REPOSYNC_BASE_DIR}/${REPOSYNC_DIR}"
     MIRROR_TARGET = "use-mirror-upload.ops.rhcloud.com"
     MIRROR_RELATIVE_REPOSYNC = "reposync/${REPOSYNC_DIR}"
+    def (MAJOR, MINOR) = [0, 0]
+    (MAJOR, MINOR) = commonlib.extractMajorMinorVersionNumbers(params.REPOSYNC_DIR)
```
sosiouxme, line 84: REPOSYNC_DIR isn't always a version number, so I think this will fail on e.g. `ci-deps` runs of this script. In that case it doesn't make sense to figure out the next anyway.

jobs/build/sync-for-ci/Jenkinsfile
```diff
@@ -127,15 +129,28 @@ node {
...
                 }
 
                 stage("push to mirror") {
                     sh "ssh -o StrictHostKeyChecking=no ${MIRROR_TARGET} -- mkdir --mode 755 -p ${MIRROR_SYNC_DIR}"
                     sh "rsync -avzh --copy-links --chmod=a+rwx,g-w,o-w --delete ${LOCAL_SYNC_DIR}/ ${MIRROR_TARGET}:${MIRROR_SYNC_DIR} "
-
                     timeout(time: 2, unit: 'HOURS') {
                         sh "ssh -o StrictHostKeyChecking=no ${MIRROR_TARGET} -- push.enterprise.sh -v ${MIRROR_RELATIVE_REPOSYNC}"
                     }
+                    if ( commonlib.ocp4Versions.get(0) == ${REPOSYNC_DIR} ) {
```
sosiouxme, line 145: same thing here

sosiouxme (changes requested):
This makes me slightly nervous about breaking CI repos. Do you have a successful hack space run against 4.5 or something?

jobs/build/sync-for-ci/Jenkinsfile
```diff
@@ -127,15 +131,28 @@ node {
 
                 stage("push to s3") {
                     commonlib.syncRepoToS3Mirror("${LOCAL_SYNC_DIR}/", "${S3_SYNC_DIR}/")
+                    if ( commonlib.ocp4Versions.get(0) == REPOSYNC_DIR ) {
```
vfreex, line 134: This assumes versions in `commonlib.ocp4Versions` are in descending order. It's ok for now but I am afraid this will bite us someday.
Better to loop over this array to determine which one is the latest.

jobs/build/sync-for-ci/Jenkinsfile
```diff
@@ -127,15 +131,28 @@ node {
...
+                        // create 4.(x+1) dir to meet CI when branch cut
+                        def S3_SYNC_DIR_NEXT = S3_SYNC_DIR.replaceAll("${REPOSYNC_DIR}","${MAJOR}.${MINOR+1}")
```
vfreex, line 136: String replacement doesn't seem to be appropriate here because you are assuming `REPOSYNC_DIR` is not repeating. I would suggest concatenating the desired path rather than replacing a substring in the original path.

jobs/build/sync-for-ci/Jenkinsfile
```diff
@@ -127,15 +131,28 @@ node {
...
                 stage("push to mirror") {
                     sh "ssh -o StrictHostKeyChecking=no ${MIRROR_TARGET} -- mkdir --mode 755 -p ${MIRROR_SYNC_DIR}"
                     sh "rsync -avzh --copy-links --chmod=a+rwx,g-w,o-w --delete ${LOCAL_SYNC_DIR}/ ${MIRROR_TARGET}:${MIRROR_SYNC_DIR} "
-
                     timeout(time: 2, unit: 'HOURS') {
                         sh "ssh -o StrictHostKeyChecking=no ${MIRROR_TARGET} -- push.enterprise.sh -v ${MIRROR_RELATIVE_REPOSYNC}"
                     }
+                    if ( commonlib.ocp4Versions.get(0) == REPOSYNC_DIR ) {
+                        MIRROR_SYNC_DIR_NEXT = MIRROR_SYNC_DIR_NEXT.replaceAll("${REPOSYNC_DIR}","${MAJOR}.${MINOR+1}")
+                        MIRROR_RELATIVE_REPOSYNC_NEXT = MIRROR_RELATIVE_REPOSYNC.replaceAll("${REPOSYNC_DIR}","${MAJOR}.${MINOR+1}")
```
vfreex, line 149: Same here.

vfreex (changes requested):

jupierce:
I think this can more easily be handled in `scheduled-jobs/build/sync-for-ci/Jenkinsfile` (the scheduled job, not the build job). Ultimately, in the scheduled job, we just need to invoke `/aos-cd-builds/build%2Fsync-for-ci` with the same `GROUP` but different `REPOSYNC_DIR`. For example, in `scheduled-jobs/build/sync-for-ci/Jenkinsfile` , I believe logic like this would get the job done:

```
        for ( String arch : arches ) {
            runFor(group, version, arch)
            // new lines follow
            if ( version == sortedVersions()[-1] ) {
                (major, minor) = commonlib.extractMajorMinorVersionNumbers(version) 
                runFor(group, "{major}.{minor+1}", arch)
            }
        }
```

Ximinhan:
rebase with https://github.com/openshift/aos-cd-jobs/pull/2879

PR #2671, opened 2021-06-25
Create only candidate PR

thiagoalessio:
PR https://github.com/openshift/cincinnati-graph-data/pull/841 should take care of the rest ;D

PR #2819, opened 2021-11-15
jobs/build/cincinnati-prs: Drop CANDIDATE_CHANNEL_ONLY

wking:
Since 936c8783f2 (#2671), fast and later promotion has been managed by the graph-data stabilization robot.  Instead of asking the user to set this property, just inspect the requested release name to determine if it's a prerelease (which won't want an errata link) or not (where we'll now return to requiring an errata link).


wking:
Rebased around #2804 with 9cb199287 -> 4b084eab2.

pipeline-scripts/release.groovy
```diff
@@ -762,17 +760,9 @@ def openCincinnatiPRs(releaseName, advisory, candidate_only=false, ghorg='opensh
...
-            }
-            if ( isSupportEUS("${major}.${minor}") ) {
-                prefixes = [ "candidate", "fast", "stable", "eus"]
-            }
-
-            if (isReleaseCandidate) {
-                // Release Candidates never go past candidate
-                prefixes = prefixes.subList(0, 1)
+                prefixes = [ "prerelease" ]
             }
```
sosiouxme, line 766: might as well make a clean sweep, we're not gonna release any more 4.1 :-D

sosiouxme (approved):

PR #2820, opened 2021-11-17
Improve email to QE regarding assemblies

joepvd:


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -655,6 +655,9 @@ def send_notification_email(self, advisories: Dict[str, int], jira_link: str):
             content += "\nNightlies:\n"
             for arch, pullspec in self.candidate_nightlies.items():
                 content += f"- {arch}: {pullspec}\n"
+        else:
+            content += "\nThis release has been doctored with, and is not directly based off a nightly.\n"
+            content += f"The definition of the assembly can be found under key {self.assembly_name} in https://github.com/openshift/ocp-build-data/blob/{self.group_name}/releases.yml\n"
```
sosiouxme, line 660: Good idea. I just like my own wording better :) 
```suggestion
            content += "\nThis release is NOT directly based on existing nightlies.\n"
            content += f"Its definition is provided by the assembly found under key '{self.assembly_name}' in https://github.com/openshift/ocp-build-data/blob/{self.group_name}/releases.yml\n"
```

joepvd:
Thanks for your word smithing, @sosiouxme!

PR #2821, opened 2021-11-19
images-health: stop posting to aos-art

joepvd:
...until  https://issues.redhat.com/browse/ART-3425 is resolved. We do
not want to teach that these messages can be ignored.


thiagoalessio (approved):
/lgtm

PR #2825, opened 2021-11-22
Custom data path in build-sync

thegreyd:
I wanted to use a custom data path for testing out not-yet-ready assembly definitions.
Was thinking we could force dry_run when a custom data path is used, to be extra safe. 
Plus use --skip-gc-tagging flag, when running in dry_run mode.


vfreex (approved):
/lgtm

PR #2830, opened 2021-11-25
[ART-3528] no need to push images

Ximinhan:
seems last etcd image left also can be deprecated since it didn't update for a long time


joepvd (approved):
/lgtm

Should we remove the branch ocp-build-data:sync-misc? With or without soak time?

Ximinhan:
I don't know, would be safe to keep it for some time

PR #2835, opened 2021-11-29
Run ocp4 release scans in parallel

jupierce:
The original implementation of scan-sources required upstream source
for each image to be cloned. This intensive (in terms of CPU and
disk space) operations was serialized to prevent exhausting resources.

scan-sources is now relatively quick (5 minutes) and does not clone
upstream source. Even with these improvements, with five releases
being scanned, a change in 4.x can wait to be detected for ~20 minutes
because of the serial scans of releases.

This change allows all scans to happen in parallel - meaning a 4.x
scan is primarily waiting for 4.x builds.

### DIFF
--- a/jobs/build/ocp4_scan/Jenkinsfile
+++ b/jobs/build/ocp4_scan/Jenkinsfile
@@ -38,11 +38,6 @@ node('covscan') {
                         defaultValue: commonlib.ocp4Versions.join(','),
                         trim: true,
                     ),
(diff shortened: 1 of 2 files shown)


sosiouxme (approved):
seems legit

Just going to note that this moves the parallelization into separate job invocations. If you run ocp_scan directly with multiple versions, they're still serialized. But no one should be doing that anyway.

/lgtm

PR #2837, opened 2021-11-30
tkn artifact dir changed

Ximinhan:
those dir now changed to staging. 
ref https://coreos.slack.com/archives/CB95J6R4N/p1638280289111300?thread_ts=1638273105.104700&cid=CB95J6R4N


joepvd (approved):
/lgtm

PR #2846, opened 2021-12-07
 [ART-3561] Fix late push image for 3.11

joepvd:
Ensure openshift-enterprise AKA openshift3/ose-control-plane gets pushed
to reg-aws.

Addresses https://issues.redhat.com/browse/ART-3561

### DIFF
--- a/jobs/build/ocp3/Jenkinsfile
+++ b/jobs/build/ocp3/Jenkinsfile
@@ -684,6 +684,12 @@ node {
                             --filter-by-os amd64
                             --push-to-defaults ${ODCS_OPT}
                         """
+                        buildlib.doozer """
+                            ${doozerOpts}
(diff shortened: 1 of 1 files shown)


jobs/build/ocp3/Jenkinsfile
```diff
@@ -684,6 +684,12 @@ node {
                             --filter-by-os amd64
                             --push-to-defaults ${ODCS_OPT}
                         """
+                        buildlib.doozer """
```
vfreex, line 687: @joepvd Could you add a comment about the hack so we won't forget the reason to do this?
Otherwise LGTM

sosiouxme (approved):
slightly kludgey but it's 3.11 :shrug: 
/lgtm

PR #2847, opened 2021-12-08
pipeline-scripts/release: Pivot to version-agnostic candidate channel

wking:
Follows up on #2819; consider reviewing that one first.

Instead of adding 4.y.z to `candidate-4.y` and `candidate-4.(y+1)`, add them to the version-agnostic, internal `candidate channel` created by openshift/cincinnati-graph-data#112.  [The graph-data stabilization bot][2] can take it from there.

[2]: https://github.com/openshift/cincinnati-graph-data/blob/ce3e909ce7275257d397f58e85b3e57c1b2cc588/hack/stabilization-changes.py


sosiouxme (approved):
/lgtm
and tested

PR #2849, opened 2021-12-09
Have two cri-o in plashet

joepvd:
See https://coreos.slack.com/archives/C01CQA76KMX/p1639060078470200

A regression is suspected in cri-o, and rhcos probably wants to pin
previous version.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -1459,6 +1459,7 @@ def buildBuildingPlashet(version, release, el_major, include_embargoed, auto_sig
                     "--include-previous-for python3-openvswitch", // this is a package prefix
                     "--include-previous-for ovn", // this is a package prefix
                     "--include-previous-for haproxy", // avoid chicken and egg issues with base image & haproxy bumps
+                    "--include-previous-for cri-o", // Fix regression in 4.10 CI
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
Looks sane to me.
Why not just do it for all packages?

joepvd:
>Why not just do it for all packages?

Because we generally do not want pins of rpms. If this was for an already shipping release, we would ship the latest rpm, even if this is not consumed by (all) the images. The exception needs to stay an exception.

cgwalters:
The larger OpenShift strategy for what we ship is not "the latest".  The flow for everything is:

commit on GH → CI → merge → internal build → nightly → integration tests → ship

In this vein, what we should ship from the rpm repos should similarly be defined by CI, not by "the latest".  This entire issue has occurred because of that gap.

So to flip this around, once RHCOS gains lockfiles, the version of crio.rpm we ship should be defined by that lockfile.

sosiouxme (approved):

sosiouxme:
@cgwalters it may be worth discussing this a bit with Justin and me. Pins make things awkward when it comes time to ship things. "Latest is best" is pretty simple to follow, and relatively verifiable; any deviation risks balls being dropped. Some of this though is due to the errata restriction that we can only ship one build of a package, and only in increasing order. It may be worth exploring how to violate that.

PR #2860, opened 2021-12-23
fix shasum didn't update in stable

Ximinhan:
seems aws sync didn't update some file if the file size didn't change, ref the solution here https://github.com/aws/aws-cli/issues/3273


thiagoalessio (approved):
/lgtm

PR #2867, opened 2022-01-13
add dev options for golang-builder job

sosiouxme:


vfreex (approved):
/lgtm

sosiouxme:
doozer fails currently when you build `--scratch` but that can be solved separately

PR #2869, opened 2022-01-16
[ART-3397]: begin cleansing references to old mirror

sosiouxme:
this doesn't attempt to get them all, just some that I noticed currently blocking build and release.


joepvd (approved):
/lgtm

PR #2882, opened 2022-01-26
Fix set-client-latest

joepvd:
...and avoid
```
java.lang.UnsupportedOperationException: not supporting tuples yet
```


thiagoalessio (approved):
/lgtm

PR #2886, opened 2022-01-27
Do not require `initialize` to set up proxy env vars

joepvd:


vfreex (approved):
/lgtm
We also need to delete proxy settings at `$JENKINS_URL/configure`.
We may also encounter errors if there's a job doesn't import commonlib/buildlib.

PR #2888, opened 2022-01-27
Fix release message promote-assembly

joepvd:


vfreex:
The first parameter of `sendReleaseCompleteMessage` is a dict.
This issue will also be fixed by https://github.com/openshift/aos-cd-jobs/pull/2894

PR #2863, opened 2022-01-07
Migrate to have gen-payload apply imagestream updates

jupierce:
Requires https://github.com/openshift/doozer/pull/546

PR #2891, opened 2022-01-27
[ART-3656] build-sync: Don't fail in case of no change

thegreyd:
Right now multiple build-syncs can be running in parallel / close to each other. 
This causes one of them to fail since a previous one updated the imagestreams
 and there's nothing to do. In that case one of our verification fails - 
with error "Image Stream did not update"
since it thinks we failed to perform an update. This change fixes that by skipping if 
there isn't an update to apply

Test run: 
- no update - https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Fbuild-sync/4/console
- update - https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Fbuild-sync/5/console

### DIFF
--- a/jobs/build/build-sync/build.groovy
+++ b/jobs/build/build-sync/build.groovy
@@ -131,6 +131,13 @@ def buildSyncApplyImageStreams() {
         writeJSON(file: currentISfilename, json: currentIS)
         artifacts.addAll(["pre-apply-${namespace}-${theStream}.json"])
 
+        // Make sure there's still an update to IS
+        def diffStatus = buildlib.oc("--kubeconfig ${buildlib.ciKubeconfig} diff --filename=${isFile}", [returnStatus: true])
+        if ( diffStatus == 0 ) {
+            echo("No difference found in generated and current IS. Skipping")
+            continue
+        }
+
         // We check for updates by comparing the object's 'resourceVersion'
         def currentResourceVersion = currentIS.metadata?.resourceVersion ?: 0
         echo("Current resourceVersion for ${theStream}: ${currentResourceVersion}")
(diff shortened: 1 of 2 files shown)


jobs/build/build-sync/build.groovy
```diff
@@ -127,8 +127,16 @@ def buildSyncApplyImageStreams() {
...
         // Get the current IS and save it to disk. We may need this for debugging.
         def currentIS = getImageStream(theStream, namespace)
-        writeJSON(file: "pre-apply-${namespace}-${theStream}.json", json: currentIS)
-        artifacts.addAll(["pre-apply-${namespace}-${theStream}.json"])
+        def currentISfilename = "pre-apply-${namespace}-${theStream}.json"
+        writeJSON(file: currentISfilename, json: currentIS)
+        artifacts.addAll([currentISfilename])
+
+        // Make sure there's still an update to IS
+        def diffStatus = buildlib.oc("--kubeconfig ${buildlib.ciKubeconfig} diff --filename=${isFile}", [capture: true, returnStatus: true])
```
vfreex, line 135: I am not sure this will do the right thing. In kubernetes, admission controllers and other operators may add additional fields (like `annotations`) to the resource. This may always produce a diff. But I could be wrong.

joepvd:
I think this can get linked to ticket https://issues.redhat.com/browse/ART-3656. 

This will create merge conflicts with in-flight work https://github.com/openshift/aos-cd-jobs/pull/2863.

/hold

jobs/build/build-sync/build.groovy
vfreex: `oc apply` is idempotent. If we don't want to it fail in case of no change, why not deleting the code that checks the resource version?

thegreyd: I agree simpler would be to just remove the verification. In the past there were cases when imagestream did not update - oc apply failed for some reason - I don't know if they still happen? But I would say since this is better than just plain failing - it does skip when there is not change to make - see the linked test run. Maybe it still produces a diff in some cases - if it does let's come back and fix that?

PR #2905, opened 2022-01-31
s/continue/return/ in stage

joepvd:


jupierce:
The `continue` statements are inside of the `lock` closure. Effectively, they are on their stack frame / separate method. It cannot use the loop context in the previous stack frame.

PR #2906, opened 2022-02-01
Remove exposed URL as per InfoSec team request

locriandev:


jupierce:
Thanks for take the initiative here. Unfortunately, the request is to scrub from all history. I'm working on that now as it requires `admin` to force push.

PR #2927, opened 2022-02-11
[ART-3781] [ART-3783] Improve check-bugs job

locriandev:


jobs/build/check-bugs/Jenkinsfile
```diff
@@ -1,8 +1,21 @@
 #!/usr/bin/env groovy
 node {
+    currentBuild.displayName = params.BUILD_VERSION
+
+    // Skip check for 3.11
+    if (params.BUILD_VERSION == "3.11") {
```
joepvd, line 6: Think the skip for 3.11 should only happen for the regression check. We do want to have the check for bugs with the `blocker+` flag set.

joepvd (commented):
Nice! One nit, then we can merge

jobs/build/check-bugs/Jenkinsfile
locriandev: fixed

joepvd:
/lgtm

:shipit:

PR #2943, opened 2022-02-16
[ART-3784] Improve output formatting for found blockers

locriandev:


scheduled-jobs/build/check-bugs/Jenkinsfile
```diff
@@ -4,11 +4,26 @@ properties( [
     disableResume(),
 ] )
 
-def commonlib = load("pipeline-scripts/commonlib.groovy")
-commonlib.ocpVersions.each { it ->
+def sortedVersions() {
+  return commonlib.ocpVersions.sort(false)
```
vfreex, line 8: This actually sorts versions by alphabetical order, which is technically incorrect.
You may correct the logic to sort versions, or don't sort at all.

vfreex (changes requested):
Generally LGTM. Could you exclude the elliott bump from this PR? It is referencing the commit of your PR, which is likely to change during review process or cause conflict if elliott is bumped by other people.

scheduled-jobs/build/check-bugs/Jenkinsfile
locriandev: I removed the sorting. By the way, the function is used in a few places, see for example https://github.com/openshift/aos-cd-jobs/blob/master/scheduled-jobs/scanning/images-health/Jenkinsfile#L24

If you think that's incorrect, we might want to fix this across the whole repo

PR #2947, opened 2022-02-19
Trigger rhcos builds when rpms may have changed

jupierce:
When we update plashets (or when we are rebuilding images due to RHEL dependency updates), trigger RHCOS builds. If there are no changes, the build will quickly exit.

### DIFF
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -156,14 +156,18 @@ node {
                     }
                 }
 
-                // Since plashets may have been rebuild, fire off sync for CI. TODO: Run for other arches
-                // if CI ever requires them.
-                // Disabled because it increases 404s from mirror. sync-for-ci completely replaces content, so if
-                // yum is retrieving content while the mirror is being updated, yum can error out. This is
-                // exacerbated by a 1 minute caching effect in the rpm mirroring pods in CI for repo manifest
-                // data: https://github.com/openshift/content-mirror/pull/1#discussion_r581380438 . We need to
-                // find a way to stack these files so that the caching effect does not result in 404s.
-                build wait: false, propagate: false, job: '/scheduled-builds/sync-for-ci', parameters: [string(name: 'ONLY_FOR_VERSION', value: params.BUILD_VERSION)]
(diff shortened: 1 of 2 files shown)


sosiouxme:
/hold

there was a limitation in p8 capacity (IIRC) such that it doesn't handle more than two concurrent builds very well; the spread-out schedule we ended up with is intended to avoid that. any thoughts on how to prevent that being a problem when we no longer follow the schedule?

jobs/build/ocp4/Jenkinsfile
```diff
@@ -156,14 +156,18 @@ node {
...
+                if ( params.ASSEMBLY == "stream" ) {
+                    // Since plashets may have been rebuilt, fire off sync for CI. This will transfer
+                    // RPMs out to mirror.openshift.com/enterprise so that they may be consumed through
+                    // CI rpm mirrors.
+                    build wait: false, propagate: false, job: '/scheduled-builds/sync-for-ci', parameters: [string(name: 'ONLY_FOR_VERSION', value: params.BUILD_VERSION)]
+
+                    // Also trigger rhcos builds for the release in order to absorb any changes from plashets or
+                    // RHEL which may have triggered our rebuild. If there are no changes to the RPMs, the build
+                    // should exit quickly. If there are changes, the hope is that by the time our images are done
+                    // building, RHCOS will be ready and build-sync will find consistent RPMs.
```
sosiouxme, line 168: that seems unlikely unless we're rebuilding a whole lot of non-RHCOS images. but in any case we will get new builds a great deal sooner than if we waited for the schedule, and `ocp4_scan` will incorporate them quickly.

jupierce:
@sosiouxme I've added a lock that will let us control the number of synchronous executions by setting the resource count for a set of named locks: `lock("rhcos-build-capacity-${jobArch}")`

sosiouxme:
/hold cancel
ppc64le has been de-scheduled so this is ready
/lgtm

it does occur to me, though, that this won't reliably get us past flakes. maybe as a refinement we should invoke this job in the `ocp4_scan` branch of logic where RHCOS has updates but isn't in sync yet.

PR #2954, opened 2022-02-28
remove trim() to avoid java.lang.NullPointerException

locriandev:
This PR should fix annoying messages like this: https://coreos.slack.com/archives/C01UQM3PM3L/p1646017487642089


thiagoalessio (approved):
/lgtm

creative solution :D

PR #2956, opened 2022-03-01
buildlib: make signing advisory lock specific

sosiouxme:
i'm really tired of waiting on signing advisories when each version has its own


vfreex (approved):
/lgtm

PR #2964, opened 2022-03-03
[ART-3861] notify #art-team when buildvm is running out of space

locriandev:
The idea here is to have the scheduled job running once a day. If the partitions of interest are used for more than 90% (this can be easily tuned in the scheduled job), a notification will be sent to our private Slack channel. 

Currently, I considered these two partitions as relevant:
`
    partitions_to_check = [
        "/dev/mapper/rhel_buildvm-root",
        "/dev/mapper/workspace-workspace"
    ]
`
If more should be added, please add in the comments.


vfreex (approved):
/lgtm
though i didn't test this

PR #2965, opened 2022-03-04
Rebuild job with custom ocp-build-data

thegreyd:
This will allow us to use rebuild job with an ocp-build-data fork
useful with the process we follow - when an assembly definition
 is in a PR and is not merged and we need to rebuild a component
and put the nvr back into the assembly definition.

hack job: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Frebuild/5/console


thegreyd:
Merging to use it for 4.10.2

PR #2925, opened 2022-02-09
[ART-3705] Validate upgrade edges from prev releases.yml

thegreyd:
https://issues.redhat.com/browse/ART-3705

Test:
- [Fail] SLACK_BOT_TOKEN='4' ./artcd --config=../config/artcd.toml -C . -vv --dry-run promote -g openshift-4.9 --assembly 4.9.20 --skip-attached-bug-check --arch=x86_64
```
ValueError: `upgrades` does not contain ['4.8.31', '4.8.30'] edge(s) defined in origin/openshift-4.8:releases.yml. These versions were found to be greater than the latest previous upgrade edge 4.8.29
```
- [Success] SLACK_BOT_TOKEN='4' ./artcd --config=../config/artcd.toml -C . -vv --dry-run promote -g openshift-4.9 --assembly 4.9.21 --skip-attached-bug-check --arch=x86_64

PR #2971, opened 2022-03-08
Promote validate 2

thegreyd:
Continuing off of https://github.com/openshift/aos-cd-jobs/pull/2925
From yuxiang's comments
> We came across some issues when preparing the releases for the next week. We had to revert your upgrade edges verification PR https://github.com/openshift/aos-cd-jobs/pull/2925 because it has a few issues. e.g.
>  - When we promoting 4.6, it checks releases.yml on the 4.5 branch, but we don’t have assembly support at that time.
>  - Functions like get_release_name doesn’t work as expected. You can use existing function like get_release_name to get the release name from assembly name.
> Could you open another PR for improvements?


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -274,6 +275,30 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+            raise ValueError(f"Group config for assembly {self.assembly} is missing the required `upgrades` field. If no upgrade edges are expected, please explicitly set the `upgrades` field to an empty string.")
+        previous_list = list(map(lambda s: s.strip(), upgrades_str.split(","))) if upgrades_str else []
+        # Ensure all versions in previous list are valid semvers.
+        if any(map(lambda version: not VersionInfo.isvalid(version), previous_list)):
+            raise VerificationError("Previous list (`upgrades` field in group config) has an invalid semver.")
+
+        major, minor = util.isolate_major_minor_in_group(self.group)
+
+        # we do not have assembly releases before 4.6
+        if minor <= 6 :
```
thegreyd, line 289: Do not attempt to read assembly before 4.6

pyartcd/tests/pipelines/test_promote.py
```diff
@@ -270,3 +272,17 @@ def test_build_release_image_from_image_stream(self, cmd_assert_async: AsyncMock
...
+    def test_validate_upgrade_edges(self, cmd_assert_async: AsyncMock, get_all_assembly_semvers_for_release: AsyncMock):
+        runtime = MagicMock(config={"build_config": {"ocp_build_data_url": "https://example.com/ocp-build-data.git"}},
+                            working_dir=Path("/path/to/working"), dry_run=False)
+        assembly_name = "4.10.99"
+        pipeline = PromotePipeline(runtime, group="openshift-4.10", assembly=assembly_name, release_offset=None, arches=["x86_64", "s390x", "ppc64le", "aarch64"])
+
+        assembly_type = assembly.AssemblyTypes.STANDARD
+        upgrades_str = "4.10.98,4.10.97,4.9.99"
+
+        with self.assertRaisesRegex(VerificationError, "does not contain ['4.9.100']"):
```
thegreyd, line 287: @vfreex I cannot get this assertion right and I don't know what's wrong :/ Could you take a look?
Running with "venv/bin/python -m pytest --verbose --color=yes tests/pipelines/test_promote.py::TestPromotePipeline::test_validate_upgrade_edges"

thegreyd:
Squashed and rebased (took a while.. phew)

vfreex:
@thegreyd Could you exclude elliott and doozer commit changes from this PR?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -307,6 +308,30 @@ async def change_advisory_state(self, advisory: int, state: str):
...
+        # we do not have assembly releases before 4.6
+        if minor <= 6:
+            return previous_list
+
+        prev_assembly_semvers = await util.get_all_assembly_semvers_for_release(major, minor - 1, build_data_path)
+
+        in_previous_list = util.sorted_semver([x for x in prev_assembly_semvers if x in previous_list], reverse=True)
+        not_in_previous_list = [x for x in prev_assembly_semvers if x not in previous_list]
+        latest_prev = in_previous_list[0]
+        greater_than_latest_prev = [x for x in not_in_previous_list if semver.compare(x, latest_prev) == 1]  # if x > latest_prev
```
vfreex, line 330: This doesn't seem to be correct.

In your code:
- `previous_list` is the list of upgrade edges defined for the current release.
- `prev_assembly_semvers` is the list of all releases defined in previous version's `releases.yml`.
- `in_previous_list` is  `prev_assembly_semvers ∩ previous_list`, sorted in descending order.
- `not_in_previous_list` is `prev_assembly_semvers \ previous_list`, aka releases in previous version's releases.yml but not listed as an upgrade edge.
- `latest_prev` is the maximum in `in_previous_list`, aka the maximum common release in both `prev_assembly_semvers` and `in_previous_list`. However, this might not be the latest previous release and even not exist.
- `greater_than_latest_prev` is the list of releases that are in not_in_previous_list but is greater than prev_assembly_semvers. This doesn't seem to be reasonable.

We want to ensure every previous-minor release listed in `previous_list` doesn't contain a non-existent release in previous version.

I assume `latest_prev` means the latest release in the previous minor version. Then `latest_prev` should be the highest version in `prev_assembly_semvers` instead of `in_previous_list`.

```suggestion
        prev_assembly_semvers = map(semver.VersionInfo.parse, prev_assembly_semvers)
        latest_prev = max(prev_assembly_semvers)  # will fail if not existing
        greater_than_latest_prev = list(filter(lambda x: x.major == latest_prev.major and x.minor == latest_prev.minor and x > latest_prev, map(semver.VersionInfo.parse, previous_list)))
```

vfreex (changes requested):
It seems to have some issues in the validate function.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -307,6 +308,30 @@ async def change_advisory_state(self, advisory: int, state: str):
...
+            raise ValueError(f"Group config for assembly {self.assembly} is missing the required `upgrades` field. If no upgrade edges are expected, please explicitly set the `upgrades` field to an empty string.")
+        previous_list = list(map(lambda s: s.strip(), upgrades_str.split(","))) if upgrades_str else []
+        # Ensure all versions in previous list are valid semvers.
+        if any(map(lambda version: not VersionInfo.isvalid(version), previous_list)):
+            raise VerificationError("Previous list (`upgrades` field in group config) has an invalid semver.")
+
+        major, minor = util.isolate_major_minor_in_group(self.group)
+
+        # we do not have assembly releases before 4.6
+        if minor <= 6:
```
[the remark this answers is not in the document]
thegreyd: good point!

pyartcd/pyartcd/pipelines/promote.py
thegreyd: @vfreex it's confusing the way I wrote it, I should've left some comments to clarify.
The usecase I had in mind when writing this was, let's take 4.9 and 4.8 as examples, with 4.9.1 as the release we want to promote and suppose

```python
prev_assembly_semvers = [4.8.0, 4.8.1, 4.8.2, 4.8.3]

# 4.8.1 edge blocked due to regression so doozer doesn't output it as a suggestion
# 4.8.3 has been promoted but is not yet in fast channel
previous_list = [4.9.0, 4.8.2, 4.8.0]  # taken from `doozer release:calc-previous`

in_previous_list = [4.8.2, 4.8.0]
not_in_previous_list = [4.8.1, 4.8.3]
latest_prev = 4.8.2
greater_than_latest_prev = [4.8.3]
```

Basically I realized that `not_in_previous_list` can have blocked edges as well. Checking only those edges that are greater the last greatest edge, I'm making sure that we don't complain about previous blocked edges. This is the case when a release has been promoted, but isn't in fast channel and so doesn't show up in release:calc-previous output, which is where we've made errors in the past. 

I realize it's not super clear. I have the idea to integrate this inside release:calc-previous so it takes into account previous releases.yml and then in the validation we just run release:calc-previous again and check that output against what's here as what should be. But I think this change should be sufficient until I get that change (plus release date validation) in?

vfreex: Could you list and document the cases you'd like to capture? It could be hard to understand, review, and maintain this part without good documentation.

The point in my previous comment is about a potential logic hole in this function.
In your example, if I make a small change on `previous_list` to `[4.9.0, 4.8.2]`, `in_previous_list` will be empty and `latest_prev` will panic. If `latest_prev` donates the latest release, which doesn't have to be promoted, in the previous version, it doesn't seem to correct to calculate `latest_prev` by evaluating `max(prev_assembly_semvers ∩ previous_list)`.

vfreex: Let's also take a step back and think about the purpose of the previous list before writing the validation. It is acted as the last resort to prevent certain regressions if Cincinnati is hacked.

e.g.
- User cannot "upgrade" from `4.10.1` to `4.9.25`.
- User cannot "upgrade" from `4.9.99` to `4.10.1` if `4.9.99` is a much newer release than `4.10.1`.

PR #2989, opened 2022-03-18
[ART-3530] create an ironic plashet

sosiouxme:
paired with https://github.com/openshift/ocp-build-data/pull/1443/files this creates a `4.11-ironic` plashet next to`4.11` and `4.11-el8`.

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -1444,7 +1444,7 @@ def buildBuildingPlashet(version, release, el_major, include_embargoed, auto_sig
     def brewTag = "rhaos-${major_minor}-rhel-${el_major}-candidate"
     def embargoedBrewTag = "--embargoed-brew-tag rhaos-${major_minor}-rhel-${el_major}-embargoed"
     if (for_ironic) {
-        productVersion = "OSE-8-RHEL-${el_major}-IRONIC"
+        productVersion = "OSE-IRONIC-${major_minor}-RHEL-${el_major}"
         brewTag = "rhaos-${major_minor}-ironic-rhel-${el_major}-candidate"
         embargoedBrewTag  = ""  // unlikely to exist until we begin using -gating tag
     }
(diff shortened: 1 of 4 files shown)


joepvd:
When transitioning to el9, how likely is it that we need to have two repositories up for a brief time simultaneously? Would it make sense to namespace the repository name with the rhel-version (both in local buildvm cache and the ose repo (plashet)?

sosiouxme:
good point @joepvd i don't know that we'll need it (ironic may be able to transition all at once without too much disruption) but a little future-proofing can't hurt.

sosiouxme:
needs a good deal more supporting work in other areas (`rebuild` job, elliott etc) but I think this piece of work is ready.

pipeline-scripts/buildlib.groovy
```diff
@@ -1440,6 +1441,13 @@ def buildBuildingPlashet(version, release, el_major, include_embargoed, auto_sig
     }
 
     def productVersion = el_major >= 8 ? "OSE-${major_minor}-RHEL-${el_major}" : "RHEL-${el_major}-OSE-${major_minor}"
+    def brewTag = "rhaos-${major_minor}-rhel-${el_major}-candidate"
+    def embargoedBrewTag = "--embargoed-brew-tag rhaos-${major_minor}-rhel-${el_major}-embargoed"
+    if (for_ironic) {
+        productVersion = "OSE-8-RHEL-${el_major}-IRONIC"
```
vfreex, line 1447: I saw in the ticket the new product version name is `OSE-IRONIC-4.11-RHEL-8`. So maybe?
```suggestion
        productVersion = "OSE-IRONIC-${major_minor}-RHEL-${el_major}"
```

sosiouxme: good catch, thanks.

sosiouxme:
another good catch, it looks like we don't have auto-signing enabled for this PV yet (I attached to our signing advisory, nothing got signed).

however we don't need that before code freeze, and I don't think we need to wait on that now.

vfreex:
@sosiouxme Yeah, but the problem is setting `--arch ${release_arch} signed` without auto-signing will probably cause doozer hanging there forever.

vfreex (approved):
/lgtm

PR #2992, opened 2022-03-22
[ART-3476] sign coreos-installer

Ximinhan:
as requested from ticket, first calculate the sha256sum.txt, then use ump_producer message-digest function get the sha256sum.txt file signed


jobs/signing/sign-artifacts/umb_producer.py
```diff
@@ -287,14 +297,17 @@ def message_digest(ctx, requestor, product, request_id, sig_keyname,
...
         release_parts = release_name.split('.')
         artifact_url = MESSAGE_DIGESTS[product].format(
             arch=arch,
             release_name_xy='.'.join(release_parts[:2]),
             release_name=release_name)
-
-    artifact = get_digest_base64(artifact_url)
+        artifact = get_digest_base64(artifact_url)
+    elif product == 'coreos-installer':
+        filepath = './sha256sum.txt'
```
vfreex, line 309: Where does this file come from?

vfreex (changes requested):

vfreex:
File `sha256sum.txt` generated by `coreos-installer_sync` job can't be read by `sign-artifacts` job. Both jobs have separate workspaces on separate nodes.

I think options are:
- Use a plugin like https://www.jenkins.io/doc/pipeline/steps/copyartifact/#copyartifacts-copy-artifacts-from-another-project
- `coreos-installer_sync` calculates the base64 hash for `sha256sum.txt`, then pass the base64 hash to `sign-artifacts` by a parameter.
- Follow our current practice to download `sha256sum.txt` from mirror (https://github.com/openshift/aos-cd-jobs/blob/9b0ed349b24fd452dabcec7c6237a4c5cb646ea2/jobs/signing/sign-artifacts/umb_producer.py#L45), though this is less secure.

jobs/build/coreos-installer_sync/Jenkinsfile
```diff
@@ -83,34 +101,42 @@ node {
...
-                script: """
-                    set -euxo pipefail
-                    cd ${workdir}
-                    ssh ${mirror} mkdir -p ${dir}
-                    scp -r ${params.VERSION} ${mirror}:${dir}/${params.VERSION}
-                    ssh ${mirror} ln --symbolic --force --no-dereference ${params.VERSION} ${dir}/latest
-                    ssh ${mirror} tree ${dir}
-                    ssh ${mirror} /usr/local/bin/push.pub.sh openshift-v4/x86_64/clients/coreos-installer -v
-                """
+    stage("sign artifacts") {
```
joepvd, line 125: Should the artifacts be signed before they are put on mirror?

jobs/signing/sign-artifacts/Jenkinsfile
```diff
@@ -363,7 +377,13 @@ node {
                                 sh "aws s3 sync --no-progress --exclude='*' --include 'sha256=*' ./ s3://art-srv-enterprise/pub/rhacs/signatures/rh-acs/"
                             }
                         }
-
+                    } else if ( params.PRODUCT == 'coreos-installer' ) {
+                        sshagent(["openshift-bot"]) {
```
joepvd, line 381: sshagent is not necessary.

jobs/build/coreos-installer_sync/Jenkinsfile
```diff
@@ -83,34 +101,42 @@ node {
...
+            echo "Would have signed artifacts"
+            return
+        } else {
+            release.signArtifacts(
+                name: params.VERSION,
+                signature_name: "signature-1",
+                dry_run: params.DRY_RUN,
+                env: params.ENV,
+                key_name: params.KEY_NAME,
+                arch: "x86_64",
```
joepvd, line 136: Not sure s390x artifacts should be signed as if they were amd64?

jobs/build/coreos-installer_sync/Jenkinsfile
```diff
@@ -83,34 +101,42 @@ node {
...
+                digest: "",
```
joepvd, line 137: Digest should not be empty

jobs/build/coreos-installer_sync/Jenkinsfile
```diff
@@ -83,34 +101,42 @@ node {
...
+                client_type: "",
```
joepvd, line 138: Not sure what this is or should be, but don't think this should be empty.

jobs/signing/sign-artifacts/Jenkinsfile
```diff
@@ -363,7 +377,13 @@ node {
...
+                            withCredentials([aws(credentialsId: 's3-art-srv-enterprise', accessKeyVariable: 'AWS_ACCESS_KEY_ID', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY')]) {
+                                sh "aws s3 sync --no-progress --exclude='*' --include 'sha256sum.txt.gpg' ./ s3://art-srv-enterprise/pub/openshift-v4/${params.ARCH}/clients/coreos-installer/${params.NAME}/"
```
joepvd, line 383: This is not where currently coreos-installer is: https://mirror.openshift.com/pub/openshift-v4/s390x/clients/

It is directly under clients, the separation of different arches is in the folder for the version: https://mirror.openshift.com/pub/openshift-v4/clients/coreos-installer/v0.13.1-1/

jobs/build/coreos-installer_sync/Jenkinsfile
joepvd: Looks like other jobs loop over artifacts by architecture, and sign that. Not sure if there is something very different from signing the `oc` client?

joepvd (changes requested):
Apart from some nits, think we need to figure out how to deal with different arches and locations on mirror.

jobs/build/coreos-installer_sync/Jenkinsfile
Ximinhan: the sign artifacts job is used to sign sha256sum.txt and we need to put it on mirror first

jobs/build/coreos-installer_sync/Jenkinsfile
Ximinhan: from sign artifacts job's logic the arch parameter is used to point which arch under https://mirror.openshift.com/pub/openshift-v4/{arch} for coreos-installer it only needed under x86_64

jobs/build/coreos-installer_sync/Jenkinsfile
Ximinhan: sign sha256sum don't need digest

jobs/build/coreos-installer_sync/Jenkinsfile
Ximinhan: same as above

jobs/signing/sign-artifacts/Jenkinsfile
Ximinhan: we only need push clients to https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/coreos-installer/

jobs/build/coreos-installer_sync/Jenkinsfile
```diff
@@ -42,6 +43,23 @@ node {
...
+                ),
+                choice(
+                    name: 'KEY_NAME',
+                    description: 'Which key to sign with\nIf ENV==stage everything becomes "test"\nFor prod we currently use "redhatrelease2"',
+                    choices: [
+                        "test",
+                        "beta2",
+                        "redhatrelease2",
+                    ].join("\n"),
+                ),
```
vfreex, line 62: I don't think those low-level signing parameters should be exposed to the user. Based on the requirement, this should always be signed with `redhatrelease2` key.

jobs/signing/sign-artifacts/umb_producer.py
```diff
@@ -43,7 +43,8 @@
 # hard-coded into the URL path.
 MESSAGE_DIGESTS = {
     'openshift': 'https://mirror.openshift.com/pub/openshift-v4/{arch}/clients/{release_stage}/{release_name}/sha256sum.txt',
-    'rhcos': 'https://mirror.openshift.com/pub/openshift-v4/{arch}/dependencies/rhcos/{release_name_xy}/{release_name}/sha256sum.txt'
+    'rhcos': 'https://mirror.openshift.com/pub/openshift-v4/{arch}/dependencies/rhcos/{release_name_xy}/{release_name}/sha256sum.txt',
+    'coreos-installer': 'https://mirror.openshift.com/pub/openshift-v4/{arch}}/clients/coreos-installer/{release_name}/sha256sum.txt'
```
vfreex, line 47: `}}` -> `}`

vfreex (changes requested):

jobs/signing/sign-artifacts/Jenkinsfile
Ximinhan: I saw it used in other product, I don't know if we need to prune them all

jobs/build/coreos-installer_sync/Jenkinsfile
Ximinhan: updated with limited choise

jobs/build/coreos-installer_sync/Jenkinsfile
```diff
@@ -83,22 +98,47 @@ node {
...
+        if (params.DRY_RUN) {
+            echo "Would have signed artifacts"
             return
         }
+        release.signArtifacts(
+            name: params.VERSION,
+            signature_name: "signature-1",
+            dry_run: params.DRY_RUN,
+            env: params.ENV,
+            key_name: params.KEY_NAME,
```
vfreex, line 132: I think parameters `ENV` and `KEY_NAME` need to be removed. They shouldn't be specified when running the job.

joepvd:
Guess this is not the ticket to change the directory structure of this artifact, but it is weird that the s390x artifacts end up in the x86_64 directory.

I think the pattern of uploading to mirror from job 1, and then downloading from mirror from the signing job to compute sha's is not a great pattern. I'd much rather see that the shasum of the shasum256.txt file gets passed along as a parameter to the signing job. I would not be surprised if it would be a requirement that the signing job does not talk to the internet.

vfreex:
Agreed with @joepvd that the current layout of the directory is confusing: all binaries are synced to the x86_64 directory https://mirror.openshift.com/pub/openshift-v4/clients/coreos-installer/latest, which can be resolved to https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/coreos-installer/latest/.

If this PR needs to split the extracted binaries into their corresponding arch-specific directories before syncing to the mirror, we need to make sure the owner knows the change so things will not be broken.

Uploading and downloading sha256sum.txt is definitely not ideal and not secure. I remember we have a JIRA to improve that.

vfreex (approved):
I think I will LGTM for now. However we didn't see this job is tested in a hack space 🤷 

/lgtm

Ximinhan:
I merged it for now although coreos-installer is not updated very often

PR #2999, opened 2022-03-24
ocp4_scan+rhcos: rebuild when inconsistent

sosiouxme:
right now, if an rhcos build does not succeed for all arches, it is not
retried and remains inconsistent until the next time ocp4 kicks off an
rhcos build when something else changes. rather than just blindly retry
failed arches (which could delay picking up other changes), we would
like ocp4_scan to kick off a build for all when it sees they are
inconsistent and there's no other reason to run ocp4.

this would not retry if _all_ rhcos builds attempted failed (no
inconsistency), but that seems unlikely to occur enough to be a problem.

### DIFF
--- a/jobs/build/ocp4_scan/Jenkinsfile
+++ b/jobs/build/ocp4_scan/Jenkinsfile
@@ -175,8 +175,20 @@ timeout(activity: true, time: 30, unit: 'MINUTES') {
                                         )
                                         echo stdout
                                     } catch(err) {
-                                        echo "Error checking for INCONSISTENT_RHCOS_RPMS:\n${err}"
-                                        echo "RHCOS builds inconsistent, not triggering build-sync"
+                                        echo "INCONSISTENT_RHCOS_RPMS:\n${err}"
+                                        echo "RHCOS builds inconsistent, not triggering build-sync."
+                                        // inconsistency probably means partial failure and we would like to retry.
+                                        // but don't kick off more if already in progress.
+                                        lock(resource: "rhcos-lock-${version}", skipIfLocked: true) {
+                                            echo "triggering a ${version} RHCOS build for consistency."
+                                            build(
+                                                wait: false,
+                                                propagate: false,
+                                                job: 'build%2Frhcos',
+                                                parameters: [string(name: 'BUILD_VERSION', value: version)]
+                                            )
+                                            currentBuild.description += "<br>triggered rhcos build: ${version}"
+                                        }
                                         return
                                     }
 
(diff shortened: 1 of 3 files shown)


sosiouxme:
No good way to test the two jobs but they at least compile and run...

jobs/build/ocp4_scan/Jenkinsfile
```diff
@@ -175,8 +175,19 @@ timeout(activity: true, time: 30, unit: 'MINUTES') {
...
+                                        echo "INCONSISTENT_RHCOS_RPMS:\n${err}"
+                                        echo "RHCOS builds inconsistent, not triggering build-sync."
+                                        // inconsistency probably means partial failure and we would like to retry.
+                                        // but don't kick off more if already in progress.
+                                        lock(resource: "rhcos-lock-${version}", skipIfLocked: true) {
+                                            echo "triggering a ${version} RHCOS build for consistency."
+                                            build(
+                                                wait: false,
+                                                propagate: false,
+                                                job: '/aos-cd-builds/build%2Frhcos',
```
thegreyd, line 187: to be consistent with how we trigger jobs, `job: 'build%2Frhcos'`

jobs/build/ocp4_scan/Jenkinsfile
```diff
@@ -175,8 +175,19 @@ timeout(activity: true, time: 30, unit: 'MINUTES') {
                                         )
                                         echo stdout
                                     } catch(err) {
-                                        echo "Error checking for INCONSISTENT_RHCOS_RPMS:\n${err}"
-                                        echo "RHCOS builds inconsistent, not triggering build-sync"
+                                        echo "INCONSISTENT_RHCOS_RPMS:\n${err}"
```
thegreyd, line 178: Nice bit of improved messaging

jobs/build/ocp4_scan/Jenkinsfile
```diff
@@ -175,8 +175,19 @@ timeout(activity: true, time: 30, unit: 'MINUTES') {
...
+                                        lock(resource: "rhcos-lock-${version}", skipIfLocked: true) {
+                                            echo "triggering a ${version} RHCOS build for consistency."
+                                            build(
+                                                wait: false,
+                                                propagate: false,
+                                                job: '/aos-cd-builds/build%2Frhcos',
+                                                parameters: [string(name: 'BUILD_VERSION', value: version)]
+                                            )
+                                        }
                                         return
```
thegreyd, line 191: `currentBuild.description += "<br>triggered rhcos build: ${version}"` would be nice

jobs/build/ocp4_scan/Jenkinsfile
sosiouxme: won't work with my hack jobs but yeah... sounds good. actually i'll go back and change the place in `ocp4` that i copied it from, since everywhere else i can see does not fully qualify unless necessary.

sosiouxme:
gonna merge then and see how it goes

PR #3003, opened 2022-03-28
[ART-3908] refactor check bugs

locriandev:


pyartcd/pyartcd/pipelines/check_bugs.py
```diff
@@ -0,0 +1,230 @@
...
+        # Find bugs
+        cmd = [
+            ELLIOTT_BIN,
+            f'--group=openshift-{version}',
+            f'--working-dir={version}-working',
+            'find-bugs',
+            '--mode=sweep'
+        ]
+        self.logger.info(f'Executing command: {" ".join(cmd)}')
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
```
joepvd, line 157: Having stderr when it fails would be great as well

pyartcd/pyartcd/pipelines/check_bugs.py
```diff
@@ -0,0 +1,230 @@
+import os
```
joepvd, line 1: some tests would be great

pyartcd/pyartcd/pipelines/check_bugs.py
```diff
@@ -0,0 +1,230 @@
...
+            'find-bugs',
+            '--mode=blocker',
+            '--report',
+            '--output=slack'
+        ]
+        self.logger.info(f'Executing command: {" ".join(cmd)}')
+
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+        out, _ = process.communicate()
+        errcode = process.returncode
```
joepvd, line 124: mark the job as failed if one or more things fail. then we know we have to look.
optional, a retry of a failed command to reduce flakes. but we can get to that if there would be too many flakes

joepvd (commented):
Nice! Like this direction!

pyartcd/pyartcd/pipelines/check_bugs.py
locriandev: done here: https://github.com/openshift/aos-cd-jobs/pull/3003/commits/7e42210bf38259cb2f3bc265ac07f4855700a287

pyartcd/pyartcd/pipelines/check_bugs.py
locriandev: Done here: https://github.com/openshift/aos-cd-jobs/pull/3003/commits/281e546c8bd9475dcb2d19aa2d7f8cfbe46d0f1c
I'd leave retries for later, if we see we actually need this

pyartcd/pyartcd/pipelines/check_bugs.py
locriandev: Done here: https://github.com/openshift/aos-cd-jobs/pull/3003/commits/4f6bd28a0cd4a104b3807e5cb38bad4a26f87492

PR #3007, opened 2022-03-29
[ART-3397] Prune use-mirror from other jobs

joepvd:


vfreex (approved):
/lgtm
I hope this is tested :)

PR #3017, opened 2022-04-04
Remove spamming notifications about failed nightlies

locriandev:
Recently, repeated failures in nightlies have triggered tons of notifications on various slack channels. See [here](https://coreos.slack.com/archives/CJARLA942/p1648851792500539) for examples.
This is frustrating our attempts to have useful, easy to find warnings to be used by release artists, patch managers, and the like.

This PR removes these messages, and is meant to be the starting point of a discussion that will eventually entail a more efficient way to schedule and handle these checks.

### DIFF
--- a/jobs/build/build-sync/Jenkinsfile
+++ b/jobs/build/build-sync/Jenkinsfile
@@ -143,16 +143,6 @@ node {
         }
         writeFile file: failCountFile, text: "${failCount}"
 
-        if (failCount > 1) {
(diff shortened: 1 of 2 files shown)


thiagoalessio:
🤔 I agree that it is very annoying, but maybe, instead of completely getting rid of them, we could make the bot always post in the same thread for the day. WDYT?

locriandev:
links to https://issues.redhat.com/browse/ART-3953

PR #3019, opened 2022-04-04
ocp4_scan: Don't assume consistent rhcos because nightly exists

joepvd:
We temporarily relaxed the requirement to have consistent RHCOS rpms because of
permanent build failures. Despite the presence of inconsistencies in the
rhcos-es, no rhcos build was triggered.

This change removes the assumption that RHCOS is consistent just because
a nightly exists. It will query the consistency, and, if not consistent,
trigger an RHCOS build.


thiagoalessio (approved):
/lgtm

Sweet!

PR #3027, opened 2022-04-12
[ART-3962] Handle Dockerfile is not found on public branch

locriandev:


scheduled-jobs/build/sync-ci-images/Jenkinsfile
```diff
@@ -124,8 +124,9 @@ node {
                                 }
                                 // Only open PRs on >= 4.6 to leave well enough alone.
                                 rc = sh script:"${buildlib.DOOZER_BIN} --assembly stream --working-dir ${doozerWorking} --group ${group} images:streams prs open --interstitial 840 --add-auto-labels ${other_args} --github-access-token ${GITHUB_TOKEN}", returnStatus: true
+                                // rc=50 is used to indicate some errors raised during PR openings                                
                                 // rc=25 is used to indicate PR openings were skipped and to try again later.
-                                if ( rc != 0 && rc != 25) {
+                                if ( rc != 0 && rc != 25 && rc != 50) {
```
sosiouxme, line 129: hm. i'm not thrilled about expecting RC codes to actually come through correctly, or about making them up... should probably be real POSIX status codes. but not feeling too pedantic today, let's get this working.

sosiouxme (approved):

scheduled-jobs/build/sync-ci-images/Jenkinsfile
locriandev: I actually agree, but couldn't find any reasonable combinations of RC to be used... I wanted to leave RC=1 free, so that we can be warned when not handled exceptions were raised. And it seems like there's no better way than using RCs to send feedback from Python scripts to Jenkins pipelines :( 
We could still find more appropriate RCs, taking from [here](https://tldp.org/LDP/abs/html/exitcodes.html). What would you suggest?

PR #3038, opened 2022-04-18
Setup CI to run unit tests for pyartcd

vfreex:


pyartcd/pyartcd/git.py
```diff
@@ -0,0 +1,20 @@
+
```
joepvd, line 1: empty line

pyartcd/pyartcd/git.py
```diff
@@ -0,0 +1,20 @@
...
+import os
+from pathlib import Path
+import shutil
+from typing import Optional
+from pyartcd.exectools import cmd_assert_async
```
joepvd, line 6: Would love to see pygit2 rather than the git command in a subshell.

tekton-pipelines/images/art-tools-tester/Containerfile
```diff
@@ -0,0 +1,53 @@
...
+
+# Download oc
+ARG OC_VERSION=latest
+RUN wget -O openshift-client-linux-"$OC_VERSION".tar.gz https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/"$OC_VERSION"/openshift-client-linux.tar.gz \
+  && tar -xzvf openshift-client-linux-"$OC_VERSION".tar.gz oc kubectl
+
+# Download Tekton CLI
+ARG TKN_VERSION=0.23.1
+RUN wget -O "tkn_${TKN_VERSION}_Linux_x86_64.tar.gz" https://github.com/tektoncd/cli/releases/download/v0.23.1/"tkn_${TKN_VERSION}_Linux_x86_64.tar.gz" \
+  && tar -xzvf ./tkn_0.23.1_Linux_x86_64.tar.gz tkn
```
joepvd, line 18: `TKN_VERSION` here as well.
Any reason we're not pulling from http://mirror.openshift.com/pub/openshift-v4/x86_64/clients/pipeline/?

joepvd (commented):
Very nice!

Any particular reason the k8s entities live here rather than in `openshift-eng/art-config`?

pyartcd/pyartcd/git.py
vfreex: Nice spot. I think a linter like flake8 needs to be added to the test. For now I would like to hold on that because it will complain a lot. So better to do that in a separate PR.

vfreex: Added flake8 test.

PR #3041, opened 2022-04-19
Fix sweep.py error codes

locriandev:
If for some reason any shell command fails, the script would exit with 1. This is treated by the Jenkins pipeline as BUILD_NOT_PERMITTED instead of RUNTIME_ERROR, causing misleading information in the job title and description. Changing return code values will fix this issue.


joepvd (approved):
/lgtm

PR #3043, opened 2022-04-19
Fix venv setup

locriandev:
Calling ```pip install -r <art-tool>/requirements.txt``` is dangerous, as it does not necessarily install latest submodule code. It can either pull from PyPI, or use a cached version of the tool if the virtual env folder wasn't wiped out. Using ```pip install -e``` makes sure latest ```setup.py``` is called for each tool.

Pyartcd can also be installed within the same folder, instead of delegating to each pipeline that's using it. This ensures consistency and avoids code duplication.

By removing the ```try-catch``` block that surrounds the virtualenv setup, we make sure that if errors are raised the job will stop, instead of running with outdated versions of our tooling (buildvm has old versions of doozer and elliott installed at system level, that will serve as unwanted fallback if the virtualenv is not created successfully)

### DIFF
--- a/jobs/build/check-bugs/Jenkinsfile
+++ b/jobs/build/check-bugs/Jenkinsfile
@@ -36,9 +36,6 @@ node {
     // Check for mock build
     commonlib.checkMock()
 
-    // Install pyartcd
-    commonlib.shell(script: "pip install --no-cache-dir --upgrade -e ./pyartcd")
-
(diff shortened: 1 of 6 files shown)


thegreyd:
Changes lgtm, to be safe could you dry-run (for example sweep job) on a hack branch?

vfreex (approved):

PR #3049, opened 2022-04-21
[ART-3915] ocp4: serialize mass rebuilds

sosiouxme:
as a simple crude measure, keep from flooding the queue with multiple mass rebuilds at the same time (prevent long delays for other tenants and any versions that aren't mass rebuilding).

this doesn't address any of the other ideas in that Jira which would probably be more effective; it's just a stopgap because we have a lot of mass rebuilds.

a [sanity test run](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-stage/job/build%252Focp4/50/console) went well.

### DIFF
--- a/jobs/build/ocp4/build.groovy
+++ b/jobs/build/ocp4/build.groovy
@@ -18,6 +18,7 @@ allImagebuildfailed = false
 
 // this plan is to be initialized but then adjusted for incremental builds
 buildPlan = [
+    activeImageCount: 1, // number of images active in this version
     dryRun: false, // report build plan without performing it
     forceBuild: false, // build regardless of whether source has changed
     buildRpms: false,
@@ -52,6 +53,12 @@ def initialize() {
     if (params.ASSEMBLY != 'stream' && buildlib.doozer("${doozerOpts} config:read-group --default=False assemblies.enabled", [capture: true]).trim() != 'True') {
         error("ASSEMBLY cannot be set to '${params.ASSEMBLY}' because assemblies are not enabled in ocp-build-data.")
     }
+
+    buildPlan.activeImageCount = (
+        buildlib.doozer("${doozerOpts} images:list", [capture: true])
+        .trim().split("\n")[-1]  // last line e.g. "219 images"
+        .split(" ")[0].toInteger()
+    )
 
     version.branch = buildlib.getGroupBranch(doozerOpts)
     version << determineBuildVersion(version.stream, version.branch)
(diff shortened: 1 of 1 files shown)


jobs/build/ocp4/build.groovy
```diff
@@ -340,11 +347,29 @@ def stageBuildImages() {
...
             """
+
+        def includedCount = commonlib.parseList(buildPlan.imagesIncluded).size()
+        def excludedCount = commonlib.parseList(buildPlan.imagesExcluded).size()
+        def massRebuild = (
+            includedCount && buildPlan.activeImageCount < includedCount * 2 ||  // includes more than half
+            excludedCount && buildPlan.activeImageCount > excludedCount * 2 ||  // excludes less than half
+            !(includedCount || excludedCount)  // rebuilding everything
+        )
+        if (massRebuild) currentBuild.description += "Mass image rebuild (more than half) - invoking serializing semaphore<br/>"
```
joepvd, line 358: Think `description` generally is too long already, the most important parts fall off the visible screen. Let's not use description as log line :P 

Think it'd suffice with `(mass)` in description after `4.11`.

joepvd (approved):
Nice!

One concern I have is that in this proposal it is random which version gets the mass rebuild semaphore. As we currently have 4 versions one one rhel version, it could take a while before the version we care about most gets rebuilt.

One simplistic approach would be to un-parallellize ocp4_scan, so versions get scanned in decreasing order. This will come at a cost of triggering product builds less often.

If the organization is waiting on a new build or nightly, it never is because ocp4_scan is waiting to get triggered. These potentially 30 mins extra delay would not hurt very much in the bottom line (because other potential delays are a magnitude larger).

jobs/build/ocp4/build.groovy
sosiouxme: since the title says `[200 images]` (or whatever) it's already pretty obvious which is a mass rebuild usually, so this was more of a logging statement (esp. for the dry run). That's how I think of the description anyway, as sort of a condensed summary of the build...

In a real run of this, I guess the real question we'd want answered when looking at it is "why is this build taking forever" and the console log would make that pretty evident since it would say "waiting to get a lock" so maybe there's no need for any sort of notice.

sosiouxme:
I think the randomness of this occasional process is more acceptable than a universal delay in the pipeline... but if we'd like to bias in favor of running the newer releases first, maybe have `ocp4_scan` run each with an incremental 1-or-2-minute delay. this could backfire though, as whatever change is triggering multiple mass rebuilds could easily land midway through the scans, in which case the older releases will still run first.

Also a useful question: is it better to have this random serialization than none at all? :)

PR #3050, opened 2022-04-21
Add banner to all jobs regarding Jenkins shutdown August 2022

DennisPeriquet:
Hopefully this message will remind anyone using the UI to plan ahead.

This has been rolled out Apr 21, 2022 ~6:15pm EST.

### DIFF
--- a/sjb/generate.py
+++ b/sjb/generate.py
@@ -257,7 +257,7 @@ if "test" in job_config:
 if action != None:
     debug("[INFO] Marking this as a {} job for the {} repo".format(action, target_repo))
 
-DEFAULT_DESCRIPTION = "<div style=\"font-size: 32px; line-height: 1.5em; background-color: yellow; padding: 5px;\">" + \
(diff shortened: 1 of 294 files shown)


jupierce:
@DennisPeriquet - I don't want to bike shed you as this is already above and beyond. But as a user of this system, my eye is likely trained to completely ignore this yellow banner. Could we add a second, independent div, red, larger font?

dperique:
Good idea. Let me see about how to do that.

I looked at that message and it didn't quite jump out at me either.

sjb/generate.py
```diff
@@ -258,6 +258,7 @@ def parse_action(action):
     debug("[INFO] Marking this as a {} job for the {} repo".format(action, target_repo))
 
 DEFAULT_DESCRIPTION = "<div style=\"font-size: 32px; line-height: 1.5em; background-color: yellow; padding: 5px;\">" + \
```
jupierce, line 260:
```suggestion
DEFAULT_DESCRIPTION = "<div style=\"font-size: 32px; line-height: 1.5em; background-color: red; padding: 5px;\">*** THIS SERVER WILL BE SHUTDOWN AUGUST 2022 ***</div><div style=\"font-size: 16px; line-height: 1.5em; background-color: yellow; padding: 5px;\">" + \
```

jupierce: Mostly the same, just break out into another div.

DennisPeriquet: thanks
Screen Shot 2022-04-21 at 6 38 30 PM

jupierce (approved):

jupierce:
Thanks, @DennisPeriquet!

PR #3058, opened 2022-05-05
Add Doozer data path param to build olm bundles

locriandev:
Both OCP4 and custom jobs allow us to specify DOOZER_DATA_PATH to test customization on our forks. This is not currently handled by the `olm_bundle` job that gets triggered afterwards, resulting in possible inconsistencies. As an example, [this build](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Folm_bundle/7534/console) failed because it picked `openshift/ocp-build-data` where the operator was disabled for 4.11.

[This one](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/dpaolell/job/olm_bundle/5/console) instead succeeded as it was pointed to the right fork

### DIFF
--- a/jobs/build/olm_bundle/Jenkinsfile
+++ b/jobs/build/olm_bundle/Jenkinsfile
@@ -42,6 +42,12 @@ pipeline {
             trim: true,
         )
         string(
+            name: 'DOOZER_DATA_PATH',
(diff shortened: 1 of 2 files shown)


thiagoalessio (approved):
/lgtm

PR #3059, opened 2022-05-05
Forward doozer data path to build-sync and olm_bundle jobs

locriandev:
Follows #3058

Now that `olm_bundle` has a `DOOZER_DATA_PATH` param, we can forward from `ocp4` and `custom` jobs that trigger it.
Note that the function `buildlib.sync_images` is called only by these 2 jobs.


thiagoalessio (approved):
/lgtm

PR #3065, opened 2022-05-18
Add quay addresses to no_proxy

joepvd:
List has been taken from https://docs.openshift.com/container-platform/4.10/installing/install_config/configuring-firewall.html


thiagoalessio (approved):
/lgtm

PR #3071, opened 2022-05-25
[ART-4061] check jira env when trigger elliott

Ximinhan:


vfreex (changes requested):
If I set `USE_JIRA`, Bugzilla bugs will be totally ignored.

I remembered on our arch call we decided to use one single elliott verb to do with both systems. e.g. Running elliott find-bugs will sweep both JIRA and Bugzilla bugs.

vfreex:
A dedicated parameter has been added to jenkins job. This doesn't seem to be needed anymore.

PR #3079, opened 2022-06-02
Allow default advisory value for candidate assembly

thegreyd:
We don't create advisories when preparing FCs. To manually set advisory values to something else than -1 (like 0) is an extra step that is unnecessary. Turn this off for candidate


sosiouxme (approved):
/lgtm
we do want advisories for RCs but i don't think there'll be a problem remembering those.

PR #3089, opened 2022-06-06
add quiet period for health job

Ximinhan:
Every Monday, our healthjob always gets stuck or fails when querying pr. Although I increased the waiting time in the doozer function, it doesn't seem to have the expected effect. I want to try to increase the interval between each job


Ximinhan:
`quietPeriod` currently not supported by our multi-branch pipeline https://issues.jenkins.io/browse/JENKINS-37588?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

PR #3091, opened 2022-06-07
build/rhcos: vary cluster resource limits

sosiouxme:
we had rhcos builds today that could have run but were waiting because two rhcos builds were already in progress.
the limit is really only 2 for ppc, the rest can handle more.


sosiouxme:
:shipit:

PR #3101, opened 2022-06-15
ocp4: Build any necessary bundles

joepvd:
In case an ocp4 build gets interrupted after the operator was built, but
before the bundle build completed, the build of the bundle will not be
retried.

This commit changes that behavior, it calls olm_bundle indiscriminately
from the ocp4 job, picking up if a previous bundle was not attempted.
This mimics the approach in pyartcd prepare-release, where the doozer
bundle command gets called for all bundles (and in the general case has
othing to do).

The custom job also calls olm_bundle. The behavior of being selective
has not been changed, as a targeted approach is generally desired
behavior there.

### DIFF
--- a/jobs/build/ocp4/build.groovy
+++ b/jobs/build/ocp4/build.groovy
@@ -448,21 +448,11 @@ def stageSyncImages() {
         return
     }
 
-    def record_log = buildlib.parse_record_log(doozerWorking)
-    def records = record_log.get('build', [])
-    def operator_nvrs = []
-    for (record in records) {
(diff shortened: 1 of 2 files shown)


thiagoalessio (approved):
/lgtm

vfreex (commented):
The reason why operator NVRs are explicitly passed to olm_bundle job is because we may rebuild operator for an assembly that has a basis event. In that case, operators will not be included to the assembly until they are pinned. Removing that will prevent them from building.

But, prepare-release will take care of it anyway. Not sure what the best approach is here... Maybe we can set up a scheduled job to periodically trigger a olm_bundle build?

PR #3113, opened 2022-06-21
Add optional parameter 'https-proxy' to covscan:images

locriandev:
Linked to https://github.com/openshift/doozer/pull/608


jobs/scanning/covscan-images/Jenkinsfile
```diff
@@ -68,6 +68,12 @@ timeout(activity: true, time: 1, unit: 'DAYS') {  // if there is no log activity
                             defaultValue: false,
                             description: "Force rebuild of the covscan repositories."
                         ),
+                        [
+                            name: 'HTTPS_PROXY',
```
vfreex, line 72: It will be a burden for user to remember which proxy is current and set this build parameter for every job run.
Suggest removing this parameter, but read the env var from the build env.

vfreex (changes requested):

jobs/scanning/covscan-images/Jenkinsfile
locriandev: Done

PR #3120, opened 2022-06-27
Revert "Add ~/.local/bin to PATH"

vfreex:
Reverts openshift/aos-cd-jobs#3118


vfreex:
It is now configured at https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/computer/(built-in)/.

PR #3121, opened 2022-06-28
set prepare release subtask done after create jira

Ximinhan:
automatically update prepare release subtask status to reduce manually update

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -186,6 +186,13 @@ class PrepareReleasePipeline:
             jira_issues = self.create_release_jira(jira_template_vars)
             jira_issue = jira_issues[0] if jira_issues else None
             jira_issue_key = jira_issue.key if jira_issue else None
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -186,6 +186,13 @@ async def run(self):
             jira_issues = self.create_release_jira(jira_template_vars)
             jira_issue = jira_issues[0] if jira_issues else None
             jira_issue_key = jira_issue.key if jira_issue else None
+            subtask = self._jira_client.get_issue(jira_issue.fields.subtasks[1].key)
+            self._jira_client.add_comment(
+                    subtask,
+                    "prepare release job : {}".format(os.environ.get("BUILD_URL")),
+                    visibility={'type': 'group', 'value': 'Red Hat Employee'}
+            )
+            self._jira_client.transition_issue(subtask, '21')
```
joepvd, line 195: Would be good to un-magic string `21`.

Maybe store it in a variable with a good name, a link to where other numbers can be found?

Ximinhan: updated

PR #3122, opened 2022-06-28
prepare-release: recognize placeholder JIRA key ART-0

vfreex:


vfreex:
Requires https://github.com/openshift/doozer/pull/614

PR #3151, opened 2022-07-27
Update prevalidation tag name for ironic in 4.12 and higher

elfosardo:
Prevalidation tag name was changed to match root rhaos-{version}-ironic-rhel


sosiouxme:
closing in favor of https://github.com/openshift/aos-cd-jobs/pull/3152

PR #3155, opened 2022-07-28
Log command invocations in promote-assembly

thegreyd:
I noticed that we weren't logging commands that were being run
as we do in prepare-release.
This helps in debugging when such commands fail.


vfreex:
`cmd_assert_async` and `cmd_gather_async` are already logging the command with `debug`. Maybe change the logging level with `info` if needed.

thegreyd:
Done

PR #3157, opened 2022-07-29
ART-3761 Use release_date from assembly definition if present

thegreyd:
https://issues.redhat.com/browse/ART-3761
Summary
- Update gen-assembly job to accept a date (required=False). Needs https://github.com/openshift/doozer/pull/626 to work
- Update prepare-release job to read release_date from assembly if present
    - It will raise an error if given date and assembly date don't match
    - If date is given through the job itself, record it in assembly definition
    - Try and update advisories with release date (using elliott advisory-commons). This will let us change assembly definition and run prepare-release to update dates in advisories.

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -744,7 +744,7 @@ update JIRA accordingly, then notify QE and multi-arch QE for testing.""")
               help="don't create advisories/jira; pick them up from ocp-build-data")
 @pass_runtime
 @click_coroutine
-async def rebuild(runtime: Runtime, group: str, assembly: str, name: Optional[str], date: Optional[str],
+async def prepare_release(runtime: Runtime, group: str, assembly: str, name: Optional[str], date: Optional[str],
                   package_owner: Optional[str], nightlies: Tuple[str, ...], default_advisories: bool):
     # parse environment variables for credentials
     jira_token = os.environ.get("JIRA_TOKEN")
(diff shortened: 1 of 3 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -377,7 +415,7 @@ async def clone_build_data(self, local_path: Path):
         await exectools.cmd_assert_async(cmd)
 
     async def update_build_data(self, advisories: Dict[str, int], jira_issue_key: Optional[str]):
-        if not advisories and not jira_issue_key:
+        if not (advisories or jira_issue_key or self.release_date):
```
ashwindasr, line 418: Just to confirm, we are returning `False` only if all 3 values, `advisories`, `jira_issue_key` and `self.release_date `are None/Empty?

ashwindasr (changes requested):

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -204,10 +219,12 @@ async def run(self):
                 )
                 self._jira_client.close_task(subtask)
 
-
         _LOGGER.info("Updating ocp-build-data...")
         build_data_changed = await self.update_build_data(advisories, jira_issue_key)
 
+        _LOGGER.info("Try and update advisory release date in case it changed...")
+        self.update_advisory_release_date()
```
vfreex, line 226: No need to call this function if the advisories are just created.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -696,8 +735,8 @@ def send_notification_email(self, advisories: Dict[str, int], jira_link: str):
               help="The name of an assembly to rebase & build for. e.g. 4.9.1")
 @click.option("--name", metavar="RELEASE_NAME",
               help="release name (e.g. 4.6.42)")
-@click.option("--date", metavar="YYYY-MMM-DD", required=True,
-              help="Expected release date (e.g. 2020-11-25)")
+@click.option("--date", metavar="YYYY-Mon-DD", required=False,
```
vfreex, line 738: I think we shouldn't allow users to specify release date with `--data` anymore if it is already defined in releases.yml. Would like to completely remove this option.

vfreex (changes requested):

PR #3161, opened 2022-08-04
Increase promote job wait_for_stable time to 12 hours with 10m intervals

thegreyd:
Right now wait_for_stable is not effective since rarely 
a release is accepted on RC in 3 hours. 3 hours is too 
short of a window to be useful. Extending the time to 
12 hours, with 10 min intervals (instead of 5), 
would be more useful.


vfreex (approved):
/lgtm

thiagoalessio (approved):
/lgtm

PR #3168, opened 2022-08-11
Update latest symlinks for 4.11 multi releases

vfreex:
4.11 multi releases have `-multi` in their names:
```
4.11.0-rc.6 ==> 4.11.0-multi-rc.6
4.11.0 ==> 4.11.0-multi
```

Because `-multi` release names don't get into Cincinnati,
we query Cincinnati graph using amd64 arch and only update the symlink
if the multi client binaries are published to mirror.


jupierce (approved):
lgtm. Thanks @vfreex !

PR #3176, opened 2022-08-18
Check for namespace existence before running get_image_stream()

locriandev:


vfreex:
I don’t think it is needed. The command will fail anyway.
Maybe just removing stderr=None from cmd_gather_async call to make the stderr  included in the exception message.

locriandev:
> I don’t think it is needed. The command will fail anyway. Maybe just removing stderr=None from cmd_gather_async call to make the stderr included in the exception message.

sounds reasonable, I updated the PR with this suggestion

PR #3181, opened 2022-08-23
Stop producing 4.7

joepvd:


joepvd:
/hold
Looks like 4.7 needs to continue to build to facilitate upgrades from 4.6 EUS, regardless of whether the release is supported.

PR #3182, opened 2022-08-23
WIP gen-assembly can create PR and run build-sync

thegreyd:
See https://github.com/openshift/aos-cd-jobs/pull/3163 for previous work on this

## Current Status
- [Job run](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Fgen-assembly/2/console)
- PR https://github.com/openshift/ocp-build-data/pull/1885

## Todo
- [ ] Display doozer log when the command is running.
- [ ] Use doozer -o flag to output to the file and read from that file
- [ ] Mention gen-assembly job url in PR description
- [ ] Gen-assembly can run build-sync (re https://github.com/openshift/aos-cd-jobs/pull/3183)

PR #3183, opened 2022-08-23
Build sync accepts doozer data path gitref

thegreyd:
This will help us run build-sync with an automated PR like this https://github.com/openshift/ocp-build-data/pull/1885
https://github.com/openshift/aos-cd-jobs/pull/3182

[Test run](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Fbuild-sync/8/console)

### DIFF
--- a/jobs/README.md
+++ b/jobs/README.md
@@ -79,6 +79,11 @@ This is so standard that when provided, `buildlib.doozer` and
 Fork of `ocp-build-data` to use instead of the [default](https://github.com/openshift/ocp-build-data)
 (e.g. when you have an assembly definition in your own fork).
 
+### DOOZER\_DATA\_GITREF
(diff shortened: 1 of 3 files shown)


vfreex (commented):
Having build-sync run against an arbitrary branch could be useful for testing.
 
But I have a concern about running build-sync against a PR. build-sync actually touches and manipulates our production data. Having build-sync run against an unmerged PR might lead to a security vulnerability that an attacker can create a PR to compromise our production data and services. Would like to hold on the build-sync change until we get more ideas about it.

vfreex:
@thegreyd Sounds good to me.
/lgtm

PR #3195, opened 2022-08-29
rhcos_sync: use installer data instead of m-os-c

sosiouxme:
We recently recognized that we really need to be getting the default
RHCOS bootimages according to the installer data file (to match what the
installer downloads by default), not according to machine-os-content
(which may not even have bootimages).

Also fixes up the "normal" case of when we need this (i.e. for public releases):
* Looks at the quay release registry instead of registry.ci
* Fixes a bug with release tag parsing for non-nightly tags


jobs/build/rhcos_sync/Jenkinsfile
```diff
@@ -119,19 +120,34 @@ node {
         }
 
         tag = params.FROM_RELEASE_TAG
-        if (params.OVERRIDE_NAME != "") {
-            name = params.OVERRIDE_NAME
-        } else {
-            name = tag
```
sosiouxme, line 125: Specifying a tag, which seemed like it ought to be how this is normally done, was full of pitfalls. Here, the default name is the same as the tag which would include the arch suffix. Rather than try to sort that out, take the default from the name in the release image itself.

jobs/build/rhcos_sync/Jenkinsfile
```diff
@@ -119,19 +120,34 @@ node {
...
+
         (major, minor) = commonlib.extractMajorMinorVersionNumbers(tag)
         ocpVersion = "$major.$minor"
 
         (arch, priv) = releaselib.getReleaseTagArchPriv(tag)
         suffix = releaselib.getArchPrivSuffix(arch, priv)
 
-        cmd = "oc image info -o json \$(oc adm release info --image-for machine-os-content registry.ci.openshift.org/ocp$suffix/release$suffix:$tag) | jq -r .config.config.Labels.version"
+        cmd = """
+            tmp=\$(mktemp -d /tmp/tmp.XXXXXX)
```
sosiouxme, line 146: I debated putting this under `.` but I didn't like the complication of having to make sure it was actually erased/overwritten from previous run - would not like to somehow ship out the wrong build, or have many of these hanging around. The use of `tmp` may raise security eyebrows, though, in which case reconsider.

Also I _really_ wish `oc image extract --file` worked like `oc adm release extract --file` or enabled stdout somehow - there should be no need for this single file to land on disk in the first place.

jobs/build/rhcos_sync/Jenkinsfile
```diff
@@ -119,19 +120,34 @@ node {
...
+            oc image extract --path /manifests/:\$tmp \$(oc adm release info --image-for installer ${image})
+            cat \$tmp/coreos-bootimages.yaml | yq -r .data.stream | jq -r .architectures.${arch}.artifacts.qemu.release
```
sosiouxme, line 148: Yes, the installer wraps the JSON... in a YAML definition of an API object. I seriously considered just using the source label and looking it up from github. There are two issues with that though:
1. Requires an access token for github and other complications
2. Requires consulting github when the data is right there in the payload... I'd rather use the source of truth.

pipeline-scripts/release.groovy
```diff
@@ -673,7 +673,7 @@ def getReleaseTagArchPriv(from_release_tag) {
         priv = true
     } else if (!nameComponents[2].isNumber()) {
         arch = commonlib.brewArchForGoArch(nameComponents[2])
-        priv = nameComponents[3] == "priv"
+        if (nameComponents.length > 3 && nameComponents[3] == "priv") priv = true
```
sosiouxme, line 676: This was the source of the bug where a normal tag like `4.12.0-ec.2-s390x` would hit a nondescript `java.lang.ArrayIndexOutOfBoundsException` - because it's expecting a nightly tag with at least 4 elements apparently; but we don't sync nightlies as dev-previews anymore and this didn't do the right thing for other tags.

vfreex (approved):
/lgtm

PR #3200, opened 2022-08-30
release.groovy: improve release tag parsing

sosiouxme:
`rhcos_sync` will now send this stable tags, make sure they are interpreted correctly.


sosiouxme:
I made a small change to qualify `commonlib` contents which surprised me by working correctly without qualification in my test. Don't want to rely on that though.

PR #3207, opened 2022-09-01
[ART-4064] promote-assembly: use arches defined in ocp-build-data

vfreex:
- Removes `ARCHES` parameter. The arches defined in ocp-build-data
  group.yml and the override in assembly definition will be used.
- Whether to validate backport bugs is now determined by Cincinnati
  graph data. Requires
https://github.com/openshift/elliott/pull/411/files to control whether
this check should be skipped.
- We shouldn't allow to promote a multi payload if arches are different
  from the arches used to build the components. Checking this in
`gen-payload` will be easier so this job doesn't include that check.

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -199,8 +199,7 @@ class PromotePipeline:
                 major, minor = util.isolate_major_minor_in_group(self.group)
                 next_minor = f"{major}.{minor + 1}"
                 logger.info("Checking if %s is GA'd...", next_minor)
-                cincinnati = CincinnatiAPI()
-                graph_data = await cincinnati.get_graph(channel=f"fast-{next_minor}")
+                graph_data = await CincinnatiAPI().get_graph(channel=f"fast-{next_minor}")
                 no_verify_blocking_bugs = False
                 if not graph_data.get("nodes"):
                     logger.info("%s is not GA'd. Blocking Bug check will be skipped.", next_minor)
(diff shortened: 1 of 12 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -193,18 +191,28 @@ async def run(self):
...
-                if self.skip_attached_bug_check:
-                    logger.info("Skip checking attached bugs.")
+            if self.skip_attached_bug_check:
+                logger.info("Skip checking attached bugs.")
+            else:
+                major, minor = util.isolate_major_minor_in_group(self.group)
+                next_minor = f"{major}.{minor + 1}"
+                logger.info("Checking if %s is GA'd...", next_minor)
+                cincinnati = CincinnatiAPI()
+                graph_data = await cincinnati.get_graph(channel=f"fast-{next_minor}")
```
sosiouxme, line 201: don't see a reason for a variable here...?
```suggestion
                graph_data = await CincinnatiAPI().get_graph(channel=f"fast-{next_minor}")
```

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -193,18 +191,28 @@ async def run(self):
...
+                no_verify_blocking_bugs = False
+                if not graph_data.get("nodes"):
+                    logger.info("%s is not GA'd. Blocker Bug check will be skipped.", next_minor)
```
sosiouxme, line 204: Once `elliott verify-attached-bugs` is not broken I'd like to have a quick look and see what happens if we don't skip this. I'd prefer we don't unless we have to, but I'm not sure we remember the reason why we did this.

sosiouxme (approved):
overall LGTM, pending https://github.com/openshift/elliott/pull/411/
/hold
until then. Also pending `elliott verify-attached-bugs` working.

PR #3211, opened 2022-09-05
update jira in promote job

Ximinhan:


joepvd:
Think this needs to be one of the last steps. Think we can call this step done if release is signed, PR to cincinnatti-graph-data is opened, etc.

Ximinhan:
OK, updated

joepvd:
/lgtm

Let's try it this week :)

PR #3225, opened 2022-09-15
Create placeholder bug before attach-cve-flaws

thegreyd:
attach-cve-flaws is a step that errors out often,
and those errors can block creating placeholder bugs.
creation of placeholder is not dependent on attach-cve-flaws
step, since it just processes attached-bugs and if no bugs
are attached, it will exit. So there should be no side effect of
doing this


vfreex (approved):
/lgtm

PR #3226, opened 2022-09-19
WIP ART-4437 Update clients at latest-{major}.{minor} on mirror

thegreyd:
https://issues.redhat.com/browse/ART-4437

Right now we don't update latest-4.y dirs when uploading
clients, this automates that.

Initially tried to create latest symlink in the existing dir
structure (https://github.com/openshift/aos-cd-jobs/pull/3226/commits/0dc4cbebc5e851efba30842ab7532e9e9e4dc06b)

But having `aws s3 sync --delete` for latest meant that
we couldn't do it in the same command, so do it separately
(https://github.com/openshift/aos-cd-jobs/pull/3226/commits/a2d44532b5bf1880684cee133f6aa525113d4a12)


vfreex:
@thegreyd I think we'd better to keep our practice and update those "symlinks" in set_client_latest job. The reason is that it determines which release is the "latest" using data from Cincinnati graph. With this change, latest link would always point to the release that is most recently promoted.

PR #3237, opened 2022-09-27
clean workspace after job complete

Ximinhan:


vfreex (changes requested):
Please do it before job starts.

Ximinhan:
it will clean doozer venv before doozer invoked

PR #3243, opened 2022-09-30
[ART-4862] Add build-microshift job

vfreex:
This job is similar to `custom` job, but it uses a different version
numbering schema for microshift.

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -11,7 +11,7 @@ from typing import Dict, Iterable, List, Optional, Tuple
 import click
 import yaml
 from doozerlib.assembly import AssemblyTypes
-from pyartcd import constants, exectools
+from pyartcd import exectools
 from pyartcd.cli import cli, click_coroutine, pass_runtime
 from pyartcd.record import parse_record_log
 from pyartcd.runtime import Runtime
@@ -24,6 +24,8 @@ PlashetBuildResult = namedtuple("PlashetBuildResult", ("repo_name", "local_dir",
 
 class BuildMicroShiftPipeline:
     """ Rebase and build MicroShift for an assembly """
+
+    SUPPORTED_ASSEMBLY_TYPES = {AssemblyTypes.STANDARD, AssemblyTypes.CANDIDATE, AssemblyTypes.PREVIEW}
 
     def __init__(self, runtime: Runtime, group: str, assembly: str,
                  ocp_build_data_url: str, logger: Optional[logging.Logger] = None):
@@ -57,7 +59,7 @@ class BuildMicroShiftPipeline:
             await load_group_config(self.group, self.assembly, env=self._doozer_env_vars)
             releases_config = await load_releases_config(Path(self._doozer_env_vars["DOOZER_WORKING_DIR"], "ocp-build-data"))
             assembly_type = get_assembly_type(releases_config, self.assembly)
-            if assembly_type not in [AssemblyTypes.STANDARD, AssemblyTypes.CANDIDATE, AssemblyTypes.PREVIEW]:
+            if assembly_type not in self.SUPPORTED_ASSEMBLY_TYPES:
                 raise ValueError(f"Building MicroShift for assembly type {assembly_type.value} is not currently supported.")
             release_name = get_release_name(assembly_type, self.group, self.assembly, None)
 
@@ -86,6 +88,11 @@ class BuildMicroShiftPipeline:
 
     @staticmethod
     def generate_microshift_version_release(ocp_version: str, timestamp: Optional[str] = None):
+        """ Generate version and release strings for microshift rpm.
+        Example version-releases:
+        - 4.12.42-202210011234
+        - 4.13.0~rc.4-202210011234
+        """
         if not timestamp:
             timestamp = datetime.utcnow().strftime("%Y%m%d%H%M")
         release_version = VersionInfo.parse(ocp_version)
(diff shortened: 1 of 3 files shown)


pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # determines OCP version
+        match = re.fullmatch(r"openshift-(\d+).(\d+)", group)
+        if not match:
+            raise ValueError(f"Invalid group name: {group}")
+        self._ocp_version = (int(match[1]), int(match[2]))
+
+        # sets environment variables for Doozer
+        self._doozer_env_vars = os.environ.copy()
+        self._doozer_env_vars["DOOZER_WORKING_DIR"] = str(self.runtime.working_dir / "doozer-working")
```
locriandev, line 44: Isn't this going to complain?
```
TypeError: unsupported operand type(s) for /: 'str' and 'str'
```
IIRC, `runtime.workdir` is a string

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -0,0 +1,164 @@
...
+    async def run(self):
+        self._slack_client.bind_channel(self.group)
+        slack_response = await self._slack_client.say(f":construction: Build microshift for assembly {self.assembly} :construction:")
+        slack_thread = slack_response["message"]["ts"]
+
+        try:
+            await load_group_config(self.group, self.assembly, env=self._doozer_env_vars)
+            releases_config = await load_releases_config(Path(self._doozer_env_vars["DOOZER_WORKING_DIR"], "ocp-build-data"))
+            assembly_type = get_assembly_type(releases_config, self.assembly)
+            if assembly_type not in [AssemblyTypes.STANDARD, AssemblyTypes.CANDIDATE, AssemblyTypes.PREVIEW]:
```
locriandev, line 60: What do you think about moving the allowed assembly types definition to the top of the file? This might help should we want to add a new type in the future

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -0,0 +1,164 @@
...
+            await self._slack_client.say(f"Hi @release-artists , microshift for assembly {self.assembly} has been successfully built.", slack_thread)
+            await self._slack_client.say(f"Please create a PR to pin the build:\n```\n{example_schema}\n```", slack_thread)
+        except Exception as err:
+            error_message = f"Error building microshift: {err}\n {traceback.format_exc()}"
+            self._logger.error(error_message)
+            await self._slack_client.say(error_message, slack_thread)
+            raise
+
+    @staticmethod
+    def generate_microshift_version_release(ocp_version: str, timestamp: Optional[str] = None):
```
locriandev, line 88: Would appreciate a comment that shows the expected output of this function (both for pre-release and non-pre-release)

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -0,0 +1,164 @@
...
+
+        if self.runtime.dry_run:
+            return [f"microshift-{version}-{release.replace('.p?', '.p0')}.el8"]
+
+        # parse record.log
+        with open(Path(self._doozer_env_vars["DOOZER_WORKING_DIR"]) / "record.log", "r") as file:
+            record_log = parse_record_log(file)
+            return record_log["build_rpm"][-1]["nvrs"].split(",")
+
+    def _generate_example_schema(self, nvrs: List[str]) -> Dict:
```
locriandev, line 124: It would help to see an example schema in the docstrings for this function

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -0,0 +1,164 @@
...
+            # Prints example schema
+            click.secho("Build completes. Please update the assembly schema in releases.yaml to pin the following NVR(s) to the assembly:\n", fg="green")
+            for nvr in nvrs:
+                click.secho(f"\t{nvr}", fg="green")
+            example_schema = yaml.safe_dump(self._generate_example_schema(nvrs))
+            click.secho(f"\nExample schema:\n\n{example_schema}", fg="green")
+
+            # Sends a slack message
+            # TODO: We might want this job to auto create a PR to ocp-build-data.
+            await self._slack_client.say(f"Hi @release-artists , microshift for assembly {self.assembly} has been successfully built.", slack_thread)
```
locriandev, line 79: Do we want to notify even if `DRY_RUN` was specified?

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -0,0 +1,164 @@
...
+            await self._slack_client.say(f"Please create a PR to pin the build:\n```\n{example_schema}\n```", slack_thread)
```
locriandev, line 80: I think it would be awesome to have the job open a PR for us, but maybe this can be a separate story :smiley_cat:

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: It is of type `pathlib.Path`.

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: I've added a static field `SUPPORTED_ASSEMBLY_TYPES` for this.

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: Done

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: Done

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: SlackClient class honors --dry-run option in its own implementation, so a slack message will not be actually sent in dry run mode.

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: That would be the next step :)

PR #3254, opened 2022-10-06
Fix env requirement for pyartcd/tox.ini

locriandev:
To prevent this error while running tox against pyartcd:
```
ERROR:   py36: could not install deps [-rrequirements-dev.txt]; v = InvocationError('/home/dpaolell/develop/github.com/openshift/aos-cd-jobs/pyartcd/.tox/py36/bin/python -m pip install -rrequirements-dev.txt', 1)
```

### DIFF
--- a/pyartcd/tox.ini
+++ b/pyartcd/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py36
+envlist = py38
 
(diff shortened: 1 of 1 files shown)


thiagoalessio (approved):
/lgtm

PR #3256, opened 2022-10-06
[ART-4931] Define GITLAB_TOKEN env var when calling doozer images:rebase

locriandev:
Ref. https://issues.redhat.com/browse/ART-4931

With https://github.com/openshift/doozer/pull/659, if `canonical_builders_from_upstream` is `auto` Doozer will need to clone https://gitlab.cee.redhat.com/ocp-release-schedule/schedule to verify whether we are before or after feature freeze for a given release. In order to do so, it will need to use an access token which is stored in Jenkins with the name `gitlab-ocp-release-schedule-schedule`. This PR makes sure we always provide this token when calling `doozer images:rebase`

### DIFF
--- a/jobs/build/golang-builder/Jenkinsfile
+++ b/jobs/build/golang-builder/Jenkinsfile
@@ -88,23 +88,25 @@ pipeline {
         }
         stage('build') {
             steps {
-                script {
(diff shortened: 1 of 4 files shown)


joepvd (approved):
/lgtm

PR #3257, opened 2022-10-07
Bump doozer

locriandev:


locriandev:
Already bumped in https://github.com/openshift/aos-cd-jobs/pull/3259

PR #3267, opened 2022-10-14
update-microshift-pocket: explicitly include microshift

vfreex:
MicroShift rpm config is now set to disabled mode. We need to include it explicitly in the doozer command line.

`--assembly` parameter is already set by `buildlib.doozer`

### DIFF
--- a/jobs/build/update-microshift-pocket/Jenkinsfile
+++ b/jobs/build/update-microshift-pocket/Jenkinsfile
@@ -63,7 +63,7 @@ node {
         """
     )
 
-    buildlib.doozer("--working-dir ${DOOZER_WORKING} --assembly ${params.ASSEMBLY} --group openshift-${version} config:plashet --base-dir ${ORG_PLASHET_DIR} --name repos --repo-subdir os -i microshift --arch x86_64 unsigned --arch s390x unsigned --arch ppc64le unsigned --arch aarch64 unsigned from-tags -t rhaos-${version}-rhel-${params.RHEL_TARGET}-candidate NOT_APPLICABLE")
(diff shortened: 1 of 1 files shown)


jobs/build/update-microshift-pocket/Jenkinsfile
```diff
@@ -63,7 +63,7 @@ node {
         """
     )
 
-    buildlib.doozer("--working-dir ${DOOZER_WORKING} --assembly ${params.ASSEMBLY} --group openshift-${version} config:plashet --base-dir ${ORG_PLASHET_DIR} --name repos --repo-subdir os -i microshift --arch x86_64 unsigned --arch s390x unsigned --arch ppc64le unsigned --arch aarch64 unsigned from-tags -t rhaos-${version}-rhel-${params.RHEL_TARGET}-candidate NOT_APPLICABLE")
+    buildlib.doozer("--working-dir ${DOOZER_WORKING} --group openshift-${version} -r microshift config:plashet --base-dir ${ORG_PLASHET_DIR} --name repos --repo-subdir os -i microshift --arch x86_64 unsigned --arch s390x unsigned --arch ppc64le unsigned --arch aarch64 unsigned from-tags -t rhaos-${version}-rhel-${params.RHEL_TARGET}-candidate NOT_APPLICABLE")
```
joepvd, line 66: Could you make it `--rpms` rather than `-r`? For scripts I think long form options are better for readability.

joepvd (approved):
/lgtm

With a bit to optionally address

jobs/build/update-microshift-pocket/Jenkinsfile
locriandev: Is there a reason to remove the `--assembly` option? Doozer `from-tags` command [seems to be using it](https://github.com/openshift/doozer/blob/master/doozerlib/cli/config_plashet.py#L488)

vfreex: @joepvd makes sense.
@locriandev `buildlib.doozer` already adds `--assembly` option.

PR #3268, opened 2022-10-14
build-microshift: Auto create PRs to pin microshift builds

vfreex:


pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -139,34 +139,68 @@ def _generate_example_schema(self, nvrs: List[str]) -> Dict:
...
+            d = {"html_url": "https://github.example.com/foo/bar/pull/1234", "number": 1234}
+            result = namedtuple('pull_request', d.keys())(*d.values())
+            return result
+        pushed = await build_data.commit_push(f"{title}\n{body}")
+        result = None
+        if pushed:
+            github_token = os.environ.get('GITHUB_TOKEN')
+            if not github_token:
+                raise ValueError("GITHUB_TOKEN environment variable is required to create a pull request")
+            owner = "openshift"
```
joepvd, line 192: Could the branch be pushed to the bot's account, and a PR opened from there?

joepvd (approved):
/lgtm

With an optional nit

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: he repo to push is actually configured at https://github.com/openshift/aos-cd-jobs/blob/94905e9c5c1a7e381cdda8b9a85cb0fea7f5254e/config/artcd.toml#L8. It can be changed to the fork, but we need some tests (and maybe also need to pre-create the fork) before we actually make that change.

The `owner` here is the merge destination.

PR #3275, opened 2022-10-21
gen-assembly: Move to Python and auto create PRs

vfreex:


vfreex:
Test job run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Fgen-assembly/33/
Test PR: https://github.com/openshift/ocp-build-data/pull/2083

joepvd (approved):
/lgtm

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -0,0 +1,233 @@
...
+
+    async def _create_or_update_pull_request(self, fn: Callable[[OrderedDict], OrderedDict]):
+        """ Create or update pull request for ocp-build-data
+        :param fn: a callback function that makes changes to releases.yml
+        """
+        self._logger.info("Creating ocp-build-data PR...")
+        # Clone ocp-build-data
+        build_data_path = self._working_dir / "ocp-build-data-push"
+        build_data = GitRepository(build_data_path, dry_run=self.runtime.dry_run)
+        ocp_build_data_repo_push_url = self.runtime.config["build_config"]["ocp_build_data_repo_push_url"]
```
vfreex, line 160: The repo to push is actually configured at https://github.com/openshift/aos-cd-jobs/blob/94905e9c5c1a7e381cdda8b9a85cb0fea7f5254e/config/artcd.toml#L8. It can be changed to the fork, but we need some tests (and maybe also need to pre-create the fork) before we actually make that change.

Currently those auto created branches always start with auto- . We can clean them up easily in the future.

jobs/build/gen-assembly/Jenkinsfile
```diff
@@ -92,55 +99,68 @@ node {
...
+                "--working-dir=./artcd_working",
+                "--config", "./config/artcd.toml",
+            ]
+
+            if (params.DRY_RUN) {
+                cmd << "--dry-run"
+            }
+            cmd += [
+                "gen-assembly",
+                "--ocp-build-data-url", params.DOOZER_DATA_PATH,
```
locriandev, line 122: For consistency within our tooling, I suggest naming this param as `data-path`, as seen for Elliott and Doozer

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -0,0 +1,233 @@
...
+            ocp_build_data_url = self.runtime.config.get("build_config", {}).get("ocp_build_data_url")
+        if ocp_build_data_url:
+            self._doozer_env_vars["DOOZER_DATA_PATH"] = ocp_build_data_url
+
+    async def run(self):
+        self._slack_client.bind_channel(self.group)
+        slack_response = await self._slack_client.say(f":construction: Generating assembly definition {self.assembly} :construction:")
+        slack_thread = slack_response["message"]["ts"]
+        try:
+            if self.arches and self.custom:
```
locriandev, line 64: `and not self.custom`?

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -0,0 +1,233 @@
...
+            cmd.append(",".join(self.arches))
+        cmd.append("release:gen-assembly")
+        cmd.append(f"--name={self.assembly}")
+        cmd.append("from-releases")
+        for nightly in candidate_nightlies:
+            cmd.append(f"--nightly={nightly}")
+        if self.custom:
+            cmd.append("--custom")
+        else:
+            if self.in_flight:
```
locriandev, line 143: We used to break the build if the user specified IN_FLIGHT_PREV or PREVIOUS for a custom release; now, if CUSTOM is flagged, we just ignore IN_FLIGHT_PREV or PREVIOUS. The behavior is the same, but the user might lose awareness about what is going on.

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -0,0 +1,233 @@
...
+            self._logger.info("Getting nightlies from Release Controllers...")
+            candidate_nightlies = await self._get_nightlies()
+            self._logger.info("Generating assembly definition...")
+            assembly_definition = await self._gen_assembly_from_releases(candidate_nightlies)
+            out = StringIO()
+            yaml.dump(assembly_definition, out)
+            self._logger.info("Generated assembly definition:\n%s", out.getvalue())
+
+            # Create a PR
+            def _merge(a, b):
```
locriandev, line 76: Looks like a utility function we used to merge two Python dictionaries. A docstring (even a concise one) would be appreciated, since recursion could make it not straight forward to understand what is going on

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -0,0 +1,233 @@
...
+                    if k not in b:
+                        # move new entry to the beginning
+                        c.move_to_end(k, last=False)
+                return c
+            pr = await self._create_or_update_pull_request(lambda releases_yaml: _merge(assembly_definition, releases_yaml))
+
+            # Sends a slack message
+            message = f"Hi @release-artists , please review assembly definition for {self.assembly}: {pr.html_url}"
+            await self._slack_client.say(message, slack_thread)
+        except Exception as err:
```
locriandev, line 91: Do we expect other types of exceptions apart from ChildProcessError?

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -0,0 +1,233 @@
...
+            if self.in_flight:
+                cmd.append(f"--in-flight={self.in_flight}")
+            for previous in self.previous_list:
+                cmd.append(f"--previous={previous}")
+            if self.auto_previous:
+                cmd.append("--auto-previous")
+        _, out, _ = await exectools.cmd_gather_async(cmd, stderr=None, env=self._doozer_env_vars)
+        return yaml.load(out)
+
+    async def _create_or_update_pull_request(self, fn: Callable[[OrderedDict], OrderedDict]):
```
locriandev, line 152: Maybe `fn` is not the best parameter name, how about `merge_function` or the like?

pyartcd/pyartcd/pipelines/gen_assembly.py
locriandev, line 76: Is there a particular need to use an internal function passed as a callable param to another function? Couldn't we just define a `merge_assemblies` member function to increase code readability?

pyartcd/pyartcd/pipelines/gen_assembly.py
```diff
@@ -0,0 +1,233 @@
...
+            self._logger.warning("[DRY RUN] Would have created pull-request with head '%s', base '%s' title '%s', body '%s'", head, base, title, body)
+            d = {"html_url": "https://github.example.com/foo/bar/pull/1234", "number": 1234}
+            result = namedtuple('pull_request', d.keys())(*d.values())
+            return result
+        pushed = await build_data.commit_push(f"{title}\n{body}")
+        result = None
+        if pushed:
+            github_token = os.environ.get('GITHUB_TOKEN')
+            if not github_token:
+                raise ValueError("GITHUB_TOKEN environment variable is required to create a pull request")
```
locriandev, line 188: If `DRY_RUN` is false, we do want to have a PR created. Then, moving this check at the very beginning of the pipeline could save us some time if things need to get fixed

locriandev (dismissed):

jobs/build/gen-assembly/Jenkinsfile
vfreex: Sounds good to me.

pyartcd/pyartcd/pipelines/gen_assembly.py
vfreex: This is a catch-all statement to capture whatever exception and send it to Slack. No need to worry about the concrete type here.

pyartcd/pyartcd/pipelines/gen_assembly.py
vfreex: I removed the callback parameter.

pyartcd/pyartcd/pipelines/gen_assembly.py
vfreex: I moved it out of the member function.

pyartcd/pyartcd/pipelines/gen_assembly.py
vfreex: I added an exception if both CUSTOM and IN_FLIGHT_PREV or PREVIOUS is specified.

pyartcd/pyartcd/pipelines/gen_assembly.py
vfreex: makes sense to move the check to the beginning.
In case of a dry run, `pushed` will be False so this will not be executed.

vfreex:
@locriandev Thanks. Could you approve the PR as well?

PR #3276, opened 2022-10-21
Stop producing 4.6 and 4.7

joepvd:


sosiouxme (approved):
/lgtm
if it turns out we need to do this for some reason, we can revert, but might as well spare ourselves the mass rebuilds we won't need

PR #3286, opened 2022-10-28
[ART-4863] Support building microshift against nightlies

vfreex:
Adds the following parameters to `build-microshift` job:
- RELEASE_PAYLOADS: Specify a list of nightlies to rebase against.
- UPDATE_POCKET: If checked, call `update-microshift-pocket` job after build completes.
- NO_REBASE: If checked, build without rebase.

Parameter `RELEASE_PAYLOADS` and `NO_REBASE` are implemented by setting environment variables for `microshift-rebase` modification script.

Requires https://github.com/openshift/ocp-build-data/pull/2121/.

Test build: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Fbuild-microshift/40/console

### DIFF
--- a/jobs/build/build-microshift/Jenkinsfile
+++ b/jobs/build/build-microshift/Jenkinsfile
@@ -27,7 +27,24 @@ node {
                         string(
                             name: "ASSEMBLY",
                             description: "The name of an assembly to rebase & build for. e.g. 4.9.1",
+                            defaultValue: "test",
                             trim: true
+                        ),
+                        string(
+                            name: 'RELEASE_PAYLOADS',
+                            description: '(Optional) List of release payloads to rebase against; can be nightly names or full pullspecs',
+                            defaultValue: "",
+                            trim: true,
+                        ),
+                        booleanParam(
+                            name: "UPDATE_POCKET",
+                            description: "Update pocket on mirror; ",
+                            defaultValue: false
+                        ),
+                        booleanParam(
+                            name: "NO_REBASE",
+                            description: "(For testing only) Do not rebase microshift code; build the current source we have in the upstream repo",
+                            defaultValue: false
                         ),
                         string(
                             name: 'OCP_BUILD_DATA_URL',
@@ -79,6 +96,14 @@ node {
                     "-g", "openshift-$params.BUILD_VERSION",
                     "--assembly", params.ASSEMBLY,
                 ]
+                if (params.RELEASE_PAYLOADS) {
+                    for (nightly in commonlib.parseList(params.RELEASE_PAYLOADS)) {
+                        cmd << "--payload" << nightly.trim()
+                    }
+                }
+                if (params.NO_REBASE) {
+                    cmd << "--no-rebase"
+                }
                 withCredentials([string(credentialsId: 'art-bot-slack-token', variable: 'SLACK_BOT_TOKEN'), string(credentialsId: 'openshift-bot-token', variable: 'GITHUB_TOKEN')]) {
                     echo "Will run ${cmd}"
                     if (params.IGNORE_LOCKS) {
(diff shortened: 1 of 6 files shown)


jobs/build/build-microshift/Jenkinsfile
```diff
@@ -27,8 +27,25 @@ node {
                         string(
                             name: "ASSEMBLY",
                             description: "The name of an assembly to rebase & build for. e.g. 4.9.1",
+                            defaultValue: "test",
                             trim: true
                         ),
+                        string(
+                            name: 'RELEASE_PAYLOADS',
+                            description: '(Optional) Release payloads to rebase against',
```
thegreyd, line 35: I think we can be clearer here since this could be interpreted as pullspec / release name / nightly name, but if we're just accepting pullspec we can specify here.
And should also mention about arches, since we need specifically [x64, arm], can we say that? Unless in future that list can change/expand?

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -56,36 +58,80 @@ def __init__(self, runtime: Runtime, group: str, assembly: str,
...
             assembly_type = get_assembly_type(releases_config, self.assembly)
             if assembly_type not in self.SUPPORTED_ASSEMBLY_TYPES:
                 raise ValueError(f"Building MicroShift for assembly type {assembly_type.value} is not currently supported.")
-            release_name = get_release_name(assembly_type, self.group, self.assembly, None)
+
+            custom_payloads = None
+            if assembly_type is AssemblyTypes.STREAM:
+                # rebase against nightlies
+                # rpm version-release will be like `4.12.0~test-202201010000.p?`
+                if not self.payloads:
```
thegreyd, line 73: Maybe I missed something, but shouldn't we check for `self.no_rebase` here?

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -56,36 +58,80 @@ def __init__(self, runtime: Runtime, group: str, assembly: str,
...
                 message += f"\nA PR to update the assembly definition has been created/updated: {pr.html_url}"
-            await self._slack_client.say(message, slack_thread)
+            if slack_client:
+                await slack_client.say(message, slack_thread)
         except Exception as err:
             error_message = f"Error building microshift: {err}\n {traceback.format_exc()}"
             self._logger.error(error_message)
-            await self._slack_client.say(error_message, slack_thread)
+            if slack_client:
+                await slack_client.say(error_message, slack_thread)
```
thegreyd, line 116: I see the fail notification now, do we plan on getting the retry mechanism?

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -210,8 +262,13 @@ async def _create_or_update_pull_request(self, nvrs: List[str]):
               help="The group of components on which to operate. e.g. openshift-4.9")
 @click.option("--assembly", metavar="ASSEMBLY_NAME", required=True,
               help="The name of an assembly to rebase & build for. e.g. 4.9.1")
+@click.option("--payload", "payloads", metavar="PULLSPEC", multiple=True,
```
thegreyd, line 265: Maybe I should've asked this earlier, but aren't `--payload` and `--no-rebase` mutually exclusive?  If that's the case, we could implement them with a single flag?

jobs/build/build-microshift/Jenkinsfile
vfreex: Currently it should be interpreted as pullspec. But I think it would be more convenient to be nightly name.
In any case, i will make it more clear in the description.

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: Makes sense. Error shouldn't be raised if no_rebase is set.

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: If we need a retry mechanism, i would like to get it in the slack_client class in a separate PR.

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: They are not mutually exclusive. When building microshift for named assemblies, we don't specify either flags. `--payload` is just for STREAM and --no-rebase is just for testing purpose.

jobs/build/build-microshift/Jenkinsfile
thegreyd: :+1:

pyartcd/pyartcd/pipelines/build_microshift.py
thegreyd: I think you mentioned doozer retries build, so we're good

pyartcd/pyartcd/pipelines/build_microshift.py
thegreyd: Ah, it wasn't clear to me that no_rebase was only useful for testing.  I see your comment so we're good

pyartcd/pyartcd/pipelines/build_microshift.py
thegreyd: :+1:

PR #3290, opened 2022-11-01
[ART-5150] poll-payload: Fix publish-rpms not being triggered

vfreex:
`publish-rpms` is a job to publish RHEL 8 worker RPMs to the mirror.

We expect to run publish-rpms whenever there is a new x86_64 nightly is created and accepted by the Release Controller.

However this is broken because commit https://github.com/openshift/aos-cd-jobs/commit/f75f7c75e0d0c038af096c2fab84f7d2c3fe3c91 removes the handlers of triggering pre-release job. As a side effect, `publish-rpms` will not be invoked as well.

Currently, starting publish-rpms is invoked in the logic of checking for release stream changes. It is not obvious enough and breaks the design of poll-payload job. It should be refactored to become a callback function like other event handlers.

This PR:
- Changes `publishRpms` to an event handler like `setDevPreviewLatest` and `startPreReleaseJob`.
- Enhances the event loop to call multiple event handlers defined in `ACTIONS`.
- Removes the arch check in `publishRPMs`. It will be hooked into the x86_64 release stream change event, so the arch check will not be needed.

### DIFF
--- a/scheduled-jobs/build/poll-payload/Jenkinsfile
+++ b/scheduled-jobs/build/poll-payload/Jenkinsfile
@@ -12,20 +12,20 @@ import groovy.json.JsonOutput
 // A map of release stream names to be polled and and actions to be taken
 @Field final ACTIONS = [
     // as part of 4.next we will manually publish only sprintly dev-previews
-    "4-dev-preview": this.&setDevPreviewLatest,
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

PR #3292, opened 2022-11-01
Revert "build signed compose for 4.12"

joepvd:
Reverts openshift/aos-cd-jobs#3288

We're missing product listings, which fails the build. Reverting for now.


sosiouxme:
agreed, for now.
/lgtm

PR #3313, opened 2022-11-10
[ART-5142]  Use patch_version defined or implied in assembly definition

vfreex:
Use patch_version defined or implied in assembly definition

This will also allow to build microshift for a custom release.

Requires https://github.com/openshift/doozer/pull/683
Also fix unit tests.


vfreex:
/hold cancel
Doozer PR merged

PR #3320, opened 2022-11-15
add allbuilds flag for elliott

Ximinhan:
at GA time we need to attach all builds to get CDN repo mapped correctly

### DIFF
--- a/jobs/build/prepare-release/Jenkinsfile
+++ b/jobs/build/prepare-release/Jenkinsfile
@@ -61,7 +61,7 @@ node {
                             defaultValue: false
                         ),
                         booleanParam(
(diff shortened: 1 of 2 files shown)


Ximinhan:
ref: https://github.com/openshift/elliott/pull/461

thegreyd:
This looks good to me but if there's only one usecase of running at candidate/ga time, then we should code that up?

Ximinhan:
@thegreyd  this will be needed during the RC period, also prevent manually do it for every release

vfreex (commented):
Generally lgtm.

vfreex (approved):
/lgtm

PR #3334, opened 2022-11-22
[ART-5345] Move kerberos keytab into Jenkins credential store

jupierce:


jupierce:
@locriandev - ptal. We need to move the keytab off the file system. Since tickets are fairly log lived and most jobs call buildlib.kinit, I think we can assume the ticket will be valid once util runs. If you don't feel confident that is the case, situations where we call util in a shell will need to be wrapped in withCredential.

locriandev:
The only place where I see `util.kinit()` being used is in the [sweep job](https://github.com/openshift/aos-cd-jobs/blob/master/pyartcd/pyartcd/pipelines/sweep.py#L41). In this case, I think we should wrap [its invocation](https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/sweep/Jenkinsfile#L98) with the new credentials

jupierce:
@locriandev  updated.

PR #3340, opened 2022-11-23
ART-5258 Trigger multi rhcos builds

thegreyd:
https://issues.redhat.com/browse/ART-5258

Release browser for unified builds: https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/storage/browser

To be enabled this will need config like https://github.com/openshift/ocp-build-data/pull/2288

Test job: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Frhcos/35/console

RHCOS Jenkins: https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com/job/build/

Doc with details to trigger remotely https://github.com/coreos/fedora-coreos-pipeline/blob/main/HACKING.md

### DIFF
--- a/jobs/build/rhcos/Jenkinsfile
+++ b/jobs/build/rhcos/Jenkinsfile
@@ -47,57 +47,85 @@ node {
             'ppc64le': 'jenkins_serviceaccount_ocp-ppc.stage.psi.redhat.com',
             's390x': 'jenkins_serviceaccount_osbs-s390x-3.prod.engineering.redhat.com.kubeconfig',
             'aarch64': 'jenkins_serviceaccount_osbs-aarch64-1.engineering.redhat.com',
+            'multi': 'jenkins_serviceaccount_ocp-virt.prod.psi.redhat.com.kubeconfig',
         ]
 
         // Disabling compose lock for now. Ideally we achieve a stable repo for RHCOS builds in the future,
         // but for now, being this strict is slowing down the delivery of nightlies.
         //lock("compose-lock-${params.BUILD_VERSION}") {
         lock(resource: "rhcos-lock-${params.BUILD_VERSION}", skipIfLocked: true) {  // wait for all to succeed or fail for this version before starting more
-        timestamps {
-            def archJobs = [:]
-            for (arch in arches) {
-                def jobArch = arch.trim() // make sure we use a locally scoped variable
-                if (!releaseArches.contains(jobArch)) {
-                    echo "Skipping ${jobArch} since ${params.BUILD_VERSION} only supports ${releaseArches}"
-                    continue
-                }
-                archJobs["trigger-${jobArch}"] = {
-                    try {
-                        lock(label: "rhcos-build-capacity-${jobArch}", quantity: 1) { // cluster capacity limited per arch
-                            withCredentials([file(credentialsId: kubeconfigs[jobArch], variable: 'KUBECONFIG')]) {
-                                // the squid proxy inhibits communication to some RHCOS clusters, so augment no_proxy
-                                sh  'export no_proxy=ocp-ppc.stage.psi.redhat.com,api.ocp-virt.prod.psi.redhat.com,$no_proxy\n' +
-                                    "oc project\n" +
-                                    "BUILDNAME=`oc start-build -o=name buildconfig/rhcos-${params.BUILD_VERSION}`\n" +
-                                    'echo Triggered $BUILDNAME\n' +
-                                    'for i in {1..240}; do\n' +
-                                    '   PHASE=`oc get $BUILDNAME -o go-template=\'{{.status.phase}}\'`\n' +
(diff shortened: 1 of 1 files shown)


jobs/build/rhcos/Jenkinsfile
```diff
@@ -47,57 +47,86 @@ node {
...
-            def archJobs = [:]
-            for (arch in arches) {
-                def jobArch = arch.trim() // make sure we use a locally scoped variable
-                if (!releaseArches.contains(jobArch)) {
-                    echo "Skipping ${jobArch} since ${params.BUILD_VERSION} only supports ${releaseArches}"
-                    continue
+
+        // Check if urls.rhcos_release_base.multi is defined in group.yml
+        def cmd = "doozer -g openshift-${params.BUILD_VERSION} config:read-group urls.rhcos_release_base.multi"
+        multi_builds_enabled = commonlib.shell(script: cmd, returnStatus: true)
```
jlebon, line 60: Minor: in those situations, I find it less confusing to do something like:

```suggestion
        multi_builds_enabled = (commonlib.shell(script: cmd, returnStatus: true) == 0)
```

so that lower code can just do e.g. `if (multi_builds_enabled)` and it's easy to read.

jobs/build/rhcos/Jenkinsfile
```diff
@@ -47,57 +47,86 @@ node {
...
-        timestamps {
-            def archJobs = [:]
-            for (arch in arches) {
-                def jobArch = arch.trim() // make sure we use a locally scoped variable
-                if (!releaseArches.contains(jobArch)) {
-                    echo "Skipping ${jobArch} since ${params.BUILD_VERSION} only supports ${releaseArches}"
-                    continue
+
+        // Check if urls.rhcos_release_base.multi is defined in group.yml
+        def cmd = "doozer -g openshift-${params.BUILD_VERSION} config:read-group urls.rhcos_release_base.multi"
```
vfreex, line 59: Needs to add `--default 0` otherwise it will raise an error if the key doesn't exist.

jobs/build/rhcos/Jenkinsfile
vfreex: `urls.rhcos_release_base.multi` is of type string. So I would suggest testing if the string is empty after calling `.trim()`.

jobs/build/rhcos/Jenkinsfile
```diff
@@ -47,57 +47,86 @@ node {
...
+        multi_builds_enabled = commonlib.shell(script: cmd, returnStatus: true)
+
+        if (multi_builds_enabled == 0) {
```
vfreex, line 62: Then this can be changed to
```suggestion
        if (!multi_builds_enabled) {
```

jobs/build/rhcos/Jenkinsfile
vfreex: Any error in execution will make the pipeline think multi is disabled.
Would suggest to to add --default '' to `cmd` to let it return empty string if the key doesn't exist.
And change `returnStatus` to `returnStdout` to let it return the value of the key.

This will also fail the pipeline if the execution fails.

```suggestion
        def multi_builds_enabled = commonlib.shell(script: cmd, , returnStdout: true).trim()
```

jobs/build/rhcos/Jenkinsfile
```diff
@@ -47,57 +47,86 @@ node {
...
+            currentBuild.displayName += "multi"
+            echo "triggering multi builds"
+            def jenkins_url = 'https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com'
+
+            try {
+                withCredentials([file(credentialsId: kubeconfigs['multi'], variable: 'KUBECONFIG')]) {
```
vfreex, line 68: We already have pretty robust implementation for triggering and waiting for single-arch rhcos builds. I think this can be simplified by cleaning up `arches` dict to only have the `multi` entry then we can reuse the logic of handling single-arch builds to handle multi build.

Would also like to see the job is triggered by `oc start-build` instead of `curl` call so we won't have to handle authentication and URL endpoint by ourselves.

vfreex (dismissed):
Great job to have this working! I have some questions about how the multi job is triggered. Would like to make it consistent with what we have now.

jobs/build/rhcos/Jenkinsfile
sosiouxme: @vfreex @joepvd the new implementation of RHCOS pipeline doesn't use openshift entities at all to run. It's just using Jenkins and defining everything as Jenkins jobs. This means we no longer depend on the the kubernetes plugin for Jenkins (which caused a lot of problems in the past), but that means we have to trigger jobs by some other means that Jenkins understands directly.

sosiouxme (approved):

PR #3357, opened 2022-11-30
add concurrency limit

Ximinhan:
fix out of memory issue


vfreex:
Based on the conversation on Slack, this is not related.

PR #3365, opened 2022-12-02
[ART-4437] Update latest symlinks for ECs

vfreex:
Since 4.13, ECs will be in cincinnati. Update latest symlinks
automatically when an EC lands.


jobs/build/set_client_latest/S3-set-v4-client-latest.sh
```diff
@@ -115,6 +115,15 @@ for arch in ${ARCHES}; do
         fi
     fi
 
+
+    if [ "$CLIENT_TYPE" == "auto" ]; then
+        if [[ "$RELEASE"  =~ "-[efr]c\.[0-9]+" ]]; then  #
```
Ximinhan, line 120: I think RC should belong to ocp

joepvd (approved):
/lgtm

PR #3373, opened 2022-12-08
Mass pruning tool for unused RHCOS AMIs in rh-dev

jupierce:


mike-nguyen:
This is great!  Is there any way to get graph data for 4.5 and prior?

jupierce:
@mike-nguyen - Not that the data is gone, but thankfully I ran this before the release controller team archived older releases. You can see that the <= 4.5 releases are mentioned in the analysis I posted: https://drive.google.com/file/d/19BVPz7qn-K_ADO8WZKSQBgXHB-81-3ND/view?usp=share_link . I'll build off this cache if we run this. I'll tag AMIs that are preserved so we don't have to do any fancy work to remember this fact in the future.

jlebon:
Would be good to add a top-level docstring documenting the approach of this tool. IIUC, it's crawling through the update graph and then collating the AMIs to keep by looking at each node and its associated `rhcos.json` metadata in the corresponding openshift/installer repo commit.

I think it's a valid approach, but I think a better approach long-term is to prune at the cosa build level since it's not just AMIs that we want to prune. The code here could build a list of cosa build IDs we want to keep, and a GC process could prune all the cosa builds (and the referenced remote resources within, which include AMIs) which aren't on that list and older than X. The larger issue for that is tracked at https://github.com/coreos/fedora-coreos-tracker/issues/99 and parts of the functionality exists in https://github.com/coreos/coreos-assembler/blob/main/src/cmd-remote-prune.

As is, this code will invalidate AMIs referenced in the `meta.json` of their associated cosa builds which isn't a huge concern, but isn't great either. I wouldn't consider it a blocker to do it this way.

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,255 @@
...
+                        images_list = installer_arch['images']
+                    if 'aws' not in images_list:
+                        print(f'  No AMIs for arch: {installer_arch_name}')
+                        continue
+                    for region_name, region_entry in images_list['aws']['regions'].items():
+                        if region_name not in installer_ami_ids:
+                            installer_ami_ids[region_name] = {}
+                            installer_ami_ids[region_name]['images'] = []
+                        ami_id = region_entry['image']
+                        installer_ami_ids[region_name]['images'].append(ami_id)
```
sosiouxme, line 157: it's not important but i just prefer brevity
```suggestion
                        ami_id = region_entry['image']
                        installer_ami_ids.setdefault(region_name, {}).setdefault('images', []).append(ami_id)
```

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,255 @@
...
+            rhcos_in_use[payload_key] = {
+                'payload_version': machine_os_rhcos_version,
+                'installer_amis': installer_ami_ids
+            }
+
+            with riu_path.open('a+') as f:
+                cache_entry = yaml.dump({
+                    payload_key: rhcos_in_use[payload_key]
+                })
+                # Write, line by line, to the cache file
```
sosiouxme, line 169: minor note that `yaml.dump` includes a line break.

for a moment i was confused thinking each entry would be one line. perhaps better to say:
```suggestion
                # Write to the cache file separated by a blank line
```

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,255 @@
...
+            image_tags = image.get('Tags', [])
+
+            print(f'  Checking AMI {image_id} ({image_name})')
+            # Build up a string we are certain will contain the RHCOS version
+            ami_info = f'{image_name} {image_description} {image_tags}'
+            matches_payload_keys = list()
+            for payload_key, payload_key_references in rhcos_in_use.items():
+                if payload_key_references['payload_version'] in ami_info:
+                    print(f'Will not clean up RHCOS AMI {image_id} ({image_name}) since it is referenced by a release controller release payload {payload_key}')
+                    matches_payload_keys.append(payload_key)
```
sosiouxme, line 211: do we actually care about release payload references to AMIs? we only use AMIs at install time, right? and i'm pretty sure we don't reference AMIs anywhere besides build metadata and the installer so don't we only need to keep things that the installer refers to?

the difference isn't going to be huge though if you want to be cautious here and keep both.

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,255 @@
...
+                'creation': image_creation,
+                'tags': image_tags,
+                'preserve': len(matches_payload_keys) > 0,
+                'used_by_payloads': matches_payload_keys,
+            }
+
+            if matches_payload_keys:
+                will_preserve.add(f'{aws_region}:{image_id}')
+
+            if not matches_payload_keys:
```
sosiouxme, line 237:
```suggestion
            else:
```
? :-)

sosiouxme (approved):
seems legit

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,255 @@
...
+                release_json_result = oc.invoke('adm', cmd_args=['release', 'info', payload_pullspec, '-o=json'],
+                                                no_namespace=True, auto_raise=False)
+                if release_json_result.status() == 0:
+                    break
+                else:
+                    print(f'Error acquiring image information: {release_json_result.err()}')
+
+            if release_json_result.status() != 0:
+                if '.ci' in payload_version or '.nightly' in payload_version:
+                    continue
```
sosiouxme, line 80: Actually... we don't want to `preserve` everything in a nightly for the indefinite future, but we _do_ want to preserve it for long enough to become a promoted release. So should there also just be a blanket exemption for AMIs until they're old enough (say, a month)? Presumably garbage collection config would cover that, perhaps that's why it's not built in here.

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,255 @@
...
+
+            release_info_str = release_json_result.out()
+            release_info = json.loads(release_info_str)
+            try:
+                machine_os_rhcos_version = release_info['displayVersions']['machine-os']['Version']
+            except KeyError:
+                # No machine-os version information in the image. In old releases, use image info for the
+                # machine-os-content image.
+                print('  Querying for machine-os-content info')
+                moc_pullspec = list(filter(lambda tag: tag['name'] == 'machine-os-content', release_info['references']['spec']['tags']))[0]['from']['name']
```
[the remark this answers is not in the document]
jupierce: We could add that, but the method itself is a fallback. In current builds, the rhcos version is in the payload metadata and doesn't require inspecting the rhcos image: https://github.com/openshift/aos-cd-jobs/pull/3373/files#diff-a3f55c4ee0db478ae07381ad1394f41bc69ea3b33c5f606ace1e946afd082c27R87

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,255 @@
...
+                if 'registry.svc' in moc_pullspec:
```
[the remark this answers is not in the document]
jupierce: I added this because some ancient CI payload did. Nothing ART promoted.

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
jupierce: > the difference isn't going to be huge though if you want to be cautious here and keep both.
I don't think we do need them. So, yes, this is just a cautionary step.

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
jupierce: Right - If we ran this again the future, older nightlies would have been garbage collected from the release controller. As written, this tool would not try to preserve them.

mike-nguyen:
I was just working on some boot image bumps and realized that this tool could mark images for deletion that are in openshift/installer pull requests that have not merged. I don't think that should affect this PR but just something to be aware of when this is setup to run.

With the new RHCOS pipeline, we won't be replicating to all regions unless we run the `cloud-replicate` job (which will only be used when creating boot images) so that should cut down on the number of snapshots on AWS.

jupierce:
@mike-nguyen . In the latest push, I've excluded images created in the last 30 days.

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,191 @@
...
+        print(f'Requires recent clone of github.com/openshift/installer in GOPATH: {str(installer_git_path)}')
+        exit(1)
+
+    http = urllib3.PoolManager()
+
+    amis_in_use: Set = set()
+
+    git_depth = 0
+    while True:
+        installer_commit: Commit = f'upstream/master~{git_depth}'
```
jlebon, line 63: This is going down `master` only. I think we need to go down from all the release branches too, no?

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
```diff
@@ -0,0 +1,191 @@
...
+            gitshow_process = subprocess.run(['git', '-C', str(installer_git_path), 'show', f'{installer_commit}:data/data/rhcos.json'], capture_output=True)
+            if gitshow_process.returncode != 0:
+                print(f'Failed to find rhcos.json at depth {git_depth}')
+                break
+
+        # Otherwise gitshow process contains a file like https://github.com/openshift/installer/blob/01adff5d629bb403a6cf49f9d44070762ef77e93/data/data/coreos/rhcos.json
+        # or older format like https://github.com/openshift/installer/blob/bfdba2d19e3fa1105bab93aa4cce8bb7f44b4b2c/data/data/rhcos.json
+        amis_in_file = list(re.findall('ami-[a-z0-9]+', str(gitshow_process.stdout)))
+        amis_in_use.update(amis_in_file)
+        git_depth += 1
```
jlebon, line 78: Instead of visiting every commit on the branch, it'd be more efficient to ask `git` for only the commits which touch the files we care about and then only visit those. Something like:

```
git log --format=%H <branch> -- data/data/coreos/rhcos.json data/data/rhcos.json
```

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
jupierce: Great call. I wasn't considering post-cut changes. Updated.

hacks/prune-rhcos-snapshots/prune-rhcos-snapshots.py
jupierce: I like it. Updated.

jupierce:
That latest push scans all branches and finds 2159 AMIs from the installer repo.

jlebon (approved):
Approach looks sane to me! One thing I double-checked was whether we've used `rhcos.json` even from 4.1.0, which we did.

mike-nguyen:
LGTM to run

PR #3376, opened 2022-12-09
[ART-5504] create and wait on new rhcos builds

sosiouxme:
[https://issues.redhat.com/browse/ART-5504](ART-5504) is part of a wider effort to interact with multi RHCOS builds like we expect. Here we lay the groundwork by moving the new build logic to `pyartcd` and having it use the Jenkins API to see when existing builds are running, create a new one if not, and wait for it (and its children) to complete so that we won't kick off duplicates (make the version lock work again).

This also provides knobs to create a new build even if one is running, and also to have a new build occur even when the RHCOS pipeline doesn't think it needs one. These will be useful for manual overrides or automated recovery from incomplete/inconsistent builds.

Finally, I think it provides improved status about what it sees and does.


jobs/build/rhcos/Jenkinsfile
```diff
@@ -25,10 +25,20 @@ node {
...
                         defaultValue: commonlib.brewArches.join(','),
                         trim: true,
                     ),
+                    booleanParam(
+                        name: 'IGNORE_EXISTING',
+                        description: '(Multi pipeline only) Create a new build even if one is already in progress.',
+                        defaultValue: false,
+                    ),
+                    booleanParam(
+                        name: 'BUILD_ALWAYS',
```
thegreyd, line 38: Why not 'force'?
Also does this include "ignore_existing" implicitly ? I would assume it does
Would we have cases where we use "ignore_existing" and not "force"?

sosiouxme: > Why not 'force'?

Exactly because of this assumption :)

`FORCE` implies "ignore conditionals, just get me a new build". I don't want that hammer here. Also it implies ignoring the mutex in this job itself, which I don't ever want to do.

We need `BUILD_ALWAYS` when latest builds do not have all arches complete, or contents are not consistent. Current RHCOS pipeline logic will not actually rebuild for this reason (build will run and abort early) because no change is needed to x86. So we want our automation to ignore that and have a new build anyway.  Except we would not want to do this if the current pipeline is simply waiting on arch builds to finish, or if there is already another build in progress (possibly a human started it, possibly a previous run started it and then died).  So I intend automation to run `BUILD_ALWAYS` without `IGNORE_EXISTING` to first check if there are any running builds, and then force a build if not. This is to keep FORCEd runs from piling up. Admittedly probably needs a better name, since it _won't_ always trigger a build.

`IGNORE_EXISTING` is more intended for humans. (I called it `--skippable` in the params to invert the logic... likely needs more naming thought too.) Using our superior situational awareness we may see that any current runs are irrelevant to our purposes and we really want to force a new one ASAP. This will allow us to do so within the context of the mutex (waiting for the build we kick off) preventing automation from kicking off more runs.

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -0,0 +1,191 @@
...
+        self.request_session.post(
+            f"{JENKINS_BASE_URL}/job/build/buildWithParameters",
+            data=params,
+        )
+
+        # wait for a related build to begin
+        initial_builds = []
+        while not initial_builds:  # may take a few seconds for the build to start
+            time.sleep(1)
+            initial_builds = self.query_existing_builds()
```
sosiouxme, line 128: occurs to me this should have a limit. all kinds of ways this can go wrong and leave us hanging. if we're waiting more than 15 minutes for a build to start, something is probably wrong and we should try again later.

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -0,0 +1,191 @@
...
+                f"{JENKINS_BASE_URL}/job/{job}/api/json?tree=builds[number,description,result]"
+            ).json()["builds"]
+            if b["number"] == number
+        ), None)
+
+    def wait_for_builds(self):
+        """Wait for all builds for this version to complete, and give status updates on stderr"""
+        builds_seen: Dict[Tuple, str] = {}
+        completed_builds: Dict[Tuple, Dict] = {}
+        while True:
```
sosiouxme, line 149: likewise here... if the pipeline isn't complete after maybe 6 hours, chances are something is wrong.

sosiouxme:
made some improvements and rebased. can't test ocp4_scan right now, not sure why it's broken in master.

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @staticmethod
+    def build_url(job: str, number: int) -> str:
+        return f"{JENKINS_BASE_URL}/job/{job}/{number}/"
+
+    def query_existing_builds(self) -> List[Dict]:
+        """Check if there are any existing builds for the given version. Returns builds in progress."""
+        builds = []
+        for job in ("build", "build-arch", "release"):
+            builds.extend(
```
thegreyd, line 101: Hmm so we are collecting all the in progress builds for all jobs and looking at those.
But what if for a build-id, 1 arch is in progress and we want to re-trigger another arch.
I guess we're not handling that case right now? This would also be easier if the main build job 
would wait on build-arch and release jobs..

pyartcd/pyartcd/pipelines/build_rhcos.py
```diff
@@ -0,0 +1,197 @@
...
+
+            # check if any previous builds have newly completed
+            for spec in builds_seen.keys():
+                if spec not in new_builds_seen and spec not in completed_builds:
+                    completed_builds[spec] = completed = self.build_result(*spec)
+                    if completed:  # silently ignore if it's somehow not there... should never happen
+                        print(f"{completed['url']} finished with {completed['result']}: '{completed['description']}'", file=sys.stderr)
+
+            # if there are no builds left running, we're done
+            if not new_builds_seen:
```
thegreyd, line 167: Wouldn't this break even if there are pending builds in builds_seen?

thegreyd:
Really nice work! This brings us closer to even more fine grained control thanks to the jenkins api 
I saw a case where the build jobs succeeded but the release job failed https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Frhcos/55/
it was ultimately rerun manually. But in a case like that, if noone is closely watching, we could be skipping out
on perfectly good builds. So I guess we need to watch all 3 jobs closely, maybe rhcos team can handle some of that
retry logic if they wait for release job in the main build job

jobs/build/rhcos/Jenkinsfile
thegreyd: Agreed! It's good to have that button for when we might need it

pyartcd/pyartcd/pipelines/build_rhcos.py
sosiouxme: you could manually re-trigger another arch, and as long as it started before the others finished, this would adopt it as part of the build and wait on it.

but for the most part i don't want us re-triggering arch builds unless we're trying to get a release and watching them very closely (and ideally not at all).

pyartcd/pyartcd/pipelines/build_rhcos.py
sosiouxme: yes, that's why `start_build` waits until a build shows up before it considers its job complete (entirely possible that jenkins could be out of executors or dies or whatever and the build is delayed or lost).

sosiouxme:
In theory we have the information here that the release job failed, and our status at least would consider it failed. We could in theory do something about it.

But that's kind of like noticing when OSBS creates a build and then brew fails to tag it. We _could_ recover from that but... I just don't want that level of detail (and if a build fails, we don't know that it was a flake - we could waste a lot of time on retries). I want a build that succeeds or fails, and to start from scratch if it fails, so I'm hoping the RHCOS pipeline takes care of all those details. This is a workaround really...

pyartcd/pyartcd/pipelines/build_rhcos.py
thegreyd: I missed that new_builds_seen is querying existing builds in each run - so makes sense now

PR #3382, opened 2022-12-13
[ART-4752] migrate operator_sdk job to python and also update jira subtask

Ximinhan:
migrate operator_sdk job to python also update the logic for pre-release
this can used for two cases one is for weekly release, usually only need to put in assembly release name and it will check whether there is new SDK build in this release and update corresponding jira when complete, it will looks like [this job]( https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Foperator-sdk_sync/87)
another use case is to publish a specific build to [pre-release](https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/operator-sdk/pre-release/) dir on mirror. QE still need this link, it may happen at pre-GA phase when QE requested, it will looks like [this job](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Foperator-sdk_sync/84/)


locriandev (changes requested):

jobs/build/operator-sdk_sync/Jenkinsfile
```diff
@@ -19,126 +19,48 @@ pipeline {
...
+            name: 'ASSEMBLY',
+            description: 'Under which ASSEMBLY directory to place the binaries.',
+            defaultValue: "",
             trim: true,
         )
         booleanParam(
             name: 'UPDATE_LATEST_SYMLINK',
             description: 'You just want to update "latest" on the highest 4.x version',
-            defaultValue: true,
+            defaultValue: false,
```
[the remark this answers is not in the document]
Ximinhan: only the highest 4.x version needs to be set as true, set it false as default can avoid mistake

jobs/build/operator-sdk_sync/Jenkinsfile
```diff
@@ -19,126 +19,48 @@ pipeline {
...
-                            sh "rm -f ./operator-sdk"
+                    sh "rm -rf ./artcd_working && mkdir -p ./artcd_working"
+                    currentBuild.displayName += " ${params.ASSEMBLY}"
+                    def cmd = [
+                        "artcd",
+                        "-v",
+                        "--working-dir=./artcd_working",
+                        "--config=./config/artcd.toml",
+                        "operator-sdk-sync",
+                        "--group=openshift-${params.ASSEMBLY.tokenize(".")[0]}.${params.ASSEMBLY.tokenize(".")[1]}",
```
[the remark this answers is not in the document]
Ximinhan: it's true, but I never see an edge case as for now, I want to simplify the parameter, if we already provided assembly name no need to pass the group value again

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,130 @@
...
+    def __init__(self, runtime: Runtime, group: str, assembly: str, updatelatest: bool) -> None:
+        _LOGGER.info("Initializing ...")
+        self.runtime = runtime
+        self.group = group
+        self.version = group.split("-")[1]
+        self.assembly = assembly
+        self.updatelatest = updatelatest
+        self.sdk = "operator-sdk"
+
+        self._jira_client = runtime.new_jira_client()
```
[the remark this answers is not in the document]
Ximinhan: update_jira will be called multi times, no need to create a client multi times

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,130 @@
...
+        for arch in archlist:
+            self._extract_binaries(arch, sdkVersion, build['extra']['image']['index']['pull'][0])
+        self._update_jira(self.parent_jira_key, 7, "operator_sdk_sync job : {}".format(os.environ.get("BUILD_URL")))
+
+    def _get_sdkversion(self, build):
+        output = subprocess.getoutput(f"oc image info --filter-by-os amd64 -o json {build} | jq .digest")
+        shasum = re.findall("sha256:\w*", output)[0]
+        cmd = f"oc image extract {self.imagePath}@{shasum} --path /usr/local/bin/{self.sdk}:. --confirm && chmod +x {self.sdk} && ./{self.sdk} version && rm {self.sdk}"
+        _LOGGER.info(cmd)
+        sdkvalue = subprocess.getoutput(cmd)
```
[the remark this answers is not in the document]
Ximinhan: cmd_gather_async don't support pipeline operation, things that can be solved by one line of code will become more cumbersome

jobs/build/operator-sdk_sync/Jenkinsfile
```diff
@@ -19,126 +19,48 @@ pipeline {
 
     parameters {
         string(
-            name: 'OCP_VERSION',
```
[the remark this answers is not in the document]
Ximinhan: the value OCP_VERSION here is represented in the dir name on mirror, as same as the assembly name in new job

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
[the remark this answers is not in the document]
vfreex: @Ximinhan you can try `shell=true` option in `cmd_gather_async` in case if you want to change to that.

jobs/build/operator-sdk_sync/Jenkinsfile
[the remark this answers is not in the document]
vfreex: Could you add another parameter for `--group`?

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,133 @@
...
+        self.version = group.split("-")[1]
+        self.assembly = assembly
+        self.updatelatest = updatelatest
+        self.sdk = "operator-sdk"
+
+        self._jira_client = runtime.new_jira_client()
+        ocp_build_data_repo = runtime.new_github_client().get_repo("openshift/ocp-build-data")
+        release_file = ocp_build_data_repo.get_contents("releases.yml", ref=group)
+        release_yaml = yaml.load(release_file.decoded_content, Loader=yaml.FullLoader)
+        self.extra_ad_id, self.parent_jira_key = self.get_ad_jira_key(assembly, release_yaml)
```
vfreex, line 27: Those lines could be moved to `run`.

vfreex: The reason is that, things in the constructor is harder to make a unit test, and we will probably add try-except in `run` to capture the exception message and send it to slack or somewhere else.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,133 @@
...
+            self._logger.info("Advisory status not in REL_PREP yet ...")
+            return
+        if advisory.errata_state == "SHIPPED_LIVE":
+            self._logger.info("Advisory status already in SHIPPED_LIVE, update subtask 9 ...")
+            self._update_jira(self.parent_jira_key, 8, "Advisory status already in SHIPPED_LIVE")
+        self._logger.info("Advisory status already in post REL_PREP, update subtask 7 ...")
+        self._update_jira(self.parent_jira_key, 6, "Advisory status already in REL_PREP")
+
+        et_builds = advisory.errata_builds
+        sdk_build = [i for i in et_builds[f'OSE-{self.version}-RHEL-8']
```
vfreex, line 42: I think you could iterate all product versions here. We will have rhel 9 builds and we don't want this to break.

vfreex: Otherwise, you could use `elliott -i $something find-builds -k image` to find the NVR.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,133 @@
...
+                     if re.search("openshift-enterprise-operator-sdk-container*", i)]
+        if not sdk_build:
+            self._logger.info("No SDK build to ship, update subtask 8 then close ...")
+            self._update_jira(self.parent_jira_key, 7,
+                              f"No SDK build to ship, operator_sdk_sync job: {os.environ.get('BUILD_URL')}")
+            return
+
+        build = koji.ClientSession(constants.BREW_SERVER).getBuild(sdk_build[0])
+        archlist = ["amd64", "arm64", "ppc64le", "s390x"]
```
vfreex, line 51: `archlist` can be get from ImageMetadata, or we can just query it through Brew buildInfo API.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,133 @@
...
+                          f"operator_sdk_sync job: {os.environ.get('BUILD_URL')}")
+
+    def get_ad_jira_key(self, assembly, release_yaml):
+        if 'group' in release_yaml['releases'][assembly]['assembly'].keys():
+            extra_ad_id = release_yaml['releases'][assembly]['assembly']['group']['advisories']['extras']
+            parent_jira_key = release_yaml['releases'][assembly]['assembly']['group']['release_jira']
+            return extra_ad_id, parent_jira_key
+        if 'assembly' not in release_yaml['releases'][assembly]['assembly']['basis'].keys():
+            raise ValueError("Can not find jira and advisory number from assembly")
+        return self.get_ad_jira_key(release_yaml['releases'][assembly]['assembly']['basis']['assembly'], release_yaml)
```
vfreex, line 66: You can use `doozer config:read-group` or `load_group_config` in util.py.

Ximinhan: I think load_group_config gets group.yml while we need extra advisory id and parent jira id

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
Ximinhan: this job is only for operator-sdk usage, this image only have rhel8 product list now, if product list changed, we will get a request

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
Ximinhan: this job is for operator-sdk usage, sdk should support full arch

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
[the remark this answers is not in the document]
Ximinhan: In the following cmd, I will need to execute multiple shell commands and connect the input and output. If the async method is used, it will split the command and combine it with the communicate() method. I still think the current method is more straightforward for this job.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
vfreex: No. `load_group_config` calls `doozer config:read-group` internally, which merges overrides in releases.yml.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,148 @@
...
+    def _extract_binaries(self, arch, sdkVersion, build):
+        output = subprocess.getoutput(f"oc image info --filter-by-os {arch} -o json {build} | jq .digest")
+        shasum = re.findall("sha256:\\w*", output)[0]
+
+        rarch = arch
+        rarch = 'x86_64' if arch == 'amd64' else rarch
+        rarch = 'aarch64' if arch == 'arm64' else rarch
+        tarballFilename = f"{self.sdk}-{sdkVersion}-linux-{rarch}.tar.gz"
+
+        cmd = f"rm -rf ./{rarch} && mkdir ./{rarch}" + \
```
vfreex, line 94: If the shell script is long, I would recommend to keep the whole script in a multi-line string:
cmd = """#!/bin/bash
# some command
"""
or have the script stored in a separate file.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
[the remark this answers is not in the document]
vfreex: exectools in pyartcd passes almost all parameters to the underlying asyncio.create_subprocess_exec function. It may not have a shell parameter.
We can always extend those functions to call asyncio.create_subprocess_shell if we want the async functionality.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
vfreex: I wouldn't assume that. In case we only built a few arches, or the arch list changes in the future, we may forget to update here.
If you want to keep this simple, I would recommend to keep this a build parameter with default value of the current archlist.

locriandev (approved):

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,148 @@
...
+    def _sync_mirror(self, arch):
+        extra_args = "--exclude '*' --include '*.tar.gz'"
+        local_dir = f"./{arch}/"
+        if self.prerelease:
+            s3_path = f"/pub/openshift-v4/{arch}/clients/operator-sdk/pre-release/"
+        else:
+            s3_path = f"/pub/openshift-v4/{arch}/clients/operator-sdk/{self.assembly}/"
+        cmd = f"aws s3 sync --no-progress --exact-timestamps {extra_args} --delete {local_dir} s3://art-srv-enterprise{s3_path}"
+        self._logger.info(cmd)
+        subprocess.run(cmd, shell=True)
```
vfreex, line 118: The exit code is not captured here.

vfreex (changes requested):

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
Ximinhan: we don't need to capture exit code, if any command fails in cmd it should fail the job and we need to check the print logs.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,148 @@
...
+        if self.updatelatest:
+            s3_path_latest = f"/pub/openshift-v4/{arch}/clients/operator-sdk/latest/"
+            cmd = f"aws s3 sync --no-progress --exact-timestamps {extra_args} --delete {local_dir} s3://art-srv-enterprise{s3_path_latest}"
+            self._logger.info(cmd)
+            subprocess.run(cmd, shell=True)
```
vfreex, line 123: same here

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,148 @@
...
+        rarch = 'x86_64' if arch == 'amd64' else rarch
+        rarch = 'aarch64' if arch == 'arm64' else rarch
+        tarballFilename = f"{self.sdk}-{sdkVersion}-linux-{rarch}.tar.gz"
+
+        cmd = f"rm -rf ./{rarch} && mkdir ./{rarch}" + \
+              f" && oc image extract {constants.OPERATOR_URL}@{shasum} --path /usr/local/bin/{self.sdk}:./{rarch}/ --confirm" + \
+              f" && chmod +x ./{rarch}/{self.sdk} && tar -c --preserve-order -z -v --file ./{rarch}/{tarballFilename} ./{rarch}/{self.sdk}" + \
+              f" && ln -s {tarballFilename} ./{rarch}/{self.sdk}-linux-{rarch}.tar.gz && rm -f ./{rarch}/{self.sdk}"
+        self._logger.info(cmd)
+        subprocess.run(cmd, shell=True)
```
vfreex, line 99: same here

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,148 @@
...
+        if arch == 'amd64':
+            tarballFilename = f"{self.sdk}-{sdkVersion}-darwin-{rarch}.tar.gz"
+            cmd = f"oc image extract {constants.OPERATOR_URL}@{shasum} --path /usr/share/{self.sdk}/mac/{self.sdk}:./{rarch}/ --confirm" + \
+                  f" && chmod +x ./{rarch}/{self.sdk} && tar -c --preserve-order -z -v --file ./{rarch}/{tarballFilename} ./{rarch}/{self.sdk}" + \
+                  f" && ln -s {tarballFilename} ./{rarch}/{self.sdk}-darwin-{rarch}.tar.gz && rm -f ./{rarch}/{self.sdk}"
+            self._logger.info(cmd)
+            subprocess.run(cmd, shell=True)
```
vfreex, line 106: same here

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
vfreex: > if any command fails in cmd it should fail the job

That is not true. There are 2 issues with using shell script in `subprocess.run`:
- Unless `check=True` is set, `subprocess.run` won't raise an exception on error.
- If multiple commands are executed in a shell script, some shell options like `set -eo pipefail` should be set otherwise the shell script won't abort or return non-zero.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,148 @@
...
+              help="The name of an assembly. e.g. 4.9.1")
+@click.option("--nvr", metavar="BUILD_NVR", required=False,
+              help="Pin specific Build NVR")
+@click.option("--prerelease", metavar="PRE_RELEASE", is_flag=True, required=False,
+              help="Use pre-release as directory name.")
+@click.option("--updatelatest", metavar="UPDATE_LATEST_SYMLINK", is_flag=True, required=False,
+              help="Update latest symlink on mirror")
+@pass_runtime
+@click_coroutine
+def tarball_sources(runtime: Runtime, group: str, assembly: str, nvr: str, prerelease: bool, updatelatest: bool):
```
vfreex, line 146: function name is wrong.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+    async def run(self):
+        if self.assembly:
+            group_config = await util.load_group_config(self.group, self.assembly, env=os.environ.copy())
+            self.extra_ad_id = group_config.get("advisories", {}).get("extras", 0)
+            self.parent_jira_key = group_config.get("release_jira")
+            advisory = Erratum(errata_id=self.extra_ad_id)
+            self._logger.info("Check advisory status ...")
+            if advisory.errata_state in ["QE", "NEW_FILES"]:
+                self._logger.info("Advisory status not in REL_PREP yet ...")
+                return
```
vfreex, line 38: If this is an error, an exception should be raised.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+            if advisory.errata_state == "SHIPPED_LIVE":
+                self._logger.info("Advisory status already in SHIPPED_LIVE, update subtask 9 ...")
+                self._update_jira(self.parent_jira_key, 8, "Advisory status already in SHIPPED_LIVE")
```
vfreex, line 41: This could cause an issue if the parent jira template changes. This could be improved by writing a keyword or label based simple search to find the sub jira.

If you don't have time to do it for now, I would suggest adding a comment to mention the caveat.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+            self._logger.info("Advisory status already in post REL_PREP, update subtask 7 ...")
+            self._update_jira(self.parent_jira_key, 6, "Advisory status already in REL_PREP")
```
vfreex, line 43: same here.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+
+            et_builds = advisory.errata_builds
+            sdk_build = [i for i in et_builds[f'OSE-{self.version}-RHEL-8'] if re.search("openshift-enterprise-operator-sdk-container*", i)]
+            if not sdk_build:
+                self._logger.info("No SDK build to ship, update subtask 8 then close ...")
+                self._update_jira(self.parent_jira_key, 7, f"No SDK build to ship, operator_sdk_sync job: {os.environ.get('BUILD_URL')}")
```
vfreex, line 49: same here.

vfreex: Also please use the function in Runtime to get the build URL so we will have a single place to change after moving to Tekton.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+                return
+            build = koji.ClientSession(constants.BREW_SERVER).getBuild(sdk_build[0])
+        elif self.nvr:
+            build = koji.ClientSession(constants.BREW_SERVER).getBuild(self.nvr)
+        else:
+            self._logger.info("no assembly or nvr provided")
+            return
```
vfreex, line 56: Shouldn't this be an exception?

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+
+        archlist = ["amd64", "arm64", "ppc64le", "s390x"]
+        sdkVersion = self._get_sdkversion(build['extra']['image']['index']['pull'][0])
+        self._logger.info(sdkVersion)
+        for arch in archlist:
+            self._extract_binaries(arch, sdkVersion, build['extra']['image']['index']['pull'][0])
+        if self.assembly:
+            self._update_jira(self.parent_jira_key, 7, f"operator_sdk_sync job: {os.environ.get('BUILD_URL')}")
```
vfreex, line 64: same here

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+        sdkversion = re.findall("v\\d.*-ocp", sdkvalue)[0]
+        return sdkversion
+
+    def _extract_binaries(self, arch, sdkVersion, build):
+        output = subprocess.getoutput(f"oc image info --filter-by-os {arch} -o json {build} | jq .digest")
+        shasum = re.findall("sha256:\\w*", output)[0]
+
+        rarch = arch
+        rarch = 'x86_64' if arch == 'amd64' else rarch
+        rarch = 'aarch64' if arch == 'arm64' else rarch
```
vfreex, line 81: Use `util.brew_arch_for_go_arch`.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,137 @@
...
+            s3_path = f"/pub/openshift-v4/{arch}/clients/operator-sdk/{self.assembly}/"
+        cmd = f"aws s3 sync --no-progress --exact-timestamps {extra_args} --delete ./{arch}/ s3://art-srv-enterprise{s3_path}"
+        self.exec_cmd(cmd)
+        if self.updatelatest:
+            s3_path_latest = f"/pub/openshift-v4/{arch}/clients/operator-sdk/latest/"
+            cmd = f"aws s3 sync --no-progress --exact-timestamps {extra_args} --delete ./{arch}/ s3://art-srv-enterprise{s3_path_latest}"
+            self.exec_cmd(cmd)
+
+    def exec_cmd(self, cmd):
+        self._logger.info(cmd)
```
vfreex, line 111: Log message should be improved, such as `About to run command %s`.

vfreex (commented):
I added more comments.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
vfreex: I am not convinced to hardcode the product version `OSE-{self.version}-RHEL-8` here. It is pretty easy to find the sdk build regardless of the product version.

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            raise ValueError("no assembly or nvr provided")
+
+        archlist = ["amd64", "arm64", "ppc64le", "s390x"]
+        sdkVersion = self._get_sdkversion(build['extra']['image']['index']['pull'][0])
+        self._logger.info(sdkVersion)
+        for arch in archlist:
+            self._extract_binaries(arch, sdkVersion, build['extra']['image']['index']['pull'][0])
+        if self.assembly:
+            self._jira_client.complete_subtask(self.parent_jira_key, 7, f"operator_sdk_sync job: {os.environ.get('BUILD_URL')}")
```
vfreex, line 62: Use `Runtime.get_job_run_url`

pyartcd/pyartcd/pipelines/operator_sdk_sync.py
```diff
@@ -0,0 +1,126 @@
...
+            if advisory.errata_state in ["QE", "NEW_FILES"]:
+                raise ValueError("Advisory status not in REL_PREP yet ...")
+            if advisory.errata_state == "SHIPPED_LIVE":
+                self._logger.info("Advisory status already in SHIPPED_LIVE, update subtask 9 ...")
+                self._jira_client.complete_subtask(self.parent_jira_key, 8, "Advisory status already in SHIPPED_LIVE")
+            self._logger.info("Advisory status already in post REL_PREP, update subtask 7 ...")
+            self._jira_client.complete_subtask(self.parent_jira_key, 6, "Advisory status already in REL_PREP")
+
+            et_builds = advisory.errata_builds
+            sdk_build = [i for i in et_builds[f'OSE-{self.version}-RHEL-8'] if re.search("openshift-enterprise-operator-sdk-container*", i)]
```
vfreex, line 45: I didn't see a compelling reason to hardcode the product version here. This can be improved by iterating all builds:
```suggestion
            sdk_build = [b for _, builds in et_builds.items() for b in builds if b.startswith('openshift-enterprise-operator-sdk-container')]
```

thiagoalessio (dismissed):
@Ximinhan please don't forget to squash all these intermediary commits before merging :pray:

vfreex (approved):
/lgtm

PR #3392, opened 2022-12-16
Microshift nightlies: Move to own assembly

joepvd:
When running a test build in assembly `test`, the microshift build will be found when creating the repository. If this is for a GA or RC release, the build will get signed. I think this should not be a side effect of running a test.

Not sure if anything depends on the tooling being in the test assembly.


vfreex:
There is no real technical difference between test or any assembly which is not explicitly defined from the perspective of doozer. I know some jenkins job will raise an error if an assembly is not explicitly defined but test assembly is whitelisted. No matter what assembly name is used, it will be signed with release key anyway after RC.

vfreex (approved):
/lgtm

PR #3394, opened 2022-12-20
Fix backup buildvm

locriandev:
- update buildvm{2} URLs
- break the pipeline in case the source host does not correspond to the Jenkins URL


joepvd (approved):
/lgtm

PR #3405, opened 2023-01-05
[ART-4864] prepare-release: Support microshift rpm

vfreex:
- Creates `microshift` dedicated advisory for 4.12+.
- Attaches microshift builds to that advisory.
- Instructs release artists to rerun `prepare-release` after `build-microshift` job finishes.

Requires https://github.com/openshift/elliott/pull/491

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -111,8 +111,10 @@ class BuildMicroShiftPipeline:
             message = f"Hi @release-artists , microshift for assembly {self.assembly} has been successfully built."
             if pr:
                 message += f"\nA PR to update the assembly definition has been created/updated: {pr.html_url}"
+                message += "\nTo publish the build to the pocket (if asked by QE), run <https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fupdate-microshift-pocket/|update-microshift-pocket> job after the PR is merged."
+                message += f"\nTo attach the build to Errata, run <https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fprepare-release/|prepare-release> job or `elliott --group {self.group} --assembly {self.assembly} --rpms microshift find-builds -k rpm --member-only --use-default-advisory microshift` after the PR is merged."
                 if assembly_type is AssemblyTypes.PREVIEW:
-                    message += "\n This is an EC release. Please run <https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fmicroshift_sync/|microshift_sync> with UPDATE_PUB_MIRROR checked after the PR is merged."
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -183,50 +190,37 @@ async def run(self):
...
+                self._jira_client.start_task(jira_issue)
+            else:
+                _LOGGER.warning("[DRY RUN ]Would have updated Jira ticket status")
 
         _LOGGER.info("Updating ocp-build-data...")
         build_data_changed = await self.update_build_data(advisories, jira_issue_key)
 
         _LOGGER.info("Sweep builds into the the advisories...")
-        for kind, advisory in advisories.items():
+        for impetus, advisory in advisories.items():
```
thegreyd, line 217: Personally I like the variable name `kind` better than `impetus` in this case, it makes it obvious what I'm looking at. I'm not sure what the intent behind chosing impetus was in our other commands

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -735,11 +747,11 @@ def send_notification_email(self, advisories: Dict[str, int], jira_link: str):
               help="Do not filter our shipped builds, attach all builds to advisory")
 @pass_runtime
 @click_coroutine
-async def rebuild(runtime: Runtime, group: str, assembly: str, name: Optional[str], date: str,
+async def prepare_release(runtime: Runtime, group: str, assembly: str, name: Optional[str], date: str,
```
thegreyd, line 750: haha thanks for fixing this!

thegreyd:
Other than some minor things, lgtm

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -111,7 +111,8 @@ async def run(self):
             message = f"Hi @release-artists , microshift for assembly {self.assembly} has been successfully built."
             if pr:
                 message += f"\nA PR to update the assembly definition has been created/updated: {pr.html_url}"
-                message += "\nTo publish the build to the pocket, run update-microshift-pocket job after the PR is merged."
+                message += "\nTo publish the build to the pocket (if asked by QE), run <https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fupdate-microshift-pocket/|update-microshift-pocket> job after the PR is merged."
+                message += "\nTo attach the build to Errata, run <https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fprepare-release/|prepare-release> job after the PR is merged."
```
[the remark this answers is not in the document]
vfreex: sure, I will add that message for instructive purpose. But in the long term the preferred way should be using prepare-release to manage everything. prepare-release shouldn't have side effect to after a rerun (if it does, that is something we need to fix).

pyartcd/pyartcd/pipelines/prepare_release.py
vfreex: We already use `kind` and `impetus` interchangeably in our code. It is confusing. Sometimes `kind` means `rpm` or `image`, but sometimes it means `rpm` `image` `extras` `metadata`. We also use `impetus` for the later.

I am not sure if everyone feel comfortable if we use `impetus` for `rpm` `image` `extras` `metadata`.

pyartcd/pyartcd/pipelines/prepare_release.py
vfreex: LOL

PR #3406, opened 2023-01-05
Create job to accept a release

locriandev:


jobs/build/accept-release/Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    def dry_run = params.CONFIRM ? '' : '[DRY_RUN]'
+    currentBuild.displayName = "${params.RELEASE_NAME} ${dry_run}"
+
+    def confirm_param = params.CONFIRM ? "--confirm" : ''
+
+    buildlib.withAppCiAsArtPublish() {
+        res = commonlib.shell(
+            returnAll: true,
```
vfreex, line 76: With `returnAll`, it will not print the logs on the console. Would like to remove that option.

locriandev: Done

jobs/build/accept-release/Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+                  --arch ${params.ARCH} \
+                  --upgrade-url ${params.UPGRADE_URL} \
+                  --upgrade-minor-url ${params.UPGRADE_MINOR_URL} \
+                  ${confirm_param}
+                """,
+        )
+        if (res.returnStatus != 0) {
+            error("Script failed. See error above")
+        }
+    }
```
vfreex, line 88: `commonlib.shell` will throw an exception by default if we don't specify `returnAll`, so no need to check the return value (it will be `null` on success).

Removing `res` and this check should solve this.

vfreex:
Basically lgtm. After you fix the return value issue, i am ok with merging this.

jobs/build/accept-release/Jenkinsfile
locriandev: Fixed

PR #3412, opened 2023-01-09
[ART-5644] Add jenkins job microshift_sync

vfreex:
This is to make MicroShift dev-preview artifacts publicly available for 4.13+.

### DIFF
--- a/jobs/build/build-microshift/Jenkinsfile
+++ b/jobs/build/build-microshift/Jenkinsfile
@@ -37,11 +37,6 @@ node {
                             trim: true,
                         ),
                         booleanParam(
-                            name: "UPDATE_POCKET",
-                            description: "Update pocket on mirror; ",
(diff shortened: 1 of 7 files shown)


vfreex:
Test build with 4.12.0-rc.6: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Fmicroshift_sync/20/console

vfreex:
@jupierce yeah, we probably need some special logic since building and managing microshift doesn't really fit our current design model. It could make it a lot of easier if we can encode the basis event in the NVR so those builds don't have to be explicitly pinned. Previously I was working on a doc about future automation improvements but i didn't really finish it. I think I could open that doc up as soon as possible, and get this issue with some tasks in the doc into jira so we can start working on them.

PR #3414, opened 2023-01-12
[ART-3880] add stage publish_azure_marketplace

Ximinhan:
trigger publish_azure_marketplace job in rhcos_sync job when new rhcos is mirrored

### DIFF
--- a/jobs/build/rhcos_sync/Jenkinsfile
+++ b/jobs/build/rhcos_sync/Jenkinsfile
@@ -207,6 +207,18 @@ node {
                 """)
             }
         }
+
+        // run publish_azure_marketplace when publishing RHCOS bootimages
+        stage("Publish azure marketplace") {
+            build(
+                job: 'build%2publish_azure_marketplace',
+                propagate: false,
+                wait: true,
+                parameters: [
+                    string(name: 'VERSION', value: ocpVersion),
+                ]
+            )
+        }
     } catch ( err ) {
         slacklib.to(ocpVersion).say("""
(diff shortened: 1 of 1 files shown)


jobs/build/rhcos_sync/Jenkinsfile
```diff
@@ -207,6 +207,18 @@ node {
                 """)
             }
         }
+
+        // run publish_azure_marketplace when publishing RHCOS bootimages
+        stage("Publish azure marketplace") {
+            build(
+                job: 'build%2publish_azure_marketplace',
+                propagate: false,
```
joepvd, line 215: Any reason to have propagate as false?

vfreex, line 215: +1 with joepvd. Unless there is a reason, I refer to remove `propagate: false`

joepvd:
also, maybe should explicitly exclude 4.7 here, in case a fix would need a bootimage bump for 4.7.

Ximinhan:
this job was added after 4.10, I think it should apply for 4.10+

jobs/build/rhcos_sync/Jenkinsfile
Ximinhan: I thought it should not be a blocker for rhcos_sync, if rhcos sync step success and publish_azure_marketplace job failed, then the rhcos_sync job should success and we only need to take a look on marketplace job and fix and rebuild it.
anyway, I try to avoid rebuilding from rhcos_sync, but making rhcos_sync failed also works for me to raise alert.

vfreex (commented):
/lgtm

PR #3418, opened 2023-01-16
Port ocp4-scan to Python

locriandev:
Ref. https://issues.redhat.com/browse/ART-3673

Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/ocp4-scan/31/console

With this PR, `ocp4-scan` takes a single OCP version param (cannot scan multiple versions). This should be acceptable as automation only scans one version at the time. Scanning multiple versions can be done manually for testing purposes, but is discouraged and should not be done for frozen releases.

If changes/inconsistencies are found, `ocp4-scan` can trigger these builds:
- rhcos
- ocp4
- build-sync

The adopted pattern here is to trigger these jobs remotely using Jenkins API. A service account has been created for this purpose, along with the relevant token, and both are stored inside Jenkins credentials store.

Since the aforementioned jobs are part of a multibranch pipeline, they cannot be configured to receive remote triggers. As a workaround, a new folder named `triggered-jobs` has been created in Jenkins; these jobs are acting like proxies between the caller `ocp4-scan` Python job, and the called jobs.

This PR also moves `ocp4-scan` to `buildvm`. This makes triggering jobs via remote API easier. To compensate resource consumption, `build-sync` has been moved to `buildvm2`

### DIFF
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -1,0 +1,60 @@
+import base64
+import logging
+import os
+
+import aiohttp
+
+logger = logging.getLogger(__name__)
+
+
+async def trigger_jenkins_job(job_path: str, params=None):
+    """
+    Trigger a job using remote API calls.
+    :param job_path: relative path to the job, starting from <buildvm hostname>:<jenkins port>
+    :param params: optional dict containing job parameters
+    """
+
+    service_account = os.environ['JENKINS_SERVICE_ACCOUNT']
+    token = os.environ['JENKINS_SERVICE_ACCOUNT_TOKEN']
+
+    # Build authorization header
+    auth = base64.b64encode(f'{service_account}:{token}'.encode()).decode()
+    auth_header = f'Basic {auth}'
+    headers = {'Authorization': auth_header}
+
+    # Build url
+    # If the job to be triggered is parametrized, use 'buildWithParameters' and send HTTP data
+    # Otherwise, just call the 'build' endpoint with empty data
+    url = f'https://buildvm.hosts.prod.psi.bos.redhat.com:8443/{job_path}'
+    if params:
+        url += f'/buildWithParameters'
+    else:
+        url += f'/build'
+
+    # Call API endpoint
+    logger.info('Triggering remote job %s', job_path)
+    async with aiohttp.ClientSession(headers=headers) as session:
+        async with session.post(url, data=params) as response:
+            response.raise_for_status()
+            await response.text()
+
+
+async def trigger_ocp4(build_version: str):
+    await trigger_jenkins_job(
+        job_path='job/triggered-builds/job/ocp4',
+        params={'BUILD_VERSION': build_version}
+    )
+
+
+async def trigger_rhcos(build_version: str, new_build: bool):
+    await trigger_jenkins_job(
+        job_path='job/triggered-builds/job/rhcos',
+        params={'BUILD_VERSION': build_version, 'NEW_BUILD': new_build}
+    )
+
+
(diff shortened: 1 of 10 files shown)


pyartcd/pyartcd/jenkins.py
```diff
@@ -0,0 +1,67 @@
...
+        raise RuntimeError
+
+    # We need to use cURL --resolve option to pin the request to local host
+    # Doing so in pure Python requires a lot of hacking (see for example http://bit.ly/3ICezXw)
+    auth_header = base64.b64encode(f'{service_account}:{token}'.encode()).decode()
+    logger.info(f'auth_header: {auth_header}')
+
+    cmd = f'curl -X POST -H "Authorization: Basic {auth_header}" ' \
+          f'--resolve buildvm.hosts.prod.psi.bos.redhat.com:8443:127.0.0.1 ' \
+          f'https://buildvm.hosts.prod.psi.bos.redhat.com:8443/{job_path}'
```
vfreex, line 34: As a workaround, we can either adding `127.0.0.1 saml.buildvm.hosts.prod.psi.bos.redhat.com` to `/etc/hosts` to make it resolvable, or have a custom resolver if using `aiohttp`. But I am ok with this if this is simpler.

pyartcd/pyartcd/jenkins.py
```diff
@@ -0,0 +1,67 @@
...
+    else:
+        cmd += '/build'
+
+    logger.info('Triggering remote job %s', job_path)
+    await exectools.cmd_assert_async(cmd)
+
+
+async def trigger_ocp4(build_version: str):
+    await trigger_jenkins_job(
+        job_path='job/triggered-builds/job/ocp4',
```
vfreex, line 51: Is this a new job or I missed something?

pyartcd/pyartcd/pipelines/ocp4_scan.py
```diff
@@ -0,0 +1,107 @@
...
+    async def _get_changes(self):
+        """
+        Check for changes by calling doozer config:scan-sources
+        Changed rpms, images or rhcos are recorded in self.changes
+        self.rhcos_changed is also updated accordingly
+        """
+
+        # Run doozer scan-sources
+        cmd = f'doozer --assembly stream --working-dir={DOOZER_WORKING} --group=openshift-{self.version} ' \
+              f'config:scan-sources --yaml --ci-kubeconfig {os.environ["KUBECONFIG"]}'
```
vfreex, line 70: If `KUBECONFIG` is required, would like to add a check at the beginning of this job to fail earlier.

pyartcd/pyartcd/jenkins.py
locriandev: Yeah, I explained this in the PR description:

> Since the aforementioned jobs are part of a multibranch pipeline, they cannot be configured to receive remote triggers. As a workaround, a new folder named triggered-jobs has been created in Jenkins; these jobs are acting like proxies between the caller ocp4-scan Python job, and the called jobs.

pyartcd/pyartcd/pipelines/ocp4_scan.py
locriandev: Added an explicit check in 6542f9863c7a76e393cf47250ec816828e90315b
The code will raise a RuntimeError if `KUBECONFIG` is not defined

pyartcd/pyartcd/jenkins.py
locriandev: Done

joepvd (approved):
/lgtm

Let's take it for a spin 💪🏼

PR #3437, opened 2023-01-20
rhcos_sync: allow alternative meta.json

sosiouxme:
I expected it to work like this already, turns out SYNC_LIST was expected to be a list of files, so this makes it act like meta.json if it ends with .json


sosiouxme:
[tested](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Frhcos_sync/52/console)

PR #3463, opened 2023-01-26
Use string representation of weekday instead

locriandev:
Small improvement: get rid of non-intuitive weekday integer representation, by using Python `strftime` format code `%A`.
As stated [in the docs](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes), this will return the weekday as locale’s full name


joepvd (approved):
/lgtm

PR #3469, opened 2023-01-30
[ART-5903] add openvswitch dependency to mirror

Ximinhan:
add openvswitch-selinux-extra-policy and openvswitch2.17 from rhel-8-fast-datapath repo to package list


vfreex (approved):
/lgtm

PR #3483, opened 2023-02-01
plashet: Allow previous for rhcos consumed rpms

joepvd:
RHCOS passes a package lock from the x86_64 build on to other arches. If the ose repo has changed in the mean while, the build will fail.

This change allows two builds from rhcos consumed rpms to be included in the internal ose repo, so the builds will not choke on that.


pipeline-scripts/buildlib.groovy
```diff
@@ -1446,6 +1446,17 @@ def buildBuildingPlashet(version, release, el_major, include_embargoed, auto_sig
                     "--include-previous-for ovn", // this is a package prefix
                     "--include-previous-for haproxy", // avoid chicken and egg issues with base image & haproxy bumps
                     "--include-previous-for cri-o", // Fix regression in 4.10 CI
+                    // Allow previous for rhcos, so consecutive arch builds have the builds from their locks available:
+                    "--include-previous-for podman",
+                    "--include-previous-for crun",
+                    "--include-previous-for skopeo",
+                    "--include-previous-for conmon",
+                    "--include-previous-for conmon-rs",
+                    "--include-previous-for openshift",
```
sosiouxme, line 1455:
```suggestion
                    "--include-previous-for openshift-hyperkube",
```

sosiouxme (approved):
not sure we got them all, or if this is even a good idea, but let's give it a shot.
/lgtm

PR #3485, opened 2023-02-02
add greenwave test check in prepare release

Ximinhan:
use the feature of https://github.com/openshift/elliott/pull/503, check greenwave test status after prepare-release complete, alert in slack if there are failed greenwave test need to check

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -82,6 +82,7 @@ class PrepareReleasePipeline:
                 raise ValueError("default_advisories cannot be set for a non-stream assembly.")
 
         self.release_date = date
+        self._slack_client = self.runtime.new_slack_client()
         self.package_owner = package_owner or self.runtime.config["advisory"]["package_owner"]
         self.working_dir = self.runtime.working_dir.absolute()
         self.default_advisories = default_advisories
@@ -267,6 +268,14 @@ class PrepareReleasePipeline:
                 self.change_advisory_state(advisory, "QE")
             except CalledProcessError as ex:
                 _LOGGER.warning(f"Unable to move {impetus} advisory {advisory} to QE: {ex}")
+
+        _LOGGER.info("Check failed greenwave tests ...")
+        self._slack_client.bind_channel(self.release_name)
+        for advisory in [advisories["image"], advisories["extras"], advisories["metadata"]]:
+            test_result = await self.check_failed_greenwave(advisory)
+            if test_result != '[]':
+                _LOGGER.warning(f"Some greenwave tests on {advisory} failed with {test_result}")
+                await self._slack_client.say(f"Some greenwave tests failed on {advisory} @release-artists")
 
     async def load_releases_config(self) -> Optional[None]:
         repo = self.working_dir / "ocp-build-data-push"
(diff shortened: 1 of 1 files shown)


vfreex (commented):
Is there a ticket (or Slack thread, email, etc) describing why we need to do this?

Also if it is just sends a warning message on greenwave failure, what do you expect a release artist to do? Without a concrete instruction, I think people will probably just ignore it.

Ximinhan:
it's an improvement alert, no request ticket
the error can be hidden in the job's log instead of messages in API, we should do a precheck on the failed test to see if it is indeed a failure or need to retrigger. it can save time for qe feedback to us

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -268,6 +269,14 @@ async def run(self):
             except CalledProcessError as ex:
                 _LOGGER.warning(f"Unable to move {impetus} advisory {advisory} to QE: {ex}")
 
+        _LOGGER.info("Check failed greenwave tests ...")
+        self._slack_client.bind_channel(self.release_name)
+        for advisory in [advisories["image"], advisories["extras"], advisories["metadata"]]:
+            test_result = get_advisory_greenwave(advisory)
+            if test_result != []:
+                _LOGGER.warning(f"Some greenwave tests on {advisory} failed with {test_result}")
```
thegreyd, line 277: Can we display the number of failing tests here

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -268,6 +269,14 @@ async def run(self):
             except CalledProcessError as ex:
                 _LOGGER.warning(f"Unable to move {impetus} advisory {advisory} to QE: {ex}")
 
+        _LOGGER.info("Check failed greenwave tests ...")
+        self._slack_client.bind_channel(self.release_name)
```
thegreyd, line 273: Would this be posted in a separate message in the release channel or as a single message in the prepare-release thread? Would prefer the second option

Ximinhan: not in thread

pyartcd/pyartcd/pipelines/prepare_release.py
Ximinhan: data.page.result_count should have this value but I think to know the number doesn't provide much value we need to check anyway

pyartcd/pyartcd/pipelines/prepare_release.py
sosiouxme: the slack thread is created in the Jenkinsfile so there's no good way to access it in python. i think pulling that into python would be a good separate change. then this could be a thread response.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -268,6 +269,14 @@ async def run(self):
...
+        for advisory in [advisories["image"], advisories["extras"], advisories["metadata"]]:
+            test_result = get_advisory_greenwave(advisory)
+            if test_result != []:
+                _LOGGER.warning(f"Some greenwave tests on {advisory} failed with {test_result}")
+                await self._slack_client.say(f"Some greenwave tests failed on {advisory} @release-artists")
```
sosiouxme, line 278: I'd really like to have a link here, so I can be lazy.
```suggestion
                await self._slack_client.say(f"Some greenwave tests failed on https://errata.devel.redhat.com/advisory/{advisory}/test_run/greenwave_cvp @release-artists")
```

sosiouxme (approved):
room for improvement but i see no reason to delay for that.

PR #3488, opened 2023-02-06
plashet: Don't include previous for rust-afterburn, buildah, skopeo

joepvd:
With the change to include previous builds for rpms that rhcos consumes, errata tool takes issue. It refuses to sign builds that are older than builds already signed.

This PR removes the offenders. These builds do not tend to change very frequently, so this tactical retreat seems fine.


thiagoalessio (approved):
/lgtm

PR #3492, opened 2023-02-08
ART-3395 Accept release job enhancements

thegreyd:
https://issues.redhat.com/browse/ART-3395

Summary:
- Use release_tool.py script instead of our hack script
- Allow rejecting a release

This should give our serviceaccount the needed roles https://github.com/openshift/release/pull/36192

Hack run: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Faccept-release/34/console

TODO: Figure out how to fetch/store script (and requirements.txt) - leaning towards a dumb solution of keeping local copy, comparing to remote and error out if outdated. We don't expect the files to be updated, should be once / twice in a quarter or less.


locriandev:
Would be nice to have a test run. Maybe we can trigger a nightly for an old version (e.g. 4.7) and reject it.
Is the `accept` functionality going to work on nightlies too, or just on named releases like before?

vfreex:
This basically LGTM. Would be nice if there is a test run but I am ok with merging this and test in production.

PR #3497, opened 2023-02-10
not check microshift advisory in promote job

Ximinhan:
microshift advisory should be empty at this step, no need to check it here otherwise elliott will return failed


joepvd (changes requested):
Think we want to have all the tests run on the microshift advisory.

The error that this pr intends to address is that it fails when moving the advisory to QE. This is a best-effort move, if it fails, the promotion continues. 

We _could_ target the exclusion better by taking out the microshift rpm for the `elliott change-state` command, but not sure if we want that. The error is harmless, and the attempt is good.

Ximinhan:
why do we need to change microshift advisory to QE before building microshfit rpm?

vfreex:
promote-assembly can be rerun any time. We won't assume microshift advisory has builds attached or not.

PR #3507, opened 2023-02-16
[ART-6052] [WIP] update build_microshift

Ximinhan:
build microshift now become a regular requirement since 4.12+, and should update release.yml and attach builds to advisory automatically.

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -14,8 +14,9 @@ import click
 from doozerlib.assembly import AssemblyTypes
 from doozerlib.util import (brew_arch_for_go_arch,
                             isolate_nightly_name_components)
-from ghapi.all import GhApi
-from ruamel.yaml import YAML
+import ruamel.yaml
+import sys
+import io
 from semver import VersionInfo
 
 from pyartcd import constants, exectools, oc, util
@@ -26,9 +27,6 @@ from pyartcd.runtime import Runtime
 from pyartcd.util import (get_assembly_basis, get_assembly_type,
                           isolate_el_version_in_release, load_group_config,
                           load_releases_config)
-
-yaml = YAML(typ="rt")
-yaml.preserve_quotes = True
 
 
 class BuildMicroShiftPipeline:
@@ -68,7 +66,7 @@ class BuildMicroShiftPipeline:
     async def run(self):
         slack_client = None
         try:
-            await load_group_config(self.group, self.assembly, env=self._doozer_env_vars)
+            group_config = await load_group_config(self.group, self.assembly, env=self._doozer_env_vars)
             releases_config = await load_releases_config(Path(self._doozer_env_vars["DOOZER_WORKING_DIR"], "ocp-build-data"))
             assembly_type = get_assembly_type(releases_config, self.assembly)
             if assembly_type not in self.SUPPORTED_ASSEMBLY_TYPES:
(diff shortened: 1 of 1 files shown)


vfreex (commented):
/hold
Needs a closer review of this process before merge.

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -265,51 +282,26 @@ def _pin_nvrs(self, nvrs: List[str], releases_config) -> Dict:
...
+        releasefile=repo.get_contents("releases.yml", ref=branch)
+        yaml=ruamel.yaml.YAML()
+        yaml.preserve_quotes = True
+        releases_yaml = yaml.load(releasefile.decoded_content)
+        self._pin_nvrs(nvrs, releases_yaml) # insert nvrs into release.yaml
+        f=io.StringIO()
+        yaml.dump(releases_yaml,f)
+        f.seek(0)
+        commit=repo.update_file(releasefile.path, "update microshift", f.read(), releasefile.sha, branch=branch) # update releases.yaml in microshift branch
+        pr=repo.create_pull(title=title,body=body,head=f"openshift:{branch}", base={self.group}) # create pr from microshift to {group} branch
```
thegreyd, line 300: What's the rationale to create pr and then merge immediately? Might as well do a direct commit like we do when creating advisories?

thegreyd: I don't think we want `build-microshift` to be the one merging/committing, since we
could be running it manually - it would be ideal for promote job to do that 
- merge PR
- attach build
- attach bugs
- move advisory to qe

Ximinhan: AFAIK direct commit is not permitted in github repo's setting ? what we do in gen-assembly is also through pr like https://github.com/openshift/ocp-build-data/pull/2613

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -84,8 +82,8 @@ async def run(self):
                 if not self.payloads:
                     raise ValueError("Release payloads must be specified to rebase against assembly stream.")
                 payload_infos = await self.parse_release_payloads(self.payloads)
-                if "x86_64" not in payload_infos or "aarch64" not in payload_infos:
-                    raise ValueError("x86_64 payload and aarch64 payload are required for rebasing microshift.")
+                if "x86_64" not in payload_infos:
+                    raise ValueError("x86_64 payload is required for rebasing microshift.")
```
vfreex, line 86: We still need aarch64 pullspec to run the rebase script.

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -121,19 +119,17 @@ async def run(self):
...
+                    pr = await self._update_release_in_pr(nvrs)
+
+                # Sends a slack message
+                if slack_client:
+                    message = f"Assembly definition has been updated with {pr}"
+                    if assembly_type is AssemblyTypes.PREVIEW:
+                        message += "\n This is an EC release. Please run <https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fmicroshift_sync/|microshift_sync> with `UPDATE_PUB_MIRROR` checked."
+                    await slack_client.say(message, slack_thread)
+
+            self._attach_builds(group_config.get("advisories").get("microshift"))
```
vfreex, line 131: This job could be run to build microshift nightlies that are never intended to release.

pyartcd/pyartcd/pipelines/build_microshift.py
vfreex: Good point. Merging PRs or attaching builds doesn't seem to be the responsibility of build-microshift job.
Merging PRs could be done by promote-assembly job.
Also Justin proposed another option to include a Brew event in the microshift NVR and adjust Doozer to implicitly include that build without a pin but it is up to you to think about how to implement.

vfreex (changes requested):

pyartcd/pyartcd/pipelines/build_microshift.py
Ximinhan: if we don't need to pin and just for the release process we can get the build nvr which this job triggered and directly attach to the advisory
or I saw the latest [microshift-4.12.6-202303012057.p0.g50997a2.assembly.4.12.6.el8](https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2398501) have assembly value in version

vfreex: >if we don't need to pin and just for the release process we can get the build nvr which this job triggered and directly attach to the advisory
If we do like this, the builds will not be part of the assembly. That will violate the design. We need to have all builds managed by assemblies so they can be tracked and rebuilt by our toolings.

PR #3510, opened 2023-02-17
Add ability to exclude packages from plashet repo creation

joepvd:
Probably should be configured in the repo config in group.yml.


sosiouxme (changes requested):
we can block inheritance, and i'd rather do that and have what we mean rather than all this version-specific stuff in code

joepvd:
Agree. Filed [ticket](https://issues.redhat.com/browse/CWFCONF-3501), and updated the PR to not exclude this package, but keep the ability to exclude packages. Feel free to close or merge.

thegreyd:
Closing this, as this has been stale

PR #3517, opened 2023-02-21
poll-payload: Use kubeconfig and registry where needed

joepvd:
This PR aims to solve the doozer exception in [this job](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/scheduled-builds/job/poll-payload/129357/console), and gain a bit more intel on failures like [these](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/scheduled-builds/job/poll-payload/129356/console).


vfreex (approved):
/lgtm

PR #3534, opened 2023-02-28
ART-6054 Set advisory dependencies for ET release driver

thegreyd:
Uses elliott methods implemented in https://github.com/openshift/elliott/pull/511

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -178,7 +178,7 @@ class PrepareReleasePipeline:
                 else:
                     _LOGGER.info("Reusing existing microshift advisory %s", advisories["microshift"])
 
-        self.set_advisory_dependencies(advisories)
+        await self.set_advisory_dependencies(advisories)
 
         jira_issue_key = group_config.get("release_jira")
         jira_issue = None
@@ -234,7 +234,7 @@ class PrepareReleasePipeline:
         # attached to the advisory
         # currently for rpm advisory and cves only
         _LOGGER.info("Sweep bugs into the the advisories...")
-        self.sweep_bugs(check_builds=True)
+        self.sweep_bugs()
 
         _LOGGER.info("Adding placeholder bugs...")
         for impetus, advisory in advisories.items():
@@ -469,7 +469,6 @@ class PrepareReleasePipeline:
     def sweep_bugs(
         self,
         advisory: Optional[int] = None,
-        check_builds: bool = False,
     ):
         cmd = [
             "elliott",
@@ -478,8 +477,6 @@ class PrepareReleasePipeline:
             "--assembly", self.assembly,
             "find-bugs:sweep",
         ]
-        if check_builds:
-            cmd.append("--check-builds")
         if advisory:
             cmd.append(f"--add={advisory}")
         else:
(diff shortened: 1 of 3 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -620,6 +621,20 @@ def _render_jira_template(fields: Dict, template_vars: Dict):
             _LOGGER.warning("Failed to render JIRA template text: %s", ex)
         return fields
 
+    def set_advisory_dependencies(self, advisories):
+        target_kind = ['rpm', 'metadata']
+        expected_blocking = {advisories['image'], advisories['metadata']}
```
vfreex, line 626: A certain advisory might not exist. Would like to skip that dependency if one of the advisory kind doesn't exist.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -620,6 +621,20 @@ def _render_jira_template(fields: Dict, template_vars: Dict):
...
+        _LOGGER.info("Setting advisory dependencies for %s", target_kind)
+        for kind in target_kind:
+            target_advisory_id = advisories[kind]
+            blocking: Optional[List] = get_blocking_advisories(target_advisory_id)
+            if not blocking:
+                raise ValueError(f"Failed to fetch blocking advisories for {target_advisory_id} ")
+            if expected_blocking.issubset(set(blocking)):
+                continue
+            for blocking_advisory_id in expected_blocking:
+                set_blocking_advisory(target_advisory_id, blocking_advisory_id, "SHIPPED_LIVE")
```
vfreex, line 636: blocking_advisory or target_advisory could also be DROPPED. I think we need to skip those advisories too.

vfreex (changes requested):
There seems to be a few issues. Also PR needs rebase.

pyartcd/pyartcd/pipelines/prepare_release.py
thegreyd: This might be a problem.. we usually drop advisories after prepare-release completes, so by this time the dependency would already be set

thegreyd: This could be a logic for drop advisory job

thegreyd: Right now I am okay with this failing and displaying a warning

thegreyd: A slack message will also be sent

pyartcd/pyartcd/pipelines/prepare_release.py
thegreyd: Added

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -232,7 +234,7 @@ async def run(self):
         # attached to the advisory
         # currently for rpm advisory and cves only
         _LOGGER.info("Sweep bugs into the the advisories...")
-        self.sweep_bugs(check_builds=True)
```
thegreyd, line 235: elliott sweep now forces this by default, so remove this

pyartcd/pyartcd/pipelines/prepare_release.py
vfreex: Sounds good to me.

vfreex (approved):

vfreex (approved):

thiagoalessio:
please don't forget to squash that when merging

PR #3548, opened 2023-03-09
move slack notification from Jenkins file to python

Ximinhan:
test job:https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fprepare-release/4/consoleFull
in this test job, all slack messages was triggered from the python script, like https://redhat-internal.slack.com/archives/C02CJ4FNYSV/p1678258108178919 and https://redhat-internal.slack.com/archives/C02CJ4FNYSV/p1678259017632839

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -22,11 +22,13 @@ from jira.resources import Issue
 from pyartcd import exectools, constants
 from pyartcd.cli import cli, click_coroutine, pass_runtime
 from pyartcd.jira import JIRAClient
+from pyartcd.slack import SlackClient
 from pyartcd.mail import MailService
 from pyartcd.record import parse_record_log
 from pyartcd.runtime import Runtime
 from pyartcd.util import (get_assembly_basis, get_assembly_type,
-                          get_release_name_for_assembly)
+                          get_release_name_for_assembly,
+                          is_greenwave_all_pass_on_advisory)
 from ruamel.yaml import YAML
 from tenacity import retry, stop_after_attempt, wait_fixed
 
@@ -36,6 +38,7 @@ _LOGGER = logging.getLogger(__name__)
 class PrepareReleasePipeline:
     def __init__(
         self,
+        slack_client: SlackClient,
         runtime: Runtime,
         group: Optional[str],
         assembly: Optional[str],
@@ -83,6 +86,7 @@ class PrepareReleasePipeline:
 
         self.release_date = date
         self.package_owner = package_owner or self.runtime.config["advisory"]["package_owner"]
+        self._slack_client = slack_client
         self.working_dir = self.runtime.working_dir.absolute()
         self.default_advisories = default_advisories
         self.include_shipped = include_shipped
@@ -261,6 +265,7 @@ class PrepareReleasePipeline:
             self.send_notification_email(advisories, jira_issue_link)
 
         # Move advisories to QE
+        self._slack_client.bind_channel(self.release_name)
         for impetus, advisory in advisories.items():
             try:
                 if impetus == "metadata":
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -759,21 +762,30 @@ def send_notification_email(self, advisories: Dict[str, int], jira_link: str):
...
+            date=date,
+            nightlies=nightlies,
+            package_owner=package_owner,
+            jira_token=jira_token,
+            default_advisories=default_advisories,
+            include_shipped=include_shipped,
+        )
+        await pipeline.run()
+        await slack_client.say(f":white_check_mark: prepare-release for {name if name else assembly} completes.")
+    except Exception as e:
```
locriandev, line 789: I think this would catch ValueError at line 772, thus preventing the build from turning red. Is there a specific reason to catch broad exceptions now?

locriandev (changes requested):

pyartcd/pyartcd/pipelines/prepare_release.py
Ximinhan: line 772 is also a kind of exception, I think the pipeline won't run, and build job will be red if the code goes to 772, cat exception on the top level is to make our error slack alert include all possible failures in pipeline as we did in jenkinsfile finally segment

Ximinhan: And I think for other job with slack alert should also have such design instead of init slack client in pipeline.run() because exception may also happened in init step.

locriandev: This code:
```
    try:
        raise ValueError()
    except Exception:
        print('all good!')
```
will not raise an exception. You're catching everything, thus the pipeline won't run but the build will not be red. 

It's ok to initialize slack before the pipeline runs; what I think we shouldn't do is catch any exception randomly. Being more specific in scope and exception type is desirable, otherwise we might not notice when things break

Ximinhan: check line 791, it will forward the error to Jenkins and make build red like https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fprepare-release/6/consoleFull

Ximinhan: to catch and through again here is for checking overall status then said "the job failed" otherwise message needs to be append under each exception

locriandev (approved):

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -759,21 +762,30 @@ def send_notification_email(self, advisories: Dict[str, int], jira_link: str):
...
-        nightlies=nightlies,
-        package_owner=package_owner,
-        jira_token=jira_token,
-        default_advisories=default_advisories,
-        include_shipped=include_shipped,
-    )
-    await pipeline.run()
+    slack_client = runtime.new_slack_client()
+    slack_client.bind_channel(assembly)
+    await slack_client.say(f":construction: prepare-release for {name if name else assembly} :construction:")
```
vfreex, line 767: Would like to move all message sending code into `run` and keep this CLI function as simple as possible.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -85,6 +87,7 @@ def __init__(
         self.release_date = date
         self._slack_client = self.runtime.new_slack_client()
         self.package_owner = package_owner or self.runtime.config["advisory"]["package_owner"]
+        self._slack_client = slack_client
```
vfreex, line 90: 2 `self._slack_client` assignments here.

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -759,21 +762,30 @@ def send_notification_email(self, advisories: Dict[str, int], jira_link: str):
...
-        date=date,
-        nightlies=nightlies,
-        package_owner=package_owner,
-        jira_token=jira_token,
-        default_advisories=default_advisories,
-        include_shipped=include_shipped,
-    )
-    await pipeline.run()
+    slack_client = runtime.new_slack_client()
+    slack_client.bind_channel(assembly)
```
vfreex, line 766: Needs to be bind to `group` instead of `assembly` (which could be something like `rc.1`).

vfreex (changes requested):
I left a few comments.

pyartcd/pyartcd/pipelines/prepare_release.py
Ximinhan: we can't catch exceptions outside run like init function, catch it here we can make overall status accurate

vfreex (approved):

PR #3550, opened 2023-03-10
[ART-6052] remove pocket

Ximinhan:
according to ART-6052, pocket is no longer used for microshfit

### DIFF
--- a/jobs/build/microshift_sync/README.md
+++ b/jobs/build/microshift_sync/README.md
@@ -2,4 +2,4 @@ Publishes a specific version of the microshift RPM to mirror.openshift.com.
 
 A yum repo will be created or updated when this job is run.
 
-If `UPDATE_PUB_MIRROR` is checked,  the artifacts will be published to a public location on the mirror: `/pub/openshift-v4/<ARCH>/microshift/ocp-dev-preview/<assembly-name>/el?/os/Packages/...`.
+The artifacts will be published to a public location on the mirror for EC: `/pub/openshift-v4/<ARCH>/microshift/ocp-dev-preview/<assembly-name>/el?/os/Packages/...`.
--- a/jobs/build/microshift_sync/README.md
+++ b/jobs/build/microshift_sync/README.md
@@ -2,7 +2,4 @@ Publishes a specific version of the microshift RPM to mirror.openshift.com.
 
 A yum repo will be created or updated when this job is run.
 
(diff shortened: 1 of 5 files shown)


jobs/build/microshift_sync/README.md
```diff
@@ -2,7 +2,4 @@ Publishes a specific version of the microshift RPM to mirror.openshift.com.
 
 A yum repo will be created or updated when this job is run.
 
-If `UPDATE_POCKET` is checked, the artifacts will be published into a "pocket". `mirror.openshift.com/pockets/microshift/` is the pocket used. A pocket is delivery vehicle that is intended to provide artifacts to a select set of consumers
-without access to every other pocket or `/enterprise/*`.
-
-If `UPDATE_POCKET` is checked,  the artifacts will be published to a public location on the mirror: `/pub/openshift-v4/<ARCH>/microshift/ocp-dev-preview/<assembly-name>/el?/os/Packages/...`.
+If `UPDATE_PUB_MIRROR` is checked,  the artifacts will be published to a public location on the mirror: `/pub/openshift-v4/<ARCH>/microshift/ocp-dev-preview/<assembly-name>/el?/os/Packages/...`.
```
vfreex, line 5: I think if `UPDATE_PUB_MIRROR` is removed, we no long need `UPDATE_PUB_MIRROR`. Otherwise it will be a no-op

vfreex (changes requested):

jobs/build/microshift_sync/README.md
Ximinhan: we only remove pocket but still publish to mirror, like https://mirror.openshift.com/pub/openshift-v4/x86_64/microshift/ocp-dev-preview/4.13.0-ec.4/, isn't it?

vfreex: If UPDATE_PUB_MIRROR is checked, this will become a noop.

Ximinhan: I don't get it, we need `UPDATE_PUB_MIRROR` option for the job or not? or we confirmed  .../microshift/ocp-dev-preview/... is no longer needed and we can remove this part of logic?

vfreex: I mean removing `UPDATE_PUB_MIRROR` build parameter and treat it as true in the logic.

Ximinhan: ok, pr updated

PR #3553, opened 2023-03-13
[ART-6052] attach microshift build and check bugs

Ximinhan:
after build microshift complete, attach the build info advisory and sweep microshift bugs in promote job


vfreex (changes requested):
We shouldn't write new logic with Groovy.

Ximinhan:
implement in https://github.com/openshift-eng/aos-cd-jobs/pull/3563

PR #3561, opened 2023-03-17
promote: Fix `KeyError: 'digest'`

vfreex:
`digest` key doesn't exist in the constructed `dest_manifest_list`. We need to fetch the manifest list again from the registry after manifest list push to obtain the digest.


joepvd (approved):
/lgtm

PR #3564, opened 2023-03-17
Trigger 9.2 rhcos pipeline for 4.13, 4.14

joepvd:


jobs/build/rhcos/Jenkinsfile
```diff
@@ -82,9 +82,10 @@ node {
                     withCredentials([file(credentialsId: kubeconfigs['multi'], variable: 'KUBECONFIG')]) {
                         // we want to see the stderr as it runs, so will not capture with commonlib.shell;
                         // but somehow it is buffering the stderr anyway and [lmeyer] cannot figure out why.
+                        rhcos_version = ["4.13", "4.14", "4.15"].contains(params.BUILD_VERSION) ? "${params.BUILD_VERSION}-9.2" : params.BUILD_VERSION
```
thegreyd, line 85: Ideally we want to be consistent here and here:https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.13/group.yml#L73

thegreyd: Maybe [build-rhcos](https://github.com/openshift-eng/aos-cd-jobs/blob/master/pyartcd/pyartcd/pipelines/build_rhcos.py) should factor that logic in instead

thegreyd (changes requested):
Would like the rhel8/9 logic to be in build_rhcos.py

thegreyd:
Closing in favor of https://github.com/openshift-eng/aos-cd-jobs/pull/3582

PR #3574, opened 2023-03-23
Upload 3.11 plashets to ocp artifacts

locriandev:
Ref. https://issues.redhat.com/browse/ART-6395

[Tetst run ](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/ocp3/5/console) broke at final `sweep` stage, because the hack space wasn't setup to resolve the job path.

[no-image-latest](https://ocp-artifacts.hosts.prod.psi.rdu2.redhat.com/pub/RHOCP/plashets/3.11/no-image-latest/) symlink has been created instead of `latest` because the test run excluded all image builds for speed purposes.


build-scripts/ocp-artifacts/publish-oc-binary.sh
```diff
@@ -46,7 +46,7 @@ pkg_tar() {
 
 OSE_VERSION=$1
 VERSION=$2
-RPM_PATH=/mnt/rcm-guest/puddles/RHAOS/plashets/${OSE_VERSION}/building/%s/os/Packages
+RPM_PATH=/mnt/data/pub/RHOCP/plashets/${OSE_VERSION}/building/%s/os/Packages
```
vfreex, line 49: Although this is not strictly required, it would be nice if we follow the same hierarchy like:
```
https://ocp-artifacts.hosts.prod.psi.rdu2.redhat.com/pub/RHOCP/plashets/4.10/stream/el7/2023-03/202303221742/
https://ocp-artifacts.hosts.prod.psi.rdu2.redhat.com/pub/RHOCP/plashets/4.10/stream/el7/latest
```

locriandev: sounds good to me, maybe subject for a new PR though since this is just addressing the migration to ocp-artifacts

build-scripts/ocp-artifacts/3.11-plashet-to-ocp-artifacts.sh
```diff
@@ -26,7 +26,7 @@ fi
 FULL_VERSION="${VERSION}-${RELEASE}"
 
 BUILD_MODE="${3}"
-BASEDIR="/mnt/rcm-guest/puddles/RHAOS"
+BASEDIR="/mnt/data/pub/RHOCP"
```
vfreex, line 29: I doubt this script is still needed.

Historically, this script is used to push files on rcm-guest to the old mirror.openshift.com. It prepares some symlinks and eventually runs `push.enterprise.sh` on rcm-guest. After mirror.openshift.com was moved to S3, we deleted related code but code including preparing symlinks (http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/plashets/3.11/{latest|no-image-latest}) is still remained.

I saw no real use of those symlinks in our pipeline. In case people from other teams are still using them, we should redirect them to ocp-artifacts anyway.

So I would suggest removing this script.

jobs/build/ocp3/Jenkinsfile
```diff
@@ -732,13 +732,13 @@ node {
 
 
             // Push the building puddle out to the correct directory on the mirrors (e.g. online-int, online-stg, or enterprise-X.Y)
-            buildlib.invoke_on_rcm_guest("3.11-plashet-to-rcm-guest.sh", SYMLINK_NAME, NEW_FULL_VERSION, BUILD_MODE)
+            buildlib.invoke_on_ocp_artifacts("3.11-plashet-to-ocp-artifacts.sh", SYMLINK_NAME, NEW_FULL_VERSION, BUILD_MODE)
```
vfreex, line 735: As mentioned above, this might not be needed anymore.

pipeline-scripts/buildlib.groovy
```diff
@@ -624,16 +624,16 @@ def args_to_string(Object... args) {
...
 /**
- * We need to execute some scripts directly from the rcm-guest host. To perform
+ * We need to execute some scripts directly from the ocp-artifacts host. To perform
  * those operations, we stream the script into stdin of an SSH bash invocation.
  * @param git_script_filename  The file in build-scripts/rcm-guest to execute
  * @param args A list of arguments to pass to script
  * @return Returns the stdout of the operation
  */
-def invoke_on_rcm_guest(git_script_filename, Object... args ) {
+def invoke_on_ocp_artifacts(git_script_filename, Object... args ) {
```
vfreex, line 633: In the era of rcm-guest, we run `push.enterprise.sh` on rcm-guest to publish a local directory to the old mirror. Now both the old mirror and rcm-guest is decommissioned. Although we host plashet repos on ocp-artifacts, I don't see a reason to invoke any commands on ocp-artifacts.

pipeline-scripts/buildlib.groovy
```diff
@@ -1567,14 +1565,14 @@ def build_plashets(doozerOpts, version, release, dryRun = false) {
...
+        if (major > 4 || major == 4 && minor >= 6) { // 4.6+ already fully migrated; no need to create repos on ocp-artifacts.
             return plashets_built
         }
-        // Continue to build and sync plashet repos out to rcm-guest until RHCOS starts consuming content from ocp-artifacts.
+        // Continue to build and sync plashet repos out to ocp-artifacts until RHCOS starts consuming content from ocp-artifacts.
         plashets_built = [:]
     }
 
-    // produce plashet repos on rcm-guest
+    // produce plashet repos on ocp-artifacts
```
vfreex, line 1575: The code before this line is for producing plashet repos on ocp-artifacts, while the code after this line is for producing plashet repos on rcm-guest. A replace for host name here is not valid.

Since 4.1-4.5 has EOL, I think we can remove the legacy code after this line.

pipeline-scripts/buildlib.groovy
```diff
@@ -1567,14 +1565,14 @@ def build_plashets(doozerOpts, version, release, dryRun = false) {
         // Populate plashets_built
         plashets_built = readYaml(file: "$working_dir/plashets_built.yaml")
         echo "plashets_built: $plashets_built"
-        if (major > 4 || major == 4 && minor >= 6) { // 4.6+ already fully migrated; no need to create repos on rcm-guest.
+        if (major > 4 || major == 4 && minor >= 6) { // 4.6+ already fully migrated; no need to create repos on ocp-artifacts.
```
vfreex, line 1568: The comment here is valid. No need to replace the host name.

pipeline-scripts/buildlib.groovy
```diff
@@ -1453,55 +1453,53 @@ def buildBuildingPlashet(version, release, el_major, include_embargoed, auto_sig
...
-     * ocp-build is a user established for us on the rcm-guest host by
-     * Red Hat PnT devops. See /home/jenkins/.ssh.config.
      */
 
     // During an image build, doozer will provide repo files pointing back to a directory named
-    // 'building' in this rcm-guest directory. Before creating 'building', let's get the
+    // 'building' in this ocp-artifacts directory. Before creating 'building', let's get the
     // repo over there.
-    def destBaseDir = "/mnt/rcm-guest/puddles/RHAOS/plashets/${major_minor}"
+    def destBaseDir = "/mnt/data/pub/RHOCP/plashets/${major_minor}"
```
vfreex, line 1473: If you look at `buildlib.build_plashets` closely, `./hacks/plashet/build-plashet.py` is to create plashet repos on ocp-artifiacts while this function is untouched to continue to do for rcm-guest.

So I would like to delete this function after 3.11 is migrated. Also I would like to avoid to run command directly on ocp-artifacts if possible.

vfreex (changes requested):

jobs/build/ocp3/Jenkinsfile
```diff
@@ -732,13 +732,13 @@ node {
...
 
-            // 3.11-plashet-to-rcm-guest.sh sets up a different puddle name on rcm-guest and the mirrors
+            // 3.11-plashet-to-ocp-artifacts.sh sets up a different puddle name on rcm-guest and the mirrors
             PLASHET = "v${NEW_FULL_VERSION}_${PLASHET}"
             final mirror_url = "https://mirror.openshift.com/enterprise/enterprise-${params.BUILD_VERSION}"
 
-            buildlib.invoke_on_rcm_guest("publish-oc-binary.sh", params.BUILD_VERSION, NEW_FULL_VERSION)
+            buildlib.invoke_on_ocp_artifacts("publish-oc-binary.sh", params.BUILD_VERSION, NEW_FULL_VERSION)
```
vfreex, line 741: oc binaries can be extracted from the current working directory. I don't think it is good to do that remotely on ocp-artifacts. After the binaries are extracted, they can be sync'd to mirror immediately.

build-scripts/ocp-artifacts/3.11-plashet-to-ocp-artifacts.sh
locriandev: Fixed in e8db8cea361bb82124bee6bf6db92f44c1e7736e

pipeline-scripts/buildlib.groovy
locriandev: `rcm-guest` will be decommissioned and no longer exist, does it make sense to keep referring to it when it's gone?

jobs/build/ocp3/Jenkinsfile
locriandev: fixed in e8db8cea361bb82124bee6bf6db92f44c1e7736e

pipeline-scripts/buildlib.groovy
locriandev: We still run `public-oc-binary.sh` though. This script seems to store `oc` for different arches on the server it runs on. We used to run it on rcm-guest, where should we run it instead?

pipeline-scripts/buildlib.groovy
locriandev: I had noticed that, but I figured I would apply the least changes as possible. When 3.11 will go (and one day it will), we can simply remove ocp3 job and related functions

jobs/build/ocp3/Jenkinsfile
locriandev: ok got it, will work on this

pipeline-scripts/buildlib.groovy
vfreex: The comment says "4.6+ fully migrated to ocp-artifacts, so there is no need to create plashet repos on rcm-guest". 
It will not make sense if "rcm-guest" is replaced with "ocp-artifacts" in this sentence.

We can completely remove the comment and the lines after `return`.

pipeline-scripts/buildlib.groovy
vfreex: I am ok with doing a minimum change.

pipeline-scripts/buildlib.groovy
joepvd: We needed a jump host in the past for one reason or the other. That reason does not exist anymore. I think we can run everything on buildvm, and then copy the resulting artifacts over to ocp-artifacts / mirror with the existing means.

pipeline-scripts/buildlib.groovy
locriandev: Fixed in 73de9020807e67ae5cf78f9801dcf9fb9c361e05

pipeline-scripts/buildlib.groovy
locriandev: Fixed in 2af7e5965cfdfc480ffebb7e902032150ee109f4

pipeline-scripts/buildlib.groovy
locriandev: I think we can remove that script. OC binaries are already being pushed to our [s3 bucket](https://s3.console.aws.amazon.com/s3/buckets/art-srv-enterprise?region=us-east-1&prefix=pub/openshift-v3/clients/3.11.213/linux/&showversions=false) by [this script](https://github.com/openshift-eng/aos-cd-jobs/blob/master/jobs/build/ocp3/s3-publish-oc-binary.sh). Since we'd have to tell `oc` consumers to change the location they're pulling from, we can instruct them to use our AWS S3 bucket. We might need to create a user for that, but this is something I'd address when the need comes.

Done it in c93e14f66ec23d9a8b15977ef4c5ec5be60ca57f

jobs/build/ocp3/Jenkinsfile
locriandev: e1fa2a15ab86458f3356fd7b4ecdfb33366ee302 removed [publish-oc-binary.sh](https://github.com/openshift-eng/aos-cd-jobs/pull/3574/commits/c93e14f66ec23d9a8b15977ef4c5ec5be60ca57f).

locriandev:
Closed in favor of https://github.com/openshift-eng/aos-cd-jobs/pull/3581

PR #3576, opened 2023-03-23
ART-6354 Setup snyk

ashwindasr:
Steps
- Get the github token by following this [doc](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation) using openshift-art-build-bot credentials on bitwarden. Generate a JWT token first and using that generate a github token
- Add a secret to the repo by using GitHub API endpoint ([docs](https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-a-repository-secret)) _(but make sure to check if a secret with the same name doesn't exist, else the new value will overwrite the old one!)_
  - You will need the repository public key, which you can get from [here](https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#get-a-repository-public-key)
  - And also the encrypt the secret value using the instructions given in the [doc](https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-a-repository-secret).

Test run: https://github.com/ashwindasr/aos-cd-jobs/actions/runs/4503123895/jobs/7925900495

The `--all-projects` argument will tell snyk to look at all projects like `pyartcd` automatically and check their requirements.txt files.

### DIFF
--- a/.github/workflows/snyk_security.yaml
+++ b/.github/workflows/snyk_security.yaml
@@ -1,0 +1,25 @@
+name: Snyk Scan
+
+on: [ push, pull_request ]
+
+jobs:
+  snyk:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Test for open source vulnerabilities and license issues.
+        uses: snyk/actions/python-3.8@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+          COMMAND: "apt update -y && apt install -y rpm && pip install -r pyartcd/requirements.txt"
+        with:
+          command: test
+          args: --all-projects
+      - name: Test for any known security issues using Static Code Analysis.
+        uses: snyk/actions/python-3.8@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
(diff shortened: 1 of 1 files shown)


ashwindasr:
Created secret `{'name': 'SNYK_TOKEN', 'created_at': '2023-03-23T16:56:25Z', 'updated_at': '2023-03-23T16:56:25Z'}
`

.github/workflows/snyk_security.yaml
```diff
@@ -0,0 +1,25 @@
...
+          command: test
+          args: --all-projects
+      - name: Test for any known security issues using Static Code Analysis.
+        uses: snyk/actions/python-3.8@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+          COMMAND: "apt update -y && apt install -y rpm && pip install -r pyartcd/requirements.txt"
+        with:
+          command: code test
+          args: --all-projects
```
thegreyd, line 25: Curious what this does?

ashwindasr: So if we add a new project in the future, it will automatically find it and check it

PR #3583, opened 2023-03-28
Include 4.14 nightlies in poll-payload

locriandev:
Ref. https://issues.redhat.com/browse/ART-5724


vfreex:
Included in https://github.com/openshift-eng/aos-cd-jobs/pull/3611

PR #3586, opened 2023-03-29
Update to use latest oc-mirror and squash RUN commands

dmc5179:
oc-mirror went GA with OpenShift 4.11. The schema for oc-mirror change significantly between 4.10 and 4.11+. While the 4.10 oc-mirror version originally installed by this Dockerfile works for the oc-mirror commands used by the build today, it may not in the future. Consider moving to a GA version of oc-mirror.

Also combined all of the RUN commands into a single RUN command to speed things up a little.


bparry02:
The 4.10 version of oc-mirror is no longer parsing the current version of the operator index as expected. We must update to 4.11+

jupierce (approved):

PR #3606, opened 2023-04-06
Revert "[ART-4155] download cli tarball from GitHub for a commit and publish"

Ximinhan:
Reverts openshift-eng/aos-cd-jobs#3602


Ximinhan:
it should be fixed by https://github.com/openshift-eng/aos-cd-jobs/pull/3607

PR #3613, opened 2023-04-07
Fix promote job quotes and publish tarballs update

ashwindasr:
Publish source tarballs for cli, installer and operator registry

`"""`  and using `${}` allows us to interpolate groovy variables into groovy strings. But if that string contains bash script `$` we need to escape it using `\$`. Thats why the `oc_mirror_extract_cmd` never got an error before, with `"""`

`'''` doesn't let us interpolate groovy variables and keeps it as a raw groovy string.

### DIFF
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -501,37 +501,60 @@ def stagePublishClient(quay_url, from_release_tag, release_name, arch, client_ty
     def CLIENT_MIRROR_DIR="${BASE_TO_MIRROR_DIR}/${arch}/clients/${client_type}/${release_name}"
     sh "mkdir -p ${CLIENT_MIRROR_DIR}"
 
-    def tools_extract_cmd = "MOBY_DISABLE_PIGZ=true GOTRACEBACK=all oc adm release extract --tools --command-os='*' -n ocp " +
-                                " --to=${CLIENT_MIRROR_DIR} --from ${quay_url}:${from_release_tag}"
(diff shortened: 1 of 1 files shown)


sosiouxme (approved):

ashwindasr:
Tested in [hack space](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fpromote-assembly/7/console)

ashwindasr:
Test [run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fpromote-assembly/11/console)

ashwindasr:
Directly injected variables to env so as to not confuse groovy and bash interpolation

ashwindasr:
Test [run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fpromote-assembly/19/console)

sosiouxme:
/lgtm again
we can merge this now, or if you'd like to gain confidence in it we could use the hack space on a release this week to make sure it works first before merging.

ashwindasr:
Yes, that would be great. I'll update my hackspace and will keep back all the code I removed.

ashwindasr:
oc extract wants an empty directory to write to, but the source tarballs were written to the same directory, which caused an issue. Moving that piece of code to after the extract command.

PR #3624, opened 2023-04-13
move mirror-binaries to python

Ximinhan:
this pr is split from https://github.com/openshift-eng/aos-cd-jobs/pull/3577 , we think pr 3577 is a little big, so decouple the changes into pieces, you can check the review history in https://github.com/openshift-eng/aos-cd-jobs/pull/3577

this pr move mirror binaries step to python


pyartcd/pyartcd/util.py
```diff
@@ -267,3 +268,28 @@ async def is_build_permitted(version: str, data_path: str = constants.OCP_BUILD_
 
     # Fallback to default
     return True
+
+
+def get_release_controller_arch(releaseStreamName):
```
thegreyd, line 273: Let's keep the `lowercase_with_underscore` style for function params

thegreyd: Also I think we would have something in doozer/elliott doing this, maybe we import that here?

pyartcd/pyartcd/util.py
```diff
@@ -267,3 +268,28 @@ async def is_build_permitted(version: str, data_path: str = constants.OCP_BUILD_
...
+            arch = goArch
+    return arch
+
+
+def get_release_controller_url(releaseStreamName):
+    arch = get_release_controller_arch(releaseStreamName)
+    return f"https://{arch}.ocp.releases.ci.openshift.org"
+
+
+def print_dir_tree(path_to_dir):
```
thegreyd, line 287: `log_dir_tree` since we're logging here?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +400,211 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                # example file names:
+                #  - openshift-client-linux-4.3.0-0.nightly-2019-12-06-161135.tar.gz
+                #  - openshift-client-mac-4.3.0-0.nightly-2019-12-06-161135.tar.gz
+                #  - openshift-install-mac-4.3.0-0.nightly-2019-12-06-161135.tar.gz
+                #  - openshift-client-linux-4.1.9.tar.gz
+                #  - openshift-install-mac-4.3.0-0.nightly-s390x-2020-01-06-081137.tar.gz
+                #  ...
+                # So, match, and store in a group, any character up to the point we find -DIGIT. Ignore everything else
+                # until we match (and store in a group) one of the valid file extensions.
+                match = re.match(r'^([^-]+)((-[^0-9][^-]+)+)-[0-9].*(tar.gz|tgz|bz|zip)$', f)
```
thegreyd, line 597: It would be handy to have unit tests to test we're getting the name that we expect.
Would suggest either extracting it out in a separate func, and then testing that

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +400,211 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                f.write(f"{shasum} opm-{platform}-{release_name}.tar.gz")
+
+    async def publish_multi_client(self, working_dir, from_release_tag, release_name, client_type):
+        subprocess.run(f"docker login -u openshift-release-dev+art_quay_dev -p {os.environ['PASSWORD']} quay.io")
+        # Anything under this directory will be sync'd to the mirror
+        BASE_TO_MIRROR_DIR = f"{working_dir}/to_mirror/openshift-v4"
+        shutil.rmtree(BASE_TO_MIRROR_DIR, ignore_errors=True)
+        RELEASE_MIRROR_DIR = f"{BASE_TO_MIRROR_DIR}/multi/clients/{client_type}/{release_name}"
+
+        for goArch in util.goArches:
```
thegreyd, line 551: `go_arch` instead of `goArch`

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +400,211 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+            path_args.append(f'--path=/usr/bin/registry/{binary}:{client_mirror_dir}')
+        extract_release_binary(operator_registry, path_args)
+        # Compress binaries into tar.gz files and calculate sha256 digests
+        os.chdir(client_mirror_dir)
+        for idx, binary in enumerate(binaries):
+            platform = platforms[idx]
+            os.chmod(binary, 0o755)
+            with tarfile.open(f"opm-{platform}-{release_name}.tar.gz", "w:gz") as tar:  # archive file
+                tar.add(binary)
+            os.remove(binary)  # remove oc-mirror
```
thegreyd, line 537: `remove oc-mirror` ?

pyartcd/pyartcd/oc.py
```diff
@@ -35,3 +39,47 @@ async def registry_login(runtime: Runtime):
...
+        except Exception as e:
+            logger.error(tracker.get_result())
+            raise e
+
+
+def extract_release_client_tools(release_pullspec: str, path_arg: str, arch: str):
+    # oc adm release extract --tools --command-os='*' -n ocp --to=<workdir> --filter-by-os=<arch> --from <pullspec>
+    with oc.tracking() as tracker:
+        try:
+            r = Result("tools")
```
thegreyd, line 76: Curious, what does "tools" mean in this context?

jobs/build/promote-assembly/Jenkinsfile
```diff
@@ -184,34 +191,6 @@ node {
...
-        }
-        release_info.content.each { arch, info ->
-            echo "Mirroring client binaries for $arch"
-            def dest_release_tag = release.destReleaseTag(release_info.name, arch)
-            // Currently a multi/heterogeneous release payload has a modified release name to workaround a Cincinnati issue.
-            // Using the real per-arch release name in $info instead of the one defined by release artists.
-            def release_name = info.metadata.version
-            if (!params.DRY_RUN) {
-                if (arch != "multi") {
-                    release.stagePublishClient(quay_url, dest_release_tag, release_name, arch, client_type)
```
thegreyd, line 206: I don't see this func getting removed.. is anything else using it?
It would be helpful to compare them side by side

pyartcd/pyartcd/oc.py
Ximinhan: it's a self-defined operation name https://github.com/openshift/openshift-client-python/blob/main/packages/openshift/result.py#L8

jobs/build/promote-assembly/Jenkinsfile
Ximinhan: I saw it in pre-release job and release job, but seems theses jobs are not in use

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: remove the binary, keep the tarball

pyartcd/pyartcd/constants.py
```diff
@@ -17,5 +17,11 @@
     "ppc64le": "registry.ci.openshift.org/ocp-ppc64le/release-ppc64le",
     "aarch64": "registry.ci.openshift.org/ocp-arm64/release-arm64",
 }
+MIRROR_CLIENTS = {
```
thegreyd, line 20: Can't tell what this is used for. Would want a more descriptive name or comment above it

pyartcd/pyartcd/constants.py
```diff
@@ -17,5 +17,11 @@
...
+    "oc": "openshift-client",
+    "installer": "openshift-installer",
+    "operator-registry": "opm",
+}
 
 OCP_BUILD_DATA_URL = 'https://github.com/openshift-eng/ocp-build-data'
+QUAY_URL = "quay.io/openshift-release-dev/ocp-release"
```
thegreyd, line 27:
```suggestion
QUAY_RELEASE_REPO_URL = "quay.io/openshift-release-dev/ocp-release"
```

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +400,206 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
         self._logger.warn("Issue %s is permitted with justification: %s", err, justification)
         return justification
 
+    async def publish_client(self, working_dir, from_release_tag, release_name, arch, client_type):
+        subprocess.run(f"docker login -u openshift-release-dev+art_quay_dev -p {os.environ['PASSWORD']} quay.io")
```
locriandev, line 404: Why not use `exectools.cmd_assert_async` instead?

pyartcd/pyartcd/util.py
```diff
@@ -267,3 +268,28 @@ async def is_build_permitted(version: str, data_path: str = constants.OCP_BUILD_
...
+            arch = goArch
+    return arch
+
+
+def get_release_controller_url(release_stream_name):
+    arch = get_release_controller_arch(release_stream_name)
+    return f"https://{arch}.ocp.releases.ci.openshift.org"
+
+
+def log_dir_tree(path_to_dir):
```
locriandev, line 287: Just printing the folder structure could be confusing and hard to detect in a long log stream; I suggest prepending it with an explatatory message such as `Printing dir tree for <path>`

pyartcd/pyartcd/util.py
```diff
@@ -267,3 +268,28 @@ async def is_build_permitted(version: str, data_path: str = constants.OCP_BUILD_
...
+    for child in os.listdir(path_to_dir):
+        child_path = os.path.join(path_to_dir, child)
+        logger.info(child_path)
+
+
+def log_file_content(path_to_file):
```
locriandev, line 293: Same as above, would be nice to log an explanatory message such as `Printing file contents for <path>` beforehand

locriandev (changes requested):

pyartcd/pyartcd/constants.py
Ximinhan: add comments

pyartcd/pyartcd/util.py
Ximinhan: add info logs

pyartcd/pyartcd/util.py
Ximinhan: add info logs

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +402,212 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+            if image_stat == 0:  # image exists
+                _, image_info = get_release_image_info_from_pullspec(cli_pull_spec)
+                # Retrieve the commit from image info
+                commit = image_info["config"]["config"]["Labels"]["io.openshift.build.commit.id"]
+                source_url = image_info["config"]["config"]["Labels"]["io.openshift.build.source-location"]
+                source_name = source_url.split("/")[-1]
+                source_name = constants.MIRROR_CLIENTS[source_name]
+                # URL to download the tarball a specific commit
+                response = requests.get(f"{source_url}/archive/{commit}.tar.gz", stream=True)
+                if response.ok:
```
joepvd, line 432: How to deal with an error? No retry?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +402,212 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+            image_stat, cli_pull_spec = get_release_image_pullspec(f"{quay_url}:{from_release_tag}", tarball)
+            if image_stat == 0:  # image exists
+                _, image_info = get_release_image_info_from_pullspec(cli_pull_spec)
+                # Retrieve the commit from image info
+                commit = image_info["config"]["config"]["Labels"]["io.openshift.build.commit.id"]
+                source_url = image_info["config"]["config"]["Labels"]["io.openshift.build.source-location"]
+                source_name = source_url.split("/")[-1]
+                source_name = constants.MIRROR_CLIENTS[source_name]
+                # URL to download the tarball a specific commit
+                response = requests.get(f"{source_url}/archive/{commit}.tar.gz", stream=True)
```
joepvd, line 431: Do we need to download this for each architecture?

pyartcd/pyartcd/oc.py
```diff
@@ -35,3 +40,55 @@ async def registry_login(runtime: Runtime):
...
+    return res, json.loads(out)
+
+
+def extract_release_binary(image_pullspec: str, path_args: List[str]):
+    # oc image extract --confirm --only-files --path=/usr/bin/..:<workdir> <pullspec>
+    cmd_args = ['extract', '--confirm', '--only-files'] + path_args + [image_pullspec]
+    common_oc_wrapper("extract_image", "image", cmd_args, False, False)
+
+
+def get_release_image_pullspec(release_pullspec: str, image: str):
```
joepvd, line 82: Please add type declarations to the output (also of the other functions).

Think the first one should be a `bool` rather than an `int`, as it makes reading the resulting code easier.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +402,212 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                    shasum = hashlib.sha256(f.read()).hexdigest()
+                with open("sha256sum.txt", 'a') as f:  # write shasum to sha256sum.txt
+                    f.write(f"{shasum}  {dir}/sha256sum.txt\n")
+        os.chdir(current_path) # return to parent path
+        util.log_dir_tree(current_path)
+
+        # Publish the clients to our S3 bucket.
+        await exectools.cmd_assert_async(f"aws s3 sync --no-progress --exact-timestamps {BASE_TO_MIRROR_DIR}/ s3://art-srv-enterprise/pub/openshift-v4/", stdout=sys.stderr)
+
+    def create_symlink(self, path_to_dir, log_tree, log_shasum):
```
joepvd, line 580: Shouldn't this by async?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +402,212 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        # we expect to publish to mirror
+        CLIENT_MIRROR_DIR = f"{BASE_TO_MIRROR_DIR}/{arch}/clients/{client_type}/{release_name}"
+        os.makedirs(CLIENT_MIRROR_DIR)
+
+        # extract release clients tools
+        extract_release_client_tools(f"{quay_url}:{from_release_tag}", f"--to={CLIENT_MIRROR_DIR}", None)
+
+        # Get cli installer operator-registory pull-spec from the release
+        for tarball in ["cli", "installer", "operator-registry"]:
+            image_stat, cli_pull_spec = get_release_image_pullspec(f"{quay_url}:{from_release_tag}", tarball)
```
joepvd, line 422: Could you extract this into a function and put it under test?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -376,6 +402,212 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                            text = await response.text()
+                            with open(outputDestMd, 'w') as f:
+                                f.write(text)
+                    else:
+                        with open(outputDest, 'w') as f:
+                            f.write(f"<html><body><p>Changelog information cannot be computed for this release. Changelog information will be populated for new releases once {prevGA} is officially released.</p></body></html>")
+                        with open(outputDestMd, 'w') as f:
+                            f.write(f"Changelog information cannot be computed for this release. Changelog information will be populated for new releases once {prevGA} is officially released.")
+        except Exception as e:
+            self._logger.error("Error generating changelog for release")
```
joepvd, line 513: Could the error be part of the log?

joepvd (commented):
Left some remarks. This is not an exhaustive review.

Think it'd be great to:
- have return types in the method signatures
- extract functionality into their own methods
- put these methods under test
- have a careful review on what methods should be async

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: there is no async function in it, and considering it will change dir I would make it single thread

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: it will be in the job's console log but not the changelog

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: let's keep the behavior as groovy dose

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: let's keep the behavior as groovy dose

pyartcd/pyartcd/oc.py
Ximinhan: the result.status is an int value https://github.com/openshift/openshift-client-python/blob/main/packages/openshift/result.py#L29

Ximinhan:
tested with 
4.13.0-rc.8 https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fpromote-assembly/33/
4.11.40 https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fpromote-assembly/35/

pyartcd/pyartcd/oc.py
```diff
@@ -35,3 +40,55 @@ async def registry_login(runtime: Runtime):
...
+    # oc adm release info --image-for=<image> <pullspec>
+    cmd_args = ['release', 'info', f'--image-for={image}', release_pullspec]
+    return common_oc_wrapper("image_info_in_release", "adm", cmd_args, False, True)
+
+
+def extract_release_client_tools(release_pullspec: str, path_arg: str, arch: str):
```
vfreex, line 88: Let's to be more clear: `arch` should be `go_arch`.
Also if this is an optional parameter, the declaration could be `arch: Optional[str] = None`.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        CLIENT_MIRROR_DIR = f"{BASE_TO_MIRROR_DIR}/{arch}/clients/{client_type}/{release_name}"
+        os.makedirs(CLIENT_MIRROR_DIR)
+
+        # extract release clients tools
+        extract_release_client_tools(f"{quay_url}:{from_release_tag}", f"--to={CLIENT_MIRROR_DIR}", None)
+
+        # Get cli installer operator-registory pull-spec from the release
+        for tarball in ["cli", "installer", "operator-registry"]:
+            image_stat, cli_pull_spec = get_release_image_pullspec(f"{quay_url}:{from_release_tag}", tarball)
+            if image_stat == 0:  # image exists
```
vfreex, line 422: Actually this check is not strictly correct. If `image_stat == 0`, we know the image must exist. However, if not, it doesn't mean the image doesn't exist. Other errors (e.g. network failure or auth error) could cause non-zero image_stat too.

Since this matches the existing logic, @Ximinhan could you add a `FIXME` comment here so we won't forget to fix this?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                _, image_info = get_release_image_info_from_pullspec(cli_pull_spec)
+                # Retrieve the commit from image info
+                commit = image_info["config"]["config"]["Labels"]["io.openshift.build.commit.id"]
+                source_url = image_info["config"]["config"]["Labels"]["io.openshift.build.source-location"]
+                source_name = source_url.split("/")[-1]
+                source_name = constants.MIRROR_CLIENTS[source_name]
+                # URL to download the tarball a specific commit
+                response = requests.get(f"{source_url}/archive/{commit}.tar.gz", stream=True)
+                if response.ok:
```
vfreex, line 431: This file is not optional. Let's raise an exception if failing to download this file (calling `response.raise_for_status()`).

jobs/build/promote-assembly/Jenkinsfile
```diff
@@ -173,7 +176,12 @@ node {
         }
         echo "Will run ${cmd}"
         buildlib.withAppCiAsArtPublish() {
-            withCredentials([string(credentialsId: 'art-bot-slack-token', variable: 'SLACK_BOT_TOKEN'), string(credentialsId: 'jboss-jira-token', variable: 'JIRA_TOKEN'), string(credentialsId: 'jenkins-service-account', variable: 'JENKINS_SERVICE_ACCOUNT'), string(credentialsId: 'jenkins-service-account-token', variable: 'JENKINS_SERVICE_ACCOUNT_TOKEN')]) {
+            withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'creds_dev_registry.quay.io', usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD'],
```
vfreex, line 179: `USERNAME` and `PASSWORD` are too vague. Each need a more descriptive name.

vfreex (changes requested):

pyartcd/pyartcd/constants.py
```diff
@@ -17,8 +17,15 @@
     "ppc64le": "registry.ci.openshift.org/ocp-ppc64le/release-ppc64le",
     "aarch64": "registry.ci.openshift.org/ocp-arm64/release-arm64",
 }
+# The client name mapping in nightly with name on mirror
+MIRROR_CLIENTS = {
```
vfreex, line 21: This doesn't see to be a constant that needs to be used in other modules. I think it is better to be moved to `promote.py`.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
         self._logger.warn("Issue %s is permitted with justification: %s", err, justification)
         return justification
 
+    async def publish_client(self, working_dir, from_release_tag, release_name, arch, client_type):
+        _, minor = util.isolate_major_minor_in_group(self.group)
+        quay_url = constants.QUAY_RELEASE_REPO_URL
+        # Anything under this directory will be sync'd to the mirror
+        BASE_TO_MIRROR_DIR = f"{working_dir}/to_mirror/openshift-v4"
```
vfreex, line 408: Better to be changed to lower case since this is not a constant.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        shutil.rmtree(f"{BASE_TO_MIRROR_DIR}/{arch}", ignore_errors=True)
+
+        # From the newly built release, extract the client tools into the workspace following the directory structure
+        # we expect to publish to mirror
+        CLIENT_MIRROR_DIR = f"{BASE_TO_MIRROR_DIR}/{arch}/clients/{client_type}/{release_name}"
```
vfreex, line 413: Better to be changed to lower case since this is not a constant.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
         self._logger.warn("Issue %s is permitted with justification: %s", err, justification)
         return justification
 
+    async def publish_client(self, working_dir, from_release_tag, release_name, arch, client_type):
+        _, minor = util.isolate_major_minor_in_group(self.group)
+        quay_url = constants.QUAY_RELEASE_REPO_URL
+        # Anything under this directory will be sync'd to the mirror
+        BASE_TO_MIRROR_DIR = f"{working_dir}/to_mirror/openshift-v4"
+        shutil.rmtree(f"{BASE_TO_MIRROR_DIR}/{arch}", ignore_errors=True)
```
vfreex, line 409: Let's not `ignore_errors` here. Would like to see an exception is raised if failing to delete the directory.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
         self._logger.warn("Issue %s is permitted with justification: %s", err, justification)
         return justification
 
+    async def publish_client(self, working_dir, from_release_tag, release_name, arch, client_type):
```
vfreex, line 404: Let's be more clear about what `arch` means here. It seems to be a Brew arch instead of Go arch.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+
+        if arch == 'x86_64':
+            # oc image  extract requires an empty destination directory. So do this before extracting tools.
+            # oc adm release extract --tools does not require an empty directory.
+            image_stat, oc_mirror_pullspec = get_release_image_pullspec(f"{quay_url}:{from_release_tag}", "oc-mirror")
+            if image_stat == 0:  # image exist
+                # extract image to workdir, if failed it will raise error in function
+                extract_release_binary(oc_mirror_pullspec, [f"--path=/usr/bin/oc-mirror:{CLIENT_MIRROR_DIR}"])
+                current_path = os.getcwd()
+                os.chdir(CLIENT_MIRROR_DIR)
```
vfreex, line 449: Avoid `os.chdir` if possible. This could lead to race condition if we run those functions with multithread or coroutine in the future.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        platforms = ['linux']
+        if arch == 'x86_64':  # For x86_64, we have binaries for macOS and Windows
+            binaries += ['darwin-amd64-opm', 'windows-amd64-opm']
+            platforms += ['mac', 'windows']
+        path_args = []
+        for binary in binaries:
+            path_args.append(f'--path=/usr/bin/registry/{binary}:{client_mirror_dir}')
+        extract_release_binary(operator_registry, path_args)
+        # Compress binaries into tar.gz files and calculate sha256 digests
+        os.chdir(client_mirror_dir)
```
vfreex, line 530: Avoid `os.chdir` if possible. This could lead to race condition if we run those functions with multithread or coroutine in the future.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        os.chdir(current_path)
+
+    async def publish_multi_client(self, working_dir, from_release_tag, release_name, client_type):
+        # Anything under this directory will be sync'd to the mirror
+        BASE_TO_MIRROR_DIR = f"{working_dir}/to_mirror/openshift-v4"
+        shutil.rmtree(f"{BASE_TO_MIRROR_DIR}/multi", ignore_errors=True)
+        RELEASE_MIRROR_DIR = f"{BASE_TO_MIRROR_DIR}/multi/clients/{client_type}/{release_name}"
+        current_path = os.getcwd()
+
+        for go_arch in util.go_arches:
```
vfreex, line 551: Arches should be read from group config then converted to go arches.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,215 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        util.log_dir_tree(RELEASE_MIRROR_DIR)
+
+        # Publish the clients to our S3 bucket.
+        await exectools.cmd_assert_async(f"aws s3 sync --no-progress --exact-timestamps {BASE_TO_MIRROR_DIR}/multi s3://art-srv-enterprise/pub/openshift-v4/multi", stdout=sys.stderr)
+
+    def create_symlink(self, path_to_dir, log_tree, log_shasum):
+        # External consumers want a link they can rely on.. e.g. .../latest/openshift-client-linux.tgz .
+        # So whatever we extract, remove the version specific info and make a symlink with that name.
+        current_path = os.getcwd() # /mnt/workspace/jenkins/working/build_promote-assembly
+        os.chdir(path_to_dir) # path_to_dir is relative path artcd_working/to_mirror/openshift-v4/aarch64/clients/ocp/4.13.0-rc.6
```
vfreex, line 586: Avoid os.chdir if possible. This could lead to race condition if we run those functions with multithread or coroutine in the future.

pyartcd/pyartcd/constants.py
Ximinhan: we will migrate client_sync job later, I think there will be other types of clients added here

pyartcd/pyartcd/oc.py
```diff
@@ -1,8 +1,13 @@
 import json
 import os
-
+import logging
 from pyartcd import exectools
 from pyartcd.runtime import Runtime
+import openshift as oc
```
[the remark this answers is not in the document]
Ximinhan: requirements already have openshift-client

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: if we don't chdir it will make symlink content in tarball incorrect like the issue https://redhat-internal.slack.com/archives/CJARLA942/p1683773060618819

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: I think it's not necessary, we also have multi in go arch list while group not

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: added else case

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: if we don't use chdir it will make symlink contents on mirror incorrect like https://redhat-internal.slack.com/archives/CJARLA942/p1683773060618819

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: same as above

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: if it's other errors which is similar to oc shell command return status not 0 then it will raise in oc lib and will not get here

pyartcd/pyartcd/pipelines/promote.py
vfreex: This is not true. symlinks have nothing to do with current directory or chdir call. What you need to do is setting the symlink link target to a relative path instead of full physical path.

pyartcd/pyartcd/pipelines/promote.py
vfreex: This is not true. symlinks have nothing to do with current directory or chdir call. What you need to do is setting the symlink link target to a relative path instead of full physical path.

pyartcd/pyartcd/pipelines/promote.py
joepvd: Unresolve. Would like to see something about the exception in the log, and not have it swallowed.

joepvd (dismissed):
There are some Unresolved remarks, setting to Changes Requested.

pyartcd/pyartcd/pipelines/promote.py
joepvd: Unresolve. We need bits of this PR covered in unit tests.

pyartcd/pyartcd/oc.py
Ximinhan: updated

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: ut added

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: what do you mean to see some exception? it will raise the exception and stop the job

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: can you be more specific

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,214 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                    f.write(f"{shasum}  {dir}/sha256sum.txt\n")
+        util.log_dir_tree(release_mirror_dir)
+
+        # Publish the clients to our S3 bucket.
+        await exectools.cmd_assert_async(f"aws s3 sync --no-progress --exact-timestamps {base_to_mirror_dir}/multi s3://art-srv-enterprise/pub/openshift-v4/multi", stdout=sys.stderr)
+
+    def create_symlink(self, path_to_dir, log_tree, log_shasum):
+        # External consumers want a link they can rely on.. e.g. .../latest/openshift-client-linux.tgz .
+        # So whatever we extract, remove the version specific info and make a symlink with that name.
+        current_path = os.getcwd() # /mnt/workspace/jenkins/working/build_promote-assembly
```
vfreex, line 584: No need to get the current directory. Artifacts to work with should be in `path_to_dir`

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,214 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                # until we match (and store in a group) one of the valid file extensions.
+                match = re.match(r'^([^-]+)((-[^0-9][^-]+)+)-[0-9].*(tar.gz|tgz|bz|zip)$', f)
+                if match:
+                    new_name = match.group(1) + match.group(2) + '.' + match.group(4)
+                    # Create a symlink like openshift-client-linux.tgz => openshift-client-linux-4.3.0-0.nightly-2019-12-06-161135.tar.gz
+                    os.symlink(f, new_name)
+                    shutil.move(new_name, f"{path_to_dir}/{new_name}")
+
+        if log_tree:
+            util.log_dir_tree(f"{current_path}/{path_to_dir}")  # print dir tree
```
vfreex, line 608: The path here doesn't seem correct. I believe `path_to_dir` is accessible.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,214 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                #  - openshift-install-mac-4.3.0-0.nightly-s390x-2020-01-06-081137.tar.gz
+                #  ...
+                # So, match, and store in a group, any character up to the point we find -DIGIT. Ignore everything else
+                # until we match (and store in a group) one of the valid file extensions.
+                match = re.match(r'^([^-]+)((-[^0-9][^-]+)+)-[0-9].*(tar.gz|tgz|bz|zip)$', f)
+                if match:
+                    new_name = match.group(1) + match.group(2) + '.' + match.group(4)
+                    # Create a symlink like openshift-client-linux.tgz => openshift-client-linux-4.3.0-0.nightly-2019-12-06-161135.tar.gz
+                    os.symlink(f, new_name)
+                    shutil.move(new_name, f"{path_to_dir}/{new_name}")
```
vfreex, line 605: This doesn't seem to be needed.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,214 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+
+        # Publish the clients to our S3 bucket.
+        await exectools.cmd_assert_async(f"aws s3 sync --no-progress --exact-timestamps {base_to_mirror_dir}/{build_arch} s3://art-srv-enterprise/pub/openshift-v4/{build_arch}", stdout=sys.stderr)
+
+    async def generate_changelog(self, release_name, client_mirror_dir, minor):
+        try:
+            # To encourage customers to explore dev-previews & pre-GA releases, populate changelog
+            # https://issues.redhat.com/browse/ART-3040
+            prevMinor = minor - 1
+            rcURL = util.get_release_controller_url(release_name)
```
vfreex, line 485: `release_name` doesn't include an arch. Do you mean the quay tag here?

vfreex: Also the arch can be passed as parameter. No need to extract from the tag name.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -375,6 +401,214 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+                shasum = hashlib.sha256(f.read()).hexdigest()
+            with open(f"{client_mirror_dir}/sha256sum.txt", 'a') as f:  # write shasum to sha256sum.txt
+                f.write(f"{shasum}  opm-{platform}-{release_name}.tar.gz\n")
+
+    async def publish_multi_client(self, working_dir, from_release_tag, release_name, arch_list, client_type):
+        # Anything under this directory will be sync'd to the mirror
+        base_to_mirror_dir = f"{working_dir}/to_mirror/openshift-v4"
+        shutil.rmtree(f"{base_to_mirror_dir}/multi", ignore_errors=True)
+        release_mirror_dir = f"{base_to_mirror_dir}/multi/clients/{client_type}/{release_name}"
+        current_path = os.getcwd()
```
vfreex, line 550: this is no longer needed.

vfreex (commented):
left a few comments.

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: the symlink is created under curent work dir

vfreex: pls directly create the symlink in "{path_to_dir}/{new_name}" rather than create and move.

vfreex (approved):
Basically LGTM

PR #3625, opened 2023-04-13
covscan: Fix a reposync issue

vfreex:
We may see the following message when running `reposync` in covscan-images job:

```
19:42:11  Not using downloaded covscan/repomd.xml because it is older than what we have:
19:42:11    Current   : Tue Mar 28 14:04:51 2023
19:42:11    Downloaded: Tue Dec 20 13:20:17 2022
```

(from https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/scanning/job/scanning%252Fcovscan-images/28/console)

This is because we use the same cache directory for rhel 7, 8, and 9 repos. This can be solved by using separate cache directories.


joepvd (approved):
/lgtm

PR #3627, opened 2023-04-13
Port olm-bundle job to Python

locriandev:
Ref. https://issues.redhat.com/browse/ART-3676

Test builds:
- https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/olm-bundle/34/console was started first; it acquired the lock right away and built the bundle
- https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/olm-bundle/35/console was started a few seconds later; it found the lock already acquired, and waited for it to be available before building the bundle

### DIFF
--- a/pyartcd/pyartcd/pipelines/olm_bundle.py
+++ b/pyartcd/pyartcd/pipelines/olm_bundle.py
@@ -1,0 +1,87 @@
+import os
+
+import click
+from aioredlock import Aioredlock, LockError
+
+from pyartcd import constants, exectools
+from pyartcd.cli import cli, pass_runtime, click_coroutine
+from pyartcd import locks
+from pyartcd.runtime import Runtime
+
+
+@cli.command('olm-bundle')
+@click.option('--version', required=True, help='OCP version')
+@click.option('--assembly', required=True, help='Assembly name')
+@click.option('--data-path', required=False, default=constants.OCP_BUILD_DATA_URL,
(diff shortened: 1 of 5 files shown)


pyartcd/pyartcd/pipelines/olm_bundle.py
```diff
@@ -0,0 +1,87 @@
...
+        redis_host=os.environ['REDIS_HOST'],
+        redis_port=os.environ['REDIS_PORT']
+    )
+    retry_policy = locks.RETRY_POLICY['olm_bundle']
+    lock_manager = Aioredlock(
+        [redis_url],
+        internal_lock_timeout=locks.LOCK_TIMEOUTS['olm-bundle'],
+        retry_count=retry_policy['retry_count'],
+        retry_delay_min=retry_policy['retry_delay_min']
+    )
```
vfreex, line 71: It would nice if these lines are moved to a separate function (e.g. `new_lock_manager` in `locks.py`) so that we can reuse it in the future.

vfreex (commented):
Thanks for the great work. This looks good to me.
I added a comment about moving some lines to a separate function.

pyartcd/pyartcd/pipelines/olm_bundle.py
locriandev: Very good idea, implemented in 96ea8bc88de32d9438e862561ad40619005862fe

PR #3629, opened 2023-04-13
Fix for installer/opm source names

cblecker:
What should be the last fix to #3622

Looking at https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.12.13/, we need updating the outputs to match. The client one looks good.

### DIFF
--- a/pipeline-scripts/release.groovy
+++ b/pipeline-scripts/release.groovy
@@ -546,9 +546,9 @@ def stagePublishClient(quay_url, from_release_tag, release_name, arch, client_ty
                 source_name="openshift-client"
                 ;;
               installer)
-                source_name="openshift-installer"
+                source_name="openshift-install"
                 ;;
-              operator-registry)
+              operator-framework-olm)
                 source_name="opm"
                 ;;
             esac


pipeline-scripts/release.groovy
```diff
@@ -546,9 +546,9 @@ def stagePublishClient(quay_url, from_release_tag, release_name, arch, client_ty
                 source_name="openshift-client"
                 ;;
               installer)
-                source_name="openshift-installer"
+                source_name="openshift-install"
                 ;;
-              operator-registry)
+              operator-framework-olm)
```
ashwindasr, line 551: We don't have `operator-framework-olm` in the release, did you mean to set the `source_name` ?

ashwindasr (changes requested):

pipeline-scripts/release.groovy
cblecker:
```
oc adm release info 4.12.13 --output=json | jq -r '.references.spec.tags[] | select (.name=="operator-registry") | .annotations'
{
  "io.openshift.build.commit.id": "dd3cc9cbf5ccc16441f427b2d1386d6ca430f82f",
  "io.openshift.build.commit.ref": "",
  "io.openshift.build.source-location": "https://github.com/openshift/operator-framework-olm"
}
```

It's the "operator-registry" image that is used to determine the SHA of the opm source code in the "operator-framework-olm" repo: https://github.com/openshift/operator-framework-olm/tree/master/staging/operator-registry/cmd/opm

So in this case, the "operator-registry" image will say it got the source from the "operator-framework-olm" repo, and we are swapping that our here with the "opm" source name so we are writing the file correctly

ashwindasr: Ahh my bad, just had a look at your earlier PR. Looks good.

ashwindasr (approved):
Tested in hackspace. All [green](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fpromote-assembly/31/console)

/lgtm

ashwindasr:
Waiting for secondary lgtm

PR #3641, opened 2023-04-19
Check for stream of rhcos job instead of version

thegreyd:
This is tripping us - we need to check for stream(4.14-9.2) instead of version (4.14)


sosiouxme (approved):
/lgtm

PR #3646, opened 2023-04-24
Bump early operator index to track 4.14

Ximinhan:
ref https://redhat-internal.slack.com/archives/CB95J6R4N/p1682313011386399


joepvd (approved):
/lgtm

the 4.13 operator index is published elsewhere in a more formal place.

PR #3651, opened 2023-04-26
Remove buildlib.determineBuildVersion()

locriandev:
This function is always returning 4.y ([see here](https://github.com/openshift-eng/aos-cd-jobs/blob/master/pipeline-scripts/buildlib.groovy#L1236)) and can be simplified. Verified from my hack space that `version.full` always equals `version.stream`.

Ref. https://issues.redhat.com/browse/ART-3672


vfreex (approved):
/lgtm

PR #3652, opened 2023-04-27
Enable comment-on-pr and bump doozer

thegreyd:
Follow up to https://github.com/openshift-eng/doozer/pull/761


thegreyd:
https://github.com/openshift-eng/doozer/pull/765 has merged and I've updated it here
We are good to merge now

PR #3663, opened 2023-05-10
Use jenkinsapi instead of triggered builds

locriandev:
Ref. https://issues.redhat.com/browse/ART-6796

Replace triggered-builds with Python's `jenkinsapi`.

To allow backlinking from the triggered job to the original build, a new `TRIGGERED_FROM` param has been added to :
- build-sync
- ocp4
- rhcos
- build-microshift

This param will be filled by the calling job.
Setting `lazy=True` when initializing Jenkins objects prevents the library from scanning all jobs. This results in:
- speeding up initialization time
- avoiding an exception that gets raised when the server receives too many requests

This has been tested from a hacked ocp4-scan and 4.7:
- [hacked ocp4-scan](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/ocp4-scan/44/console)
- [triggered ocp4 build](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Focp4/42323/)


pyartcd/pyartcd/jenkins.py
```diff
@@ -1,71 +1,107 @@
...
-    :param job_path: relative path to the job, starting from <buildvm hostname>:<jenkins port>
-    :param params: optional dict containing job parameters
+    If 'BUILD_URL' env var was provided, add a TRIGGERED_FROM
+    field to the build params to allow backlinking to the original job
     """
 
-    service_account = os.environ['JENKINS_SERVICE_ACCOUNT']
-    token = os.environ['JENKINS_SERVICE_ACCOUNT_TOKEN']
-
     build_url = os.environ.get('BUILD_URL', '')
```
thegreyd, line 34: I guess we are exploring other ways to automatically get the parent build..
I would enforce this, so complain if build_url isn't set - since I can't
think of a <not from a jenkins job invokation> of this func

pyartcd/pyartcd/jenkins.py
```diff
@@ -1,71 +1,107 @@
...
+
+from pyartcd import constants
 
 logger = logging.getLogger(__name__)
 
 
-async def trigger_jenkins_job(job_path: str, params: dict = {}):
+class Jobs(Enum):
+    BUILD_SYNC = 'aos-cd-builds/build%2Fbuild-sync'
+    MICROSHIFT = 'aos-cd-builds/build%2Fbuild-microshift'
```
thegreyd, line 14: Would like to keep the names canonical - so BUILD_MICROSHIFT

pyartcd/pyartcd/jenkins.py
```diff
@@ -1,71 +1,107 @@
...
+    logger.info('Started new build for job %s',
+                build.baseurl.replace(constants.JENKINS_SERVER_URL, constants.JENKINS_UI_URL))
+
+    logger.info('Waiting for build to complete...')
+    build.block_until_complete()
+    logger.info(f'Completed with status: {build.get_status()}')
+    return build.get_status()
+
+
+def start_build_non_blocking(job_name: str, params: dict):
```
thegreyd, line 62: Any strong reason to keep the 2 funcs separate vs `blocking: bool` param?

pyartcd/pyartcd/jenkins.py
```diff
@@ -1,71 +1,107 @@
...
-    async with aiohttp.ClientSession(headers=headers) as session:
-        async with session.post(url, data=params) as response:
-            response.raise_for_status()
-            await response.text()
-
-
-async def trigger_ocp4(build_version: str):
-    await trigger_jenkins_job(
-        job_path='job/triggered-builds/job/ocp4',
+def start_ocp4(build_version: str, blocking: bool = False):
```
thegreyd, line 77: Don't see a dry_run .. did we support it before? We probably are skipping triggering this if it was set in the parent.
This looks good, but I'm also thinking if there's a better way to do it instead of having 1 func for each job.

Maybe something like 
`start_build(Jobs.RHCOS, params)` 

and then we have associated param dict defined

allowed_params = {
    Jobs.RHCOS: {'BUILD_VERSION', 'NEW_BUILD', 'DRY_RUN'}
   ...
}

and then we can do some ignore case and see if params passed are in allowed_params in start_build.
What do you think?

thegreyd: I could even be convinced that we don't need allowed_params, since any error / silly thing
we do in calling start_build(..) would be apparent and result in obvious failure

pyartcd/pyartcd/pipelines/ocp4_scan.py
```diff
@@ -64,7 +64,7 @@ async def run(self):
             # Inconsistency probably means partial failure and we would like to retry.
             #  but don't kick off more if already in progress.
             self.logger.info('Triggering a %s RHCOS build for consistency', self.version)
-            await trigger_rhcos(self.version, True)
+            jenkins.start_rhcos(self.version, True)
```
thegreyd, line 67: Would make blocking a keyword param to make it more clear

pyartcd/pyartcd/jenkins.py
locriandev: Done in https://github.com/openshift-eng/aos-cd-jobs/pull/3663/commits/91e809f1c6701514533552a1fcdb8115bcfc1df8

pyartcd/pyartcd/jenkins.py
locriandev: Originally, I had a different implementation between the 2, so I had to keep them separate. I realize now they are pretty much the same, so makes sense to unify them.

Done in 085cc2e19d55d00fa9359c00c9eabde6d6de606b

pyartcd/pyartcd/jenkins.py
locriandev: I prefer to have a clear interface to call each job, without having to get into details like choosing enum from `Jobs` field, etc. Each job has a calling interface with separate params, and will be maintained independently from the others

pyartcd/pyartcd/pipelines/ocp4_scan.py
locriandev: Done in 12fb9f899a3c5950ba6c997eb78d4cfe8da52cf8

pyartcd/pyartcd/jenkins.py
thegreyd: I'm okay with that :+1:

thegreyd:
Overall lgtm

vfreex (approved):

vfreex (approved):

pyartcd/pyartcd/jenkins.py
```diff
@@ -1,74 +1,161 @@
...
+
+    logger.info('Starting new build for job: %s', job_name)
+    queue_item = jenkins.get_job(job_name).invoke(build_params=params)
+    build_url = block_until_building(queue_item, watch_building_delay)
+    logger.info('Started new build at %s', build_url)
+
+    # If blocking==True, wait for the build to complete; get its status and return it
+    if blocking:
+        logger.info('Waiting for build to complete...')
+        build_result = block_until_complete(queue_item, watch_complete_delay)
```
Ximinhan, line 120: since we already have the build number, we should use build.is_building() to check the status, the queue item will be destroyed after it triggers the build.

locriandev: Right. Fixed in 5b6922ee3814d47449511d6e3018e70d4bee3c4d

Test build: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/test-remote/70/console

pyartcd/pyartcd/jenkins.py
```diff
@@ -1,74 +1,147 @@
...
+
+    # If blocking==True, wait for the build to complete; get its status and return it
+    if blocking:
+        logger.info('Waiting for build to complete...')
+        build = Build(
+            url=build_url.replace(constants.JENKINS_UI_URL, constants.JENKINS_SERVER_URL),
+            buildno=int(build_url.split('/')[-1]),
+            job=job
+        )
+        build.block_until_complete()
```
Ximinhan, line 107: would you mind try
>>> b.job.jenkins.requester.post_and_confirm_status(f"{b.baseurl}/submitDescription",params={"Submit": "submit", "description":f"build by:{job.baseurl}"},data="",valid=[200])
to get the downstream job set in the description?
and can you add a wrapper to support starting a batch of builds or just return the build object? my use case looks like https://github.com/openshift-eng/aos-cd-jobs/pull/3563/files#diff-de01749278991bbde16a03048265c175ddc0aaf9a68f3f878396c59b821e4573R432-R433 I may need to invoke a list of jobs in a loop and monitoring the results in parallel

locriandev: Fixed the tracking part. Regarding the loop of parallel builds, let's address it in a new PR

PR #3677, opened 2023-05-17
Do not empty out ocp4 description at start

locriandev:
ocp4-scan is trying to initially set the build description by linking the upstream build  that caused it. Jenkinsfile is overwriting it, so we lose links between builds.


vfreex (approved):
lgtm

PR #3681, opened 2023-05-17
Inline sweep job

locriandev:
Sweep job is just used to move bugs to ON_QA. It is a straight elliott invocation, and can be inlined in the jobs currently using it (ocp3, ocp4).


joepvd (approved):
/lgtm

PR #3683, opened 2023-05-18
[ART-6761] update default arch for microshift

Ximinhan:
stop syncing microshift for s390x and ppc64 arch


vfreex:
This is ok-ish, though ideally this should be controlled by a config.

PR #3691, opened 2023-05-22
Provide ocp3 bug sweep with JIRA_TOKEN env var

locriandev:
Same fix as https://github.com/openshift-eng/aos-cd-jobs/pull/3690


thiagoalessio (approved):
/lgtm

PR #3692, opened 2023-05-22
Fix poll-for-failed-nightlies description

locriandev:
Avoid having descriptions like
```
null Notifying for 4.7.0-0.nightly
```


joepvd (approved):
/lgtm

PR #3693, opened 2023-05-23
Pass all params when calling build_sync from ocp4_scan

thegreyd:


pyartcd/pyartcd/pipelines/ocp4_scan.py
```diff
@@ -70,7 +70,11 @@ async def run(self):
                 return
 
             self.logger.info('Triggering a %s build-sync', self.version)
-            jenkins.start_build_sync(build_version=self.version, blocking=False)
+            jenkins.start_build_sync(
+                build_version=self.version,
+                assembly="stream",
```
thegreyd, line 75: Actual fix

PR #3700, opened 2023-06-01
Fix poll payload with EC

Ximinhan:
ref: https://redhat-internal.slack.com/archives/CB95J6R4N/p1685458972839399
in this function context
latestRelease.name for ec is 4.14.0-ec.1
releaseStream for ec is 4-dev-preview-s390x
and set client latest job only accept channel parameter, for ec it should only have candidate channel 
the error in the job https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/scheduled-builds/job/poll-payload/131591/console is it can't extract version from value 4-dev-preview-s390x


joepvd (approved):
/lgtm

PR #3724, opened 2023-06-12
fix sync mirror args

Ximinhan:
In our promote job we called with `delete_old=false` https://github.com/openshift-eng/aos-cd-jobs/blob/master/jobs/build/promote-assembly/Jenkinsfile#L201 but groovy treated it as position args and set include_only=false while delete_old=true as default. So --delete parameter is added.
this temporary fix is to set the variable in the correct order
It won't be an issue after migrating this function to python


Ximinhan:
this fix should be not needed after https://github.com/openshift-eng/aos-cd-jobs/pull/3725 merged

vfreex (approved):

PR #3725, opened 2023-06-12
[ART-7040] move sync rhcos step to python

Ximinhan:
move sync rhcos srpms step to python.
the sync rhocs srpms step is trying to solve an issue that RHCOS is not using a source from a standard released repo so we need to publish the unreleased srpms.
basically there are two way to handle it:
1 we can no do the sync step and enable this step when this issue happened to make RHCOS correct
2 we can also run this sync by default with filters to prevent this issue happen

this pr is using the 2 method, the filter is set to run the sync against all 4.13+ and not GA releases.

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -25,6 +25,7 @@ from pyartcd import constants, exectools, util, jenkins
 from pyartcd.cli import cli, click_coroutine, pass_runtime
 from pyartcd.exceptions import VerificationError
 from pyartcd.jira import JIRAClient
+from pyartcd.s3 import sync_dir_to_s3_mirror
 from pyartcd.oc import get_release_image_info, get_release_image_pullspec, extract_release_binary,\
     extract_release_client_tools, get_release_image_info_from_pullspec, extract_baremetal_installer
 from pyartcd.runtime import Runtime
@@ -362,6 +363,8 @@ class PromotePipeline:
                         await self.publish_client(self._working_dir, f"{release_name}-{arch}", data["content"][arch]['metadata']['version'], arch, client_type)
                     else:
                         await self.publish_multi_client(self._working_dir, f"{release_name}-{arch}", data["content"][arch]['metadata']['version'], data['content'], client_type)
+        # sync rhcos
+        await self.sync_rhcos_srpms(assembly_type, data)
         json.dump(data, sys.stdout)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -370,6 +373,28 @@ def _get_image_stream_name(assembly_type: assembly.AssemblyTypes, arch: str):
         go_arch_suffix = go_suffix_for_arch(arch)
         return f'4-dev-preview{go_arch_suffix}' if assembly_type == assembly.AssemblyTypes.PREVIEW else f'release{go_arch_suffix}'
 
+    async def sync_rhcos(self, assembly_type, data):
```
joepvd, line 376: name probably should be `sync_rhcos_srpms`

vfreex (dismissed):
/hold
The `config:rhcos-srpms` was a temporary measure to publish sources for unreleased RHCOS content. This is no longer needed at this moment because 4.13, 4.14, and 4.15 are all moved to 9.2 EUS. So this step can be completely removed.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -370,6 +374,28 @@ def _get_image_stream_name(assembly_type: assembly.AssemblyTypes, arch: str):
         go_arch_suffix = go_suffix_for_arch(arch)
         return f'4-dev-preview{go_arch_suffix}' if assembly_type == assembly.AssemblyTypes.PREVIEW else f'release{go_arch_suffix}'
 
+    async def sync_rhcos_srpms(self, assembly_type, data):
+        # Sync potential pre-release source on which RHCOS depends. See ART-6419 for details.
+        major, minor = util.isolate_major_minor_in_group(self.group)
+        if (major > 4 or (major == 4 and minor >= 13)) and assembly_type in [assembly.AssemblyTypes.CANDIDATE, assembly.AssemblyTypes.STANDARD]:
```
vfreex, line 380: Currently it is no longer needed for 4.13 and 4.14 since they have moved to 9.2 EUS. In case we need it in the future, maybe we need to turn it into a switch in ocp-build-data.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -370,6 +374,28 @@ def _get_image_stream_name(assembly_type: assembly.AssemblyTypes, arch: str):
...
+                        "--group", self.group,
+                        "--assembly", self.assembly,
+                        "config:rhcos-srpms",
+                        "--version", data["content"][arch]['rhcos_version'],
+                        "--arch", arch,
+                        "-o", src_output_dir,
+                    ]
+                    await exectools.cmd_assert_async(cmd, env=self._doozer_env_vars)
+            # Publish the clients to our S3 bucket.
+            await sync_dir_to_s3_mirror(src_output_dir, "/pub/openshift-v4/sources/packages/", "", "", False, False)
```
vfreex, line 395: For better readability, could you use keyword arguments instead of magical constants `False, False`?

pyartcd/pyartcd/pipelines/promote.py
vfreex, line 380: `assembly.AssemblyTypes.PREVIEW` is also needed. It was missed in the existing code though.

Ximinhan: updated

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -374,6 +378,28 @@ def _get_image_stream_name(assembly_type: assembly.AssemblyTypes, arch: str):
         go_arch_suffix = go_suffix_for_arch(arch)
         return f'4-dev-preview{go_arch_suffix}' if assembly_type == assembly.AssemblyTypes.PREVIEW else f'release{go_arch_suffix}'
 
+    async def sync_rhcos_srpms(self, assembly_type, data):
+        # Sync potential pre-release source on which RHCOS depends. See ART-6419 for details.
+        major, minor = util.isolate_major_minor_in_group(self.group)
+        if (major > 4 or (major == 4 and minor >= 13)) and assembly_type in [assembly.AssemblyTypes.CANDIDATE, assembly.AssemblyTypes.PREVIEW]:
```
joepvd, line 384: Please remove the version logic. Type preview or Candidate is enough.

joepvd (changes requested):
One comment inline. And please rebase your work so the commits make sense without using "the special merge button".

pyartcd/pyartcd/pipelines/promote.py
Ximinhan: updated

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -374,6 +377,28 @@ def _get_image_stream_name(assembly_type: assembly.AssemblyTypes, arch: str):
         go_arch_suffix = go_suffix_for_arch(arch)
         return f'4-dev-preview{go_arch_suffix}' if assembly_type == assembly.AssemblyTypes.PREVIEW else f'release{go_arch_suffix}'
 
+    async def sync_rhcos_srpms(self, assembly_type, data):
+        # Sync potential pre-release source on which RHCOS depends. See ART-6419 for details.
+        major, minor = util.isolate_major_minor_in_group(self.group)
+        if assembly_type in [assembly.AssemblyTypes.CANDIDATE, assembly.AssemblyTypes.STANDARD]:
```
vfreex, line 383: I think the decision to is to do this only for `CANDIDATE` and `PREVIEW`.

joepvd (approved):
/lgtm

PR #3726, opened 2023-06-12
Move build-sync fail counter to Redis

locriandev:
Tested on a 4.14 build-sync: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/build-sync/13/console

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_sync.py
+++ b/pyartcd/pyartcd/pipelines/build_sync.py
@@ -9,7 +9,7 @@ import yaml
 from pyartcd.cli import cli, pass_runtime, click_coroutine
 from pyartcd.oc import registry_login
 from pyartcd.runtime import Runtime
-from pyartcd import exectools, constants
+from pyartcd import exectools, constants, redis, util
 from pyartcd.util import branch_arches
 from doozerlib.util import go_suffix_for_arch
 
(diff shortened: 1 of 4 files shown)


pyartcd/pyartcd/pipelines/build_sync.py
```diff
@@ -322,4 +322,65 @@ async def build_sync(runtime: Runtime, version: str, assembly: str, publish: boo
...
+    fail_count_name = f'{version}-assembly.{assembly}.count'
+
+    try:
+        await pipeline.run()
+
+        #  All good: reset fail count to 0
+        await redis.set_value(fail_count_name, 0)
+        runtime.logger.info('Fail count "%s" set to 0', fail_count_name)
+
+    except (RuntimeError, ChildProcessError):
```
joepvd, line 334: I think the redis setter should not be in the `try` block.
Because a failure to set the counter should be handled differently than a failure to run the pipeline. 
Also: How should the error / job result look like when the pipeline succeeds but setting count to `0` fails?

pyartcd/pyartcd/redis.py
```diff
@@ -0,0 +1,38 @@
...
+
+import aioredis
+
+# Redis instance template, to be renderes with env vars
+redis = Template('${protocol}://:${redis_password}@${redis_host}:${redis_port}')
+
+
+def redis_url(use_ssl=True):
+    if not os.environ.get('REDIS_SERVER_PASSWORD', None):
+        raise RuntimeError('Please define REDIS_SERVER_PASSWORD env var')
```
joepvd, line 12: Maybe this should be a less generic or even custom exception. This might be a case where we allow an exception to be targetedly swallowed. Maybe even a RedisException or RedisRuntimeError or so?

pyartcd/pyartcd/pipelines/build_sync.py
locriandev: `(RuntimeError, ChildProcessError)` will not catch the case where we fail to update the counter. This means, the job will break (and we'll know, which is good). Since this is an edge case, I think it's fine to leave the exception unhandled. It's something we want to be notified about ASAP

vfreex (approved):

PR #3729, opened 2023-06-13
update s3 lib

Ximinhan:
sync_repo_to_s3_mirror function is not flexible especially when remove_old is not True, make it a wrap for base function sync_dir_to_s3_mirror

### DIFF
--- a/pyartcd/pyartcd/s3.py
+++ b/pyartcd/pyartcd/s3.py
@@ -1,9 +1,34 @@
 import os
-
 from pyartcd import exectools
 
 
 async def sync_repo_to_s3_mirror(local_dir: str, s3_path: str, dry_run: bool = False, remove_old: bool = True):
+    # Sync is not transactional. If we update repomd.xml before files it references are populated,
+    # users of the repo will get a 404. So we run in three passes:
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/s3.py
```diff
@@ -13,33 +38,16 @@ async def sync_repo_to_s3_mirror(local_dir: str, s3_path: str, dry_run: bool = F
...
-        '--exclude', '*', '--include', '*/repomd.xml', local_dir, full_s3_path
-    ]
-    await exectools.cmd_assert_async(cmd, env=env)
-
-    # For most repos, clean up the old rpms so they don't grow unbounded. Specify remove_old=false to prevent this step.
-    # Otherwise:
-    # 3. Everything should be synced in a consistent way -- delete anything old with --delete.
+    if exclude:
+        base_cmd += ['--exclude', exclude]
+    if include_only:
```
locriandev, line 48: Should probably be an `elif` clause?

locriandev (changes requested):

pyartcd/pyartcd/s3.py
Ximinhan: updated

locriandev (approved):

PR #3739, opened 2023-06-16
Correctly initialize fail counter on first failure

locriandev:
```
fail_count = int(await redis.get_value(fail_count_name)) + 1
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

This happens when first build-sync run for a version-assembly pair fails. In this case, we need to set it to 0 before incrementing


vfreex (approved):
lgtm

PR #3748, opened 2023-06-19
Doozer load releases

locriandev:
Load releases config using doozer (fetches data using GitHub API) instead of relying on filesystem


joepvd (approved):
/lgtm

PR #3770, opened 2023-07-03
Do not have green failed build-sync runs

locriandev:
In certain cases, build-sync pipeline was terminating prematurely while catching a failure (mostly due to doozer non-0 return code). This caused some job runs to look green on Jenkins, even though they actually failed. This PR fixes this by re-raising the exception after updating the failure counter, and before exiting the pipeline.


vfreex (approved):
lgtm

PR #3773, opened 2023-07-05
[ART-6670] optimize create microshift pr method and auto merge pr

Ximinhan:
* auto merge microshift pin pr

### DIFF
--- a/pyartcd/pyartcd/pipelines/build_microshift.py
+++ b/pyartcd/pyartcd/pipelines/build_microshift.py
@@ -139,7 +139,7 @@ class BuildMicroShiftPipeline:
             # Sends a slack message
             message = f"Hi @release-artists , microshift for assembly {self.assembly} has been successfully built."
             if pr:
-                message += f"\nA PR to update the assembly definition has been created/updated: {pr.html_url}"
+                message += f"\nA PR to update the assembly definition has been created and merged: {pr.html_url}"
                 message += "\nReview and merge the PR before you proceed.\n"
             if assembly_type in [AssemblyTypes.CANDIDATE, AssemblyTypes.STANDARD]:
                 message += '\n'.join([
@@ -325,9 +325,11 @@ class BuildMicroShiftPipeline:
             existing_prs = api.pulls.list(state="open", base=base, head=head)
             if not existing_prs.items:
                 result = api.pulls.create(head=head, base=base, title=title, body=body, maintainer_can_modify=True)
+                api.pulls.merge(owner=owner, repo=repo, pull_number=result.number, merge_method="squash")
             else:
                 pull_number = existing_prs.items[0].number
                 result = api.pulls.update(pull_number=pull_number, title=title, body=body)
+                api.pulls.merge(owner=owner, repo=repo, pull_number=pull_number, merge_method="squash")
         else:
             self._logger.warning("PR is not created: Nothing to commit.")
         return result


pyartcd/requirements.txt
```diff
@@ -11,6 +11,7 @@ Jinja2
 jira >= 3.4.1  # https://github.com/pycontribs/jira/issues/1486
 openshift-client
 pygit2 == 1.10.1 # https://github.com/libgit2/pygit2/issues/1176
+PyGithub >= 1.57
```
vfreex, line 14: pyartcd uses ghpai as the GitHub API client.

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -276,50 +276,23 @@ def _pin_nvrs(self, nvrs: List[str], releases_config) -> Dict:
...
-            if not existing_prs.items:
-                result = api.pulls.create(head=head, base=base, title=title, body=body, maintainer_can_modify=True)
-            else:
-                pull_number = existing_prs.items[0].number
-                result = api.pulls.update(pull_number=pull_number, title=title, body=body)
-        else:
-            self._logger.warning("PR is not created: Nothing to commit.")
-        return result
+        repo = self.runtime.new_github_client().get_repo("openshift-eng/ocp-build-data")
+        microref = repo.create_git_ref(f"refs/heads/{branch}", repo.get_git_ref(f"heads/{self.group}").object.sha)  # create a microshift branch
```
vfreex, line 283: This could fail if the branch already exists.

vfreex (commented):
Having the ocp-build-data cloned locally then pushed to remote could prevent issues like race condition, although it might not be a big concern as it is unlikely to run build-microshift for an assembly in parallel.

Ximinhan:
> Having the ocp-build-data cloned locally then pushed to remote could prevent issues like race condition, although it might not be a big concern as it is unlikely to run build-microshift for an assembly in parallel.

I think github pr check should do the conflict check

pyartcd/requirements.txt
Ximinhan: I want to use a more straightforward method instead of pulling the file, making changes, then pushing.

pyartcd/pyartcd/pipelines/build_microshift.py
Ximinhan: check added

pyartcd/requirements.txt
vfreex: I mean use ghapi is recommended instead of PyGithub.

vfreex:
> I think github pr check should do the conflict check
It is not related to the conflict check. It is about reading-modifying-writing the remote branch `auto-pin-microshift-{self.group}-{self.assembly}` without a lock.

A race condition could happen if multiple build-microshift builds run for the same assembly. Think about the following sequence:

- Build A: Started.
- Build A: Created a branch `auto-pin-microshift-{self.group}-{self.assembly}` from `commit_a`.
- Build A: Read content of releases.yml from `commit_a` in `openshift-{self.group}`. Let's call it `content_a`.
- Build B: Started.
- Build B: Saw `auto-pin-microshift-{self.group}-{self.assembly}` already existed. Delete branch branch `auto-pin-microshift-{self.group}-{self.assembly}`.
- Build B: Created a branch `auto-pin-microshift-{self.group}-{self.assembly}` from `commit_b`. `commit_b` is newer than `commit_a`.
- Build B: Read content of releases.yml from `commit_a` in `openshift-{self.group}`. Let's call it `content_b`.
- Build B: Updated releases.yml `auto-pin-microshift-{self.group}-{self.assembly}` with modified content based on `content_b`.
- Build A: Updated releases.yml `auto-pin-microshift-{self.group}-{self.assembly}` with modified content based on `content_a`. Note, `auto-pin-microshift-{self.group}-{self.assembly}` is based on `commit_b` instead of `commit_a`. The commit history is incorrect!
- Build B: Merged `auto-pin-microshift-{self.group}-{self.assembly}` into `openshift-{self.group}`. At this moment of merge, `auto-pin-microshift-{self.group}-{self.assembly}` has updated releases.yml based on `content_a`. So `content_b` is lot.
- Build A: Would fail to merge `auto-pin-microshift-{self.group}-{self.assembly}` into `openshift-{self.group}`.

With the cloning to local approach, no race condition will happen because the changes are made into the local branch. Regardless of the order of pushing to remote, the remote branch will always have the commit history, so merging will either succeed without a data loss, or fail due to a conflict.

vfreex:
But, we can avoid this by using a lock.

Ximinhan:
> > I think github pr check should do the conflict check
> 
> It is not related to the conflict check. It is about reading-modifying-writing the remote branch `auto-pin-microshift-{self.group}-{self.assembly}` without a lock.
> 
> A race condition could happen if multiple build-microshift builds run for the same assembly. Think about the following sequence:
> 
> * Build A: Started.
> * Build A: Created a branch `auto-pin-microshift-{self.group}-{self.assembly}` from `commit_a`.
> * Build A: Read content of releases.yml from `commit_a` in `openshift-{self.group}`. Let's call it `content_a`.
> * Build B: Started.
> * Build B: Saw `auto-pin-microshift-{self.group}-{self.assembly}` already existed. Delete branch branch `auto-pin-microshift-{self.group}-{self.assembly}`.
> * Build B: Created a branch `auto-pin-microshift-{self.group}-{self.assembly}` from `commit_b`. `commit_b` is newer than `commit_a`.
> * Build B: Read content of releases.yml from `commit_a` in `openshift-{self.group}`. Let's call it `content_b`.
> * Build B: Updated releases.yml `auto-pin-microshift-{self.group}-{self.assembly}` with modified content based on `content_b`.
> * Build A: Updated releases.yml `auto-pin-microshift-{self.group}-{self.assembly}` with modified content based on `content_a`. Note, `auto-pin-microshift-{self.group}-{self.assembly}` is based on `commit_b` instead of `commit_a`. The commit history is incorrect!
> * Build B: Merged `auto-pin-microshift-{self.group}-{self.assembly}` into `openshift-{self.group}`. At this moment of merge, `auto-pin-microshift-{self.group}-{self.assembly}` has updated releases.yml based on `content_a`. So `content_b` is lot.
> * Build A: Would fail to merge `auto-pin-microshift-{self.group}-{self.assembly}` into `openshift-{self.group}`.
> 
> With the cloning to local approach, no race condition will happen because the changes are made into the local branch. Regardless of the order of pushing to remote, the remote branch will always have the commit history, so merging will either succeed without a data loss, or fail due to a conflict.

In the above case, when updating file it will be applying patch changes against release.yml's shasum, the second update operation of updating releases.yml by build A will fail, for this case build B interrupts Build A and should successfully commit its changes, build A is failed at updating file since the content already changed by build B's commit. 

If Build A update releases.yml happened before Build B update releases.yml then Build B will fail at update file, and then pr will be created and merged but delete pin branch will fail because branch from commit_a was deleted.


The workflow delete branch/create branch/update file/create pull/merge operations all have commit checks on the server. Considering the race condition jobs are doing the same thing, It may doesn't matter invoked by which Build as long as they are in the right order(only one branch exist after multi jobs run create branch, file is only updated once after multi jobs run update file, pr can be merged once after multi jobs run merge) and in `openshift-{self.group}` branch will only have one merge commit in the commit history.

pyartcd/requirements.txt
Ximinhan: ghapi is valuable for cloning data from remote to local env and pushing data to remote, but I think we can simplify the workflow as the pr did

joepvd: I think ghapi is capable of more than file operations on repositiories. I prefer to stick to a single dependency in the same area if possible. Have not seen a good rationale why have two deps would give a better solution. Could you investigate again?

vfreex:
@Ximinhan 

> If Build A update releases.yml happened before Build B update releases.yml then Build B will fail at update file, and then pr will be created and merged but delete pin branch will fail because branch from commit_a was deleted.

Good point. I haven't tested if the API will fail if the sha of the file is changed. If so, I am ok with this approach.

pyartcd/requirements.txt
vfreex: No, ghapi is a client library for GitHub API. It doesn't clone a git repo.

Ximinhan:
checked the [git lib](https://github.com/openshift-eng/aos-cd-jobs/blob/master/pyartcd/pyartcd/git.py) basically, it was a wrapper of the workflow using shell command instead of github python API. This part can be optimized later.

vfreex:
@Ximinhan we are not talking about the same thing.
I was talking about ghapi https://ghapi.fast.ai/, while you were talking about pyartcd.git module https://github.com/openshift-eng/aos-cd-jobs/blob/master/pyartcd/pyartcd/git.py

Ximinhan:
> @Ximinhan we are not talking about the same thing. I was talking about ghapi https://ghapi.fast.ai/, while you were talking about pyartcd.git module https://github.com/openshift-eng/aos-cd-jobs/blob/master/pyartcd/pyartcd/git.py

I want to use PyGithub to replace our current usage of ghapi and pyartcd/git.py lib in the repo, since PyGithub is powerful to handle all git operations in a convenient way

Ximinhan:
Okay, I will use ghapi to merge the pr.

PR #3774, opened 2023-07-05
Publish baremetal installer binary on mirror

locriandev:
Ref. https://issues.redhat.com/browse/ART-7207

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -25,7 +25,8 @@ from pyartcd import constants, exectools, util, jenkins
 from pyartcd.cli import cli, click_coroutine, pass_runtime
 from pyartcd.exceptions import VerificationError
 from pyartcd.jira import JIRAClient
-from pyartcd.oc import get_release_image_info, get_release_image_pullspec, extract_release_binary, extract_release_client_tools, get_release_image_info_from_pullspec
+from pyartcd.oc import get_release_image_info, get_release_image_pullspec, extract_release_binary,\
+    extract_release_client_tools, get_release_image_info_from_pullspec, extract_baremetal_installer
 from pyartcd.runtime import Runtime
 from ruamel.yaml import YAML
 from semver import VersionInfo
@@ -424,6 +425,9 @@ class PromotePipeline:
             else:
                 self._logger.error(f"Error get {tarball} image from release pullspec")
 
+        # ART-7207 - upload baremetal installer binary to mirror
+        self.publish_baremetal_installer_binary(from_release_tag, client_mirror_dir)
+
         # Starting from 4.14, oc-mirror will be synced for all arches. See ART-6820 and ART-6863
         major, minor = util.isolate_major_minor_in_group(self.group)
         if major > 4 or minor >= 14 or build_arch == 'x86_64':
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -461,6 +465,47 @@ async def publish_client(self, working_dir, from_release_tag, release_name, buil
...
+            self._logger.error('Error getting baremetal-installer binary from %s', release_pullspec)
+            return
+
+        # Get baremetal image pullspec
+        res, image_info = get_release_image_info_from_pullspec(baremetal_installer_pullspec)
+        if res:
+            self._logger.error('Error retrieving image info for %s', baremetal_installer_pullspec)
+            return
+
+        # Check rhel version (used for archive naming
```
Ximinhan, line 482: in the ticket description, I thought it should always be rhel8, maybe I am wrong, can you double confirm?

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -461,6 +465,47 @@ async def publish_client(self, working_dir, from_release_tag, release_name, buil
...
+        display_name = image_info['config']['config']['Labels']['io.k8s.display-name']
```
joepvd, line 483: This feels brittle, because it is a label that one of the base images sets, and does not happen to get overridden by later layers. Not sure what is the best way to query this info without pulling in data sources of another kind.

joepvd (commented):
lgtm, with two comments. Let's wait with merging to see if the implementation matches the spec.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -424,6 +425,9 @@ async def publish_client(self, working_dir, from_release_tag, release_name, buil
             else:
                 self._logger.error(f"Error get {tarball} image from release pullspec")
 
+        # ART-7207 - upload baremetal installer binary to mirror
+        self.publish_baremetal_installer_binary(from_release_tag, client_mirror_dir)
```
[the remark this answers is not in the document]
locriandev: I think `publish` here stands for publish to mirror, which is an expression we use widely. I would not care about changing it

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -461,6 +465,47 @@ async def publish_client(self, working_dir, from_release_tag, release_name, buil
         # Publish the clients to our S3 bucket.
         await exectools.cmd_assert_async(f"aws s3 sync --no-progress --exact-timestamps {base_to_mirror_dir}/{build_arch} s3://art-srv-enterprise/pub/openshift-v4/{build_arch}", stdout=sys.stderr)
 
+    def publish_baremetal_installer_binary(self, from_release_tag: str, client_mirror_dir: str):
+        release_pullspec = f'{constants.QUAY_RELEASE_REPO_URL}:{from_release_tag}'
+        image_stat, baremetal_installer_pullspec = get_release_image_pullspec(release_pullspec, 'baremetal-installer')
+
+        if image_stat:  # != 0, image does not exist
+            self._logger.error('Error getting baremetal-installer binary from %s', release_pullspec)
+            return
```
vfreex, line 474: `get_release_image_pullspec` already has `check=True` so no need to check the exist code.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -461,6 +465,47 @@ async def publish_client(self, working_dir, from_release_tag, release_name, buil
...
+
+        # Get baremetal image pullspec
+        res, image_info = get_release_image_info_from_pullspec(baremetal_installer_pullspec)
+        if res:
+            self._logger.error('Error retrieving image info for %s', baremetal_installer_pullspec)
+            return
```
vfreex, line 480: Same here

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -461,6 +465,47 @@ async def publish_client(self, working_dir, from_release_tag, release_name, buil
...
+        if 'RHEL 8' in display_name:
+            rhel_version = 'rhel8'
+        else:
+            rhel_version = 'rhel9'
+
+        # oc adm release extract --command=openshift-baremetal-install -n=ocp <release-pullspec>
+        res, _ = extract_baremetal_installer(release_pullspec, client_mirror_dir)
+        if res:
+            self._logger.error('Error extracting baremetal-installer binary from %s', release_pullspec)
+            return
```
vfreex, line 493: Same here.

pyartcd/pyartcd/pipelines/promote.py
locriandev: Right, but what we want to do here is exit from `publish_baremetal_installer_binary` function, because if `oc adm release info` fails, nothing else can be done. `check_status` in `common_oc_wrapper` will only log a warning

pyartcd/pyartcd/pipelines/promote.py
locriandev: Same as above: we need to exit from `publish_baremetal_installer_binary` if  `oc image info` fails

pyartcd/pyartcd/pipelines/promote.py
locriandev: Same as above: we need to exit from `publish_baremetal_installer_binary` if `oc adm release extract` fails

pyartcd/pyartcd/pipelines/promote.py
vfreex: If `oc adm release info` fails, an exception will be raised and the pipeline will abort unless you capture the exception.

pyartcd/pyartcd/pipelines/promote.py
locriandev: Fixed with https://github.com/openshift-eng/aos-cd-jobs/pull/3774/commits/113c9a074dbe1daf09aa37634fff6cdb72f5e5e4

pyartcd/pyartcd/pipelines/promote.py
locriandev: Fixed with https://github.com/openshift-eng/aos-cd-jobs/pull/3774/commits/113c9a074dbe1daf09aa37634fff6cdb72f5e5e4

pyartcd/pyartcd/pipelines/promote.py
locriandev: Fixed with https://github.com/openshift-eng/aos-cd-jobs/pull/3774/commits/113c9a074dbe1daf09aa37634fff6cdb72f5e5e4

vfreex (approved):
LGTM

pyartcd/pyartcd/pipelines/promote.py
locriandev: According to [this comment](https://issues.redhat.com/browse/ART-7207?focusedId=22586790&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22586790), we decided to hardcode the rhel8 suffix into the tarball name. This will eventually change once baremetal-installer moves to rhel9. I'll add a comment in `ocp-build-data` as a reminder

PR #3783, opened 2023-07-10
[ART-6667] only create defined advisory

Ximinhan:
ref: https://issues.redhat.com/browse/ART-6667
only create placeholder advisory defined in assembly

### DIFF
--- a/pyartcd/pyartcd/pipelines/prepare_release.py
+++ b/pyartcd/pyartcd/pipelines/prepare_release.py
@@ -146,53 +146,13 @@ class PrepareReleasePipeline:
             if release_config:
                 advisories = group_config.get("advisories", {}).copy()
 
-            if self.release_version[2] == 0:  # GA release
-                if "rpm" not in advisories:
-                    _LOGGER.info("rpm advisory is not defined.")
-                elif advisories["rpm"] < 0:
-                    advisories["rpm"] = self.create_advisory("RHEA", "rpm", "ga")
-                else:
-                    _LOGGER.info("Reusing existing rpm advisory %s", advisories["rpm"])
-                if "image" not in advisories:
-                    _LOGGER.info("image advisory is not defined.")
-                elif advisories["image"] < 0:
-                     advisories["image"] = self.create_advisory("RHEA", "image", "ga")
-                else:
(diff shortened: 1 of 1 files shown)


joepvd (commented):
Getting confused while reading all the chained `if`s. 
Can't we have:
- set `impetus = 'RHEA` if GA, `RHBA` otherwise
- and loop through the advisories in the assembly definition, and create 'em if not exist? 

Or what is the real reason for all this logic?

Ximinhan:
> Getting confused while reading all the chained `if`s. Can't we have:
> 
> * set `impetus = 'RHEA` if GA, `RHBA` otherwise
> * and loop through the advisories in the assembly definition, and create 'em if not exist?
> 
> Or what is the real reason for all this logic?

I guess the previous approach is caused by each advisory needing different args for creating

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -146,39 +146,13 @@ async def run(self):
...
-                    advisories["metadata"] = self.create_advisory("RHBA", "image", "metadata")
-                else:
-                    _LOGGER.info("Reusing existing metadata advisory %s", advisories["metadata"])
-            # microshift advisory is present since 4.12
-            if self.release_version >= (4, 12):
-                if advisories.get("microshift", 0) <= 0:
-                    advisories["microshift"] = self.create_advisory("RHBA", "rpm", "microshift")
-                else:
-                    _LOGGER.info("Reusing existing microshift advisory %s", advisories["microshift"])
+            is_ga = True if self.release_version[2] == 0 else False
```
vfreex, line 149:
```suggestion
            is_ga = self.release_version[2] == 0
```

pyartcd/pyartcd/pipelines/prepare_release.py
```diff
@@ -146,39 +146,13 @@ async def run(self):
...
+            for ad in advisories:
+                if advisories[ad] < 0:
+                    if ad == "microshfit":
+                        advisories[ad] = self.create_advisory("RHBA", "rpm", ad)
```
vfreex, line 153: RHEA for GA as well?

vfreex (approved):
LGTM

PR #3785, opened 2023-07-11
Skip plashets builds for test assembly (unless explicitly requested)

locriandev:


jobs/build/custom/Jenkinsfile
```diff
@@ -192,6 +192,12 @@ node {
             }
 
             stage("repo: ose 'building'") {
+                if (!params.UPDATE_REPOS && params.ASSEMBLY && params.ASSEMBLY == 'test') {
```
vfreex, line 195: I think we can skip plashet creation for all STREAM typed assemblies, which includes `stream`, `test`, `microshift` and probably others.

locriandev: Done with https://github.com/openshift-eng/aos-cd-jobs/pull/3785/commits/5f0e08931447b89b2eda724eb72250637cedcea7

joepvd: hmm but then `type: stream, name: test` _should_ look at the `name: stream` assembly

vfreex: @locriandev I mean checking the type of assembly so that we don't need to hardcode the assembly names here. Otherwise LGTM.

vfreex: It can be done by reading releases.yml. But if that is too complex for a simple PR like this, I am ok with your current change since it is already an improvment.

vfreex (approved):

jobs/build/custom/Jenkinsfile
joepvd: I would like to have an option to not have outdated repository when using `test` assembly.

PR #3809, opened 2023-07-20
[rebuild] Provide token and data path to all doozer invokes

thegreyd:
## Summary
- Provide github token to rebuild job (this not being there right now breaks the job completely)
- Provide env vars and credentials for doozer to record builds into db
- Provide data path in rebuild job to all doozer invokes

## Test run
https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Frebuild/19/console

### DIFF
--- a/jobs/build/rebuild/Jenkinsfile
+++ b/jobs/build/rebuild/Jenkinsfile
@@ -105,7 +105,10 @@ node {
                 cmd << params.DISTGIT_KEY
             }
             sshagent(["openshift-bot"]) {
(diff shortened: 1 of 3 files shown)


ashwindasr:
Does this rebuild job rebuild images as well? Then should we consider adding the art DB credentials here as well, to match [this](https://github.com/openshift-eng/aos-cd-jobs/pull/3810/files)

ashwindasr (approved):
lgtm

PR #3810, opened 2023-07-20
Add ENV, credentials to log builds to DB

ashwindasr:
Problem: doozer isn't logging to the DB anymore since the ENV values weren't being set

Since the doozer in [buildlib.groovy](https://github.com/openshift-eng/aos-cd-jobs/blob/8f253d1997cc98818b4e9e8d9c201eafabd95022/pipeline-scripts/buildlib.groovy#L167-L186) is not being used anymore, migrating the env values to the main Jenkins file that are necessary to store build logs to DB


ashwindasr:
If we can pass this once for all doozer invokes that would be the best

sosiouxme (approved):
/lgtm
(however ideally we would have a centralized way to invoke `artcd`)

PR #3815, opened 2023-07-21
Make slack readable again

joepvd:


ashwindasr:
Will this still post the error in slack linking to the job if it fails? 🤔

PR #3817, opened 2023-07-24
[ART-1618] [ART-2800] New signing mechanism

vfreex:
This PR adds a new signing mechanism to let promote-assembly job directly send signing requests through UMB.

This eliminates the intermediate step that pulling artifacts from the internet.

Basically this PR gives pyartcd the ability to sign things with pure Python implementation. So that we will not need a dedicated sign-artifiacts job and also eliminate the need to hand over artifacts between jobs. Concretely, it
comes with a class AsyncUMBClient aiming to simplify send and receive UMB messages using request/response model. (Existing python library is hard to use).

```python
uri = "stomp+ssl://umb.stage.api.redhat.com:61612"
cert_file = "ssl/nonprod-openshift-art-bot.crt"
key_file = "ssl/nonprod-openshift-art-bot.key"  
async with AsyncUMBClient(uri, cert_file, key_file) as umb_client:
    # send a message
    await umb_client.send(topic, request_body)
    # receive messages
    receiver = await umb_client.subscribe(consumer_queue, subscription_name)
    async for message in receiver.iter_messages():
        print(message.headers["message-id"])
```

`class AsyncSignatory` sits on top of `class AsyncUMBClient` , which can be called from a python script to sign artifacts:

```python
uri = "stomp+ssl://umb.stage.api.redhat.com:61612"
cert_file = "ssl/nonprod-openshift-art-bot.crt"
key_file = "ssl/nonprod-openshift-art-bot.key"
async with AsyncSignatory(uri, cert_file, key_file, sig_keyname="beta2", requestor="yuxzhu") as signatory:
    pullspec = "quay.io/openshift-release-dev/ocp-release:4.11.31-x86_64"
    digest = "sha256:cc10900ad98b44ba432bc0d99e7d4fffb5498fd6844fc3b6a0a3552ee6d64059"
    # sign a release payload
    with open("signature-1", "wb") as sig_file:
        await signatory.sign_json_digest("openshift", "4.11.31", pullspec, digest, sig_file)
    # sign a message digest
    with open("sha256sum.txt", "rb") as in_file, open("sha256sum.txt.gpg", "wb") as sig_file:
        await signatory.sign_message_digest("openshift", "4.11.31", in_file, sig_file)
```

promote-assembly is updated to use `AsyncSignatory`. `sign-artifacts` job is not used but I haven't removed it.


pyartcd/pyartcd/runtime.py
```diff
@@ -3,7 +3,7 @@
 from pathlib import Path
 from typing import Any, Dict, Optional
 
-import toml
+import tomli
```
vfreex, line 6: This change is not related but let me explain why it is here.

When I attempted to store UMB brokers configuration in artcd.toml, I met some bugs in
`toml` library. It can't correctly parse configs with nested dicts and arrays.

Considering `toml` is not actively maintained and `tomli` has been adopted as a standard library in Python 3.11, I think we should move to `tomli`.

At the end, I decided to write a URI parser to configure UMB brokers using syntax like `stomp+ssl://example.com:12345`, so this became an irrelevant change.

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -409,11 +424,133 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        key_file = os.environ["SIGNING_KEY"]
+        uri = constants.UMB_BROKERS[self.signing_env]
+        sig_keyname = "redhatrelease2" if client_type == 'ocp' else "beta2"
+        self._logger.info("About to sign artifacts with key %s", sig_keyname)
+        json_digest_sig_dir = self._working_dir / "json_digests"
+        message_digest_sig_dir = self._working_dir / "message_digests"
+        base_to_mirror_dir = self._working_dir / "to_mirror/openshift-v4"
+
+        lock_name = f'signing-lock'
+        lock_policy = locks.LOCK_POLICY['compose']
```
locriandev, line 462: Why not defining its own policy? Unless there's reasons to assimilate compose and signing locks, I'd rather keep them separate

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -409,11 +424,133 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        lock_manager = locks.new_lock_manager(
+            internal_lock_timeout=lock_policy['lock_timeout'],
+            retry_count=lock_policy['retry_count'],
+            retry_delay_min=lock_policy['retry_delay_min']
+        )
+        async with await lock_manager.lock(lock_name):
+            async with AsyncSignatory(uri, cert_file, key_file, sig_keyname=sig_keyname) as signatory:
+                tasks = []
+                for _, release_info in release_infos.items():
```
locriandev, line 471:
```suggestion
                for release_info in release_infos.values():
```

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -409,11 +424,133 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+        """ Sign a JSON digest claim
+        :param signatory: Signatory
+        :param pullspec: Pullspec of the payload
+        :param digest: SHA256 digest of the payload
+        :param sig_path: Where to save the signature file
+        """
+        self._logger.info("Signing json digest for payload %s with digest %s...", pullspec, digest)
+        sig_path.parent.mkdir(parents=True, exist_ok=True)
+        with open(sig_path, "wb") as sig_file:
+            if self.runtime.dry_run:
```
locriandev, line 503: I would move this check on the very beginning of the function. We probably don't need to `mkdir`, and even if the context manager will probably close the `sig_file` on a return statement, we can probably avoid creating it in the first place

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -409,11 +424,133 @@ def _reraise_if_not_permitted(self, err: VerificationError, code: str, permits:
...
+    async def _sign_message_digest(self, signatory: AsyncSignatory, release_name, input_path: Path, sig_path: Path):
+        """ Sign a message digest
+        :param signatory: Signatory
+        :param input_path: Path to the message digest file
+        :param sig_path: Where to save the signature file
+        """
+        self._logger.info("Signing message digest file %s...", input_path.absolute())
+        sig_path.parent.mkdir(parents=True, exist_ok=True)
+        with open(input_path, "rb") as in_file, open(sig_path, "wb") as sig_file:
+            if self.runtime.dry_run:
```
locriandev, line 522: Same as above, I would rather move the `dry_run` check right at the beginning of the function

pyartcd/pyartcd/umb_client.py
```diff
@@ -0,0 +1,375 @@
...
+        fut = self._futures.get(id)
+        if not fut:
+            return
+        fut.get_loop().call_soon_threadsafe(fut.set_exception, err)
+
+    def on_disconnected(self):
+        self.__print("on_disconnected")
+        for _, r in self._receivers.items():
+            # close the event loop of the sender thread
+            r._loop.call_soon_threadsafe(r.close)
```
locriandev, line 149: It would look better if we didn't have to access a protected member (and an important one!). How about setting a getter `@property`? Besides, don't we always have only one `asyncio` loop running? Do we really need to store a reference to it within  `AsyncUMBReceiver`? Wondering if we could just use `asyncio.get_event_loop()` where we need it

pyartcd/pyartcd/umb_client.py
```diff
@@ -0,0 +1,375 @@
...
+
+        for id, _ in self._futures.items():
```
locriandev, line 151:
```suggestion
        for id in self._futures.keys():
```

pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -87,8 +95,25 @@ def __init__(self, runtime: Runtime, group: str, assembly: str,
             self._elliott_env_vars["ELLIOTT_DATA_PATH"] = self._ocp_build_data_url
             self._doozer_env_vars["DOOZER_DATA_PATH"] = self._ocp_build_data_url
 
+    def check_environment_variables(self):
```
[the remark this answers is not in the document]
vfreex: Good suggestions! Will get it into the next push.

pyartcd/pyartcd/pipelines/promote.py
vfreex: This is actually a copy-pasta XD. Will fix it in the next push.

pyartcd/pyartcd/umb_client.py
vfreex: I think I need to add more comments here to explain it.

stomp.py calls those Listener methods in a separated thread (started and managed by itself). That thread doesn't have a event loop.

There are 2 event loops here:
- the main event loop
- the event loop in the sender thread

I meant to close the event loop in the sender thread, so the code here is incorrect. I will fix it later.

pyartcd/pyartcd/pipelines/promote.py
vfreex: It was designed like this only because it was easier to test and a lot of functions in standard libraries accepting stream-like objects.

Nowadays forgetting to close a file is no longer an issue since Python uses reference counting and it will close the file when the reference count goes to `0`.

pyartcd/pyartcd/umb_client.py
vfreex: I added a new method `close_soon_threadsafe` to avoid using protected memeber.

vfreex:
Committed some minor fixes and tested with 4.11.46 promotion: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Fpromote-assembly/106/consoleFull

pyartcd/pyartcd/umb_client.py
locriandev: `id` shadows builtin symbol, maybe we can find a more descriptive name?

locriandev:
LGTM

pyartcd/pyartcd/umb_client.py
vfreex: I renamed it to `future_id`.

locriandev (approved):

vfreex:
Thanks for the review.
Rebased. Merging...

PR #3820, opened 2023-07-25
ART-6671 Trigger microshift_sync

locriandev:
Ref. https://issues.redhat.com/browse/ART-6671

### DIFF
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -22,6 +22,7 @@ class Jobs(Enum):
     RHCOS = 'aos-cd-builds/build%2Frhcos'
     OLM_BUNDLE = 'aos-cd-builds/build%2Folm_bundle'
     SYNC_FOR_CI = 'scheduled-builds/sync-for-ci'
+    MICROSHIFT_SYNC = 'aos-cd-builds/build%2Fmicroshift_sync'
 
 
 jenkins_client: Optional[Jenkins] = None
@@ -218,6 +219,17 @@ def start_sync_for_ci(version: str, blocking: bool = False):
         job_name=Jobs.SYNC_FOR_CI.value,
         params={
             'ONLY_FOR_VERSION': version
+        },
+        blocking=blocking
+    )
+
+
+def start_microshift_sync(version: str, assembly: str, blocking: bool = False):
+    return start_build(
+        job_name=Jobs.MICROSHIFT_SYNC.value,
+        params={
+            'BUILD_VERSION': version,
+            'ASSEMBLY': assembly
         },
         blocking=blocking
     )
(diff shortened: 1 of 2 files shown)


pyartcd/pyartcd/jenkins.py
```diff
@@ -221,3 +222,14 @@ def start_sync_for_ci(version: str, blocking: bool = False):
         },
         blocking=blocking
     )
+
+
+def start_microshift_sync(version: str, assembly: str, blocking: bool = False):
```
thegreyd, line 227: Is there a dry_run? It might be worth it to have a dry_run

thegreyd:
A test job in hackspace would be nice to verify this doesn't break on anything

pyartcd/pyartcd/jenkins.py
locriandev: The job has a `DRY_RUN`, but it's not needed when triggered by automation. You can always use it when starting a run by hand, but the Python interface won't need it

locriandev:
> A test job in hackspace would be nice to verify this doesn't break on anything

I agree, but it's a simple change and it's mostly about having it run in production. Possibly, a fix will be easy (same for a revert in case or emergencies)

locriandev:
> @locriandev If you have the jira ticket in title, the PR would get auto-linked to that ticket, fyi

Does not seem to be working though

pyartcd/pyartcd/pipelines/build_microshift.py
```diff
@@ -141,7 +142,9 @@ async def run(self):
             if pr:
                 message += f"\nA PR to update the assembly definition has been created/updated: {pr.html_url}"
             if assembly_type in [AssemblyTypes.PREVIEW, AssemblyTypes.CANDIDATE]:
-                message += f"\n This is a {assembly_type.name} release. Please run <https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fmicroshift_sync/|microshift_sync> to publish the build to mirror."
+                jenkins.start_microshift_sync(version=version, assembly=self.assembly)
```
vfreex, line 145: The `BUILD_VERSION` parameter expects `<major>.<minor>`, like `4.14`. However the `version` returned by `generate_microshift_version_release(release_name)` is the rpm version, which includes the `.z` part, like `4.14.1`.

Let's use value from `major, minor = util.isolate_major_minor_in_group(self.group)` instead.

vfreex (commented):
One issue, otherwise LGTM

pyartcd/pyartcd/pipelines/build_microshift.py
locriandev: Fixed in https://github.com/openshift-eng/aos-cd-jobs/pull/3820/commits/fcbbe7322089d052618b9ccaf065e5f4537b738a

PR #3822, opened 2023-07-26
[ART-7236] call advisory drop from pyartcd

Ximinhan:
drop advisory job only contains two elliott cmd, so migrate the job to python is to call them from pyartcd.


vfreex (approved):
LGTM

PR #3827, opened 2023-07-26
Add warning for PIN_BUILDS and set default to true

ashwindasr:
Scan sources will automatically run otherwise and can cause mass rebuilds, if we are testing.

### DIFF
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -56,7 +56,7 @@ node {
                     ),
                     booleanParam(
                         name: 'PIN_BUILDS',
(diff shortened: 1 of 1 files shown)


locriandev (approved):

jobs/build/ocp4/Jenkinsfile
```diff
@@ -56,8 +56,8 @@ node {
                     ),
                     booleanParam(
                         name: 'PIN_BUILDS',
-                        description: 'Build only specified rpms/images regardless of whether source has changed',
-                        defaultValue: false,
+                        description: 'Build only specified rpms/images regardless of whether source has changed. WARNING: Enable this if you don't want scan sources to run, like if you're testing for new release 4.x for branch cut',
```
thegreyd, line 59:
```suggestion
                        description: 'Build only specified rpms/images regardless of whether source has changed. WARNING: Disable this to use scan-sources as the source',
```

thegreyd (approved):

PR #3836, opened 2023-07-27
[ocp4 images:rebuild] remove stray '' for exclude

ashwindasr:
--images='' will cause doozer `--exclude` option to break


thegreyd:
Let's adjust the unit test too https://github.com/openshift-eng/aos-cd-jobs/blob/225e22f0c2784554cbc7cad2e7238022cedd5554/pyartcd/tests/pipelines/test_ocp4.py#L568

ashwindasr:
ready for another look

thegreyd (approved):

PR #3843, opened 2023-07-31
Fix release state usage

locriandev:
-  `poll-for-failed-nightlies`: improve var naming
- `microshift_sync`: remove unnecessary `software_lifecycle` check
- `ocp3`: related with https://github.com/openshift-eng/ocp-build-data/pull/3302. Currently, only x86 plashets are being built


ashwindasr:
Ahh so this is how it works, we went with best effort last week. lgtm

PR #3846, opened 2023-08-01
delete microshift auto pin branch after merged

Ximinhan:
pin microshift pr is auto-merged, its not useful to keep the ephemeral branch


vfreex (approved):
LGTM, although the duplicated code can be avoided by moving it out of if-else.

Ximinhan:
right, the pr is auto-merged now, maybe the existing pr check is not necessary anymore

PR #3860, opened 2023-08-03
WIP [ART-7190] Switch to openshift-eng/art-tools

ashwindasr:
Switch to `openshift-eng/art-tools`. Ref [ART-7190](https://issues.redhat.com/browse/ART-7190)

Gen-assembly test [run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fgen-assembly/20/).
Prepare-release test [run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fprepare-release/2/console)

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -133,8 +133,8 @@ def setup_venv(use_python38=false) {
     commonlib.shell(script: "rm -rf ${VIRTUAL_ENV}")
 
     // Used by tools that don't use buildlib.doozer() / .elliott()
-    DOOZER_BIN = "${VIRTUAL_ENV}/bin/python3 art-tools/doozer/doozer"
-    ELLIOTT_BIN = "${VIRTUAL_ENV}/bin/python3 art-tools/elliott/elliott"
+    DOOZER_BIN = "${VIRTUAL_ENV}/bin/python3 art-tools/art-tools/doozer/doozer"
+    ELLIOTT_BIN = "${VIRTUAL_ENV}/bin/python3 art-tools/art-tools/elliott/elliott"
 
     if (use_python38) {
         commonlib.shell(script: """
@@ -149,7 +149,7 @@ def setup_venv(use_python38=false) {
     }
 
     env.VIRTUAL_ENV = "${VIRTUAL_ENV}"
-    env.PATH = "${VIRTUAL_ENV}/bin:${env.WORKSPACE}/art-tools/elliott:${env.WORKSPACE}/art-tools/doozer:${env.PATH}"
+    env.PATH = "${VIRTUAL_ENV}/bin:${env.WORKSPACE}/art-tools/art-tools/elliott:${env.WORKSPACE}/art-tools/art-tools/doozer:${env.PATH}"
 
     commonlib.shell(script: "pip install --upgrade pip")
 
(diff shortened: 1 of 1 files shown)


ashwindasr:
Do not merge before @ashwindasr runs the code to sync doozer and elliott to art-tools

thegreyd:
Can you include a test job?

pipeline-scripts/buildlib.groovy
```diff
@@ -164,7 +164,9 @@ def setup_venv(use_python38=false) {
         commonlib.shell(script: "rm -rf art-tools/elliott ; cd art-tools; git clone https://github.com/${where[0]}/elliott.git; cd elliott; git checkout ${where[1]}")
```
thegreyd, line 164: doozer_commit and elliott_commit params would stop working, and I think we want to have art_tools_commit in place of them (see promote / prepare release / build-sync accepting them)

pipeline-scripts/buildlib.groovy
```diff
@@ -164,7 +164,9 @@ def setup_venv(use_python38=false) {
...
     }
 
-    commonlib.shell(script: "pip install -e art-tools/elliott/ -e art-tools/doozer/ -e pyartcd/")
+    commonlib.shell(script: "cd art-tools; rm -rf art-tools; git clone https://github.com/openshift-eng/art-tools.git; cd art-tools; git checkout master")
```
thegreyd, line 167: How does this work? Are we not having art-tools as a submodule to aos-cd-jobs (sorry if i forgot a previous discussion) ?

ashwindasr:
> Can you include a test job?

Apologies, have added now

pipeline-scripts/buildlib.groovy
ashwindasr: I see, let me see if I can understand whats happening. But when we ran the prepare release test job as dry run, it [worked](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Fprepare-release/2/console).

thegreyd: For prepare-release the change was recent and the job run is older (13 days ago) so it doesn't have the elliott_commit param. To see it break you would have to run it with a `elliott_commit` value like `ashwindasr@some_branch_name`

pipeline-scripts/buildlib.groovy
thegreyd: I like our current pattern of keeping it as a submodule, and let's keep it on the root level so it's not `art-tools/art-tools`?

ashwindasr: Ahh regarding that, I wanted to keep this PR reversible. If something happens and we need to roll back, we just need to change the paths to the old doozer and elliott submodules. I was thinking if after a couple of releases we don't have any issues, I'll swap in the whole art-tools folder in the root level. What do you think?

pipeline-scripts/buildlib.groovy
ashwindasr: Understood, will start working on those as well and will add to this PR

pipeline-scripts/buildlib.groovy
thegreyd: What will go wrong though? :)
If we've tested things and it works I don't see us rolling back. And if in case we do need to rollback, we'll revert this PR, and add submodules on it - does not sound much effort to me. I would say let's go forward and not think of going back, we fix issues and they come up

ashwindasr: Makes sense. I was trying to follow a defensive approach, but with the brainpower of our team, I think we can push forward

ashwindasr:
Closing in favor of https://github.com/openshift-eng/aos-cd-jobs/pull/3881

PR #3861, opened 2023-08-04
Signing: Fix unexpected "Connection lost" error

vfreex:
Local variable `id` was renamed to `future_id` for better readability, but this place was forgotten.


vfreex:
Tested with https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Fpromote-assembly/110/

locriandev (approved):

PR #3862, opened 2023-08-04
[ART-7409] Update gen-assembly tekton job for ART OSD

vfreex:
- Bump base image and dependencies for artcd image
- Update gen-assembly job to use pyartcd
- Add externalsecrets for secrets managed on AWS Secrets Manager


ashwindasr:
Tried to test it but my pod wouldn't spawn `0/7 nodes are available: 2 Insufficient cpu, 2 node(s) had untolerated taint {node-role.kubernetes.io/infra: }, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }. preemption: 0/7 nodes are available: 2 No preemption victims found for incoming pod, 5 Preemption is not helpful for scheduling.`

ashwindasr:
Looks like the dynamic scaling worked, the pod threw the same error but I waited for a couple of minutes and the pod was spawned. Testing

ashwindasr:
The auto PR that gets created says created by job run `None` we can maybe add a link to the pipeline run URL? I think we should be able to get some values from the ENV variables.

PR #3864, opened 2023-08-07
Signing: manifests inside a manifest list should also be signed

vfreex:
When signing a heterogeneous (multi) payload, manifests inside the manifest list should also be signed. This is missed in https://github.com/openshift-eng/aos-cd-jobs/pull/3817.

### DIFF
--- a/pyartcd/pyartcd/pipelines/promote.py
+++ b/pyartcd/pyartcd/pipelines/promote.py
@@ -472,11 +472,23 @@ class PromotePipeline:
         async with await lock_manager.lock(lock_name):
             async with AsyncSignatory(uri, cert_file, key_file, sig_keyname=sig_keyname) as signatory:
                 tasks = []
+                json_digests = []
                 for release_info in release_infos.values():
+                    version = release_info["metadata"]["version"]
                     pullspec = release_info["image"]
                     digest = release_info["digest"]
+                    json_digests.append((version, pullspec, digest))
+                    # if this payload is a manifest list, iterate through each manifest
+                    manifests = release_info.get("manifests", [])
+                    for manifest in manifests:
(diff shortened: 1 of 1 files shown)


pyartcd/pyartcd/pipelines/promote.py
```diff
@@ -472,11 +472,23 @@ async def sign_artifacts(self, release_name: str, client_type: str, release_info
...
                 for release_info in release_infos.values():
+                    version = release_info["metadata"]["version"]
                     pullspec = release_info["image"]
                     digest = release_info["digest"]
+                    json_digests.append((version, pullspec, digest))
+                    # if this payload is a manifest list, iterate through each manifest
+                    manifests = release_info.get("manifests", [])
+                    for manifest in manifests:
+                        if manifest["platform"]["os"] != "linux":
+                            raise ValueError("Unsupported OS %s in manifest list %s", manifest["platform"]["os"], release_info["image"])
```
joepvd, line 485: Out of curiosity: Why is this?

joepvd (approved):
/lgtm

pyartcd/pyartcd/pipelines/promote.py
joepvd: Maybe worthy of a comment

vfreex: This is actually a copy-pasta from https://github.com/openshift-eng/aos-cd-jobs/blob/b0063c829947db4c5950035ee993be603ce4e9c2/pyartcd/pyartcd/pipelines/promote.py#L369.
For now the manifest always has `platform.os==linux`. The check is not necessary but kinda defensive programming.

PR #3868, opened 2023-08-08
Search brew logs only in artcd_working

locriandev:
`find-and-compress-brew-logs.sh` is finding an empty brew-logs folder within a `doozer-working` dir. It is not clear who creates that, as `pyartcd` should set doozer working dir within `artcd_working`; nonetheless, this prevents the correct archiving of brew logs within Jenkins.

Updating the script to look only into `artcd_working` should do the trick.

### DIFF
--- a/jobs/build/ocp4/find-and-compress-brew-logs.sh
+++ b/jobs/build/ocp4/find-and-compress-brew-logs.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euxo pipefail
 
-BREWLOGS=`find . -name brew-logs -type d`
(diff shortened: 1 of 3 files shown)


joepvd (approved):
/lgtm
/approve

locriandev:
That extra `doozer_working` dir was most likely being created by [plashets.py](https://github.com/openshift-eng/aos-cd-jobs/blob/master/pyartcd/pyartcd/plashets.py#L221), where doozer working dir was hardcoded. Should be fixed by https://github.com/openshift-eng/aos-cd-jobs/pull/3868/commits/05c2a4b8766ed8cb09b092a6c5e768654d7a66e0

PR #3869, opened 2023-08-08
ART-7242 port github-activity-lock into pyartcd

locriandev:
https://github.com/openshift-eng/aos-cd-jobs/commit/d0604a6b210954e81357ebea9145fa1d994cea9b contains the real change. 

What we currently call `github-activity-lock` is currently shared among `rebuild`, `ocp4` and `ocp4-scan` pipelines. What this PR does is simply move the lock from Jenkins to Redis, and from Groovy into pyartcd.

There's no change in locking logic: once we're ready, we can narrow down the lock scope, wrapping for example only distgit operations.

### DIFF
--- a/pyartcd/pyartcd/pipelines/ocp4.py
+++ b/pyartcd/pyartcd/pipelines/ocp4.py
@@ -861,11 +861,13 @@ class Ocp4Pipeline:
               help='Do not build plashets (for example to save time when running multiple builds against test assembly)')
 @click.option('--mail-list-failure', required=False, default='aos-art-automation+failed-ocp4-build@redhat.com',
               help='Failure Mailing List')
+@click.option('--ignore-locks', is_flag=True, default=False,
+              help='Do not wait for other builds in this version to complete (use only if you know they will not conflict)')
 @pass_runtime
 @click_coroutine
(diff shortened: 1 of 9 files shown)


pyartcd/pyartcd/pipelines/ocp4.py
```diff
@@ -887,4 +889,24 @@ async def ocp4(runtime: Runtime, version: str, assembly: str, data_path: str, da
...
+
+    else:
+        # Create a Lock manager instance
+        lock_policy = locks.LOCK_POLICY['ocp4']
+        lock_manager = locks.new_lock_manager(
+            internal_lock_timeout=lock_policy['lock_timeout'],
+            retry_count=lock_policy['retry_count'],
+            retry_delay_min=lock_policy['retry_delay_min']
+        )
+        lock_name = f'github-activity-lock-{version}'
```
thegreyd, line 904: This looks like it can be a constructor? Also keeping the lock name at one single place

locriandev: Makes sense. I plan to open a refactoring PR to make it all clearer. I prefer not to pollute this change with the next steps though

thegreyd (approved):

PR #3874, opened 2023-08-09
Refactor pyartcd locks handling

locriandev:


pyartcd/pyartcd/locks.py
```diff
@@ -1,54 +1,69 @@
...
+    Lock.SIGNING: {
+        'retry_count': 36000,
+        'retry_delay_min': 0.1,
+        'lock_timeout': 60 * 60 * 1,  # 1 hour
     }
 }
 
 
-class LockManager(Aioredlock):
+class _LockManager(Aioredlock):
```
vfreex, line 66: I think it is ok to keep this type public since the type itself should be visible to other modules.
I guess your purpose is to discourage instantiating it directly. Maybe new_lock_manager can be moved into the class as a static method? Something like:
```python
class LockManager:
  @staticmethod
  def from_lock(policy: Lock):
    return ...
```

vfreex:
Generally LGTM. One nit about instantiating the class.

pyartcd/pyartcd/locks.py
locriandev: Done with https://github.com/openshift-eng/aos-cd-jobs/pull/3874/commits/0cef36c15b13544714f48a35426110e9b583fafe

pyartcd/pyartcd/locks.py
```diff
@@ -1,49 +1,64 @@
...
+
+# Defines the pipeline locks managed by Redis
+class Lock(enum.Enum):
+    def __str__(self):
+        return f'{self.value}-lock'
+
+    OLM_BUNDLE = 'olm-bundle'
+    MIRRORING_RPMS = 'mirroring-rpms'
+    COMPOSE = 'compose'
+    GITHUB_ACTIVITY = 'github-activity'
```
thegreyd, line 17: Can we keep these as `github-activity-{version}` ?
That way it's clear what we are expecting to go where in the string. And then the caller can use it by `.format(version=self.version)`

locriandev: Makes sense. I'll apply the same pattern everywhere else

locriandev: Done in https://github.com/openshift-eng/aos-cd-jobs/pull/3874/commits/2481f7fce81826afcf78ad994259dcfb54c9dac8

thegreyd (approved):

PR #3876, opened 2023-08-09
Improve redis wrapper module

locriandev:
- add `get_keys()`: returns a list of keys matching a given pattern
- add `delete_key()`: removes given key from the DB
- add `handle_connection()` wrapper function to handle connection/disconnection from Redis instance
- add logging

The new functions are not yet used in the pipeline; I've used them to clean up the DB from obsolete keys (named assembly fail counters) so I'd like to keep them for future reuse.

### DIFF
--- a/pyartcd/pyartcd/redis.py
+++ b/pyartcd/pyartcd/redis.py
@@ -1,7 +1,11 @@
+import logging
 import os
+from functools import wraps
 from string import Template
 
 import aioredis
+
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

Think interactions with redis should be not very often. Otherwise I'd say that a redis connection object should be created, possibly made available in runtime. But this is fine :+1:

PR #3878, opened 2023-08-09
Remove github-activity-lock when finished

locriandev:
Locks are currently not being destroyed on time, see https://redis-ui-redis.apps.artc2023.pc3z.p1.openshiftapps.com/


thiagoalessio (approved):
/lgtm

PR #3884, opened 2023-08-10
Prevent long lines getting wrapped in ruamel.yaml

vfreex:
A recent change in ruamel.yaml causes long lines getting wrapped. Explicitly setting `yaml.width = 4096` should solve this issue.


joepvd (approved):
/lgtm
/approve

PR #3898, opened 2023-08-23
WIP Extend lock if needed

thegreyd:
Right now we acquire locks for what we think is the duration of the job (example 8 hours for mass rebuild lock)
In case if the job is aborted, a lock isn't released until the timeout is reached.

Instead we can acquire a lock for a short duration, and keep extending it if we are still processing

### DIFF
--- a/pyartcd/pyartcd/locks.py
+++ b/pyartcd/pyartcd/locks.py
@@ -25,7 +25,7 @@ LOCK_POLICY = {
     Lock.OLM_BUNDLE: {
         'retry_count': 36000,
         'retry_delay_min': 0.1,
-        'lock_timeout': 60 * 60 * 2,  # 2 hours
+        'lock_timeout': 60 * 30 # 30 minutes
     },
     # mirror RPMs: give up after 1 hour
     Lock.MIRRORING_RPMS: {
--- a/pyartcd/pyartcd/pipelines/olm_bundle.py
+++ b/pyartcd/pyartcd/pipelines/olm_bundle.py
@@ -2,6 +2,7 @@ import os
 
 import click
 from aioredlock import LockError
+import atexit
 
 from pyartcd import constants, exectools
 from pyartcd.cli import cli, pass_runtime, click_coroutine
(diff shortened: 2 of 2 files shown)


pyartcd/pyartcd/pipelines/olm_bundle.py
```diff
@@ -60,27 +60,35 @@ async def olm_bundle(runtime: Runtime, version: str, assembly: str, data_path: s
...
-            runtime.logger.info(f'Successfully built:\n{", ".join(bundle_nvrs)}')
+        runtime.logger.info(f'Successfully built:\n{", ".join(bundle_nvrs)}')
 
+    try:
+        # Try to acquire olm-bundle lock for build version
+        async with await lock_manager.lock(lock_name) as lock:
+            task = asyncio.create_task(run())
+            while not task.done():
+                try:
+                    await asyncio.wait_for(task, timeout=60*15)  # Wait for 15 minutes
```
locriandev, line 88: This should probably be a module level constant in `locks.py` and use a clear name

pyartcd/pyartcd/locks.py
```diff
@@ -25,7 +25,7 @@ class Lock(enum.Enum):
     Lock.OLM_BUNDLE: {
         'retry_count': 36000,
         'retry_delay_min': 0.1,
-        'lock_timeout': 60 * 60 * 2,  # 2 hours
+        'lock_timeout': 60 * 30 # 30 minutes
```
locriandev, line 28: Why changing this? I think we still want to keep the pre-defined locks lifetime, or we'll be releasing resources before we actually finished using them.

pyartcd/pyartcd/pipelines/olm_bundle.py
```diff
@@ -60,27 +60,35 @@ async def olm_bundle(runtime: Runtime, version: str, assembly: str, data_path: s
...
+                except asyncio.TimeoutError:
+                    print("Extending lock")
+                    await lock.extend()
```
locriandev, line 91: This should probably account for the case where the lock lifetime is exceeded (could happen for example when a task hangs,  we should stop extending the lock and give up)

ashwindasr:
If https://github.com/openshift-eng/aos-cd-jobs/pull/3925 is merged, please migrate this PR to art-tools, if not urgent.
If urgent, please create a duplicate PR to art-tools.

PR #3906, opened 2023-08-25
Simplify starting of jenkins jobs

locriandev:


pyartcd/pyartcd/jenkins.py
```diff
@@ -185,15 +186,15 @@ def start_ocp4(build_version: str, assembly: str, rpm_list: list,
         params['BUILD_IMAGES'] = 'none'
 
     return start_build(
-        job_name=Jobs.OCP4.value,
+        job_name=Jobs.OCP4,
```
thegreyd, line 189: Think these params would need change `job_name -> job`

locriandev: Right, fixed

PR #3911, opened 2023-08-28
[ART-7593] Trigger SAST scanning for rpms

ashwindasr:
Have the build rpm function trigger the scan after successful RPM build.

- Test [run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ashwin-aos-cd-jobs/job/build%252Focp4/58/)

### DIFF
--- a/pyartcd/pyartcd/jenkins.py
+++ b/pyartcd/pyartcd/jenkins.py
@@ -317,13 +317,23 @@ def update_description(description: str, append: bool = True):
     set_build_description(build, description)
 
 
-def start_scan_osh(build_nvrs: list, email: Optional[str] = "", **kwargs):
-    params = {
-        'BUILD_NVRS': ','.join(build_nvrs)
-    }
+def start_scan_osh(build_nvrs: Optional[list] = None,
+                   rpm_nvrs: Optional[list] = None,
+                   email: Optional[str] = "", **kwargs):
+    params = {}
+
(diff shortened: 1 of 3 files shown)


pyartcd/pyartcd/jenkins.py
```diff
@@ -317,13 +317,23 @@ def update_description(description: str, append: bool = True):
     set_build_description(build, description)
 
 
-def start_scan_osh(build_nvrs: list, email: Optional[str] = "", **kwargs):
-    params = {
-        'BUILD_NVRS': ','.join(build_nvrs)
-    }
+def start_scan_osh(build_nvrs: Optional[list] = None,
```
thegreyd, line 320: This can be `image_nvrs` now

thegreyd: Or we don't necessarily need to differentiate it here
container nvrs will always have `-container` and rpm won't
So we can just pass all nvrs to the job and the job then can sort them

ashwindasr: Hmm if I make the change now, it will break the scans that will be triggered by the `4.13` mass rebuild. Can I work on that as a refinement and raise the PR after the mass rebuild(s) are finished?

thegreyd: Okay let's revisit this. For now this change is good

PR #3912, opened 2023-08-28
ART-7623 Let images rebase and build within the scope of mass rebuild lock

locriandev:
Currently, `mass-rebuild-serializer` lock only protects `doozer images:build`. Rebase happens earlier, and it's protected only by a version-specific lock (currently `github-activity-lock-{version}`. This could lead to such a scenario:

1.  `4.n` and `4.m` both need mass rebuilds. Rebase activities for them run in parallel
2.  `4.n` rebase completes; mass rebuild serializer lock is acquired, `4.n` images start being built
3.  `4.m` rebase completes; mass rebuild serializer lock cannot be acquired; `4.m` is put on hold
4.  `4.n` images build process completes after X hours
5. `4.m` mass rebuild starts, but finds content that was produced many hours before

In order to avoid running mass rebuilds on outdated content, `doozer images:rebase` and `doozer images:build` should happen in sequence, without long waits in between.

### DIFF
--- a/pyartcd/pyartcd/pipelines/ocp4.py
+++ b/pyartcd/pyartcd/pipelines/ocp4.py
@@ -543,9 +543,7 @@ class Ocp4Pipeline:
             jenkins.start_rhcos(build_version=self.version.stream, new_build=False)
 
     async def _rebase_images(self):
-        if not self.build_plan.build_images:
-            self.runtime.logger.info('Not rebasing images')
-            return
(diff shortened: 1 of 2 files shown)


ashwindasr:
If https://github.com/openshift-eng/aos-cd-jobs/pull/3925 is merged, please migrate this PR to art-tools, if not urgent.
If urgent, please create a duplicate PR to art-tools.

vfreex (commented):
/lgtm

PR #3920, opened 2023-08-30
Allow multiple ironic packages versions in plashets

elfosardo:
Metal Platform Team needs to pin specific packages for precise tests and updates.
With this PR we introduce a new list of packages specific for ironic containers that are allowed to be present in multiple versions in the plashets.
We'll start covering only the main packages but we may need to expand the list in the future, separating it fromthe main list of packages is the most sensible solution.

### DIFF
--- a/pyartcd/pyartcd/plashets.py
+++ b/pyartcd/pyartcd/plashets.py
@@ -31,6 +31,14 @@ previous_packages = [
     "python3-openvswitch",
 ]
 
+ironic_previous_packages = [
+    "openstack-ironic",
+    "openstack-ironic-inspector",
+    "openstack-ironic-python-agent",
+    "python-ironic-lib",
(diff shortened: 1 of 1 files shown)


joepvd (commented):
/lgtm
/approve

vfreex:
This affects all OpenShift versions and has caused issues in our pipeline to get packages signed and tracked.
I saw ironic has a `-prevalidation` Brew tag for each OCP version, e.g. `rhaos-4.14-ironic-rhel-9-prevalidation`. Is it possible to tag packages for different use cases (production-ready, testing) into different Brew tags? We can set up separate plashet repos for each Brew tag.

vfreex:
I created https://github.com/openshift-eng/aos-cd-jobs/pull/3928 to restrict this for 4.15+.

PR #3926, opened 2023-08-31
Fix RHACS s3 target dir

locriandev:
Current job is nesting folders in s3, see [this run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fclient_sync/30/console).

 Test run from hack space removed the extra folder: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/client-sync/29/console

### DIFF
--- a/jobs/build/client_sync/Jenkinsfile
+++ b/jobs/build/client_sync/Jenkinsfile
@@ -131,7 +131,7 @@ node {
 
     stage("Sync to mirror") {
         if (params.LOCATION.startsWith("rcm-guest")) {
-            source_path = params.LOCATION.substring(10)
+            source_path = "${params.LOCATION.substring(10)}/"
             commonlib.shell(
                 script:
                 """


jobs/build/client_sync/Jenkinsfile
```diff
@@ -110,6 +110,7 @@ node {
                 error 'SET_LATEST is not supported for rhacs'
             }
             s3_target_dir = "/pub/rhacs/support-packages/${dest_version}"
+            dest_version = "${dest_version}/*"
```
vfreex, line 113: Changing `dest_version` to something like `1.2.3/*` could be misleading.

I looked the code and found this issue could happen for all kinds of binaries besides RHACS. Furthermore, if `SET_LATEST` is check, `latest` directory will contain nesting as well.

It could be resolved by updating the first argument of `commonlib.syncDirToS3Mirror` call to something like `${local_dest_dir}/`.

vfreex (changes requested):

jobs/build/client_sync/Jenkinsfile
locriandev: ROSA clients did not show this issue: [this run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fclient_sync/29/parameters/) for ROSA correctly uploaded binaries to https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/rosa/1.2.25/. [Latest](https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/rosa/latest/) folder was also updated correctly.

I think the problem might be caused by how the source folder is organized in RHACS

locriandev: Also note that `SET_LATEST` cannot be set to true for RHACS clients. I've tried other approaches, such as setting `local_dir` to `local_dest_dir` in `syncDirToS3Mirror`, but this is the only one that worked so far, and also ensures nothing changes for clients other than RHACS.

vfreex: OK, I think I figured out what happened.

My assumption of `It could be resolved by updating the first argument of commonlib.syncDirToS3Mirror call to something like ${local_dest_dir}/.` is wrong. I thought the behavior of `aws sync` works with directories similar to `rsync`, but it does not.

The problem is the `rsync` command used in rhacs:

```sh
mkdir -p /mnt/workspace/jenkins/working/aos-cd-builds_build_client_sync/4.1
rsync -rlp --info=progress2 exd-ocp-buildvm-bot-prod@spmm-util:staging/rhacs/support-packages/4.1 /mnt/workspace/jenkins/working/aos-cd-builds_build_client_sync/4.1
```
The above commands created nested local directories:
```
/mnt/workspace/jenkins/working/aos-cd-builds_build_client_sync/4.1/4.1/ppc64le/{some_file}
```

When syncing directories, whether the source path has a trailing `/` or not works differently if the target directory B already exists:
- `rsync A B` will cause nested directory `B/A` to be created.
- `rsync A/ B` will copy the content of directory `A` to target directory B.
- `aws s3 sync A B` has the same result as `aws s3 sync A/ B`. <-- I thought `aws s3` works like `rsync`, but it is not.
- If target directory B doesn't exist, `rsync A B` behaves the same as `rsync A/ B`.


So when

locriandev: All right, so what's your take on this one? Maybe just adding a trailing / to `dest_version` will work? Or shall we go with current change?

vfreex: @locriandev I think we need to add trailing / to the first argument of `rsync` https://github.com/openshift-eng/aos-cd-jobs/blob/203695a4815c61acf0538b1b61d8a42ec99de484/jobs/build/client_sync/Jenkinsfile#L141C82-L141C93

locriandev:
@vfreex updated the change. Tested with:
- [rhacs](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/client-sync/32/console), created https://mirror.openshift.com/pub/rhacs/support-packages/4.1/
- [rosa](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/client-sync/33/console), created https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/rosa/1.2.26/

vfreex (approved):
/approve
/lgtm

PR #3932, opened 2023-09-04
ART-7628 Add job to clean up unreleased locks from on Redis

locriandev:
Catching the Jenkins user interruption in pyartcd is not deterministic: SIGTERM will wait for threads to complete gracefully, but a SIGKILL will be sent if they don't within a predefined timeout. In this case, no handling procedure will be executed, leaving resources uncleaned.

The proposed solution is to handle the interruption within the Jenkinsfile itself. Then, a new job is called which will be responsible for unlocking all locks left unattended by the interrupted job. `ocp4` will fill the `LOCKS` param according to what locks might have been created. The job will unlock them, if they exist.


thiagoalessio (approved):
/lgtm

PR #3935, opened 2023-09-06
ART-7628 Store Jenkins job ID as lock value

locriandev:
By setting the lock identifier to the ID of the job that created it, we can prevent cases where aborted jobs cancel a lock that was created by a previous (and still running) job.

Tested with:
- nominal scenario: [ocp4](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/ocp4/234/console) was aborted, [cleanup-locks](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/cleanup-locks/15/console) was triggered and deleted `build-lock-4.8`
- [ocp4-235](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/ocp4/235/console) created `build-lock-4.8` and was running; [ocp4-236](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/ocp4/236/console) was started, got blocked on the lock, then got canceled; [cleanup-locks](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/cleanup-locks/16/console) was triggered, but `build-lock-4.8` was not deleted as it belonged to the previous ocp4 run

### DIFF
--- a/jobs/build/ocp4/Jenkinsfile
+++ b/jobs/build/ocp4/Jenkinsfile
@@ -152,6 +152,8 @@ node {
         currentBuild.description = ""
     }
 
+    def job_id = "ocp4-${currentBuild.number}"
+
     try {
 
         sshagent(["openshift-bot"]) {
@@ -194,8 +196,8 @@ node {
 
                 // Needed to detect manual builds
                 wrap([$class: 'BuildUser']) {
-                        builderEmail = env.BUILD_USER_EMAIL
-                    }
+                    builderEmail = env.BUILD_USER_EMAIL
+                }
 
                 buildlib.withAppCiAsArtPublish() {
                     withCredentials([
@@ -213,7 +215,7 @@ node {
                                 file(credentialsId: 'art-jenkins-ldap-serviceaccount-private-key', variable: 'RHSM_PULP_KEY'),
                                 file(credentialsId: 'art-jenkins-ldap-serviceaccount-client-cert', variable: 'RHSM_PULP_CERT'),
                             ]) {
-                        withEnv(["BUILD_USER_EMAIL=${builderEmail?: ''}", "BUILD_URL=${BUILD_URL}", "JOB_NAME=${JOB_NAME}", 'DOOZER_DB_NAME=art_dash']) {
+                        withEnv(["BUILD_USER_EMAIL=${builderEmail?: ''}", "BUILD_URL=${BUILD_URL}", "JOB_NAME=${JOB_NAME}", "JOB_ID=${job_id}", 'DOOZER_DB_NAME=art_dash']) {
                             sh "rm -rf ./artcd_working && mkdir -p ./artcd_working"
                             sh(script: cmd.join(' '), returnStdout: true)
                         }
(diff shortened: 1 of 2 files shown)


thegreyd:
/hold
This PR should move to art-tools now

thegreyd:
@locriandev This looks good overall and paves a way forward -- the fact that we can set and retrieve a value to the lock is great! Thinking about the pattern of aborted job kicking off the cleanup job -- we want to retrofit other jobs that deal with locks as well (ocp4_scan, rebuild, promote, olm_bundle) and more that may come in the future. I'm thinking about the alternative of having cleanup_locks job run periodically and it using jenkinsapi to determine if the job_id associated with the lock is alive or not and then cleaning up based on that. We don't have to do it necessarily with this PR but thinking that would be a way forward, let me know what you think

Ximinhan:
we can set timeout to the lock and use [thread_local](https://aioredis.readthedocs.io/en/latest/api/high-level/#aioredis.lock.Lock.__init__) params to check the value

locriandev:
> @locriandev This looks good overall and paves a way forward -- the fact that we can set and retrieve a value to the lock is great! Thinking about the pattern of aborted job kicking off the cleanup job -- we want to retrofit other jobs that deal with locks as well (ocp4_scan, rebuild, promote, olm_bundle) and more that may come in the future. I'm thinking about the alternative of having cleanup_locks job run periodically and it using jenkinsapi to determine if the job_id associated with the lock is alive or not and then cleaning up based on that. That way the cleanup job will deal with any new jobs / locks added, provided we follow a lock naming pattern for the job to check. We don't have to implement it necessarily with this PR but thinking that would be a way forward, let me know what you think

I think that can be done, we'd need to look into jenkinsapis and find a more precise way of defining job IDs to be assigned as lock values. But yeah, I agree that'd be a win as we wouldn't have to implement cleanup procedures for old and new jobs. I'd like to merge this PR and at least have the clean up in place for `ocp4`, which is where most aborted runs will live, then take this enhancement in a new ticket

locriandev:
/unhold

Opened https://github.com/openshift-eng/art-tools/pull/38 to art-tools

vfreex:
I like the idea to store some customized value so that we will know which build holds the lock. There is just one small thing: I would like to make pyartcd more independent from Jenkins. So that jenkins-specific stuffs can be avoided if possible. Thinking Runtime.get_job_run_name could be a better choice as the identifier.

locriandev:
> I like the idea to store some customized value so that we will know which build holds the lock. There is just one small thing: I would like to make pyartcd more independent from Jenkins. So that jenkins-specific stuffs can be avoided if possible. Thinking Runtime.get_job_run_name could be a better choice as the identifier.

The lock value must be set by Jenkins, as this is where the clean up procedure is initiated from. Therefore, `pyartcd` must stay agnostic with respect to the ldentifier, that gets passed in as an env var.

jobs/build/ocp4/Jenkinsfile
```diff
@@ -152,6 +152,8 @@ node {
         currentBuild.description = ""
     }
 
+    def job_id = "ocp4-${currentBuild.number}"
```
joepvd, line 155: Can we get a full link? 
I'd like to be able to click on the reference, also when the lock is claimed from, say, a hack space.
Or maybe `$BUILD_URL` should be taken.

PR #3950, opened 2023-09-14
ART-7624 Send qe email during promote - Change config key 

thegreyd:
Should go with https://github.com/openshift-eng/art-tools/pull/52


locriandev (commented):
/lgtm

PR #3963, opened 2023-09-20
ART-7628 Provide promote with BUILD_URL env var

locriandev:
`BUILD_URL` env var is needed by `pyartcd` to set the lock value equal to the Jenkins build path


vfreex (commented):
/lgtm
/approve

PR #3979, opened 2023-09-28
Scheduled job to run scan-osh for each version

ashwindasr:
Running with default option to trigger scans progressively and storing value in Redis. Adding `CHECK_TRIGGERD` flag so that we do not trigger scans for ones that we already built. Since for now we are triggering there for those particular NVRs as well

_Requesting lgtm only. Want to check before merge._


locriandev (commented):
/lgtm

PR #3985, opened 2023-10-05
[gen-assembly] Remove limit arches warning

thegreyd:
This is no longer the case after https://github.com/openshift-eng/art-tools/pull/84
Looking for lgtm and approve


joepvd (approved):
/lgtm

PR #3994, opened 2023-10-13
Clear aws cache for mirror_path as well

ashwindasr:
Clear aws cache for mirror_path in case of microshift force rebuild for a particular version, as `rpm_list` was not getting updated.

Ref [slack](https://redhat-internal.slack.com/archives/CB95J6R4N/p1697138145592669?thread_ts=1697122268.631459&cid=CB95J6R4N)


dhellmann:
This looks like the right thing to do to me. It might be overly broad, but I think the impact of that is just a cache reset and there would be no data loss, so that's OK. And of course if we rebuild an RPM with exactly the same name we would want that anyway.

I'm not familiar with modern jenkins, so I will leave the formal review to someone who can comment on syntax, style, etc.

jupierce (approved):

PR #4003, opened 2023-10-18
fix golang builder job

Ximinhan:
golang builder job is not using pyartcd, the current doozer log already compressed by commonlib.safeArchiveArtifacts


vfreex (commented):
/lgtm
/approve

PR #4032, opened 2023-10-30
ART-7854 Remove art-tools submodule

locriandev:
Needs https://github.com/openshift-eng/art-tools/pull/153

Instead of cloning `art-tools` as a git submodule, we can directly install its dependencies from git. Consequences of this fact are:
- we will deploy all `main` changes in production instantaneously; this means particular care will be required when merging new PRs into `art-tools`
- bumping PRs will be no longer neeeded
- we can use run a job against any `art-tools` version (on a a specific fork/branch) as long as `ART_TOOLS_COMMIT` parameter is defined for that job

Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/check-bugs/287/console


sosiouxme:
As part of this, could we have commonlib install via an install script in art-tools, like https://github.com/openshift-eng/aos-cd-jobs/pull/3990/files#diff-9934365fc8ff460eae1ce59bd7b19bf6c6aa808cd45f4f5c16aefe2ed224b27aR163 ?

Because that way we can change how it installs (to e.g. add `artcommon`) without instantly breaking jenkins.

sosiouxme (dismissed):
As part of this, could we have commonlib install via an install script in art-tools, like it now does after https://github.com/openshift-eng/aos-cd-jobs/pull/3990?

Because that way we can change how it installs without instantly breaking jenkins.

Alternatively, maybe make a top-level art-tools module that pulls in the other three and install that from pip.

locriandev:
> As part of this, could we have commonlib install via an install script in art-tools, like it now does after #3990?
> 
> Because that way we can change how it installs without instantly breaking jenkins.
> 
> Alternatively, maybe make a top-level art-tools module that pulls in the other three and install that from pip.

I have updated the install script and turned `artcommon` in a library with its own install script. Please see https://github.com/openshift-eng/art-tools/pull/153 for details

vfreex (commented):
/lgtm
/approve

PR #4046, opened 2023-11-08
ART-5914 Trigger build sync automatically

locriandev:
Automatically trigger build sync after gen-assembly and comment the link of the build sync URL on the auto generated PR

### DIFF
--- a/jobs/build/gen-assembly/Jenkinsfile
+++ b/jobs/build/gen-assembly/Jenkinsfile
@@ -85,6 +85,11 @@ node {
                             description: "Take no action, just echo what the job would have done.",
                             defaultValue: false
                         ),
+                        booleanParam(
+                            name: "TRIGGER_BUILD_SYNC",
+                            description: "Automatically trigger build sync against the automatically created PR",
+                            defaultValue: true
+                        ),
                         commonlib.mockParam(),
                     ]
                 ],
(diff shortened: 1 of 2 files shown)


jobs/build/gen-assembly/Jenkinsfile
```diff
@@ -85,6 +85,11 @@ node {
                             description: "Take no action, just echo what the job would have done.",
                             defaultValue: false
                         ),
+                        booleanParam(
+                            name: "TRIGGER_BUILD_SYNC",
+                            description: "Automatically trigger build sync against the automatically created PR",
```
Ximinhan, line 90: trigger from pr's branch should use it's ability of DOOZER_DATA_GITREF, If I understand correctly this param indicates to do extra logic(comment) against the branch? I think for build_sync job it should be independent

Ximinhan, line 90: trigger from pr's branch should use it's ability of DOOZER_DATA_GITREF, If I understand correctly this param indicates to do extra logic(comment) against the branch? I think for build_sync job it should be independent

locriandev: I don't think I get your comment, could you explain better?

This particular PR does not change anything about DOOZER_DATA_GITREF. art-tools `gen-assembly` pipeline will instruct `build-sync` with DOOZER_DATA_GITREF because it needs to run against a non-standard branch.

joepvd (approved):
/lgtm

PR #4064, opened 2023-12-12
[scan-osh] Update scheduled job to once an hour

ashwindasr:
Update scheduled job to once an hour


scheduled-jobs/build/scan-osh/Jenkinsfile
```diff
@@ -17,7 +17,7 @@ node {
         // Trigger SAST scan for ${version}
         versionJobs["scan-v${version}"] = {
             try {
-                timeout(activity: true, time: 300, unit: 'MINUTES') {
+                timeout(activity: true, time: 1, unit: 'HOURS') {
```
thegreyd, line 20: This is the timeout after which a job will be aborted

PR #4070, opened 2024-01-08
support extract file from el9

Ximinhan:
fix coreos-installer sync job failure: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fcoreos-installer_sync/17/
if coreos-installer rpm is based on el9 then cpio can't directly extract it due to  [RHEL 9 uses a payload compression (zstd) which is not be available on RHEL 7](https://bugzilla.redhat.com/show_bug.cgi?id=2058426#c3) so the pr add a check, use zstd decompress content for el9 rpm.
test job: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fcoreos-installer_sync/4/console

### DIFF
--- a/jobs/build/coreos-installer_sync/extract.sh
+++ b/jobs/build/coreos-installer_sync/extract.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -uo pipefail
+set -euo pipefail
 
 WORKDIR="$1"
 cd "$WORKDIR"
--- a/jobs/build/coreos-installer_sync/extract.sh
+++ b/jobs/build/coreos-installer_sync/extract.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -uo pipefail
 
 WORKDIR="$1"
 cd "$WORKDIR"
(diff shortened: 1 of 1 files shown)


jobs/build/coreos-installer_sync/extract.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -uo pipefail
```
Ximinhan, line 3: `rpm2cpio "${rpm}" | zstd -d | cpio -idm --quiet ./usr/bin/coreos-installer` will return 1 without message but content extracted successfully, so remove -e to not break

vfreex, line 3: I will be skeptical to skip the `e` option. It could result in silent error without notice. Also I would like to see why the exit code is 1 on a successful extraction.

Ximinhan: if with `-e` it failed like https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/ximhan/job/build%252Fcoreos-installer_sync/3/console

thiagoalessio: +1 agree with @vfreex , let's not skip `e`.

thiagoalessio (dismissed):

jobs/build/coreos-installer_sync/extract.sh
Ximinhan: ok, updated in https://github.com/openshift-eng/aos-cd-jobs/pull/4070/commits/e1809ad7db1a521a0dd6f66bdfea5ebc480d517e

thiagoalessio (approved):
/lgtm

PR #4085, opened 2024-02-07
Chunk slack response in nag upstream

thegreyd:
Slack throws an error when text exceeds 3000 chars
https://api.slack.com/reference/block-kit/blocks#section
so chunk response every 2000 chars

Test build
https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Fnag-upstream/5/console

https://redhat-internal.slack.com/archives/C06625LHXQS/p1707324740747989


locriandev (commented):
/lgtm

PR #4086, opened 2024-02-08
remove validate-rhsa

Ximinhan:
The http://sfm2.prodsec.redhat.com/ server is no longer operational, we can't use it anymore and their service move to a place where only errata can access, so deprecate this step for now.

https://redhat-internal.slack.com/archives/GDBRP5YJH/p1707370865294469

### DIFF
--- a/jobs/build/promote-assembly/Jenkinsfile
+++ b/jobs/build/promote-assembly/Jenkinsfile
@@ -235,39 +235,40 @@ node {
         }
     }
 
-    stage("validate RHSAs") {
-        if (release_info.type == "custom" || release_info.type == "preview") {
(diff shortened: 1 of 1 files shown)


jobs/build/promote-assembly/Jenkinsfile
```diff
@@ -235,39 +235,40 @@ node {
         }
     }
 
-    stage("validate RHSAs") {
-        if (release_info.type == "custom" || release_info.type == "preview") {
-            echo "Skipping RHSA check for custom (hotfix) or preview release."
-            return
-        }
+    // SFM2 central server is no longer operational
```
thegreyd, line 238: Instead of commenting it out I am good to remove it

PR #4087, opened 2024-02-08
[ART-8905] Fixes for OCP 3.11 pipeline

thiagoalessio:
Test run: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/talessio-aos-cd-jobs/job/build%252Focp3/13/console

### DIFF
--- a/jobs/build/ocp3/Jenkinsfile
+++ b/jobs/build/ocp3/Jenkinsfile
@@ -647,10 +647,10 @@ node {
                         sh "tito tag --debug --accept-auto-changelog --keep-version --debug"
 
                     }
-                    if (!IS_TEST_MODE) {
-                        sh "git push"
-                        sh "git push --tags"
-                    }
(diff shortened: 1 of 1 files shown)


jobs/build/ocp3/Jenkinsfile
```diff
@@ -647,10 +647,10 @@ node {
                         sh "tito tag --debug --accept-auto-changelog --keep-version --debug"
 
                     }
-                    if (!IS_TEST_MODE) {
```
thiagoalessio, line 650: I commented these out because our buildvm jenkins user is no longer able to push tags to https://github.com/openshift/openshift-ansible ... I'm not suggesting that this is a good fix nor that it should be merged as-is, this was just enough to make the pipeline proceed and succeed.

But now we need to find out if the tagging is still necessary, and if so, ask the owners for writing permissions again.

locriandev (commented):
/hold

PR #4090, opened 2024-02-29
Try running build-sync on buildvm1

joepvd:


jobs/build/build-sync/Jenkinsfile
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env groovy
 
-node('covscan') {
+node {
```
locriandev, line 3: We had moved build-sync to buildvm2 to account for ocp4 migration to buildvm1, and to keep a reasonable resource balance between the 2. Should we move something else to buildvm2?

joepvd:
/close

the issue does not seem tied to buildvm2.

PR #4095, opened 2024-03-05
changed description for art-tools-commit

mbiarnes:


pipeline-scripts/commonlib.groovy
```diff
@@ -97,7 +97,7 @@ def mockParam() {
 def artToolsParam() {
     return [
         name: 'ART_TOOLS_COMMIT',
-        description: 'Override the art-tools submodule; Format is ghuser@commitish e.g. jupierce@covscan-to-podman-2',
+        description: 'Override the art-tools submodule; Format is ghuser@branch-name e.g. jupierce@covscan-to-podman-2',
```
joepvd, line 100: While we are at it, art-tools is not a (git) submodule anymore.

joepvd: commit-ish is more correct (but also more confusing because unknown). Here the dictionary definition: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish

You can actually do `joepvd@<sha>`, or `joepvd@<tag>`. Maybe you can leave `commit-ish` and add `branch-name` as an example?

PR #4096, opened 2024-03-07
Provide prepare-release with GITLAB_TOKEN env var

locriandev:


joepvd (commented):
/lgtm
Please self-approve depending on if you care enough to fix indentation :)

PR #4117, opened 2024-03-25
golang-builder - Drop release suffix

thegreyd:
Doozer recently started adding rhel suffix when rebasing, so don't add here


joepvd (approved):
/lgtm

PR #4118, opened 2024-03-26
Remove sending email code from Jenkinsfile

Ximinhan:
related to https://github.com/openshift-eng/art-tools/pull/497

now the send umb step is a deprecated step, QE won't look at it and we didn't succeed sent it


vfreex:
/lgtm
Though the title could be "Remove sending email code from Jenkinsfile"

PR #4132, opened 2024-04-18
ocp4: use rsync --copy-links

locriandev:
to support https://github.com/openshift-eng/art-tools/pull/555


jobs/build/ocp4/Jenkinsfile
```diff
@@ -198,6 +198,8 @@ node {
                 ]
                 if (params.SKIP_PLASHETS) {
                     cmd << "--skip-plashets"
+                } else {
+                    cmd << "--copy-links"
```
thegreyd, line 202:
```suggestion
                    cmd << "--copy-links" // FIXME: should be removed when remote vm ocp-artifacts gets access to up to date brewroot
```

PR #4140, opened 2024-04-30
Add login for quay.io/openshift/ci (DPTP's QCI) registry

jupierce:
DPTP is moving to use QCI as the source of truth for their CI images. To support this, ART should mirror images both to registry.ci.openshift.org and QCI.
When mapping from registry.ci.openshift.org location to a QCI location, use the format:
 `quay.io/openshift/ci:<namespace>_<imagestream>_<tag_name>`


joepvd (approved):
/lgtm

PR #4147, opened 2024-05-06
Archive and cleanworkspace tweak

thegreyd:


pipeline-scripts/buildlib.groovy
```diff
@@ -647,8 +647,8 @@ def watch_brew_task_and_retry(name, taskId, brewUrl) {
     }
 }
 
-def cleanWorkspace() {
-    cleanWs(cleanWhenFailure: false, notFailBuild: true)
+def cleanWorkspace(cleanWhenFailure=true) {
+    cleanWs(cleanWhenFailure: cleanWhenFailure, notFailBuild: true)
```
Ximinhan, line 651: I think we need to check the log when job failed

Ximinhan (changes requested):

pipeline-scripts/buildlib.groovy
thegreyd: cleanworkspace doesn't remove logs.
`cleanWhenFailure=true` is unecessary in our case because we archive all the needed files
from workspace for success and failure. So we want to clean the workspace up after failure
as well.

Ximinhan (approved):

PR #4149, opened 2024-05-07
Filter out eol versions when running scheduled-jobs

thegreyd:
To mark a version eol we have 2 places, one is our group.yml config and another
is ocp4Versions constant in this repo. We have marking version as eol in config, but
jobs still run for all versions in ocp4Versions. This is so that we filter these eol versions
out for jobs that don't need to run for eol versions.

Test:
https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Focp4_scan/17/console

### DIFF
--- a/scheduled-jobs/build/ocp4_scan/Jenkinsfile
+++ b/scheduled-jobs/build/ocp4_scan/Jenkinsfile
@@ -11,7 +11,11 @@ node {
     ] )
 
     def versionJobs = [:]
-    for ( version in commonlib.ocp4Versions() ) {
+    // ocp4_scan relies on `freeze_automation` config to determine whether to build or not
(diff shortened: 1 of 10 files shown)


scheduled-jobs/build/ocp4_scan/Jenkinsfile
```diff
@@ -11,7 +11,7 @@ node {
     ] )
 
     def versionJobs = [:]
-    for ( version in commonlib.ocp4Versions ) {
+    for ( version in commonlib.ocp4Versions() ) {
```
thegreyd, line 14: This needs to change since ocp4_scan / ocp4 is controlled by `freeze_automation` even for eol releases we need to build

thegreyd:
https://github.com/openshift-eng/art-docs/pull/173

PR #4151, opened 2024-05-08
Promote: Retry to parse json

thegreyd:
readJson flakes from time to time for no reason.
If it fails we fail silently, so send message to slack.


joepvd (approved):
Indeeed, let's not recurse, but just curse twice
/lgtm

PR #4167, opened 2024-05-28
Fail if target directrory for butane sync exists on mirror

joepvd:


jobs/build/butane_sync/Jenkinsfile
```diff
@@ -43,6 +48,18 @@ node {
         if (!params.VERSION) {
             error "VERSION must be specified"
         }
+        if (!params.FORCE) {
+            commonlib.shell(
+                script: """
+                    if curl --fail --silent https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/butane/{params.VERSION}/ >/dev/null 2>&1; then
```
joepvd, line 54:
```suggestion
                    if curl --fail --silent --location https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/butane/{params.VERSION}/** >/dev/null 2>&1; then
```

PR #4171, opened 2024-06-03
Notify when no unresolved threads are found

locriandev:
1. Nofity when no unresolved ART threads are found, so we know the job didn't break and we're actually good
2. add mock param
3. make the notification channel parametric as well, to enable testing (defaults to `#team-art`)


vfreex (commented):
/lgtm
/approve

PR #4172, opened 2024-06-04
Disable content_set check in ocp4/rebuild jobs

vfreex:
The certificate has expired. Temporarily disable it to unblock build. The check will likely be moved to the validator.
 Please enter the commit message for your changes. Lines starting


joepvd (approved):
/lgtm

PR #4181, opened 2024-06-13
fix rm error

ashwindasr:
Was failing with

```
ERROR: failed shell command: 
                set -e
            ...           rm -r output
```
possibly because `output` is a directory


thegreyd:
I don't think this will fix it, the failure comes due to conflicting in package resolution

ashwindasr:
/hold

To make sure that removing stuff is the right way to go

joepvd (approved):
/lgtm

PR #4186, opened 2024-06-28
sigstore: Make all default

joepvd:


jobs/signing/sigstore-sign/Jenkinsfile
```diff
@@ -42,8 +42,9 @@ node {
                     ),
                     choice(
                         name: 'SIGN_RELEASE_IMAGES',
-                        description: 'Sign the release images? Not until OTA-1267 is done.',
+                        description: 'Sign the release images?',
                         choices: ["no", "yes", "only"].join("\n"),
+                        defaultValue: "yes",
```
sosiouxme, line 47: IIRC `defaultValue` does nothing on choices, I would just reorder the choices:
```suggestion
                        choices: ["yes", "only", "no"].join("\n"),
```

sosiouxme (approved):
/lgtm

PR #4190, opened 2024-07-03
use build-in label for build-sync and olm_bundle

Ximinhan:
the covscan label seems unnecessary and misleading


locriandev (approved):
/lgtm

PR #4199, opened 2024-07-12
[ART-10184] sync to cloudflare

ashwindasr:
- Sync to both AWS s3 and Cloudflare r2 buckets

### DIFF
--- a/pipeline-scripts/commonlib.groovy
+++ b/pipeline-scripts/commonlib.groovy
@@ -593,31 +593,36 @@ def invalidateAwsCache(s3_path) {
 def syncRepoToS3Mirror(local_dir, s3_path, remove_old=true, timeout_minutes=60, issue_cloudfront_invalidation=true, dry_run=false) {
     try {
         checkS3Path(s3_path)
-        withCredentials([aws(credentialsId: 's3-art-srv-enterprise', accessKeyVariable: 'AWS_ACCESS_KEY_ID', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY')]) {
-            retry(3) {
-                timeout(time: timeout_minutes, unit: 'MINUTES') { // aws s3 sync has been observed to hang before
-                    // Sync is not transactional. If we update repomd.xml before files it references are populated,
-                    // users of the repo will get a 404. So we run in three passes:
-                    // 1. On the first pass, exclude files like repomd.xml and do not delete any old files. This ensures that we  are only adding
-                    // new rpms, filelist archives, etc.
-                    def opts = '--no-progress --exact-timestamps'
-                    if (dry_run) opts += ' --dryrun'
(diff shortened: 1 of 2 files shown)


ashwindasr:
Ohh wait, I forgot to update logic to use different tokens

/hold

ashwindasr:
actually looks good, secrets `s3-art-srv-enterprise` and `s3-art-srv-enterprise-cloudflare` exist for the aws creds, and secrets `s3-art-srv-enterprise-endpoint` and `s3-art-srv-enterprise-cloudflare-endpoint` exist for the endpoints

pipeline-scripts/commonlib.groovy
```diff
@@ -593,31 +593,36 @@ def invalidateAwsCache(s3_path) {
...
+                            // Otherwise:
+                            // 3. Everything should be sync'd in a consistent way -- delete anything old with --delete.
+                            shell(script: "aws s3 sync ${opts} --delete ${local_dir} s3://art-srv-enterprise${s3_path}")
+                        }
                     }
                 }
-            }
-            if (issue_cloudfront_invalidation) {
-                invalidateAwsCache(s3_path)
+                if (issue_cloudfront_invalidation) {
```
jupierce, line 620: This can be outside the loop. There is no caching with CloudFront on CloudFlare.

pipeline-scripts/commonlib.groovy
```diff
@@ -635,14 +640,18 @@ def syncDirToS3Mirror(local_dir, s3_path, delete_old=true, include_only='', time
...
+        for secretKeyId in ('s3-art-srv-enterprise', 's3-art-srv-enterprise-cloudflare') {
+            withCredentials([
+            aws(credentialsId: secretKeyId, accessKeyVariable: 'AWS_ACCESS_KEY_ID', secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'),
+            string(credentialsId: "${secretKeyId}-endpoint", variable: 'AWS_ENDPOINT_URL')]) {
+                retry(3) {
+                    timeout(time: timeout_minutes, unit: 'MINUTES') { // aws s3 sync has been observed to hang before
+                        shell(script: "aws s3 sync --no-progress --exact-timestamps ${extra_args} ${local_dir} s3://art-srv-enterprise${s3_path}")
+                    }
+                }
+                if (issue_cloudfront_invalidation) {
```
jupierce, line 652: This can be outside the loop.

pipeline-scripts/commonlib.groovy
ashwindasr: If we keep outside, then we would have to use `withCredentials` again right, so added `secretKeyId == 's3-art-srv-enterprise'` to the if

pipeline-scripts/commonlib.groovy
ashwindasr: same as above

PR #4204, opened 2024-07-15
Sync to cloudflare

ashwindasr:
Follows https://github.com/openshift-eng/aos-cd-jobs/pull/4201

Alongside https://github.com/openshift-eng/art-tools/pull/779

In this PR:
- Using `env.CLOUDFLARE_ENDPOINT` to handle groovy string interpolation, in bash scripts, we can skip `.env`
- Syncing to cloudflare uses the `--profile` param, hence syncing to AWS will work since its set as default.


ashwindasr:
/hold

need to rebase

PR #4217, opened 2024-08-14
Use rsync to clean up artcd_working

locriandev:
I randomly noticed that a few ocp4 builds were stuck for a while on `rm -rf ./artcd_working`. Using `rsync` instead should make the operation quicker.

Test build: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/dpaolell/job/ocp4/2/console

### DIFF
--- a/jobs/build/rhcos/Jenkinsfile
+++ b/jobs/build/rhcos/Jenkinsfile
@@ -86,7 +86,8 @@ node {
                 currentBuild.displayName += "multi"
                 echo "triggering multi builds"
                 def jenkins_url = 'https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com'
-                commonlib.shell(script: "pip install -e ./art-tools/pyartcd && rm -rf ./artcd_working && mkdir -p ./artcd_working")
+                commonlib.shell(script: "pip install -e ./art-tools/pyartcd")
+                buildlib.init_artcd_working_dir()
 
                 def dryrun = params.DRY_RUN ? '--dry-run' : ''
                 def run_multi_build = {
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -125,6 +125,18 @@ def initialize_openshift_dir() {
     echo "Initialized env.OPENSHIFT_DIR: ${env.OPENSHIFT_DIR}"
 }
 
+def init_artcd_working_dir() {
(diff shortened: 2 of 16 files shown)


jobs/build/rhcos/Jenkinsfile
```diff
@@ -86,7 +86,8 @@ node {
                 currentBuild.displayName += "multi"
                 echo "triggering multi builds"
                 def jenkins_url = 'https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com'
-                commonlib.shell(script: "pip install -e ./art-tools/pyartcd && rm -rf ./artcd_working && mkdir -p ./artcd_working")
+                commonlib.shell(script: "pip install -e ./art-tools/pyartcd")
```
thegreyd, line 89: do we need this, i think we're doing this in buildllib init?

pipeline-scripts/buildlib.groovy
```diff
@@ -125,6 +125,15 @@ def initialize_openshift_dir() {
     echo "Initialized env.OPENSHIFT_DIR: ${env.OPENSHIFT_DIR}"
 }
 
+def init_artcd_working_dir() {
+    sh """
+    mkdir /tmp/artcd_working
```
thegreyd, line 130: Would like to rename this to empty_dir or similar to indicate our intention here and that this is not meant to be a backup dir

pipeline-scripts/buildlib.groovy
```diff
@@ -125,6 +125,15 @@ def initialize_openshift_dir() {
...
+    rsync -a --delete /tmp/artcd_working ./artcd_working
```
thegreyd, line 131: Can we include a brief comment here, since it is a roundabout way to delete files, for someone who is not familiar with how this is working (me from a few minutes ago :D)

jobs/build/rhcos/Jenkinsfile
locriandev: Saw we have this in other places too, e.g. [here](https://github.com/openshift-eng/aos-cd-jobs/blob/master/jobs/build/promote-assembly/Jenkinsfile#L120). Let's tackle this change in a separate PR

pipeline-scripts/buildlib.groovy
locriandev: Agreed, updated

PR #4234, opened 2024-08-28
ose-operator-registry move to rhel9

Ximinhan:
fix early-operator-index job can't resolve ose-operator-registry pullspec


joepvd (approved):
/lgtm

PR #4241, opened 2024-08-30
remote output reading logic

Ximinhan:
we used to have some logic in Jenkins rely on output, now all steps are in python, the output is not useful anymore


joepvd (approved):
/lgtm

PR #4246, opened 2024-09-12
add option to skip get-nightlies

ashwindasr:
Along side https://github.com/openshift-eng/art-tools/pull/924


ashwindasr:
/approve

will run a test with https://github.com/openshift-eng/art-tools/pull/924 before merging that one in

PR #4247, opened 2024-09-12
update gen-assembly NIGHTLIES param help

ashwindasr:
related to https://github.com/openshift-eng/art-tools/pull/925

the automation checks the presence of "priv" in the nightly name to figure out where to look (private / public release controller)

### DIFF
--- a/jobs/build/gen-assembly/Jenkinsfile
+++ b/jobs/build/gen-assembly/Jenkinsfile
@@ -39,7 +39,7 @@ node {
                         ),
                         string(
                             name: "NIGHTLIES",
-                            description: "(Optional) List of nightlies to match with <code>doozer get-nightlies</code> (if empty, find latest)",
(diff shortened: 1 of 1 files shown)


jobs/build/gen-assembly/Jenkinsfile
```diff
@@ -39,7 +39,7 @@ node {
                         ),
                         string(
                             name: "NIGHTLIES",
-                            description: "(Optional) List of nightlies to match with <code>doozer get-nightlies</code> (if empty, find latest)",
+                            description: "(Optional for public nightlies) List of nightlies to match with <code>doozer get-nightlies</code> (if empty, find latest). If preparing from private nightlies, provide the amd64 nightly as parameter.",
```
thegreyd, line 42: why only amd64? I think it should be "provide all nightlies for desired arches"

ashwindasr: the code will auto fetch the other arches.

thegreyd: ah. Then I'd suggest "provide the amd64 nightly to match"

ashwindasr: updated

PR #4252, opened 2024-09-24
add required env GOOGLE_APPLICATION_CREDENTIALS

Ximinhan:


joepvd:
Think this job does not need access to the konflux db. Let's not give it the credentials. 

/hold

PR #4271, opened 2024-10-17
limit logrotator to 15 days and clean workspace

Ximinhan:
the scan-for-kernel-bugs job now take 15GB on buildvm because it store all history builds and save art-tools in artifacts for each job, this pr clean them to save some spaces


joepvd (approved):
/lgtm

PR #4273, opened 2024-10-22
Dump versions at init

thegreyd:
Test run: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fcheck-bugs/176/console

This is so that we know which version of oc binary is used in a job run

### DIFF
--- a/pipeline-scripts/buildlib.groovy
+++ b/pipeline-scripts/buildlib.groovy
@@ -9,6 +9,13 @@ GITHUB_URLS = [:]
 GITHUB_BASE_PATHS = [:]
 GITHUB_BASE = "git@github.com:openshift"
 
+// dump important tool versions to console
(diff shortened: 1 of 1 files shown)


pipeline-scripts/buildlib.groovy
```diff
@@ -9,6 +9,13 @@ GITHUB_URLS = [:]
 GITHUB_BASE_PATHS = [:]
 GITHUB_BASE = "git@github.com:openshift"
 
+// dump important tool versions to console
+def dump_versions() {
+    sh "oc version --client"
```
thegreyd, line 14:
```suggestion
    sh "ls -l $(which oc)"
    sh "oc version --client"
```

PR #4292, opened 2024-11-19
Adjust poll-payload actions

joepvd:
The poll-payload job runs on a schedule and runs other jobs when there is a new green payload. This proposal wants to change the following things:
- start triggering items for 4.19
- stop publishing rpms for releases that are GA. The released repositories are New Enough for the two known consumers:
  - Microshift CI
  - Bring Your Own RHEL customers


vfreex (commented):
/lgtm
/approve

PR #4298, opened 2024-11-27
Enable telemetry sink for build-sync

lgarciaaco:
Enable otel telemetry data sink by default, but easily disable it via parameters. We can then start sending data to splunk and get some insight into our pipelines.

Screenshot 2024-11-27 at 09 52 58
Screenshot 2024-11-27 at 09 52 49
Screenshot 2024-11-27 at 09 45 07

### DIFF
--- a/jobs/build/build-sync/Jenkinsfile
+++ b/jobs/build/build-sync/Jenkinsfile
@@ -165,6 +165,9 @@ node() {
         }
         if (params.TELEMETRY_ENABLED) {
             env.TELEMETRY_ENABLED = "1"
+            if (params.OTEL_EXPORTER_OTLP_ENDPOINT && params.OTEL_EXPORTER_OTLP_ENDPOINT != "") {
+                env.OTEL_EXPORTER_OTLP_ENDPOINT = params.OTEL_EXPORTER_OTLP_ENDPOINT
+            }
         }
         cmd += [
             "build-sync",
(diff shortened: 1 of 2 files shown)


lgarciaaco:
Depends on: https://github.com/openshift-eng/art-tools/pull/1157

jobs/build/build-sync/Jenkinsfile
```diff
@@ -37,6 +37,8 @@ node() {
                 commonlib.mockParam(),
                 commonlib.ocpVersionParam('BUILD_VERSION', '4'),
                 commonlib.artToolsParam(),
+                commonlib.enableTelemetryParam(),
+                commonlib.telemetryEndpointParam(),
```
locriandev, line 41: We're not using the `OTEL_EXPORTER_OTLP_ENDPOINT`, right? If not blank, we should probably pass it on to build-sync so we can override

vfreex (commented):
/lgtm

jobs/build/build-sync/Jenkinsfile
lgarciaaco: If we set it there is an env var with the same name, build-sync will pick it up then

vfreex: @lgarciaaco I would prefer to make it explicit to set the env var for OTEL_EXPORTER_OTLP_ENDPOINT. WDYT?

locriandev (commented):
/lgtm

PR #4314, opened 2024-12-13
ART-11093 ADD ignore-non-x86-nightlies parameter

Ximinhan:
require https://github.com/openshift-eng/art-tools/pull/1104


joepvd (approved):
/lgtm

PR #4337, opened 2025-01-21
[ART-11852] add create microshift flag for gen assembly job

Ximinhan:
add a gen-microshift parameter for gen-assembly command, unless it set true we will not create microshift entry in assembly config


rayfordj (approved):
/lgtm

PR #4356, opened 2025-02-25
olm_bundle_konflux: Add PLR_TEMPLATE_COMMIT parameter

vfreex:
Similar to the ocp4-konflux job, this parameter can be used to override Pipeline Run template commit from openshift-priv/art-konflux-template; Format is ghuser@commitish e.g. jupierce@covscan-to-podman-2'

pyartcd already supports `--plr-template`. We just need to expose the option to Jenkins.


joepvd (approved):
/lgtm

PR #4374, opened 2025-03-06
Run functional-tests in a job

thegreyd:
Needs https://github.com/openshift-eng/art-tools/pull/1414

Test run:
https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/sidsharm-aos-cd-jobs/job/build%252Ffunctional-tests/14/console

### DIFF
--- a/jobs/build/functional-tests/Jenkinsfile
+++ b/jobs/build/functional-tests/Jenkinsfile
@@ -45,8 +45,12 @@ node() {
                 string(credentialsId: 'jboss-jira-token', variable: 'JIRA_TOKEN'),
                 file(credentialsId: 'konflux-gcp-app-creds-prod', variable: 'GOOGLE_APPLICATION_CREDENTIALS'),
             ]) {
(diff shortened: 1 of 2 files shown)


locriandev:
lgtm. but also I wonder if it's worth giving the user freedom to set the make targets, instead of simply running `functional-elliott` and `functional-doozer` which is what this job is trying to achieve

thegreyd:
@locriandev The "make targets" parameter is intended to be exactly that :) - run any make target available in the Makefile

locriandev:
> @locriandev The "make targets" parameter is intended to be exactly that :) - run any make target available in the Makefile. Only the default is functional-elliott and functional-doozer, but nothing to stop you from running any make task.

I meant exactly the opposite, as in "do we really need that freedom?" :)

locriandev:
anyway, 
/lgtm

locriandev (commented):
/lgtm

PR #4375, opened 2025-03-11
Notify about jobs that exceed 15% fail rate

locriandev:


thegreyd:
I think it's good to have a view of all failures here (even if tiny number) - until we have something more sophisticated like a dashboard.

PR #4379, opened 2025-03-19
Prioritize konflux scans  based on the OCP version

locriandev:


thegreyd:
Don't exactly understand what's happening here .. how does changing the Jenkinsfile name affect things?

locriandev:
> Don't exactly understand what's happening here .. how does changing the Jenkinsfile name affect things?

This PR creates 2 Jenkinsfiles:
- `-lp` for Low Priority, to build OCP <= 4.16
- `-hp` for High Priority, to build OCP >= 4.17

We can then have two separate scheduled jobs to build releases at different paces depending on their version. This is to avoid overloading buildvm, and can be reverted once we succeed in limiting Doozer memory allocation (found a probable issue with RPM and repodata caching), or once we stop building in OSBS

locriandev:
/unhold
After the PR merges, manual configuration will be done in the scheduled-jobs folder in Jenkins

PR #4385, opened 2025-03-27
build-fbc: Use registry config file for pull secret

vfreex:
This is to allow to build FBC for Brew built OLM bundles.

Currently username and password is used for registry auth, but it has a limitation that `oc` command will use the same login for all registries. It makes impossible to use Konflux image registry and Brew registry at the same time. By switching to registry config file, we use different credentials for different registries.

Test build https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/hack/job/yuxzhu/job/aos-cd-builds/job/build%252Fbuild-fbc/29/console (A konflux FBC build for OLM bundle from Brew registry)


joepvd (approved):
/lgtm

PR #4398, opened 2025-04-29
Call pyartcd pipelines at log-level=info

thegreyd:
-vv (debug) results in loads of noisy debug logs in console


locriandev (commented):
/lgtm

PR #4463, opened 2025-08-27
ART-13792: update nightly operator index image

mbiarnes:


joepvd:
/hold

This is not for before branch cut.

PR #4669, opened 2026-04-27
seed-lockfile: clean workspace on failure

joepvd:
## Summary
- Wrap the seed-lockfile stage in `try/finally` so the `terminate` stage (artifact archiving + `cleanWorkspace()`) runs even when the build fails
- Previously a failed build would leave the workspace dirty

## Test plan
- [ ] Verify successful builds still archive artifacts and clean workspace
- [ ] Verify failed builds now also clean the workspace

🤖 Generated with [Claude Code](https://claude.com/claude-code)


joepvd:
/unhold
added some more jobs with the same pattern

PR #4676, opened 2026-04-29
Fix parameter syntax in syncRepoToS3Mirror call

fgallott:
https://github.com/openshift-eng/aos-cd-jobs/pull/4674 broke the sync-for-ci job

example: https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Fsync-for-ci/298526/console

doc: https://docs.groovy-lang.org/latest/html/documentation/#_mixing_named_and_positional_parameters

PR #4677, opened 2026-04-29
Fix syncRepoToS3Mirror to use positional parameters

fgallott:
PR #4676 attempted to fix the parameter syntax by changing `=` to `:`, but this created a new issue: Groovy interpreted `remove_old: false, dry_run: params.DRY_RUN` as a Map literal `[remove_old:false, dry_run:false]` which was passed as the third positional parameter, causing the AWS CLI command to fail with "Unknown options".

The `syncRepoToS3Mirror` function uses positional parameters with defaults, not named parameters. This PR fixes the call to use positional parameters consistently with the rest of the codebase (see `jobs/build/microshift_sync/Jenkinsfile` for similar usage).

**Changes:**
- Updated the call to pass all 6 parameters positionally: `local_dir`, `s3_path`, `remove_old=false`, `timeout_minutes=60`, `issue_cloudfront_invalidation=true`, `dry_run=params.DRY_RUN`

Fixes the sync-for-ci job that was broken by #4676.

### DIFF
--- a/jobs/build/sync-for-ci/Jenkinsfile
+++ b/jobs/build/sync-for-ci/Jenkinsfile
@@ -141,7 +141,7 @@ node {
                      * Instead, pay the relatively minor storage costs for reposync indefinitely. ART can
                      * clean up old directories when they are EOL.
                      */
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

PR #4731, opened 2026-06-12
fix(release-from-fbc): place --dry-run before subcommand

fbladilo:
## Summary
- `--dry-run` is a global option on the `artcd` CLI group (defined in `pyartcd/cli.py`), not on the `release-from-fbc` subcommand
- Click requires parent group options to appear **before** the subcommand name
- The Jenkinsfile was appending `--dry-run` after `release-from-fbc`, causing `Error: No such option: --dry-run`
- Moves the `--dry-run` flag insertion to before the `release-from-fbc` subcommand in the command list

## Test plan
- [x] Verified `--dry-run` is defined as a global option in `pyartcd/cli.py:60`
- [x] Audited all `dry_run` usage in `release_from_fbc.py` — all mutating operations (git push, MR creation, approval rules, CI triggers) are properly guarded
- [ ] Trigger a DRY_RUN=true build of release-from-fbc to verify the fix

Fixes failure in [build #104](https://art-jenkins.apps.prod-stable-spoke1-dc-iad2.itup.redhat.com/job/aos-cd-builds/job/build%252Frelease-from-fbc/104/)

PR #4732, opened 2026-06-12
fix(layered-products): place --dry-run before subcommand

fbladilo:
## Summary
- Same class of bug as [#4731](https://github.com/openshift-eng/aos-cd-jobs/pull/4731) (release-from-fbc)
- `--dry-run` is a global option on the `artcd` CLI group, not on the `build-layered-products` subcommand
- Click requires parent group options to appear **before** the subcommand name
- The Jenkinsfile was appending `--dry-run` after `build-layered-products`, which would cause `Error: No such option: --dry-run`
- Moves the `--dry-run` flag insertion to before the subcommand, matching the pattern used by all other working jobs (build-fbc, ocp4-konflux, gen-assembly, etc.)

## Test plan
- [x] Verified `--dry-run` is defined only as a global option in `pyartcd/cli.py:60`, not on `build-layered-products` subcommand
- [x] Audited all other Jenkinsfiles — this was the only remaining job with the bug
- [ ] Trigger a DRY_RUN=true build of layered-products to verify the fix

### DIFF
--- a/jobs/build/layered-products/Jenkinsfile
+++ b/jobs/build/layered-products/Jenkinsfile
@@ -119,6 +119,13 @@ node {
                 "-v",
                 "--working-dir=./artcd_working",
                 "--config=./config/artcd.toml",
+            ]
+
+            if (params.DRY_RUN) {
+                cmd << "--dry-run"
+            }
+
+            cmd += [
                 "build-layered-products",
                 "-g",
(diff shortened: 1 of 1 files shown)


fgallott (approved):
/lgtm

rayfordj (approved):
/lgtm
/approve

PR #4746, opened 2026-07-08
publish-rpms: skip unavailable packages in dnf download

adobes1:
Add `--skip-unavailable` to the `dnf download` command so packages absent from the configured repos (e.g. `runc` in OCP 5.0) are silently skipped rather than failing the job.


joepvd (approved):
/lgtm

PR #4747, opened 2026-07-08
publish-rpms: use --skip-broken instead of --skip-unavailable

adobes1:
`--skip-unavailable` (introduced in https://github.com/openshift-eng/aos-cd-jobs/pull/4746) is a dnf5 option and not supported in dnf4. `--skip-broken` is the dnf4 equivalent that handles packages not found in the configured repos, allowing the job to succeed when packages like `runc` are absent in OCP 5.0 repos.


joepvd (approved):
/lgtm

PR #4748, opened 2026-07-09
ART-21532: Pass version param to build-plashets job

kopero2000:
## Summary
- Pass the `VERSION` parameter from the Jenkins job to the `artcd` command when building plashets
- If the group is `golang` it is obligatory parameter, because the `MAJOR` and `MINOR` version have to be substituted in `group.yaml`
- `art-tools` repository has to be prepared for this change

### DIFF
--- a/jobs/build/build-plashets/Jenkinsfile
+++ b/jobs/build/build-plashets/Jenkinsfile
@@ -93,6 +93,11 @@ node {
                 currentBuild.description = ""
             }
 
+            // Validate that VERSION is provided for golang group
+            if (group == "golang" && !params.VERSION) {
+                error("VERSION parameter is required when building plashets for golang group")
+            }
+
         }
 
         stage("Build plashets") {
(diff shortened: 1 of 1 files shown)


jobs/build/build-plashets/Jenkinsfile
```diff
@@ -73,6 +73,11 @@ node {
                             description: 'Transform symlink into referent file/dir',
                             defaultValue: false,
                         ),
+                        booleanParam(
+                            name: 'OVERRIDE_OCP_VERSION',
+                            description: 'Substitute the ${MAJOR} and ${MINOR} in group.yaml based on the parameter VERSION',
+                            defaultValue: false,
+                        ),
```
thegreyd: i don't think this is needed. the artcd command should accept both version and group params and then the python code can validate and complain if both are passed for anything other than the golang branch. the golang branch is special since right now it is the only branch that uses the pattern of major.minor variables being passed at runtime. so we won't be using these params for anything else, and it is okay for the python code to know about this and raise error in any other case

PR #4766, opened 2026-08-06
ART-21533: Run periodic conforma-verify jobs in parallel

adobes1:
Remove `--serial` flag so all OCP versions run their conforma-verify jobs concurrently instead of sequentially.


lgarciaaco:
Do you have a test run against this change?

adobes1:
Closing, there's a lot of intermittent errors in the Konflux pipeline when we create this amount of pipelineruns at once.

PR #4816, opened 2026-09-08
fix(rhcos_sync): sign RHEL-10 stream container images (ART-23308)

lgarciaaco:
## Summary

OCP 5.x `coreos-bootimages.yaml` ships a second key `.data.streams` (a map of stream-name → stream-JSON) alongside the existing `.data.stream`. The `rhel-10` entry under `.data.streams` contains the RHEL 10 RHCOS kubevirt container disk images per arch.

Previously only `.data.stream` was extracted and passed to `sign-rhcos-containers`, leaving every RHEL-10 kubevirt image silently unsigned. This caused `SignatureValidationFailed` when HCP KubeVirt on IBM Z tried to pull the RHEL-10 boot image.

## Changes

1. **Extraction step**: after writing `rhcos-${arch}.json` from `.data.stream`, iterate `.data.streams.*` and write `rhcos-${arch}-${stream_name}.json` for each (no-op on releases without extra streams).
2. **Sign stage**: after the existing `sign-rhcos-containers` call, loop over `rhcos-${arch}-*.json` and sign each extra stream file.

## Testing

Retrigger `rhcos_sync` with `SIGN_ONLY=true` against `5.0.0-rc.1-s390x` and verify:
```bash
skopeo inspect --raw docker://quay.io/openshift-release-dev/ocp-v4.0-art-dev:sha256-79badc5fad21d22814d077560c10a90abd7d1896d415548281afbe3b7ea75861.sig
```
Should return a manifest with signature layers (not `manifest unknown`).

## Related
- Jira: [ART-23308](https://redhat.atlassian.net/browse/ART-23308)
- Bug: [OCPBUGS-86850](https://redhat.atlassian.net/browse/OCPBUGS-86850)

### DIFF
--- a/jobs/build/rhcos_sync/Jenkinsfile
+++ b/jobs/build/rhcos_sync/Jenkinsfile
@@ -119,7 +119,7 @@ node {
         # Extract any additional per-stream boot image data (e.g. rhel-10 in OCP 5.x).
         # .data.streams is a map of stream-name -> stream-JSON; each may contain container
         # images that also need to be cosigned.
-        yq -r '(.data.streams // {}) | keys[]' \$tmp/coreos-bootimages.yaml 2>/dev/null | while read stream_name; do
+        yq -r '(.data.streams // {}) | keys[]' \$tmp/coreos-bootimages.yaml | while read stream_name; do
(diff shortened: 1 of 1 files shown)


joepvd (approved):
/lgtm

PR #4834, opened 2026-09-17
[ART-23708] prevent error: '/run/user/984' directory set by  does not…

pruan-rht:
… exist.

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED


shruti-rh (approved):
/lgtm
/approved

PR #4852, opened 2026-09-17
restrict build system to konflux only

ashwindasr:


ashwindasr:
/override ci/prow/art-pre-commit-check

Created PR directly from GitHub

PR #4856, opened 2026-09-21
acm-2.17 build trigger added

MacdonEm:
Adds build trigger for acm-2.17.

Ticket: https://redhat.atlassian.net/browse/HYPBLD-872


pipeline-scripts/commonlib.groovy
```diff
@@ -29,6 +29,7 @@ ocp5Versions = [
 
 nonOCPGroups = [
     "acm-2.16",
+    "acm-2.17",
```
ashwindasr, line 32: I don't see an acm-2.17 branch through https://github.com/openshift-eng/ocp-build-data/tree/acm-2.17

ashwindasr (changes requested):

pipeline-scripts/commonlib.groovy
MacdonEm: I was not able to create a branch in the repo. I have a new branch in a fork but cannot merge it yet until there's a "landing" branch to merge into. This can wait for after that occurs 👍🏻

ashwindasr: its created - thanks!

